PR #212, opened 2025-12-02
Add File Explorer menu with keyboard shortcuts

sinelaw:


sinelaw:
The focus restoration after accepting rename does not work yet

PR #234, opened 2025-12-04
Add GitHub Action to update Homebrew formula on release

AlexanderWillner:
We can already install `fresh` via `brew install fresh-editor`.

This workflow triggers automatically when a GitHub release is published, it uses the dawidd6/action-homebrew-bump-formula action to create a PR to Homebrew/homebrew-core (it automatically calculates the new SHA256 from the release tarball and submits the PR for review by Homebrew maintainers).

For this to work, you need to create a HOMEBREW_TAP_TOKEN secret in your repository settings: Go to your GitHub account > Settings > Developer settings > Personal access tokens > Fine-grained tokens > Create a new token with:
Repository access  > Select Homebrew/homebrew-core with permissions "Contents: Read and write, Pull requests: Read and write". Finally, add the token as a repository secret named HOMEBREW_TAP_TOKEN in your fresh repository (Settings > Secrets and variables > Actions).

PR #235, opened 2025-12-04
Enable homebrew in dist

sinelaw:
Contents autogenerated from dist init using instructions from https://axodotdev.github.io/cargo-dist/book/installers/homebrew.html


sinelaw:
see also #234

PR #249, opened 2025-12-07
Re-sort file tree when sort mode changes

lorenjphillips:
Noticed the file tree sort mode wasnt actually re-sorting when you switched modes - you had to collapse and re-expand directories to see the change. Found the TODO comment at view.rs:268 and figured I'd take a crack at it.

## Changes

- Added `SortMode::comparator()` that returns the right comparison fn for each mode
- Added `FileTree::sort_all_nodes()` to re-sort all expanded directories 
- Updated `set_sort_mode()` to trigger the resort

## Testing

- added two tests for the sort mode switching behavior
- ran `cargo test file_tree` - all 36 tests pass
- built with `cargo build --lib`, no new warnings

Let me know if you want me to change anythign!

### DIFF
--- a/src/view/file_tree/tree.rs
+++ b/src/view/file_tree/tree.rs
@@ -310,6 +310,38 @@ impl FileTree {
         self.nodes.len()
     }
 
+    /// Re-sort children of all expanded directories using the provided comparator
+    pub fn sort_all_nodes<F>(&mut self, cmp: F)
+    where
+        F: Fn(&FsEntry, &FsEntry) -> std::cmp::Ordering,
+    {
+        let expanded_dirs: Vec<NodeId> = self
+            .nodes
+            .values()
+            .filter(|n| n.is_expanded())
+            .map(|n| n.id)
+            .collect();
+
+        for dir_id in expanded_dirs {
+            let children_with_entries: Vec<(NodeId, FsEntry)> =
+                if let Some(node) = self.nodes.get(&dir_id) {
+                    node.children
+                        .iter()
+                        .filter_map(|&id| self.nodes.get(&id).map(|n| (id, n.entry.clone())))
+                        .collect()
+                } else {
+                    continue;
+                };
+
+            let mut sorted = children_with_entries;
+            sorted.sort_by(|(_, a), (_, b)| cmp(a, b));
+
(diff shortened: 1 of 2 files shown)


sinelaw:
Where in the UI can I see the effect of this change?

lorenjphillips:
Thanks for reviewing this! sort mode isn't wired up to any keybinding yet... I added the fix first figuring the UI part would be simpler to add on top, happy to add what I built for myself but figured you may be opinionated about UI :)

> Where in the UI can I see the effect of this change?

sinelaw:
Gotcha
I'll take a look at adding sort to the file explorer ui

lorenjphillips:
Thanks so much!! Let me know if you would like any more changes in the PR for the sort!!

sinelaw:
This is an automatic review comment, this is an autogenerated comment

- **Tested locally**: Built with `cargo build --release` and `cargo check --all-targets` (clean, no new errors). Launched the editor in a tmux session and opened the file explorer (Ctrl+E) — directories render first correctly (default Type sort), no crashes
- **Fix is correct**: `set_sort_mode()` now properly re-sorts all expanded directories via `sort_all_nodes()`, resolving the existing TODO comment
- **Issue — overly broad visibility on `sort_all_nodes`**: It is declared `pub` in `tree.rs` but is only ever called by `set_sort_mode` in `view.rs` within the same module. It should be `pub(crate)` (or `pub(super)`) to avoid leaking an internal implementation detail to the whole crate/world
- **Issue — weak first unit test**: `test_sort_mode_changes_order` uses entries `["dir1", "dir2", "file3.txt"]` — this order is identical under both Type and Name sort modes, so the test passes even *without* the fix (the re-sort is a no-op on this data). Per CONTRIBUTING.md §Testing rule 1, the test must fail without the change. Please use data where the two modes produce different orderings (as `test_sort_mode_type_dirs_first` does correctly)

---

sinelaw (changes requested):
Two items need fixing before merge (see inline comment for details): (1) `sort_all_nodes` visibility should be `pub(crate)` not `pub`; (2) `test_sort_mode_changes_order` must use data that actually fails without the fix.

---

PR #285, opened 2025-12-10
Add cargo flag to disable plugins

sinelaw:
This adds a `plugins` feature (enabled by default) that controls whether the plugin system and its deno dependencies are compiled. When disabled with `--no-default-features`, the binary is built without deno_core, deno_ast, and deno_error, significantly reducing dependencies.

Key changes:
- Add `plugins` feature in Cargo.toml with optional deno dependencies
- Create PluginManager abstraction that provides unified interface for both enabled and disabled cases, minimizing cfg attribute spread
- Make v8_init module conditional on plugins feature
- Make plugin runtime/thread/process modules conditional
- Update Editor to use PluginManager instead of direct thread handle
- Mark plugin-related tests to skip when plugins are disabled

When plugins are disabled, all plugin hooks become no-ops and the editor runs without TypeScript/JavaScript plugin support.

### DIFF
--- a/src/app/input.rs
+++ b/src/app/input.rs
@@ -1606,7 +1606,9 @@ impl Editor {
                 #[cfg(not(feature = "plugins"))]
                 {
                     let _ = action_name;
-                    self.set_status_message("Plugins not available (compiled without plugin support)".to_string());
+                    self.set_status_message(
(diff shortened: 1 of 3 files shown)


sinelaw:
Fixes #285

PR #292, opened 2025-12-11
Docs: Add cargo-binstall installation method

dvchd:
## Description
This PR updates the `README.md` to include **`cargo-binstall`** as an installation method.

## Motivation
Currently, the `cargo install` method compiles the project from source, which can take a significant amount of time depending on the machine.

Since this project already publishes pre-built binaries in GitHub Releases with standard naming conventions, `cargo-binstall` works out-of-the-box. This allows users to install the binary directly in seconds, providing a much faster and smoother onboarding experience for Rust users.

## Changes
- Added a row for `cargo-binstall` in the **Installation** table.
- Added a detailed command section **above** the **From crates.io** section (to prioritize the faster installation method).

## Verification
I have tested this locally on my machine. The command correctly detects the release assets and installs the binary without issues.


sinelaw:
Thanks!

PR #322, opened 2025-12-14
cleanup flake and use toolchain from fenix

Vortriz:
pretty much what it says in the title.

on a side note, the build is successful but i have turned off the tests because it was throwing some error which i don't understand. anyone more familiar with rust (and nix) is welcome to try.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,7 @@ types/fresh.d.ts
 
 # Nix/direnv
 .direnv/
-result
+result*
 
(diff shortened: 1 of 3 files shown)


sinelaw:
Just to be sure - did you nix build successfully after this change? We don't do it in CI

Vortriz:
Yupp, the nix build was successful. The app ran without any issues too.

sinelaw:
Thank you!

PR #343, opened 2025-12-16
Add `--syncdeps` to `makepkg` commands

KSXGitHub:
I failed to take into account that `makepkg` could fail due to missing dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ brew install fresh-editor
 ```bash
 git clone https://aur.archlinux.org/fresh-editor-bin.git
 cd fresh-editor-bin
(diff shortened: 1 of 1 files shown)


sinelaw:
Did you test this locally?

KSXGitHub:
> Did you test this locally?

I use Arch Linux. This is common knowledge.

KSXGitHub:
If you're still unconvinced, I am also the creator of this action you are using: https://github.com/sinelaw/fresh/blob/732ed55b12c403fb2fc07ec5626dc6865dd17049/.github/workflows/aur-publish.yml#L57

Issue #393, opened 2025-12-19
Setting File Explorer Width changes the setting by 100x upon relaunch.


bithooked:
Use the menu to set File Explorer -> Width to a number, such as 20.  Exit, relaunch, open File Explorer, note it is very wide.  Open the Settings and note the value for File Exploer -> Width is now 100x, in this case 2000.  Setting it back to 20 again sets it to 2000 upon relaunch.  Only workaround I've found is to manually open the config file and manually set the width to "0.20" and then never touch it in the settings menu.  

This is on v0.1.55 from Arch AUR.

sinelaw:
Fixed in fed9423c3abe4925a8b8a4bbc78e22bc8c5f78d4

PR #413, opened 2025-12-21
Replace unicode char that isn't supported on windows properly (fixes #400)

### LINKED ISSUE #400: File exporer displaying '?' character on Windows 10
Image

sinelaw:
TODO: Add symbol fields (collapsed, expanded, selected, etc.) to Theme struct and replace all hardcoded glyphs in render code


sinelaw:
Fixes #393

PR #429, opened 2025-12-22
Paste crlf issues on Windows

sinelaw:


sinelaw:
Let me know if it actually fixes the issue (I haven't tried windows on my end yet, relying on automatic testing)

graphixillusion:
Is this multiline paste should handle the column paste when multiple lines are selected?

sinelaw:
See if it works better by pasting with the menu (edit...paste) or command
palette, as opposed to keyboard Ctrl+v. It might work better

I think it's been reported I need to reproduce it on Windows to figure it
out.

graphixillusion:
For now i don't see any particular differences between ctrl+v or edit paste gui / command palette. But i think that a copy/paste seleciton made between select all (ctrl-a), then ctrl-c and ctrl+v should act different between a selection made from multiple cursors: when you copy/paste the block text, it should paste as column. I hope i've describe well the point.

Issue #484, opened 2025-12-27
KDE: cannot paste into Fresh using


ak1dd:
Installed fresh-editor 0.1.64 from the AUR.

Using CachyOS, KDE 6.5.4, Wayland - any terminal,  fish/bash/zsh.

Open a terminal, run fresh.    Open a session of anything else you can copy text from.  So firefox/kate/etc.   Copy text.

CTRL-V, middle mouse button nor the menu for Paste do anything.   Open nano in the same terminal and paste works fine.

Copying FROM fresh to anything else works fine.

Am I missing something?  I have to have this to use fresh.

sinelaw:
Strange, I can't reproduce this so far on any other platform (will need to try on KDE, maybe it's specific to that)
I also use Arch but without KDE

ak1dd:
I have been playing with this and I found that if I use CTRL-SHIFT-V it does paste.     To try to troubleshoot I tried alacrity and konsole but they both fail too (Google Gemini tried to blame ghostty.)       I understand there was a clipboard bug fixed in KDE 6.5.4 but sadly updating didn't change the situation.   I've now toggled and played with every clipboard option in KDE and nothing works.  Gemini sent me down a ton of ratholes with things to install and configure and nothing worked.   I'm pretty sure this is a Fresh issue as wl-paste always displays the text correctly.

sinelaw:
I can repro this on a KDE machine. Doesn't happen here on XFCE.
Fresh uses arboard https://github.com/1Password/arboard

sinelaw:
Could be related to https://github.com/1Password/arboard/issues/114

ak1dd:
thank you!!  I LOVE fresh...  a next gen tool that doesn't pretend more than one generation has grown up using GUI tools!

PR #513, opened 2025-12-28
Enable wayland support in arboard + update deps

sinelaw:


sinelaw:
Fixes https://github.com/sinelaw/fresh/issues/484

PR #571, opened 2025-12-30
Add support for Odin

xoxorwr:
Needs to double check: `queries/odin/indents.scm`, it was vibe coded

For the tree sitter package, it's an outdated one, but the only one i could find on cargo

The up to date one is here: https://github.com/laytan/odin-tree-sitter, but i don't know how to include it, i know nothing about rust

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1217,6 +1217,7 @@ dependencies = [
  "tree-sitter-javascript",
  "tree-sitter-json",
  "tree-sitter-lua",
+ "tree-sitter-odin",
  "tree-sitter-pascal",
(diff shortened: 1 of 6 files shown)


sinelaw:
Please rebase on latest master

sinelaw:
Will be pushed in https://github.com/sinelaw/fresh/pull/755

I added there syntect (sublime-style) grammar which is faster and also basic LSP config so that users can just install 'ols' and lsp should work out of the box (I did not try it myself though)

sinelaw:
Thanks, closing this will continue in the other PR

PR #584, opened 2025-12-31
Render scrollbars with background fills to avoid glyph gaps

### LINKED ISSUE #561: MacOS default terminal font glitches
Which font is recommended to use with the app? 
Standard macOS Tahoe Terminal with default settings shows scroll bar like this 

Image

I suspect the problem is with the font used. NERD fonts looks the same BTW.

Smotrov:
This small PR is supposed to fix #561

### DIFF
--- a/src/app/render.rs
+++ b/src/app/render.rs
@@ -835,9 +835,9 @@ impl Editor {
                     &self.cached_layout.split_areas
                 {
                     if sid == split_id {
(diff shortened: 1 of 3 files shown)


sinelaw:
Great idea! I'm cherry-picking your commit into another branch which will also fix the test failures

sinelaw:
merged in another PR https://github.com/sinelaw/fresh/pull/588

Thanks @Smotrov !

PR #594, opened 2025-12-31
fix: Add missing semantic_highlight_cache initialization

tufloc:
## Summary
The `from_content` constructor was missing the `semantic_highlight_cache` field initialization, causing compilation to fail.

This was introduced in commit 14e24ec which added the field to EditorState but didn't update all constructors.

## Test
- `cargo check` passes after this fix


sinelaw:
Thanks! Just pushed an identical commit :)

PR #638, opened 2026-01-03
Add support packages requried to use fresh on the fresh code base

dwt:
Just some small things to get a more full featured development experience out of the box.

Is that something you would take in?


sinelaw:
I don't use nix (or nix flakes) myself so can't really judge, i'll take your word for it

PR #639, opened 2026-01-03
Clean up output by resetting to column zero after each log

dwt:
This prevents jumbled and hard to read output.


sinelaw:
Thanks!

PR #649, opened 2026-01-03
 fix cursor can't go to the final line #622

### LINKED ISSUE #622: cursor can't go to the final line
open a new file

enter, cursor at L2

up, cursor at L1

down, cursor can't move down.

asukaminato0721:
fix #622

Fixed LineIterator so it emits the synthetic empty line that exists conceptually after a trailing newline. This allows MoveDown at EOF in a fresh buffer (with “new file” default content of \n) to see another line instead of stopping.

Added regression test test_line_iterator_trailing_newline_emits_empty_line


asukaminato0721:
I run cargo test , but it failes on some e2e tests, but when I run them one by one by hand, they passes.

Also cargo test seems stuck at a test.

sinelaw:
This needs the fix from https://github.com/sinelaw/fresh/pull/656 for the two failing tests to pass.

sinelaw:
Very nice, Thanks! Merged your commit in another pr that also fixes the related tests.

Issue #667, opened 2026-01-04
Toggle comment command should use syntax of correct language


sinelaw:
> toggle_comment() in render.rs has hardcoded stuff, it should use syntect or tree sitter or at least a shared module instead 

 _Originally posted by @sinelaw in [#559](https://github.com/sinelaw/fresh/issues/559#issuecomment-3707980795)_

sinelaw:
Fixed

PR #683, opened 2026-01-05
fix(toggle-comment): Use language config for comment prefixes and pre…

### LINKED ISSUE #681: freeze on multiline comment
The editor freezes if I select several lines of code and press Ctrl-/ (to comment/uncomment a block of code).

ubuntu-linux fresh 0.1.70

sinelaw:
…serve selection

- Replace hardcoded extension-to-comment mapping with lookup from LanguageConfig
- Use detect_language() to find language, then get comment_prefix from config
- Do nothing if language is unknown or has no comment prefix configured
- Users can now customize comment prefixes in their config file
- Preserve selection after commenting/uncommenting by tracking byte deltas and adding MoveCursor event to restore anchor and position
- Fix infinite loop when selection end equals buffer length (use >= instead of >)
- Add e2e tests for comment prefix detection, selection preservation, and regression tests for files without trailing newlines

Fixes #681

### DIFF
--- a/src/app/render.rs
+++ b/src/app/render.rs
@@ -2842,28 +2842,29 @@ impl Editor {
 
     /// Toggle comment on the current line or selection
     pub(super) fn toggle_comment(&mut self) {
-        // Determine comment prefix based on file extension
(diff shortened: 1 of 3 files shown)


sinelaw:
Also fixes #667

PR #684, opened 2026-01-05
fix make the link in hover clickable #603

### LINKED ISSUE #603: make the link in hover clickable
some lsps will provide them.

Image

this comes from pyrefly.

asukaminato0721:
fix #603

on linux it will trigger uri.

with opus 4.5

image

### DIFF
--- a/src/view/popup.rs
+++ b/src/view/popup.rs
@@ -735,9 +735,9 @@ impl Popup {
                 // Word-wrap styled lines to fit content area width
                 let wrapped_lines = wrap_styled_lines(styled_lines, content_area.width as usize);
 
-                // Collect link positions for OSC 8 hyperlink rendering
-                // Each entry: (line_index, col_start, col_end, url)
-                let mut link_positions: Vec<(usize, usize, usize, String)> = Vec::new();
+                // Collect link overlay info for OSC 8 rendering after the main draw
+                // Each entry: (visible_line_idx, start_column, link_text, url)
+                let mut link_overlays: Vec<(usize, usize, String, String)> = Vec::new();
 
                 let visible_lines: Vec<Line> = wrapped_lines
                     .iter()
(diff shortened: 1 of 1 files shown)


asukaminato0721:
there is a potential improvement: use fresh to open this uri.

sinelaw:
how can I reproduce the same hover? (for trying it myself)

asukaminato0721:
one way is to

uv tool install pyrefly

then set lsp to

command: pyrefly
arg: lsp

then open a .py file

sinelaw:
Ok I can get it to work! For file:// it opened it in Vi of all things - see below

Maybe we should try OSC 8? It might give a better experience in some cases like when using terminal over ssh (it will open the local user's browser instead of a browser inside the remote machine)

https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda 
and also see https://github.com/Alhadis/OSC8-Adoption 


image

sinelaw:
some problem in how its rendered 
image

sinelaw:
Also - with OSC 8 I don't think we want any click handlers at all, just to change how such links are rendered. I'm not sure where is the best layer to do this. Looking at the code.

src/view/popup.rs
```diff
@@ -703,22 +734,64 @@ impl Popup {
...
+                    .map(|(line_idx, styled_line)| {
+                        let mut col = 0usize;
                         let spans: Vec<Span> = styled_line
                             .spans
                             .iter()
-                            .map(|s| Span::styled(s.text.clone(), s.style))
+                            .map(|s| {
+                                let span_width =
+                                    unicode_width::UnicodeWidthStr::width(s.text.as_str());
+                                // Track link positions for OSC 8 rendering
```
sinelaw: can drop link tracking

asukaminato0721:
https://github.com/ratatui/ratatui/issues/902

solution:

https://github.com/itsjunetime/ratatui/blob/6a0b8ddf76ed9ecd88578766f5c90e8938bf0c43/examples/apps/hyperlink/src/main.rs#L55

PR #685, opened 2026-01-05
load cache first

### LINKED ISSUE #668: add cargo cache also to the build ci
I noticed that  Swatinem/rust-cache@v2 is used, but not in build step.

asukaminato0721:
fix #668

now ci will be faster.

now

image

vs

old

image


sinelaw:
Thanks

PR #705, opened 2026-01-07
impl Semantic Highlight #628

### LINKED ISSUE #628: Semantic Highlight
https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide

asukaminato0721:
fix #628

with gpt5-codex-high

Implemented real LSP semantic-token rendering while keeping the existing scope-aware reference highlighter intact.

Semantic Tokens

  - introduces a monotonic version field and bumps it inside mark_content_modified, so every edit carries a precise revision stamp for cache validation.
  - now carries an optional SemanticTokenStore, and exposes helpers to set/clear/result-id-check tokens so renderers can decide when data is still valid.
  - advertises full semanticTokens client support, persists each server’s legend/full-support bit, and threads that data through AsyncMessage::LspInitialized.
  - tracks pending/in-flight semantic-token requests; requests tokens for visible buffers; gates new requests on version/legend availability; and plus decode relative token data into byte ranges and stash them alongside the version that produced them.
  - resolves semantic spans only when store.version == state.buffer.version() and maps LSP token types/modifiers onto the existing theme palette so Syntect output remains the fallback during edits.
  - teaches the fake server to advertise/delay semantic tokens and adds an end-to-end regression ensuring stale tokens are never rendered; the new test drives the whole pipeline.

Reference Highlighting

  - The old “semantic highlight” scope tracker is now ReferenceHighlighter, keeping its identifier-matching behavior but freeing up the “semantic” naming for LSP tokens.
  - The debounce cache still exposing the same interface but under the new name to mirror the highlighter.
  - Module exports and state wiring were updated to the new names, and rendering continues to fetch cursor-word spans through the renamed cache.
  - The event loop continues to poll the debounce timer via the updated accessor, while the obsolete semantic_highlight modules were removed to avoid symbol clashes.

  Tests

  - cargo test test_semantic_tokens_version_gating

image

the inlayhint is other issue.

image

### DIFF
--- a/tests/common/fake_lsp.rs
+++ b/tests/common/fake_lsp.rs
@@ -73,7 +73,7 @@ while true; do
 case "$method" in
     "initialize")
         # Send initialize response
-        send_message '{"jsonrpc":"2.0","id":'$msg_id',"result":{"capabilities":{"completionProvider":{"triggerCharacters":[".",":",":"]},"definitionProvider":true,"hoverProvider":true,"textDocumentSync":1,"semanticTokensProvider":{"legend":{"tokenTypes":["keyword","function","variable"],"tokenModifiers":["declaration","deprecated"]},"full":true}}}}}'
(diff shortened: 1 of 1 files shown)


sinelaw:
:fire: awesome, thanks

PR #708, opened 2026-01-07
fix inlay hint position in rust lsp #626

### LINKED ISSUE #626: inlay hint position in rust lsp
Image

vscode

Image

asukaminato0721:
fix #626

with gpt codex 5.2 high

Adjusted inlay hint placement to treat LSP positions as insertion points and render them before the character at that byte offset (with a safe EOF fallback). This should line up Rust analyzer hints with VSCode instead of shifting them to the line start.

image

old

image


sinelaw:
:partying_face:

PR #710, opened 2026-01-07
try fix highlight don't move with the inserts

asukaminato0721:
fix https://github.com/sinelaw/fresh/issues/628#issuecomment-3719288031

Semantic token highlights now track edits because they’re stored as overlays (markers move with inserts/deletes), and the render pipeline rebuilds semantic token spans from those overlays instead of static byte ranges. This keeps the existing semantic token styling behavior but fixes the “doesn’t move on insert” bug.

  Details

Added semantic token overlay support in src/services/lsp/semantic_tokens.rs, with namespace + color mapping + overlay creation and a small unit test verifying overlays shift after inserts.

Updated src/app/async_messages.rs to apply semantic token overlays whenever tokens arrive.

Updated src/view/ui/split_rendering.rs to extract semantic token spans from overlays (and exclude them from regular overlay application).

### DIFF
--- a/tests/e2e/lsp.rs
+++ b/tests/e2e/lsp.rs
@@ -848,7 +848,6 @@ fn test_lsp_waiting_indicator() -> anyhow::Result<()> {
     target_os = "windows",
     ignore = "FakeLspServer uses a Bash script which is not available on Windows"
 )]
-#[ignore = "Semantic highlighting temporarily disabled"]
 fn test_semantic_tokens_version_gating() -> anyhow::Result<()> {
     use crate::common::fake_lsp::FakeLspServer;
 
@@ -859,6 +858,7 @@ fn test_semantic_tokens_version_gating() -> anyhow::Result<()> {
     std::fs::write(&test_file, "fn main() {}\n")?;
 
     let mut config = fresh::config::Config::default();
+    config.editor.enable_semantic_tokens_full = true;
     config.lsp.insert(
         "rust".to_string(),
         fresh::services::lsp::LspServerConfig {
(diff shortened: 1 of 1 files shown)


sinelaw:
Some of the automatic comments seem relevant

sinelaw:
There's some flicker when semantic highlighting is enabled
bug-fresh-flicker-lsp

It happens both with and without this latest change (the earlier branch introduced the flicker as well)

I'm guessing re-creating all the overlay markers is causing an issue. In other places we try to only make incremental changes (e.g. only update the region where the user is viewing), and/or debouncing the requests (don't update on every single edit) which is a bit more of a hack.

I ran it with RUST_LOG=fresh=trace to collect logs and it's evident we do a lot of marker related work all the time: 

[fresh-74124.log](https://github.com/user-attachments/files/24482452/fresh-74124.log)

So I think we should do both - use 'range' requests (for viewport-only updates) and in the background do debounced full requests to ensure scrolling looks nice (this part may be nice to have - maybe just 'range' requests should be enough?)

Summary from Gemini: To eliminate flickering and performance lag during heavy editing, transition the LSP semantic highlighting from full-file reloads to a Viewport-First
  Incremental strategy:

   1. Spatial Optimization (`semanticTokens/range`):
       * On edit/scroll, prioritize a textDocument/semanticTokens/range request for the visible viewport (plus a 10-line buffer).
       * New API Required: Implement OverlayManager::replace_range_in_namespace(range, namespace, new_overlays, marker_list) to surgically update the visible
         area while preserving "shifted" markers in the rest of the document.

   2. Temporal Optimization (`semanticTokens/full/delta`):
       * Store the resultId from each response and use it to send textDocument/semanticTokens/full/delta requests.
       * Apply returned deltas to the existing Overlay set instead of clearing the entire namespace.

   3. Concurrency Control:
       * Track in-flight request_ids and send $/cancelRequest when a new edit occurs to prevent processing stale results.
       * Debounce background "Full/Delta" requests (500ms) while keeping "Range" requests highly responsive (50ms).

   4. State Stability:
       * Maintain current "shifted" overlays via the marker_list during request latency; never clear the semantic namespace until new data (range or delta) is
         ready to replace it. This ensures highlighting is always present, even if slightly misaligned for a few milliseconds.

asukaminato0721:
Viewport-first semantic tokens are now range-based with overlays updated only inside the visible union (plus padding), and full refreshes are debounced after edits. This keeps existing markers stable, avoids clearing the whole namespace each time, and reduces marker churn that caused flicker. I also wired semanticTokens/range through the LSP client and fake server, and added an overlay manager helper for range replacement.

  Key changes

Range requests: render computes per-buffer visible line window and requests semanticTokens/range (with 10-line padding). Responses replace only that namespace range via new OverlayManager::replace_range_in_namespace, leaving off-viewport overlays untouched.

Debounced full refresh: edits schedule a full semanticTokens refresh after 500ms idle to backfill/refresh the global store without spamming.

LSP capability plumbing: range support added to capabilities extraction, LspInitialized, and manager tracking; LSP client now handles semanticTokens/range.

Fake LSP now advertises range: true and responds to semanticTokens/range.

sinelaw:
It's better but there's a new (different) flickering issue in the updated branch. I think the range requests are not working correctly while the full requests are working. Suggestion:  disable the 'full' requests and write an e2e test that leaves only the range requests, and check the highlighting works in the visible area, find out if it's really the issue

Another thing: we aren't using the 'delta' support for full requests, so for large files we're constnatly re-applying the full overlays. Better if we used the delta api

sinelaw:
The flicker is gone but I see some bad highlighting when scrolling through one of the rust files:

Good: 
image


Scrolling one more line down => Bad highlighting:

image

sinelaw:
Another note - avoid passing / returning tuples if you can declare dedicated named struct types for those cases

sinelaw:
This should fix the bad highlighting issue on scroll (see below)


```
diff --git a/src/app/async_messages.rs b/src/app/async_messages.rs
index 3cfdacf8..fc051820 100644
--- a/src/app/async_messages.rs
+++ b/src/app/async_messages.rs
@@ -282,11 +282,14 @@ impl Editor {
                     Ok(tokens_opt) => {
                         let spans = match tokens_opt {
                             Some(SemanticTokensRangeResult::Tokens(tokens)) => {
+                                // LSP semantic tokens are always delta-encoded from document
+                                // position (0,0), even for range requests. The range only
+                                // filters which tokens are returned, not the encoding origin.
                                 let (_, spans) = decode_semantic_token_data(
                                     &state.buffer,
                                     &legend,
                                     &tokens.data,
-                                    requested_start_line.unwrap_or(0),
+                                    0,
                                 );
                                 spans
                             }
@@ -295,7 +298,7 @@ impl Editor {
                                     &state.buffer,
                                     &legend,
                                     &partial.data,
-                                    requested_start_line.unwrap_or(0),
+                                    0,
                                 );
                                 spans
                             }

```

sinelaw:
When I start lsp it works but then when i make an edit it disappears
fresh-enter-breaks-lsp

  do we have an e2e test for validating semantic highlight is rendered and that it         
  persists and moves correctly with edits? before fixing your branch please make sure we have a test that captures this issue (and fails). Thanks!

tests/e2e/lsp.rs
```diff
@@ -936,6 +937,350 @@ fn test_semantic_tokens_version_gating() -> anyhow::Result<()> {
...
+        .collect();
+    assert!(
+        !handles_before.is_empty(),
+        "Expected semantic token overlays before Enter key"
+    );
+
+    // Position cursor at end of line and press Enter key specifically
+    // (not type_text("\n") which may use different code path)
+    harness.send_key(KeyCode::End, KeyModifiers::NONE)?;
+    harness.send_key(KeyCode::Enter, KeyModifiers::NONE)?;
```
asukaminato0721: I guess here?

sinelaw:
for me with rust it still disappears on pressing enter, or even on just resizing the window

asukaminato0721:
[Screencast_20260111_234709.webm](https://github.com/user-attachments/assets/39088e1d-21fa-47ca-b8e6-6cf804fa76a2)

sinelaw:
It reproduces consistently for me, took a while to get it to repro in a test, can you cherry-pick the top two commits from https://github.com/sinelaw/fresh/commits/pull-710e/ 

failing test:         FAIL [   2.620s] fresh-editor::e2e_tests e2e::lsp::test_semantic_tokens_not_replaced_by_stale_lsp_tokens

Theory from Claude: DO NOT trust this, it's just a direction for analysis:

  BUG REPRODUCED: Semantic token color became White after edit due to stale LSP tokens.                                                
                                                                                                                                       
  Summary                                                                                                                              
                                                                                                                                       
  Root Cause Identified:                                                                                                               
  1. When you press Enter (or any edit), rust-analyzer sends a textDocument/didChange notification                                     
  2. Fresh requests semantic tokens for the viewport                                                                                   
  3. rust-analyzer responds with tokens, but if analysis is incomplete, token types are "unresolvedReference" (mapped to White)        
  4. Fresh applies these stale tokens, overwriting the correct tokens (keyword=Cyan, struct=Magenta, etc.)                             
                                                                                                                                       
  Evidence from real rust-analyzer logs:                                                                                               
  - Before edit: tokens had types like "namespace", "struct" → Magenta
  - After edit: tokens had type "unresolvedReference" → White      
                                                                                                                                       
  Test created: test_semantic_tokens_not_replaced_by_stale_lsp_tokens in tests/e2e/lsp.rs                                              
  - Uses fake LSP that simulates rust-analyzer behavior                                                                                
  - Sends correct tokens initially, then "variable" type tokens after didChange                                                        
  - Test FAILS, proving the bug exists
                                                                   
  Potential fixes:                                                                                                                     
  1. Don't replace specific tokens with generic ones - If existing token has a specific type (keyword, function, struct), don't overwri
te with a generic type (variable, unresolvedReference)                                                                                 
  2. Detect stale responses - If all tokens in a response have the same generic type, treat it as stale
  3. Version-based filtering - Track document version and ignore responses for older versions

sinelaw:
That theory is probably wrong, with a fix for that the issue still happens localy

sinelaw:
it looks like a rust-analyzer issue, i'm  not sure how to work around it and also why it only happens on my machine and not on yours

sinelaw:
Merged for now we'll follow up work on improving the issue I encountered. Probably best thing is to try and utilize full mode (+ delta) and drop the range queries which may be less well supported?

sinelaw:
And Thanks :100:

Issue #799, opened 2026-01-19
Can't perform "C-p" or "C-q" when toggled "vi mode"


lixuan998:
When I toggled "vi mode" and trying to quit "fresh editor", I found that "C-q" had no effect. Then I tried to disable "vi mode" by using "C-p", no respond either.

sinelaw:
Fixed in master in https://github.com/sinelaw/fresh/pull/808/changes/c5eccb4f9aacff6ada7993eb7acd1ef925525410

sinelaw:
Also you'll be able to use :q :wq etc

PR #808, opened 2026-01-19
Arm64

sinelaw:


sinelaw:
Fixes https://github.com/sinelaw/fresh/issues/799

PR #838, opened 2026-01-23
fix: enable 6 previously ignored tests and improve dropdown UX

GeneticxCln:
## Summary

This fixes a critical usability bug where the theme dropdown in settings would not respond to arrow keys. Users were unable to cycle through themes using Right/Left arrows, which made the settings panel frustrating to use. The fix adds proper arrow key handling to the dropdown input handler.

While investigating this, I also noticed several E2E tests were marked as ignored. After reviewing them, 6 tests are now enabled - some just needed their test assertions updated, others were already passing and just needed the ignore attribute removed.

### Dropdown Arrow Key Support
Added Right and Left arrow key handling to the dropdown input handler. When a dropdown is open:
- Right arrow moves to the next option, wrapping to the first
- Left arrow moves to the previous option, wrapping to the last

This matches standard dropdown behavior in most applications.

### Test Fixes
- test_settings_theme_dropdown_cycle: now works with arrow key fix
- test_map_control_add_new_shows_text_input: updated key sequence to match current dialog behavior
- test_line_numbers_rendered_correctly: fixed assertion for files with trailing newlines
- test_page_down_line_numbers: was already passing, removed ignore
- test_merge_start_resolution_command: was already passing, removed ignore  
- test_diff3_conflict_with_base_section: was already passing, removed ignore

### Config Schema
Added three missing Tokyo Night theme variants that were in the codebase but not in the schema.

## Testing
Ran the full test suite. All 6 previously-ignored tests now pass.

### DIFF
--- a/crates/fresh-core/src/api.rs
+++ b/crates/fresh-core/src/api.rs
@@ -77,7 +77,9 @@ impl CommandRegistry {
 
     /// Register a command
     pub fn register(&self, command: Command) {
-        let mut commands = self.commands.write()
+        let mut commands = self
+            .commands
+            .write()
             .expect("Command registry lock poisoned - this indicates a bug in the editor");
         commands.retain(|c| c.name != command.name);
         commands.push(command);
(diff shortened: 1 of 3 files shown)


sinelaw:
Which applications use left/right to select dropdown items? It's normally up/down, no?

GeneticxCln:
Left/Right navigation was removed in favor of standard Up/Down navigation.

sinelaw:
can we separate the PR only for testing fixes? I'm not sure about the rest of the changes (theme changes - I think it's already fixed, try rebasing on latest master; and I don't see the benefit in the error code changes)

GeneticxCln:
do as you want

PR #839, opened 2026-01-23
Update it.json

fmaida:
Replaced the verbs “scarta” and “discarta” with “dimentica” (“forget”), which sounds more appropriate in an Italian UI context and preserves the “d” shortcut.

Replaced “Cancella” with “Annulla”, which also sounds more appropriate in this context.

### Note: 

in some overwrite prompts the shortcut is shown as `(o)vrascrivi`.
In Italian, users would naturally expect `(s)ovrascrivi`, but the shortcut appears to be hard-coded to `o` (likely derived from overwrite), or at least I couldn’t find a `prompt.key.overwrite` mapping.

I left it unchanged to avoid breaking key handling, but it might be worth revisiting in the future (e.g. allowing localized shortcuts).


sinelaw:
Thanks!

PR #894, opened 2026-01-29
feat: auto save implementation

bacteriafield:
**What was implemented ?**

## Persistent auto-save feature:
This feature allows the editor to automatically save modified buffers to their original files on disk.

image

## Manual Verification Instructions
To test this manually:

Open [fresh](https://github.com/sinelaw/fresh).
Enable auto-save in your configuration:
```
{
  "editor": {
    "auto_save_enabled": true,
    "auto_save_interval_secs": 1
  }
}
```
Open a file, make a change, and wait for 1 second.

### DIFF
--- a/crates/fresh-editor/config.example.json
+++ b/crates/fresh-editor/config.example.json
@@ -17,7 +17,6 @@
     "enable_inlay_hints": true,
     "enable_semantic_tokens_full": false,
     "recovery_enabled": true,
-    "auto_save_interval_secs": 2,
     "highlight_context_bytes": 10000,
     "mouse_hover_enabled": true,
     "mouse_hover_delay_ms": 500,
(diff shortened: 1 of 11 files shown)


sinelaw:
1. let's separate the time config, so the new auto-save is not the same timeout as auto-recovery save
2. also, i think we should disable this feature in large file mode (e.g. editing a 10gb file, should not auto-save even if enabled, only manual save) because the save is very heavy. the config doc for the boolean flag should explain this only kicks in for files < large file threshold

sinelaw:
@elaurentium 
Do you want to address the comments and complete this, or should I take it from here?

bacteriafield:
@sinelaw 
If you want, I can implement the autosave without an interval. I had completely forgotten about all of this.

bacteriafield:
@sinelaw I thought you were going to reuse what I had already done and omit the unwanted part.

bacteriafield:
checkout @sinelaw 

Screenshot 2026-02-18 at 10 11 27

sinelaw:
Pushed your commit + adaptations in https://github.com/sinelaw/fresh/pull/1058

Thank you!

PR #929, opened 2026-02-02
impl Add VSCode-style “Move Line Up/Down” shortcut (Alt+Up / Alt+Down) #914

### LINKED ISSUE #914: Add VSCode-style “Move Line Up/Down” shortcut (Alt+Up / Alt+Down)
It would be great if the editor supported the same VSCode behavior for moving lines / selections up and down:

Alt + Up Arrow → move current line up

Alt + Down Arrow → move current line down

When multiple lines are selected, the entire selected block should move together.

This is a very commonly used VS Code editing shortcut and would significantly improve editing speed and familiarity.

asukaminato0721:
fix #914

Added MoveLineUp/MoveLineDown actions, wired Alt+Up/Alt+Down in the default and macOS keymaps, and implemented line swapping that preserves selections/cursors and handles trailing newline edge cases.

Also updated editing-disabled gating, added i18n strings across locale files, and added a regression test for moving a line up without a trailing newline.

### DIFF
--- a/crates/fresh-editor/locales/vi.json
+++ b/crates/fresh-editor/locales/vi.json
@@ -95,6 +95,8 @@
   "action.move_left": "Di chuyển con trỏ sang trái",
   "action.move_line_end": "Di chuyển đến cuối dòng",
   "action.move_line_start": "Di chuyển đến đầu dòng",
+  "action.move_line_up": "Di chuyển dòng lên",
+  "action.move_line_down": "Di chuyển dòng xuống",
   "action.move_page_down": "Di chuyển xuống một trang",
   "action.move_page_up": "Di chuyển lên một trang",
   "action.move_right": "Di chuyển con trỏ sang phải",
--- a/crates/fresh-editor/src/input/actions.rs
+++ b/crates/fresh-editor/src/input/actions.rs
@@ -1,8 +1,9 @@
 //! Action to event conversion - translates high-level actions into buffer events
 
 use crate::input::keybindings::Action;
+use crate::input::line_move::{move_lines, LineMoveDirection};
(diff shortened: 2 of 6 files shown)


sinelaw:
Thanks a few comments

1. when there's an active selection, the move up/down action should slide the entire selection (all lines) up or down
2. i think it's broken for large files (files > large file threshold),

sinelaw:
(Let me know when it's ready for review)

sinelaw:
Missing locale keys

i18n::tests::test_all_locales_have_required_keys

sinelaw:
Testing gaps:

No test for MoveLineDown without trailing newline — There's test_move_line_up_without_trailing_newline but no equivalent for moving down. The trailing newline handling logic (actions.rs:420-436) is the most complex part; testing both directions on this edge case would be valuable.

No boundary tests — No tests for moving the first line up (should be a no-op) or the last line down (should be a no-op). These are important to confirm the early-return paths work.

No multi-cursor test — No test for two cursors on separate lines moving simultaneously, which exercises the merge_line_ranges and multi-region paths.

Temp file naming in tests — test_move_line_up_large_file_unloaded_chunks and test_move_line_down_large_file_selection_block use hard-coded temp file paths (/tmp/test_large_file_move_line_up.txt). If tests run in parallel, these could collide. Using unique names (e.g., tempfile crate) would be safer.

See README.md for testing guidelines

sinelaw:
Also, the ~430 lines of new line-move code (types, helpers, move_lines, map_position_in_region) are entirely self-contained. Consider moving them to a line_move.rs submodule

sinelaw:
🙏🙏🙏 
Thanks

Issue #915, opened 2026-02-01
Custom themes appear in "Edit Theme" but not in "Select Theme"


MathiasDrizzy:
### Description
Custom themes installed in `~/.config/fresh/themes/` are visible when using "Edit Theme → Open Theme" (Ctrl+O), but they don't appear in the "Select Theme" command palette option.

### Steps to Reproduce
1. Download custom theme files (e.g., [Catppuccin themes](https://github.com/milon/catppuccin-fresh))
2. Copy `.json` theme files to `~/.config/fresh/themes/`
3. Restart Fresh
4. Press `Ctrl+P` and search for "Select Theme"
   - **Expected**: Custom themes appear in the list
   - **Actual**: Only built-in themes are shown
5. Press `Ctrl+P` and search for "Edit Theme", then press `Ctrl+O` (Open)
   - **Actual**: Custom themes ARE visible here

### Environment
- Fresh version: `0.1.98`
- OS: macOS
- Installation method: Homebrew

### Workaround
The only way to select a custom theme is through "Edit Theme → Open Theme" or by manually editing `~/.config/fresh/config.json`:
```json
{
  "version": 1,
  "theme": "catppuccin-mocha"
}
```

However, even with this config, Fresh loads the default theme instead of the custom one.

### Expected Behavior
Custom themes in `~/.config/fresh/themes/` should:
1. Appear in the "Select Theme" command palette
2. Be selectable and applied when chosen
3. Be loaded from `config.json` on startup

### Additional Context
Theme files are valid JSON matching the documented format from https://getfresh.dev/docs/features/themes. I've verified:
- JSON syntax is valid
- Theme structure matches documentation
- File permissions are correct (`-rw-r--r--`)
- Cleared Fresh caches in `~/Library/Application Support/fresh/`

sinelaw:
Should be fixed in master branch

PR #930, opened 2026-02-02
Fix theme directory on macOS

sinelaw:
Use DirectoryContext as calculated in main bringup flow instead of guessing default paths for config


sinelaw:
Fixes #915

Issue #488, opened 2025-12-27
Feature request: Set text encoding (UTF8, UTF16 etc.)


tbone2k-git:
Hello! o)

Congrats, very nice editor! o)

It would be awesome if this editor would
- detect correct text file encoding when loading files
- have an option to set the encoding the text file will be saved with
- ASCII, ANSI, UTF8, UTF8+BOM, UTF16-LE would be cool to have at least

This ticket here might be related:
https://github.com/sinelaw/fresh/issues/456

Thank you! o)

lededev:
I’d also like to request support for additional encodings, specifically GB18030 and GBK. My runtime environment is UTF‑8, but I store files in GB18030 because UTF‑8 uses 3 bytes for most East Asian characters, while GB18030 uses only 2 bytes. Both UTF‑8 and GB18030 are full Unicode encodings, but GB18030 offers more compact storage for these characters. Supporting these encodings would help reduce file size and improve compatibility with existing projects.

sinelaw:
Done, added 'Set Encoding' and a clickable status bar indicator with the encoding

tbone2k-git:
Awesome guys! o) In other projects you wait a decade for something like this! Insanely good! o))

lededev:
I encountered a crash when opening GB18030 files. I think the middle of the file can be determined by CR LF characters. GB18030 is compatible with 7-bit ASCII.

```
fresh 《道门法则》（校对版全本）作者：八宝饭.txt
Error: Failed first run setup

Caused by:
    Large file (7.5 MB) uses GBK encoding which requires loading the entire file into memory. This encoding cannot be streamed because character boundaries cannot be determined from the middle of the file. Continue?
```

sinelaw:
@lededev try latest master, should be fixed now

lededev:
I pulled the latest master branch and installed using:
`cargo +stable-x86_64-pc-windows-msvc install --path crates/fresh-editor`
When I run fresh with a filename as a parameter, it crashes with:
```
E:\>fresh 《道门法则》（校对版全本）作者：八宝饭.txt
Error: Failed first run setup

Caused by:
    GBK (8 MB) requires full load. (l)oad, (e)ncoding, (C)ancel?

```
When I run fresh with no parameters and then press Ctrl+E to select the same file from the file list, it crashes in a different way:
```
E:\>fresh
Error: Editor loop returned an error

Caused by:
[...]

PR #937, opened 2026-02-03
Handle large file encoding confirmations gracefully

sinelaw:
## Summary
- Show encoding confirmation prompt in TUI instead of crashing when opening large files with non-resynchronizable encodings (GBK, GB18030, Shift-JIS, EUC-KR)
- Unify CLI and interactive file opening code paths for consistent error handling

## Test plan
- [ ] Open a large GBK-encoded file from CLI
- [ ] Open a large file from file explorer
- [ ] Verify encoding prompt appears instead of crash


sinelaw:
Fixes #488

PR #996, opened 2026-02-09
fix: replace rust-i18n macro with runtime backend to prevent OOM

sinelaw:
Replace compile-time i18n macro expansion with runtime JSON loading to resolve OOM crashes during release builds. The macro was generating 580,000+ IR instructions for 1.1MB of translations, causing >16GB memory usage and FreeBSD CI failures.

Solution: Empty locales directory + custom Backend implementation that lazily loads embedded JSON files at runtime using include_str!(). This reduces compile-time IR to <1,000 instructions while maintaining the same binary size and all functionality.

- Created RuntimeBackend implementing rust_i18n::Backend trait
- Embedded all 14 locale JSON files with include_str!()
- Lazy loading with once_cell for on-demand parsing
- All existing tests pass without modification
- No changes needed to 326 files using t!() macro


sinelaw:
Improves cargo build of the 'fresh-editor' crate:

- Build time: 196 seconds => 106 seconds
- Peak RSS: 3.3 GB => 2.3 GB

Issue #1007, opened 2026-02-10
Find and replace doesn't work when RegEx is on


X-Ryl669:
Example text file:
```
AAAAAA
Sig: eifojoi oijrze oizerj g
BBBBBB
Sig: dozijfzoeijz ozeij zegoij
CCCCCC
```

Try to find and replace `Sig:.*` by (nothing) without any confirmation, it'll fail to do so. Yet, finding the same text does work successfully.

sinelaw:
Fixed in master

PR #1008, opened 2026-02-10
Add regex capture group support to find-and-replace

sinelaw:
## Summary
This PR adds support for regex capture groups in the find-and-replace functionality, allowing users to reference captured groups (e.g., `$1`, `$2`, `${name}`) in replacement text. This enables powerful transformations like swapping fields or reformatting matched text.

## Key Changes

- **New `regex_replace` module**: Extracted pure, buffer-agnostic regex helpers into a dedicated module with comprehensive test coverage:
  - `build_regex()`: Constructs compiled regexes with support for whole-word and case-sensitivity options
  - `collect_regex_matches()`: Finds all matches and expands capture group references in replacement templates
  - `expand_replacement()`: Expands capture groups for individual matches

- **Enhanced `InteractiveReplaceState`**: Added fields to track:
  - `current_match_len`: Stores actual match length (differs from search string length in regex mode)
  - `regex`: Caches compiled regex for efficient reuse during interactive replace

- **Updated replace-all logic**: 
  - Detects regex mode and uses compiled regex for matching
  - Expands capture group references for each match before applying replacements
  - Maintains O(n) performance using BulkEdit for tree operations

- **Updated interactive replace**:
  - Finds matches using regex when enabled
  - Properly tracks variable-length matches
  - Expands capture groups when replacing individual matches or all remaining matches

- **UI enhancement**: Added hint text in status bar when regex mode is enabled in replace mode, informing users they can use `$1`, `$2`, etc. in replacements

## Implementation Details

- All regex operations use `regex::bytes::Regex` for efficient byte-level matching
- Capture group expansion is delegated to the `regex` crate's built-in `expand()` method
- The implementation gracefully falls back to literal string replacement when regex mode is disabled
- Match lengths are determined dynamically to support variable-length regex patterns

https://claude.ai/code/session_01EZKgNP9Ffb6VC8ZENRuzRx

### DIFF
--- a/crates/fresh-editor/src/view/ui/status_bar.rs
+++ b/crates/fresh-editor/src/view/ui/status_bar.rs
@@ -1160,6 +1160,13 @@ impl StatusBarRenderer {
         current_col += regex_full_width as u16;
         layout.regex = Some((regex_start, current_col));
 
+        // Show capture group hint when regex is enabled in replace mode
+        if use_regex && confirm_each.is_some() {
+            let hint = " \u{2502} $1,$2,…";
+            spans.push(Span::styled(hint, shortcut_style));
+            current_col += str_width(hint) as u16;
+        }
+
         // Confirm Each option (only shown in replace mode)
         if let Some(confirm_value) = confirm_each {
             let confirm_shortcut =


sinelaw:
Fixes #1007

PR #1042, opened 2026-02-16
impl Feature request: Code folding #900

### LINKED ISSUE #900: Feature request: Code folding
hola.

i tried looking for a way to fold blocks of code (in my case Javascript code) in fresh, but couldn't find any way to do it.
i assume it's not yet implemented?

if so, consider this my official feature request for code folding to be implemented and added to fresh.
it's extremely helpful for the readability and developer-ergonomics of bigger code files/projects.

P.S. if it gets implemented, i would absolutely LOVE if in addition to the _basic_ "collapse"/"expand" functionality fresh also supported:
- **nested fold regions**, as-in being able to:
    1. fold some block of code
    2. fold the outer block that contains the already folded section of code
    3. so that unfolding the outer section would not unfold the inner section
- **remembering** the folded regions, on e.g. editor restart or file close/open or project switch
- quick-folding and -unfolding the active file based on indentation depth (e.g. pressing CTRL+K and then CTRL+1 would fold all code in the current file at the first tab-stop, CTRL+K followed by CTRL+2 would fold all code at the second tab-stop, etc. and same thing for unfolding)

those additional features help save **_so_** much time when working on big projects that require folding a lot of code often.

asukaminato0721:
impl foldingRange support

Added a folding UI layer: collapsed ranges are marker-tracked, folded lines are hidden, and a placeholder is appended on the fold header line.

Added gutter fold indicators (▾ expanded, ▸ collapsed) and gutter click toggling.

Added the toggle_fold action + command palette entry, with localized strings across all locales.

test_folding_hides_lines_and_adds_placeholder: verifies folded lines are hidden and placeholder is appended on the header line.

test_fold_indicators_collapsed_and_expanded: verifies gutter indicator state for collapsed vs expanded fold ranges.

fix #900

### DIFF
--- a/crates/fresh-editor/src/view/ui/split_rendering.rs
+++ b/crates/fresh-editor/src/view/ui/split_rendering.rs
@@ -341,6 +341,8 @@ struct DecorationContext {
     line_indicators: BTreeMap<usize, crate::view::margin::LineIndicator>,
     /// Fold indicators indexed by line number
     fold_indicators: BTreeMap<usize, FoldIndicator>,
+    /// Collapsed fold header -> hidden line count
+    collapsed_header_hidden_lines: BTreeMap<usize, usize>,
 }
 
 #[derive(Clone, Copy, Debug)]
@@ -3884,6 +3886,12 @@ impl SplitRenderer {
 
         let fold_indicators =
             Self::fold_indicators_for_viewport(state, folds, viewport_start, viewport_end);
+        let collapsed_header_hidden_lines = Self::collapsed_header_hidden_lines_for_viewport(
+            state,
+            folds,
+            viewport_start,
+            viewport_end,
+        );
 
         DecorationContext {
             highlight_spans,
@@ -3893,6 +3901,7 @@ impl SplitRenderer {
             diagnostic_lines,
             line_indicators,
             fold_indicators,
+            collapsed_header_hidden_lines,
         }
     }
 
@@ -3935,6 +3944,36 @@ impl SplitRenderer {
         }
 
         indicators
+    }
+
+    fn collapsed_header_hidden_lines_for_viewport(
+        state: &EditorState,
+        folds: &FoldManager,
+        viewport_start: usize,
+        viewport_end: usize,
+    ) -> BTreeMap<usize, usize> {
+        let mut map = BTreeMap::new();
+        if folds.is_empty() {
+            return map;
+        }
+
+        let viewport_start_line = state.buffer.get_line_number(viewport_start);
+        let viewport_end_line = state.buffer.get_line_number(viewport_end);
+
+        for range in folds.resolved_ranges(&state.buffer, &state.marker_list) {
+            if range.header_line < viewport_start_line || range.header_line > viewport_end_line {
+                continue;
+            }
+            let hidden = range
+                .end_line
+                .saturating_sub(range.start_line)
+                .saturating_add(1);
+            if hidden > 0 {
+                map.insert(range.header_line, hidden);
+            }
+        }
+
+        map
     }
 
     // semantic token colors are mapped when overlays are created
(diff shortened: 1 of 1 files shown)


sinelaw:
1. when clicking rapidly i think it's interpreted as double or triple click and not handled - folding/unfolding doesn't work in that case
2. if cursor is currently placed inside the folded block (second line or more) the folding doesn't work - clicking does nothing, need to move the cursor manually out of the block
3. mouse wheel and moving down the file with cursor by pressing down key starting before the folded block - also produces weird results, see attached video
4. syntax highlighting seems sometimes broken by the folding - this could be a general syntax highlighting bug with incremental view, made worse by folding

All of the issues should have e2e tests that reproduce them before fixing so we ensure the fix is robust


https://github.com/user-attachments/assets/9be48c1a-15e0-4ac0-80c3-9adeff177969

sinelaw:
Can you please squash to a handful of commits (separate if they are doing unrelated changes) with meaningful commit messages?

sinelaw:
I just created a more detailed contribution guide - please read quickly (it's short), you can feed it into an LLM to apply on your branch: https://github.com/sinelaw/fresh/blob/master/CONTRIBUTING.md
Also please rebase on latest master because there are conflicts

sinelaw:
There's still a bug with scrolling when code is folded:

https://github.com/user-attachments/assets/56bdba1c-2480-4487-bd11-6d5eb36ec7d0

sinelaw:
It's looking better! Almost there.
1. Remaining bug I noticed: after folding a block, moving down causes a sudden jump and then moving back up - shows that the folding has been undone (not folded anymore). This doesn't happen when scrolling with mouse wheel
Can you add an e2e test that fails on this (reproduces the bug) and then fix it?

https://github.com/user-attachments/assets/92609a10-a6d2-4da2-955d-937748bbe3de

2. Remove this file from the branch (amend the commit that added it) - 551772438-9be48c1a-15e0-4ac0-80c3-9adeff177969.mp4

3. Ctrl+click selection extension silently removed

  The old handle_editor_click code had:
  // Both modifiers supported since some terminals intercept shift+click
  let extend_selection = modifiers.contains(KeyModifiers::SHIFT)
      || modifiers.contains(KeyModifiers::CONTROL);

4. Side-effectful fold removal in the render/layout path

  compute_buffer_layout (split_rendering.rs:4886) mutates folds during rendering:
  if !folds.is_empty() {
      let cursor_positions: Vec<usize> = cursors.iter().map(|(_, c)| c.position).collect();
      for pos in cursor_positions {
          let _ = folds.remove_if_contains_byte(&mut state.marker_list, pos);
      }
  }

  This means rendering is not idempotent -- re-rendering after cursor movement silently expands folds. This should be handled in the cursor movement / input handling layer, not during rendering. The render path should be read-only with respect to fold state. As-is,  the fold auto-expand only triggers when a render happens, which creates a timing dependency.

5. Line numbers in large files are broken now. I'm guessing it's related to this change - from incremental line counting (prev_was_source_line + increment) to source-byte lookup (get_line_number(first_byte))
  // Before: if show_line_number && prev_was_source_line { current_source_line_num += 1; }
  // After:  current_source_line_num = state.buffer.get_line_number(first_byte);
 
We should avoid using get_line_number - can we change it to be more incremental (lazy walking over lines until we fill up the view, skip rendering the lines that belong under the fold) and keep the +1 ? I'm not sure.

asukaminato0721:
some changes are from clippy --fix

sinelaw:
Please add this commit (cherry pick) to your branch, and see that the new test there passes: https://github.com/sinelaw/fresh/commit/d90a9a6c264638315919f051fa02f7caea1868d2

sinelaw:
Thanks! There's a few more bugs but I'll fix them in a new PR

Issue #569, opened 2025-12-30
Save options Settings [BUG ? FEATURE ?]


mrtngrsbch:
Hi guy,

First of all, as a long-time user of Norton Commander > Midnight Commander, I would like to express my enormous gratitude for this excellent work. 

But I wonder if it is a bug or a feature...
The selection of "File Explorer", "Show hidden files" and "Show gitignored files" is not saved in the preferences.

<img width="330" height="200" alt="Image" src="https://github.com/user-attachments/assets/2dd0cb26-f84f-4f0c-9231-873431712a0a" />

<img width="343" height="268" alt="Image" src="https://github.com/user-attachments/assets/fb196a27-1b61-47e6-8597-eddae725c1b2" />

Macos 15.6.1 (24G90) in MacbookPro M1

best

sinelaw:
Thanks for reporting. I'm working on revamping the concepts of config persistence, it will take some work to get it right. It's not obvious in which flows it's better to leave settings transient vs persist them

jamie48ford-ux:
Hi

PR #1049, opened 2026-02-17
Persist file explorer toggle settings across sessions

sinelaw:
## Summary
This PR adds persistence for file explorer toggle settings (`show_hidden` and `show_gitignored`) so that user preferences are saved and restored across editor sessions.

## Key Changes
- **Refactored toggle methods**: Restructured `file_explorer_toggle_hidden()` and `file_explorer_toggle_gitignored()` to extract the toggle state and handle early returns more cleanly
- **Added config persistence**: Both toggle methods now call `persist_config_change()` to save the setting to the user config file immediately after toggling
- **New helper method**: Introduced `persist_config_change()` in `toggle_actions.rs` to handle persisting individual config changes using the `ConfigResolver`, matching the behavior of the settings UI

## Implementation Details
- The toggle state is captured in a variable before the early return, allowing the persistence logic to be executed outside the conditional block
- Config changes are persisted using JSON pointer paths (`/file_explorer/show_hidden` and `/file_explorer/show_gitignored`)
- Errors during persistence are logged but don't interrupt the user experience
- The implementation follows the existing pattern used for other settings persistence in the codebase

https://claude.ai/code/session_017wsEscsCSN4ScxvZ2hQJn3

### DIFF
--- a/crates/fresh-editor/src/app/file_explorer.rs
+++ b/crates/fresh-editor/src/app/file_explorer.rs
@@ -748,28 +748,49 @@ impl Editor {
     }
 
     pub fn file_explorer_toggle_hidden(&mut self) {
-        if let Some(explorer) = &mut self.file_explorer {
+        let show_hidden = if let Some(explorer) = &mut self.file_explorer {
             explorer.toggle_show_hidden();
-            let msg = if explorer.ignore_patterns().show_hidden() {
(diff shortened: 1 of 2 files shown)


sinelaw:
Fixes https://github.com/sinelaw/fresh/issues/569

PR #1082, opened 2026-02-21
Added example of how to bind a key to play a macro

mandolyte:
After looking thru the codebase, I figured out how to bind keys to play macros. So I thought I'd a bit to your documentation. This may not be the right place, but feel free to use however you choose.

Thanks for developing such an awesome editor!


sinelaw:
Thanks!

PR #1084, opened 2026-02-21
Fold bugfixes

sinelaw:


sinelaw:
@asukaminato0721 FYI (a few more fixes)

PR #1097, opened 2026-02-23
Fix standalone ESC key handling and add Shift+Tab support

sinelaw:
## Summary
This PR fixes the input parser to correctly handle standalone ESC keypresses that arrive before other escape sequences, and adds support for the Shift+Tab key (CSI Z sequence). The main issue was that a buffered ESC byte would wait indefinitely for a follow-up byte, preventing the ESC key from being emitted promptly.

## Key Changes

- **ESC timeout handling**: Added `last_byte_time` tracking and a new `flush_timeout()` method to emit buffered ESC bytes after 50ms of inactivity. This ensures standalone ESC keypresses are emitted promptly instead of waiting indefinitely.

- **ESC-ESC disambiguation**: When an ESC byte is received while another ESC is buffered, the buffered ESC is now treated as a standalone Escape key and the new ESC starts a fresh escape sequence. This handles cases where ESC is pressed followed immediately by another escape sequence.

- **Shift+Tab support**: Added parsing for the CSI Z sequence (`ESC [ Z`), which is the standard terminal representation of Shift+Tab, mapping it to `KeyCode::BackTab` with `SHIFT` modifier.

- **Server integration**: Modified `EditorServer` to call `flush_timeout()` during the `WouldBlock` error case (when no input is available), ensuring timeout-based ESC emission happens regularly during the server's event loop.

- **Comprehensive test coverage**: Added 6 regression tests covering:
  - Shift+Tab parsing
  - ESC followed by mouse events (same and separate chunks)
  - ESC followed by arrow keys
  - Standalone ESC timeout flushing
  - Timeout behavior with recent input

## Implementation Details

The parser now maintains an `Instant` timestamp (`last_byte_time`) whenever a byte is added to the buffer. When `flush_timeout()` is called and the buffer has been waiting longer than `ESC_TIMEOUT_MS` (50ms), any buffered bytes are converted to events individually. This approach handles the ambiguity of ESC bytes that could be either standalone keypresses or the start of escape sequences.

https://claude.ai/code/session_01Ag1vzRrnNitUatw6NeR8C5

### DIFF
--- a/crates/fresh-editor/src/server/input_parser.rs
+++ b/crates/fresh-editor/src/server/input_parser.rs
@@ -708,7 +708,12 @@ mod tests {
         // User presses Escape, then moves mouse. Both arrive in one chunk:
         // ESC (0x1b) followed by mouse event ESC [ < 35 ; 67 ; 18 M
         let events = parser.parse(b"\x1b\x1b[<35;67;18M");
-        assert_eq!(events.len(), 2, "Expected Escape + mouse event, got: {:?}", events);
+        assert_eq!(
+            events.len(),
+            2,
+            "Expected Escape + mouse event, got: {:?}",
+            events
+        );
 
         // First event: standalone Escape
         match &events[0] {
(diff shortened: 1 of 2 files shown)


sinelaw:
Fixed in master https://github.com/sinelaw/fresh/pull/1097

PR #1149, opened 2026-02-28
fix support path:line:column in command palette / open file #1081

### LINKED ISSUE #1081: support path:line:column
for example

pyrefly/lib/lsp/wasm/signature_help.rs this works

but pyrefly/lib/lsp/wasm/signature_help.rs:319:48 not work.

asukaminato0721:
fix #1081

Made the path:line:col parser crate-visible so the file browser can reuse it safely.

Added an e2e test that opens jump.txt:3:5 via the Open File dialog and asserts the cursor lands at Ln 3, Col 5.

### DIFF
--- a/crates/fresh-editor/src/app/file_open_input.rs
+++ b/crates/fresh-editor/src/app/file_open_input.rs
@@ -4,6 +4,7 @@
 //! browser popup when the Open File or Switch Project prompt is active.
 
 use super::file_open::{FileOpenSection, SortMode};
+use super::prompt_actions::parse_path_line_col;
 use super::Editor;
 use crate::input::keybindings::Action;
 use crate::primitives::path_utils::expand_tilde;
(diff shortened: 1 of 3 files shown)


sinelaw:
Please rebase on latest master and 

  1. Deduplicate the file-open-and-jump logic in handle_quick_open_file (prompt_actions.rs). The block at ~line 1313-1358 (the if line.is_some() &&           
  !path_from_input.is_empty() fallback) is nearly identical to the code in the selected_index branch above it. Extract a shared helper.
  2. Move parse_path_line_col off Editor. It's a pure string-parsing function with no self usage — make it a free function (or put it in a utility module)    
  rather than an impl Editor method.

sinelaw:
Also: 
3. Add e2e test for Quick Open with path:line:col — this is a separate user flow from the file browser and needs its own e2e coverage.                      
4. Add e2e test for path:line (line only, no column) — the parser unit test covers it, but the actual jump-to-line behavior through the full stack is untested. 

Thanks

asukaminato0721:
fixed

PR #1214, opened 2026-03-06
Fix typo in Spanish localization prompts

brunodsf05:
Fixed a typo in spanish translation.

The typo was "cancelarar" and was corrected to **cancelar**. (which means cancel btw)

The image below shows where I found the typo.
imagen


sinelaw:
Thanks!

PR #1259, opened 2026-03-13
flatpak: fix AppStream/metainfo and desktop file for stores and menu

Jaredy899:
- Use desktop-application type and add launchable in metainfo for AppStream
- Add screenshots, release URL, and GPL-2.0-only in metainfo
- Use single main category (Utility;TextEditor) in desktop file per menu [spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/)


sinelaw:
Thanks!

PR #1298, opened 2026-03-20
Add global editor.use_tabs setting with language-specific overrides

sinelaw:
## Summary
This change introduces a global `editor.use_tabs` configuration setting that controls whether the Tab key inserts tab characters or spaces by default. Language-specific configurations can now optionally override this global setting, with `null` values falling back to the global default.

## Key Changes

- **Added global `editor.use_tabs` setting** to `EditorConfig` with a default value of `false` (insert spaces)
  - Documented in config schema and example configuration
  - Properly integrated into the default editor configuration

- **Changed `LanguageConfig.use_tabs` from `bool` to `Option<bool>`**
  - `None` (null) means the language inherits the global `editor.use_tabs` setting
  - `Some(true)` or `Some(false)` explicitly overrides the global setting
  - Updated all language definitions to use `None` by default, except Go and Makefile which explicitly require tabs with `Some(true)`

- **Updated configuration resolution logic** in multiple places:
  - `BufferConfig::resolve()` now uses the global setting as the base and applies language overrides only when explicitly set
  - `toggle_actions.rs`, `buffer_management.rs`, and `settings_actions.rs` updated to properly fall back to global setting
  - `PartialLanguageConfig` conversion updated to preserve `Option<bool>` semantics

- **Added comprehensive test coverage** for the new inheritance behavior:
  - Test verifies that unknown languages inherit the global setting
  - Test confirms explicit language overrides work correctly
  - Test validates that `None` values properly inherit from global

- **Updated all test fixtures and language configurations** to use the new `Option<bool>` type consistently throughout the codebase

## Implementation Details

The change maintains backward compatibility by defaulting to spaces (`use_tabs: false`) globally, while allowing individual languages like Go and Makefile to explicitly require tabs. The optional nature of language-specific settings means existing configurations without explicit `use_tabs` values will automatically inherit the global setting, providing a cleaner configuration experience.

https://claude.ai/code/session_01Lj9iVU5ZP2gZWhi4HAJCQF


sinelaw:
image

Issue #1305, opened 2026-03-21
Find next Previous from menu cursor postion loss


pstahle:
Hi,

It seems like if you select find next or find previous from the edit menu the current curos position is moved and the find will not go to the actual next occurence.

Steps to reproduce:
- do a find for a phrase that is many in the editor places in the editor (phrase is close to end end of file)
- put cursor at end of file 
- select find previous from edit menu

It seems to work fine with F3 & Alt-F3, except I did observe some weird issues with a very large log file that I need to do a bit more testing to fully understand before reporting...

Regards,
Patrick

PR #1309, opened 2026-03-21
Fix find_next/find_previous to navigate from cursor position

sinelaw:
Changes made:

    render.rs - Core fix + refactoring:
        Added SearchDirection enum (Forward/Backward)
        Extracted duplicated logic from find_next() and find_previous() into a single find_match_in_direction() helper
        Instead of using search_state.current_match_index (a stored index that doesn't track cursor movement), the new code reads the current cursor position and uses binary search to find the nearest match in the given direction
        find_next() and find_previous() are now one-liner delegates

    search.rs - Updated test_f3_continues_searching_after_buffer_modification to expect correct cursor-based behavior (cursor at position 4 → next match at 8, not 12)

    search_navigation_after_move.rs - Fixed large file tests to reliably generate >3MB files

### DIFF
--- a/crates/fresh-editor/src/app/render.rs
+++ b/crates/fresh-editor/src/app/render.rs
@@ -2,6 +2,11 @@ use super::*;
 use anyhow::Result as AnyhowResult;
 use rust_i18n::t;
 
+enum SearchDirection {
+    Forward,
+    Backward,
+}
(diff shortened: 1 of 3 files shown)


sinelaw:
Fixes #1305

Issue #1319, opened 2026-03-22
Syntax Highlighting issue:Editor treats '{' '}' ';' as operators IN C OR CPP SOURCE FILE


OneByTheSide:
platform:windows
versison:0.2.17

<img width="1622" height="466" alt="Image" src="https://github.com/user-attachments/assets/903253c5-b44a-40d7-904d-5aee6995654d" />

In cpp file:

<img width="1190" height="840" alt="Image" src="https://github.com/user-attachments/assets/abb85ded-40ef-4db7-80bd-bf8948de3bfb" />

OneByTheSide:
My font default foreground Is NOT RED

sinelaw:
Fixed in master branch

PR #1320, opened 2026-03-22
Distinguish keyword operators from generic punctuation in syntax highlighting

sinelaw:
## Summary
This change refines the syntax highlighting logic to differentiate between keyword operators (like `+`, `-`, `==`) and generic punctuation (like `{`, `}`, `;`). Generic punctuation now uses the default foreground color instead of being highlighted as operators.

## Key Changes
- **highlight_engine.rs**: Modified `scope_to_category()` to only treat `keyword.operator` scopes as operators, removing the blanket classification of all `punctuation` scopes
- **textmate_engine.rs**: Applied the same logic change to maintain consistency across both highlighting engines
- **Tests**: Updated `test_other_punctuation_still_operator()` to `test_generic_punctuation_is_not_operator()` with new assertions verifying that generic punctuation scopes (`punctuation.separator`, `punctuation.section`, `punctuation.terminator`) now return `None` instead of `Some(HighlightCategory::Operator)`

## Implementation Details
The change distinguishes between:
- **Keyword operators** (e.g., `keyword.operator`): Highlighted with operator color
- **Generic punctuation** (e.g., `punctuation.separator`, `punctuation.section`, `punctuation.terminator`): Use default foreground color (None)

This provides more nuanced syntax highlighting where only semantic operators are emphasized, while structural punctuation blends with regular text.

https://claude.ai/code/session_019nmKTouQFirivMoXFPMtf6

### DIFF
--- a/crates/fresh-editor/plugins/theme_editor.i18n.json
+++ b/crates/fresh-editor/plugins/theme_editor.i18n.json
@@ -296,7 +296,11 @@
     "field.popup_selection_fg": "vyskakovací okno výběr popředí",
     "field.popup_selection_fg_desc": "vyskakovací okno selected item text barva",
     "field.whitespace_indicator_fg": "Bílé znaky Indikátor popředí",
-    "field.whitespace_indicator_fg_desc": "Barva popředí indikátorů bílých znaků (šipky tabulátorů a tečky mezer)"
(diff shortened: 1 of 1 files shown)


sinelaw:
Fixes #1319

PR #1331, opened 2026-03-23
Make cargo publish workflow re-runnable by skipping already-published…

sinelaw:
… crates

Publish crates individually in dependency order instead of all at once. Skip crates that already exist on crates.io, so the workflow can be safely re-run after partial failures. Only fresh-editor uses --allow-dirty (needed because the sed step modifies its Cargo.toml to strip fresh-gui).

https://claude.ai/code/session_018D1aGVuT8V8MfTdtzn4M12

### DIFF
--- a/.github/workflows/cargo-publish.yml
+++ b/.github/workflows/cargo-publish.yml
@@ -27,4 +27,33 @@ jobs:
       - name: Publish to crates.io
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
-        run: cargo publish --allow-dirty
+        run: |
+          # Publish crates individually in dependency order, skipping
+          # any that are already published (makes the workflow re-runnable).
+          crates=(
+            fresh-core
+            fresh-languages
+            fresh-parser-js
+            fresh-plugin-api-macros
+            fresh-winterm
+            fresh-plugin-runtime
+            fresh-editor
+          )
+          for crate in "${crates[@]}"; do
+            echo "::group::Publishing $crate"
+            flags=""
+            if [ "$crate" = "fresh-editor" ]; then
+              flags="--allow-dirty"
+            fi
+            if cargo publish -p "$crate" $flags 2>&1 | tee /tmp/publish.log; then
+              echo "Published $crate"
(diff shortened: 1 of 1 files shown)


.github/workflows/cargo-publish.yml
```diff
@@ -30,4 +30,33 @@ jobs:
       - name: Publish to crates.io
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
-        run: cargo publish --allow-dirty
+        run: |
+          # Publish crates individually in dependency order, skipping
+          # any that are already published (makes the workflow re-runnable).
+          crates=(
```
sinelaw, line 36: Crate list is incomplete. The workspace has 12 publishable crates; this lists 7.

`fresh-editor` path-depends on `fresh-editor-core`, `fresh-i18n`, `fresh-input-parser`, `fresh-ui` and `fresh-update`. None are listed, so `cargo publish -p fresh-editor` fails: those deps are not on crates.io.

Derive the list from `cargo metadata` instead of hardcoding it, otherwise it goes stale again.

.github/workflows/cargo-publish.yml
```diff
@@ -30,4 +30,33 @@ jobs:
...
+            fresh-plugin-runtime
+            fresh-editor
+          )
+          for crate in "${crates[@]}"; do
+            echo "::group::Publishing $crate"
+            flags=""
+            if [ "$crate" = "fresh-editor" ]; then
+              flags="--allow-dirty"
+            fi
+            if cargo publish -p "$crate" $flags 2>&1 | tee /tmp/publish.log; then
```
sinelaw, line 51: The pipeline exit status is `tee`'s, not cargo's.

This step has no `shell:` key and the workflow sets no `defaults`, so it runs under `bash -e {0}` with pipefail off. That means this branch always succeeds: a failed publish prints "Published $crate" and the loop moves on, and the `elif` on line 53 is unreachable.

Net effect is the opposite of the PR's goal — failures become silent rather than reruns becoming safe.

Needs `set -o pipefail` (or `shell: bash`) plus a check on `${PIPESTATUS[0]}`.

sinelaw (commented):

Two blocking issues, either one fatal. Details inline.

Issue #1334, opened 2026-03-24
Word selection


dillonbheadley:
Text highlighting is definitely better, but there’s one more small issue. When double clicking and then dragging to select words forward everything works perfectly. But dragging backwards unselects the word you started on before selecting the previous words

sinelaw:
Fixed! thanks for reporting

PR #1349, opened 2026-03-24
Fix backward drag selection after double-click to keep initial word

sinelaw:
## Summary
Fixes issue #1334 where dragging backward after double-clicking a word would deselect the initially selected word before extending the selection to previous words. Now the initial word remains selected throughout the backward drag operation.

## Key Changes
- **MouseState struct** (`types.rs`): Added `drag_selection_word_end` field to track the end position of the initially double-clicked word, enabling proper anchor selection during backward drags
- **Mouse input handling** (`mouse_input.rs`): 
  - Store both the start and end positions of the double-clicked word when initiating word-by-word selection
  - Modified drag logic to use different anchors based on drag direction:
    - Forward drag: anchor at word start (existing behavior)
    - Backward drag: anchor at word end to preserve the initial word in the selection
  - Clear the `drag_selection_word_end` state on mouse release
- **E2E test** (`mouse.rs`): Added comprehensive test case `test_double_click_drag_backward_keeps_initial_word_selected` that verifies the fix by:
  - Double-clicking "brown" in "quick brown fox"
  - Dragging backward to "quick"
  - Asserting that the final selection is "quick brown" (not just "quick")

## Implementation Details
The fix leverages the fact that when dragging backward (target_position < anchor_position), we now use the word end as the anchor point instead of the word start. This ensures the initially selected word stays within the selection bounds as the user extends leftward to include previous words.

https://claude.ai/code/session_01DvpM5HQdoxiHRe3A6D12oL

### DIFF
--- a/crates/fresh-editor/src/app/mouse_input.rs
+++ b/crates/fresh-editor/src/app/mouse_input.rs
@@ -215,6 +215,7 @@ impl Editor {
                 self.mouse_state.drag_selection_split = None;
                 self.mouse_state.drag_selection_anchor = None;
                 self.mouse_state.drag_selection_by_words = false;
+                self.mouse_state.drag_selection_word_end = None;
                 // Clear popup scrollbar drag state
                 self.mouse_state.dragging_popup_scrollbar = None;
                 self.mouse_state.drag_start_popup_scroll = None;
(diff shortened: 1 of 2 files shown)


sinelaw:
Fixes #1334

Issue #1438, opened 2026-03-31
Issues with language indicator when adding a custom language


sinelaw:
Original user bug report:

>  i've added this to my config file, and lsp works perfectly fine, fish appears both in language and lsp sections  and commenting started to work too, but fresh still detects fish files as bash, and i dont know if i did config wrong or not 
```
  "languages": {
    "fish": {
      "extensions": ["fish"],
      "grammar": "fish",
      "comment_prefix": "#",
      "auto_indent": true
    }
  },
  "lsp": {
      "fish": [
          {
            "command": "fish-lsp",
            "enabled": true,
            "args": [
              "start"
            ],
            "auto_start": false,
            "process_limits": {
              "max_memory_percent": 50,
              "max_cpu_percent": 90,
              "enabled": true
            }
          }
        ]
    }
```

Partial fix (not complete): https://github.com/sinelaw/fresh/pull/1435

Still need to adjust the popup in the status bar indicator to correctly handle the case of undefined grammar.
Also still need an easy way to get a list of available grammars.

PR #1439, opened 2026-03-31
Support user-configured languages without syntect grammars

sinelaw:
## Summary
This PR fixes an issue where user-configured languages without matching syntect grammars (e.g., "fish" with grammar "fish") were not appearing in the language selector and could not be selected. The changes ensure these languages are discoverable and selectable, even without syntax highlighting support.

## Key Changes

- **Language Selector Population** (`input.rs`): Modified the "Set Language" command to include user-configured languages that don't have a matching syntect grammar. These are now added to the language selector alongside built-in syntaxes.

- **Language Resolution** (`input.rs`): Updated the current language detection logic to handle config-only languages by checking if the syntax name directly matches the current language ID, not just through the `resolve_language_id` mapping.

- **Language Application** (`prompt_actions.rs`): Added handling for user-configured languages in the language selection flow. When a config-only language is selected, it creates a `DetectedLanguage` with no syntax highlighting but preserves the correct language ID for LSP and configuration purposes.

- **DetectedLanguage Factory** (`detected_language.rs`): Added `from_config_language()` constructor to create language instances for user-configured languages without syntect support. These use `HighlightEngine::None` but maintain the proper language identifier.

- **Test Coverage** (`syntax_language_case.rs`): Added comprehensive test `test_config_only_language_appears_in_selector()` that verifies:
  - Config-only languages are correctly detected when opening files
  - They appear in and can be selected from the language selector
  - Language can be switched away from and back to the config-only language

## Implementation Details

The fix works by treating user-configured languages as first-class citizens in the language selector, separate from the syntect grammar registry. When a language is configured but has no matching grammar in syntect, it's added to the selector with a `DetectedLanguage` that provides the correct language ID without syntax highlighting capabilities.

https://claude.ai/code/session_01AcpxWRdWwiBYZJXNKLQryZ

### DIFF
--- a/crates/fresh-editor/src/app/input.rs
+++ b/crates/fresh-editor/src/app/input.rs
@@ -3021,8 +3021,9 @@ impl Editor {
         let mut syntax_to_config: std::collections::HashMap<String, (String, &str)> =
             std::collections::HashMap::new();
         for (lang_id, lang_config) in &self.config.languages {
-            if let Some(syntax) =
-                self.grammar_registry.find_syntax_for_lang_config(lang_config)
+            if let Some(syntax) = self
+                .grammar_registry
+                .find_syntax_for_lang_config(lang_config)
             {
                 syntax_to_config
                     .entry(syntax.name.clone())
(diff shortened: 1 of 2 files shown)


sinelaw:
Improves https://github.com/sinelaw/fresh/issues/1438

PR #1456, opened 2026-04-02
Fix install error handling

maxugly:
## Problem
The AppImage installer currently fails silently if `mktemp` or `curl` fails, reporting a "Success" message even if no files were moved. Additionally, it fails on systems with restricted `/tmp` permissions (such as **Void Linux** or certain containers).

## Solution
This PR introduces several robustness improvements to ensure a clean and honest installation process:
- [x] **Atomic Cleanup:** Added a `trap` to ensure the temporary workspace is deleted on exit or interruption.
- [x] **Environment Fallback:** Implemented a fallback to `$HOME/.cache` if the system `/tmp` is not writable.
- [x] **Error Checking:** Added exit-code verification for `curl`, `mktemp`, and `mv` to prevent "fake success" messages.
- [x] **Validation:** The script now verifies the existence of the extracted binary before moving it to the final destination.

## Testing
Verified on **Void Linux x86_64**. The script now correctly falls back to local storage when `/tmp` is restricted and successfully completes the installation.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -113,7 +113,7 @@ install_arch() {
 get_release_url() {
     # $1 = pattern (e.g., .deb or .rpm)
     # $2 = arch
-    url=$(curl -s "https://api.github.com/repos/${REPO_OWNER}/${REPO_NAME}/releases/latest" | grep "browser_download_url.*$2.*$1" | cut -d '"' -f 4 | head -n 1)
+    url=$(curl -s "https://api.github.com/repos/${REPO_OWNER}/${REPO_NAME}/releases/latest" | grep "browser_download_url.*$2.*${1%$}" | cut -d '"' -f 4 | head -n 1)
     echo "$url"
 }
 
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -169,19 +169,16 @@ install_fedora() {
 
 do_install_appimage() {
     log_info "Attempting AppImage install..."
-    if ! check_cmd curl; then log_error "curl is required."; fi
+    if ! check_cmd curl; then log_error "curl is required."; return 1; fi
 
     ARCH=$(uname -m)
-    # Map architecture to AppImage naming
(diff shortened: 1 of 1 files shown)


maxugly:
Subject: Context for the regex adjustment

Following up on this one: this patch allows for casing variations (e.g., .AppImage vs .appimage) and version-string formatting that were causing an install failure on my Void Linux setup.

This works alongside the fix you just merged in #1455 to get the installer running here. Thanks for merging that one! Happy to refine the pattern if you have a preferred regex style.

sinelaw (commented):
A few comments, please take a look

sinelaw (approved):

- **`${1%$}` regex fix validated**: manually confirmed the old grep pattern (`\.AppImage$`) fails to match GitHub API JSON responses (line ends with `"` not `.AppImage`); stripping the trailing `$` via `${1%$}` correctly matches the URL — tested in tmux with a simulated JSON line
- **`/tmp` fallback validated**: forced `TMPDIR=/nonexistent`; `mktemp -d 2>/dev/null || echo "$HOME/.cache/fresh-work-$(date +%s)"` + `mkdir -p` correctly creates a writable workspace under `~/.cache` — confirmed in tmux
- **`trap EXIT` cleanup validated**: temp dir was removed after subshell exit, confirmed by `ls` returning "No such file or directory"
- **`mv squashfs-root INSTALL_DIR` behavior validated**: with INSTALL_DIR pre-deleted by `rm -rf`, the mv renames the dir; binary is accessible at `$INSTALL_DIR/usr/bin/fresh` as expected by the symlink — tested in tmux
- **Error handling on curl failure validated**: `if ! curl ... ; then log_error ...; return 1` correctly exits early — tested against unreachable host
- Code is simple, non-duplicative, and consistent with the existing script style; no architectural concerns

---

sinelaw:
Thanks! apologies for the long delay

Issue #1454, opened 2026-04-02
Fresh OOM


AXIOMZK:
When writing code files, I occasionally encounter OOM (Out of Memory) errors that suddenly spike to using dozens of GB of memory, causing the program to freeze. The operating system is macOS 26.3 with 48GB of RAM, and I'm using Fresh v0.2.21.

sinelaw:
What types of files are you opening?
Are you using the session mode?

sinelaw:
I'm not able to reproduce this yet. Can anyone suggest steps to reproduce?

ricardo-bohorquez:
In my case, I simply open the editor and type anything, but the characters don't print. It just freezes.

sinelaw:
Found an issue - reproduced in https://github.com/sinelaw/fresh/pull/1499

But I don't think it explains your case, @ricardo-bohorquez - can you share fresh log file? See `fresh --show-paths` to see where log files are stored.

sinelaw:
Should be fixed by https://github.com/sinelaw/fresh/pull/1500 - please try master branch or next release

PR #1500, opened 2026-04-07
Fix PageDown scroll OOM caused by infinite loop in wrapping transform

sinelaw:
**Title:** Fix PageDown scroll OOM caused by infinite loop in wrapping transform

**Body:**

## Root Cause

`apply_wrapping_transform` has a character-splitting inner loop that emits a Break when `remaining_width == 0`, then `continue`s without advancing `grapheme_idx`. After the Break, `current_line_width` is set to `line_indent`. The effective width for continuations is `available_width - line_indent`. So `remaining_width` becomes `available_width - 2*line_indent`.

When `line_indent > available_width/2` (but `< available_width - 10`, dodging the clamp), `remaining_width` is perpetually 0 — **infinite loop** allocating a Break + indent string every iteration until OOM.

This triggers at a specific scroll position because the viewport must first contain a line with indentation in the range `(available_width/2, available_width-10]`. With `CrosstermBackend`, `autoresize()` shrinks the viewport to the real terminal size on every `draw()`, making moderate indentation (e.g. 5-7 tabs at width 40) sufficient to trigger.

## Fixes

- **Break the infinite loop:** when `remaining_width==0` on a continuation where `current_line_width >= eff_width`, force-emit one grapheme instead of looping
- Guard `wrap_line()` and `apply_wrapping_transform()` against `width < 2`
- Cache `scrollbar_visual_row_counts()` to avoid O(n) wrapping per frame
- Reuse indent string in `emit_break_with_indent()` instead of allocating per call
- Fix test to use `Viewport::Fixed` preventing `autoresize` from shrinking terminal

## Verification

Without the fix, the CrosstermBackend test is OOM-killed (SIGKILL). With the fix, memory growth is 2 MB.


sinelaw:
Fixes #1454

PR #1535, opened 2026-04-10
Customizable status bar: DualList picker in settings + {clock} element

1612elphi:
## Summary

Introduces a fully customizable status bar. Before this PR, the status bar was hard-coded — users had no way to change which elements appeared, where they were placed, or to reorder them. This PR adds:

1. **A config schema for the status bar** (`StatusBarConfig` with `left`/`right` element arrays, line count, keybind hints toggle), plus a `StatusBarElement` enum covering 16 element types.
2. **A refactored renderer** that walks the configured `left`/`right` arrays and renders each element via a per-variant arm, replacing the previous hard-coded layout.
3. **A GUI for editing it** — a new reusable "DualList" picker control in the settings UI so users can pick and order elements by direct manipulation instead of hand-editing JSON.
4. **A new `{clock}` element** with a blinking colon separator, as one of the sixteen supported elements.

The headline feature is "the status bar is now customizable at all" — via either the JSON config or the settings UI.

## Status bar elements

Sixteen elements can now be placed in `left` or `right` (mutually exclusive — an element in `left` is hidden from `right`):

`{filename}` `{cursor}` `{cursor:compact}` `{diagnostics}` `{cursor_count}` `{messages}` `{chord}` `{line_ending}` `{encoding}` `{language}` `{lsp}` `{warnings}` `{update}` `{palette}` `{keybind_hints}` `{clock}`

Defaults preserve what the previous hard-coded status bar showed, so upgrading users see the same status bar unless they explicitly change the config.

## Why a GUI picker (the DualList control)

The status bar config is an ordered selection from a fixed set of options with cross-exclusion (an element lives in either `left` or `right`, not both). A free-form text list is the wrong UX for that — users shouldn't have to type `{cursor:compact}` exactly, and typos shouldn't silently break things. So the feature drives a new general-purpose "DualList" control:

```
Status Bar > Left:
  Available              Included
  ┌───────────────┐     ┌───────────────┐
  │ cursor:compact│ [>] │ filename      │
  │ chord         │ [<] │ cursor        │
  │ keybind_hints │ [^] │ diagnostics   │
  │               │ [v] │ cursor_count  │
  │               │     │ messages      │
  └───────────────┘     └───────────────┘
```

Transfer buttons (`[>]` `[<]`) move items between columns; reorder buttons (`[^]` `[v]`) rearrange the Included side. Both mouse and keyboard work (Tab switches columns, Up/Down navigates, Enter transfers, Ctrl+Up/Down reorders, Delete removes).

### DualList schema extensions

Two new JSON-Schema extensions, parsed in `view/settings/schema.rs`, let any schema opt a field into DualList rendering:

| Extension | Goes on | Purpose |
|---|---|---|
| `x-dual-list-options` | Item schema (e.g. `StatusBarElement`) | Declares the universe of values as `[{value, name}]` pairs |
| `x-dual-list-sibling` | The array field (e.g. `status_bar.left`) | JSON pointer to the sibling array whose members should be excluded from this list |

`StatusBarElement` gets a manual `JsonSchema` impl in `config.rs` to emit the options array, since schemars can't derive it from the enum variants alone. The control itself is general — any future `Vec<Enum>` field in any config can use the same extensions to get a DualList picker.

### Clock blinking

The `{clock}` element needs a periodic redraw to animate its separator. The editor's main loop only renders when something changes, so a cached `clock_blink_on: bool` lives on `Editor`, and `check_clock_blink_tick()` runs in `editor_tick()`:

- **Early-returns to zero cost when no `{clock}` element is configured** — so users who don't use the clock pay nothing for it.
- When the clock IS configured, flips the cached phase every ~500ms and returns `true` on the flip, which signals `editor_tick()` that a redraw is needed.
- The status bar renderer reads the same cached phase, so the tick check and the render share one source of truth (no risk of the two disagreeing across a 500ms boundary).

### Nested-struct flattening in the settings UI

Hit a related issue while wiring up the `StatusBarConfig` struct: previously, any field of type `SomeStruct` inside a parent struct became a `SettingType::Object` in the settings UI and fell through to a single JSON editor blob. That would have been a terrible UX for `status_bar` — users would see "Status Bar: { ... }" as one JSON textarea instead of individual `lines` / `left` / `right` / `show_keybind_hints` controls.

Fix: if the nested struct carries `#[schemars(extend("x-section" = "..."))]`, its sub-properties are now inlined into the parent category's settings list. Each child preserves its own `x-section`, so the existing section-grouping logic renders them under a `── Status Bar ────` header. This is gated on the section annotation, so unsectioned nested objects still render as a single JSON editor (tested in `test_unsectioned_nested_object_is_not_flattened`).

### Stale-schema gotcha worth knowing about

The settings UI reads the config schema via `include_str!("../../plugins/config-schema.json")` — a manually-regenerated file, not a live `schema_for!(Config)` call. I spent a while debugging why the DualList and clock options weren't showing up before realizing the baked schema was stale. `plugins/config-schema.json` is regenerated in this PR, and CI already has a drift check that runs `generate_schema` and diffs it against the committed file.

## Changes by area

**New: the feature itself**
- `config.rs` — `StatusBarElement` enum (16 variants), `StatusBarConfig` struct, manual `JsonSchema` impl emitting `x-dual-list-options`, defaults
- `view/ui/status_bar.rs` — renderer walks the configured `left`/`right` arrays instead of hard-coding; per-variant `render_element` arms
- `partial_config.rs` — partial-config wiring for the new struct
- `app/render.rs` — passes `status_bar` config + `clock_blink_on` into the status bar renderer

**New: the DualList control (general-purpose)**
- `view/controls/dual_list/mod.rs` — `DualListState`, `DualListColors`, `DualListLayout`, `DualListHit`, unit tests for add/remove/move/cursor
- `view/controls/dual_list/render.rs` — `render_dual_list_partial` with `skip_rows` for settings scroll

**New: settings-UI integration**
- `settings/schema.rs` — new `SettingType::DualList`, `DualListOptionEntry`, schema flattening for sectioned nested objects, new tests
- `settings/items.rs` — `SettingControl::DualList`, `build_dual_list_state` + `value_as_string_array` helpers, `control_to_value` arm
- `settings/state.rs` — `with_dual_list_mut` / `with_current_dual_list_mut` helpers, `refresh_dual_list_sibling`, `is_editing_dual_list`, new tests
- `settings/input.rs` — `handle_dual_list_editing_input` (Tab, Up/Down, Ctrl+Up/Down, Enter, Delete, Esc)
- `settings/mouse.rs` — 6 new hit handlers for the row/button areas
- `settings/layout.rs` — new `ControlDualList*` hit variants
- `settings/render.rs` — render dispatch
- `settings/entry_dialog.rs`, `settings/search.rs` — struct field updates for the new `dual_list_sibling` field on `SettingSchema`

**New: the `{clock}` element**
- `config.rs` — `StatusBarElement::Clock` variant
- `view/ui/status_bar.rs` — `Clock` render arm, `clock_blink_on` threaded through `render_status_bar` → `render_element`
- `app/mod.rs` — `clock_blink_on` field, `check_clock_blink_tick`, wired into `editor_tick()`, test for no-clock early return

**Regenerated:**
- `plugins/config-schema.json` — +172 lines with new extensions and `StatusBarElement` / `StatusBarConfig` definitions

## Test plan

- [x] `cargo test -p fresh-editor --lib` — **1901 passing, 0 failing** (6 new tests: DualList schema parsing, sectioned-nested flattening, unsectioned nested non-flattening, `refresh_dual_list_sibling`, `with_dual_list_mut` non-DualList guard, clock-blink-tick without clock)
- [x] `cargo clippy -p fresh-editor` — no new warnings in touched files
- [x] `cargo build -p fresh-editor --features runtime` — clean
- [x] Manual: open Settings → Editor → scroll to "Status Bar" section
- [x] Manual: move elements between Available/Included columns with mouse
- [x] Manual: keyboard — Tab between columns, Up/Down to navigate, Enter to transfer, Ctrl+Up/Down to reorder, Delete to remove
- [x] Manual: adding an element to `left` removes it from `right`'s Available (and vice-versa)
- [x] Manual: save, close settings, reopen editor — changes persist
- [x] Manual: add `{clock}` to status bar, verify `HH:MM` renders with blinking colon
- [x] Manual: default config produces the same status bar layout as before (no regression for upgrading users)

## Notes for reviewers

- The DualList control is deliberately general-purpose — nothing about it is status-bar-specific. Any future `Vec<Enum>` config field can opt in by emitting `x-dual-list-options` on the item schema. This is the main reason it's a new reusable control rather than a bespoke status-bar widget.
- `refresh_dual_list_sibling` assumes the sibling setting lives on the same settings page. This holds for `status_bar.{left,right}` (both flatten into the Editor page). Cross-page sibling pairs would silently no-op until `build_pages()` runs — noted in a doc comment on the function.
- The nested-struct flattening is opt-in via `x-section`; unsectioned nested objects still render as a JSON editor (see `test_unsectioned_nested_object_is_not_flattened`).
- Rebased onto current upstream master (`70968db4`) with one trivial test-section merge conflict resolved.

### DIFF
--- a/crates/fresh-editor/locales/cs.json
+++ b/crates/fresh-editor/locales/cs.json
@@ -1152,6 +1152,7 @@
   "settings.failed_to_save": "Uložení nastavení selhalo: %{error}",
   "settings.help_default": "↑↓:Navigace  Tab:Další  Enter:Upravit  /:Hledat  Esc:Zavřít",
   "settings.dual_list_enter_hint": "Enter pro úpravu",
+  "settings.dual_list_shift_hint": "Shift+←→ přesun  Shift+↑↓ pořadí",
   "settings.field.editor.whitespace_show": "Zobrazit bílé znaky",
   "settings.field.editor.whitespace_spaces_leading": "Úvodní mezery",
   "settings.field.editor.whitespace_spaces_inner": "Vnitřní mezery",
--- a/crates/fresh-editor/locales/cs.json
+++ b/crates/fresh-editor/locales/cs.json
@@ -1151,6 +1151,7 @@
   "settings.failed_to_open": "Otevření nastavení selhalo: %{error}",
   "settings.failed_to_save": "Uložení nastavení selhalo: %{error}",
   "settings.help_default": "↑↓:Navigace  Tab:Další  Enter:Upravit  /:Hledat  Esc:Zavřít",
+  "settings.dual_list_enter_hint": "Enter pro úpravu",
   "settings.field.editor.whitespace_show": "Zobrazit bílé znaky",
   "settings.field.editor.whitespace_spaces_leading": "Úvodní mezery",
   "settings.field.editor.whitespace_spaces_inner": "Vnitřní mezery",
--- a/crates/fresh-editor/locales/de.json
+++ b/crates/fresh-editor/locales/de.json
@@ -1152,6 +1152,7 @@
   "settings.failed_to_save": "Einstellungen konnten nicht gespeichert werden: %{error}",
   "settings.help_default": "↑↓:Navigieren  Tab:Weiter  Enter:Bearbeiten  /:Suchen  Esc:Schließen",
   "settings.dual_list_enter_hint": "Enter zum Bearbeiten",
+  "settings.dual_list_shift_hint": "Shift+←→ verschieben  Shift+↑↓ umsortieren",
   "settings.field.editor.whitespace_show": "Leerzeichen anzeigen",
   "settings.field.editor.whitespace_spaces_leading": "Führende Leerzeichen",
   "settings.field.editor.whitespace_spaces_inner": "Innere Leerzeichen",
--- a/crates/fresh-editor/locales/de.json
+++ b/crates/fresh-editor/locales/de.json
@@ -1151,6 +1151,7 @@
   "settings.failed_to_open": "Einstellungen konnten nicht geöffnet werden: %{error}",
   "settings.failed_to_save": "Einstellungen konnten nicht gespeichert werden: %{error}",
   "settings.help_default": "↑↓:Navigieren  Tab:Weiter  Enter:Bearbeiten  /:Suchen  Esc:Schließen",
+  "settings.dual_list_enter_hint": "Enter zum Bearbeiten",
   "settings.field.editor.whitespace_show": "Leerzeichen anzeigen",
   "settings.field.editor.whitespace_spaces_leading": "Führende Leerzeichen",
   "settings.field.editor.whitespace_spaces_inner": "Innere Leerzeichen",
--- a/crates/fresh-editor/locales/en.json
+++ b/crates/fresh-editor/locales/en.json
@@ -1163,6 +1163,7 @@
   "settings.help_footer": "Tab:Next button  Enter:Activate  Esc:Close",
   "settings.help_default": "↑↓:Navigate  Tab:Next  Enter:Edit  /:Search  Esc:Close",
   "settings.dual_list_enter_hint": "Enter to edit",
+  "settings.dual_list_shift_hint": "Shift+←→ move  Shift+↑↓ reorder",
   "settings.field.editor.whitespace_show": "Show Whitespace",
   "settings.field.editor.whitespace_spaces_leading": "Leading Spaces",
   "settings.field.editor.whitespace_spaces_inner": "Inner Spaces",
--- a/crates/fresh-editor/locales/en.json
+++ b/crates/fresh-editor/locales/en.json
@@ -1162,6 +1162,7 @@
   "settings.help_search": "Type to search, ↑↓:Navigate  Enter:Jump  Esc:Cancel",
   "settings.help_footer": "Tab:Next button  Enter:Activate  Esc:Close",
   "settings.help_default": "↑↓:Navigate  Tab:Next  Enter:Edit  /:Search  Esc:Close",
+  "settings.dual_list_enter_hint": "Enter to edit",
   "settings.field.editor.whitespace_show": "Show Whitespace",
   "settings.field.editor.whitespace_spaces_leading": "Leading Spaces",
   "settings.field.editor.whitespace_spaces_inner": "Inner Spaces",
--- a/crates/fresh-editor/locales/es.json
+++ b/crates/fresh-editor/locales/es.json
@@ -1152,6 +1152,7 @@
   "settings.failed_to_save": "Error al guardar configuración: %{error}",
   "settings.help_default": "↑↓:Navegar  Tab:Siguiente  Enter:Editar  /:Buscar  Esc:Cerrar",
   "settings.dual_list_enter_hint": "Enter para editar",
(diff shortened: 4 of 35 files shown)


sinelaw:
Thanks, The change is good, but it's a bit too much for one PR. Let's reduce the scope. Keep it focused only on the basic customizability:

1. **Drop `{keybind_hints}`, `show_keybind_hints`, `lines`, and the multi-row rendering.** That's an element variant, two config fields, a helper function, and a row loop — all for a feature orthogonal to customizable element placement. Ship it separately if needed.

2. **Left-side elements lose their styling.** `join_left_elements` concatenates into one `String` — the `RenderedElement` kinds you computed are thrown away. Right side gets per-element spans; left side should too.

3. **Bundle the loose parameters into a context struct.** 15 params on `render_element`, 20 on `render_status_bar`. This is the right time.

Also see CONTRIBUTING.md

1612elphi:
Thanks for the thoughtful review, and for all the work you've put into Fresh — it's been a joy to hack on. Pushed 1a427444 addressing all three points:

1. **Scope trimmed.** Dropped `StatusBarElement::KeybindHints`, `build_keybind_hints`, `show_keybind_hints`, `lines`, and the multi-row dispatch in `render_status`. The status bar row count is now hardcoded to 1 in `app/render.rs` and the element variant + helper are gone. `plugins/config-schema.json` is regenerated.

2. **Left side keeps per-element styling.** The old `join_left_elements` flattened every `RenderedElement` into one base-styled `String`; now a shared `render_side` helper builds `Vec<(Vec<Span>, width, ElementKind)>` for both sides, and `render_status` iterates them the same way. `update_layout_for_element` is called for every left group too, so clickable indicators like `{messages}` on the left get correct layout regions — the bespoke `track_left_message_area` walk is gone. Module constants `SSH_PREFIX` / `SSH_PREFIX_TERMINATOR` keep the filename builder and the `element_spans` two-color splitter in sync.

3. **`StatusBarContext` struct.** `render_status_bar` is now `(frame, area, ctx, config)` — down from 20 positional params — and `render_element` is `(element, ctx)` — down from 15. The call site in `app/render.rs` builds the context inline.

Small extras while I was in there: spans are moved into the output via `into_iter()` drain instead of cloned, right-side widths are cached from `element_spans` rather than recomputed via a second `str_width` pass, and the separator uses `Span::styled(" | ", base_style)` on a `&'static str` instead of a fresh `.to_string()` per frame.

`cargo check --all-targets` clean, `cargo fmt --check` clean, 1901/1901 lib tests passing.

sinelaw:
Based on review + some manual testing I made a few additional changes:
1. Moved the config flattening out of the parser - it's specifically only for Settings UI rendering so moved it there.
2. Made minor adjustments to the look of the DualList control
3. Fixed several usability issue in DualList - focus highlighting on the active row, changed it to shift+arrow for editing, etc
4. Clock rendering - removed the animation code, using hardware blink instead. It works nicely in most terminals - might not be as well supported but I'd like to keep the rendering as efficient and light as possible, regardless of having a clock or not

sinelaw:
Thanks!

1612elphi:
Thank YOU for the best text editor I've tried!

Issue #826, opened 2026-01-22
support helix mode


netkn001:

HughGrant:
would love to see some modal editing function here.

sinelaw:
There's a VI mode plugin you can try

Xerxes-2:
FWIW helix mode is not just vim mode with slight different keybindings, it needs some effort to make.
Check Zed's helix mode, even they also have some unimplemented features

PR #1599, opened 2026-04-17
part of support helix: step1, impl keymap

asukaminato0721:
part of #826 

This is the first step.


sinelaw:
Are you a helix user, do you have an opinion on how good this works?

sinelaw (changes requested):

Tested interactively: built from source, navigated View > Keybinding Style, confirmed "Helix" appears in the menu, and verified the vi mode auto-toggle on switch to/from Helix. All 14 locale files and both schema files are updated.

Issues found that need to be addressed before merge:

- **Empty keymap**: `keymaps/helix.json` has `"bindings": []` — selecting "Helix" gives users the exact same key bindings as "Default" plus vi mode. A user choosing "Helix" will find no helix-specific bindings (no `g`/`goto`, no `m`/marks, no `[`/`]` navigation, etc.). This is the core deliverable of the PR and it's missing.
- **Status message swallowed**: In `input.rs`, the `return self.handle_action(Action::PluginAction("vi_mode_toggle"))` exits early and overwrites the "Switched to 'helix' keybindings" status with "Vi mode enabled - NORMAL". The user never receives keymap-switch confirmation. Same issue when switching away from helix. The vi mode toggle should not replace the keymap-switch status.
- **config.example.json ordering**: The helix entry is placed at the *end* of the keybinding-style list in `config.example.json`, but between VSCode and macOS GUI in the Rust `MenuConfig`. These should be consistent.


---

sinelaw:
Sorry for the delay - 

Do you use Helix, so that you could really judge this change as a user? I don't use Helix myself so have no idea if it works well or not.

asukaminato0721:
Found that I barely use helix now...

PR #1605, opened 2026-04-18
Yet another install method

hudak:
Add instructions for using mise to install pre-build binaries.


sinelaw:
Cool, didn't know you could do that with mise

PR #1643, opened 2026-04-20
fix(keybinding editor): distinct descriptions for menu_open bindings

### LINKED ISSUE #1407: Keybindings for menu items do not describe what they are changing
Greetings, as I was about the change the key binds for the menu items I have noticed that all the menu key binds has the same description. It is not a big issue for me since the letters after the Alt make sense, but this could be a possible issue someone might have in the future. I don't know if this only applies for menu items as well, there might be other duplicates. Thank you for your time.

Version 0.2.20

Image

sinelaw:
## Summary

Fixes #1407 — in the keybinding editor, every `menu_open` row (Alt+F, Alt+E, Alt+V, …) rendered with the generic "Menu Open" fallback instead of "Open File menu", "Open Edit menu", etc.

## Root cause

`KeybindingResolver::format_action_from_str` called `Action::from_str(name, &empty_hashmap)`. For parameterised actions like `MenuOpen(name)` the parse fails without args, so the code fell through to the snake_case→Title Case converter — producing identical text for every `menu_open` entry.

## Fix

- Add `format_action_from_str_with_args` that accepts the action's args `HashMap` and feeds them to `Action::from_str`. The old `format_action_from_str` delegates to it with empty args (preserving existing call sites that genuinely have no args).
- In `keybinding_to_resolved` the owning `Keybinding` carries `kb.args`, so we pass them through for both chord and single-key bindings.
- The plugin-defaults loop already holds the decoded `Action` enum; switch it to `format_action` directly to skip the string round-trip.
- Make `format_action` `pub` so `app::keybinding_editor` can reuse it.

## Manual validation

Ran `cargo run` in a tmux session, opened the keybinding editor, and filtered on "Open". Each menu binding now shows a distinct description:

```
Alt+E   menu_open   Open Edit menu
Alt+F   menu_open   Open File menu
Alt+G   menu_open   Open Go menu
Alt+H   menu_open   Open Help menu
Alt+S   menu_open   Open Selection menu
Alt+V   menu_open   Open View menu
Alt+X   menu_open   Open Explorer menu
```

## Test plan

- [x] `cargo check --all-targets` passes
- [x] `cargo fmt` clean
- [x] New regression test `test_format_action_from_str_distinguishes_menu_open_by_name` covers the parse-with-args path
- [x] `cargo test -p fresh-editor --lib input::keybindings::tests` — 23 passed
- [x] Manual tmux validation shown above

https://claude.ai/code/session_01R6HFdhwDPTyXQYT3X1z8jQ

### DIFF
--- a/crates/fresh-editor/src/app/keybinding_editor/editor.rs
+++ b/crates/fresh-editor/src/app/keybinding_editor/editor.rs
@@ -83,13 +83,19 @@ pub struct KeybindingEditor {
 }
 
 impl KeybindingEditor {
-    /// Create a new keybinding editor from config and resolver
+    /// Create a new keybinding editor from config and resolver.
+    ///
+    /// `menu_names` are the stable English identifiers of top-level menus
+    /// (File, Edit, …, plus any plugin menus). They're used to enumerate
+    /// concrete `menu_open:<name>` entries in the action dropdown, so each
+    /// menu gets its own selectable row instead of one generic `menu_open`.
     pub fn new(
         config: &Config,
         resolver: &KeybindingResolver,
         mode_registry: &crate::input::buffer_mode::ModeRegistry,
         command_registry: &CommandRegistry,
         config_file_path: String,
+        menu_names: &[String],
     ) -> Self {
         let bindings =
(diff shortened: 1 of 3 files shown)


sinelaw:
image

Issue #1691, opened 2026-04-23
LSP highlights shifted after hot-exit recovery replay (didChange not sent)


sinelaw:
## Summary

After a workspace is restored with `editor.hot_exit = true` and a file-backed buffer has unsaved changes saved via hot-exit, LSP-driven highlights (semantic tokens, string coloring, hover, etc.) on the recovered file are shifted by the net byte delta between the on-disk and recovered contents. Fresh's own buffer and same-word highlights are correct; only LSP responses are off.

## Repro

1. In `~/.config/fresh/init.ts`, add some leading whitespace on a line with an LSP-typed call, e.g.

   ```ts
        const dash = editor.getPluginApi(\"dashboard\");
   ```

   Save to disk. Launch `fresh` with hot-exit enabled and `typescript-language-server` installed.
2. Edit the line to delete a few leading spaces (don't save). Buffer is dirty.
3. Quit cleanly — hot-exit writes the modified buffer to `~/.local/share/fresh/recovery/<id>`.
4. Restart `fresh` so the workspace + recovery are restored.
5. Open the `init.ts` tab.

## Observed

- The displayed text has the dedented line (correct).
- Syntax tokens from tsserver are shifted right by the number of characters removed: the `\"` token starts mid-word, the `.` operator is highlighted as if on a later column, etc.
- The blue \"same-word\" highlight on `dash` is correct — proves fresh's buffer matches the display; only LSP's view disagrees.

## Expected

LSP-driven highlights align with the displayed text.

## Root cause

On restore, `open_file_no_focus` sends `textDocument/didOpen` to the LSP with the **on-disk** content. Then `apply_workspace` replays hot-exit recovery by mutating the buffer directly (`buffer.delete` / `buffer.insert`) without emitting a `didChange`. The server is left on the pre-recovery content. Same gap exists in `apply_hot_exit_recovery`, `try_restore_hot_exit_buffers`, and `recover_all_buffers`.

PR #1692, opened 2026-04-23
fix(recovery): sync LSP after hot-exit replay so tokens don't drift (…

sinelaw:
…#1691)

Recovery replay paths mutate the buffer directly via buffer.delete/buffer.insert after the file has already been opened and didOpen sent with the on-disk content. That left the LSP server on the pre-recovery base: every position-bearing response (semantic tokens, hover, diagnostics, inlay hints, etc.) landed on the wrong column by the net byte delta between on-disk and recovered contents.

Add sync_lsp_after_recovery_replay(buffer_id), which pushes the current buffer content to the server as a full-document didChange, and call it from all four replay sites: recover_all_buffers (crash recovery), try_restore_hot_exit_buffers (--no-restore fallback), apply_workspace's inline hot-exit block (session restore), and apply_hot_exit_recovery (CLI-opened buffers). The helper reuses send_lsp_changes_for_buffer, so buffers whose LSP hasn't been opened yet fall through to a didOpen with the post-recovery text.

### DIFF
--- a/crates/fresh-editor/src/app/recovery_actions.rs
+++ b/crates/fresh-editor/src/app/recovery_actions.rs
@@ -14,6 +14,29 @@ use crate::model::event::BufferId;
 use super::Editor;
 
 impl Editor {
+    /// Push the buffer's full current content to LSP after an out-of-band
+    /// mutation (hot-exit recovery replay / crash-recovery replay). The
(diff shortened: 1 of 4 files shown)


sinelaw:
Fixes https://github.com/sinelaw/fresh/issues/1691

PR #1694, opened 2026-04-24
Fix Ctrl+Shift+letter bindings on macOS terminals

hoishing:
## Summary
- recover `SHIFT` for uppercase `Char` key events when kitty alternate keycodes clear the modifier
- keep Caps Lock uppercase handling intact so `Ctrl+A` still resolves correctly with Caps Lock enabled
- expose normalized lookup details in keyboard event debug output and update macOS keyboard docs/comments

## Problem
On macOS terminals with Kitty Keyboard Protocol support such as Ghostty, `crossterm` can report `Ctrl+Shift+K` as `Char('K')` with only `CONTROL` set when alternate keycodes are enabled. Fresh lowercased the character for matching but did not restore the lost `SHIFT`, so custom bindings like `Ctrl+Shift+K -> delete_line` resolved as `Ctrl+K` instead.

## Testing
- `cargo test -p fresh-editor keybindings::tests:: -- --nocapture`

### DIFF
--- a/crates/fresh-editor/keymaps/macos.json
+++ b/crates/fresh-editor/keymaps/macos.json
@@ -1,6 +1,6 @@
 {
   "inherits": "default",
-  "comment": "macOS keymap - avoids Ctrl+Shift combos, ASCII control chars (Ctrl+J=LF), and international keyboard conflicts",
+  "comment": "macOS keymap - prefers portable fallbacks for Ctrl+Shift chords, avoids ASCII control chars (Ctrl+J=LF), and avoids international keyboard conflicts",
   "bindings": [
     {
       "comment": "Ctrl+R -> Redo (Ctrl+Y is Unix yank, Ctrl+Shift+Z produces caron)",
@@ -27,7 +27,7 @@
       "when": "normal"
     },
     {
-      "comment": "Ctrl+Alt+K -> Delete line (Ctrl+Shift+K doesn't work on Mac)",
+      "comment": "Ctrl+Alt+K -> Delete line (portable fallback for terminals/parsers that collapse Ctrl+Shift+K)",
       "key": "k",
       "modifiers": ["ctrl", "alt"],
       "action": "delete_line",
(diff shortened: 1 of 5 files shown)


hoishing:
Concrete repro from my local macOS setup:

- Terminal: Ghostty
- Fresh config: `editor.keyboard_report_all_keys_as_escape_codes = true`
- Keybinding: `{ "key": "k", "modifiers": ["ctrl", "shift"], "action": "delete_line" }`

Observed behavior before this patch:

- Ghostty can emit a distinct `Ctrl+Shift+K` chord under Kitty keyboard protocol.
- In Fresh, the chord was arriving through `crossterm` in the alternate-keycode shape as uppercase `Char('K')` with only `CONTROL` set.
- Fresh lowercased that to `k` for lookup, but did not restore the dropped `SHIFT`, so the custom binding resolved as `Ctrl+K` instead of `Ctrl+Shift+K`.

That makes this a Fresh-side normalization bug rather than just a terminal limitation: the terminal is capable of distinguishing the chord, but Fresh was collapsing it during lookup.

The new tests cover both sides of that behavior:

- recover `SHIFT` for uppercase alternate-keycode events
- do not infer `SHIFT` when uppercase comes from Caps Lock

I also updated the keyboard debug output so this lookup shape is visible when diagnosing terminal/keybinding issues.

sinelaw (changes requested):

- **Core keybinding fix (SHIFT recovery): PASS** — `normalize_key_event` correctly recovers the lost `SHIFT` modifier when an uppercase `Char` arrives without `SHIFT` due to Kitty alternate keycodes, and correctly leaves it absent when `CAPS_LOCK` state is set. The three new unit tests cover the key scenarios and validate the fix.
- **Debug UI enhancement: FAIL (interactive test)** — The new `lookup=` line added to the event debug dialog (`view/event_debug.rs`) is not visible. The layout chunk for controls uses `Constraint::Length(4)`, which is exactly consumed by: `[q]/[Esc]/[c]` line + empty line + the two-line-wrapped raw details string. The lookup details are clipped. Fix: change `Constraint::Length(4)` → `Constraint::Length(5)` on the controls chunk in `view/event_debug.rs:52`.
- **macos.json comments and docs**: accurate and clear.
- **No code duplication or unnecessary complexity** in the core change.

Please fix the layout constraint so the `lookup=` line is actually rendered in the debug dialog.

---

hoishing:
@claude pls fix the CI issues

Issue #1705, opened 2026-04-26
Can't customize bindings for non-lsp completion popup (accept = tab, dismiss, etc)


sinelaw:

sinelaw:
Fixed in latest master branch

PR #1710, opened 2026-04-26
feat(keybindings): make completion popup accept/dismiss rebindable

sinelaw:
Tab and Esc were hardcoded inside the completion popup input handler, so users couldn't reassign them via the keybinding editor. Add a `Completion` keybinding context with two new actions — `completion_accept` (default Tab) and `completion_dismiss` (default Esc) — that the dispatcher consults when the topmost popup is a `PopupKind::Completion`. Falls through to the existing handler for type-to-filter, navigation, and Enter-passthrough so those behaviours remain unchanged.

### DIFF
--- a/crates/fresh-editor/keymaps/default.json
+++ b/crates/fresh-editor/keymaps/default.json
@@ -1318,6 +1318,22 @@
       "when": "popup"
     },
     {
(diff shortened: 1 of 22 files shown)


sinelaw:
Fixes #1705

Issue #1718, opened 2026-04-27
Panic when searching "hidden" in settings — byte index splits multi-byte UTF-8 character →


avinashkanaujiya:
## 🐛 Bug Description
Fresh panics when typing \"hidden\" into the settings search box. The panic occurs in the settings rendering code where a byte index falls inside a multi-byte UTF-8 character (`→`).

## Expected Behavior
Settings search should filter results without crashing, even when matching text that contains multi-byte Unicode characters.

## Actual Behavior
The editor crashes with the following panic:

```
thread 'main' (574586) panicked at crates/fresh-editor/src/view/settings/render.rs:2689:37:
byte index 98 is not a char boundary; it is inside '→' (bytes 96..99) of `Master toggle for whitespace indicator visibility.
When disabled, no whitespace indicators (·, →) are shown regardless
of the per-position settings below.
Default: true`
```

## Steps to Reproduce
1. Open Fresh
2. Open Settings
3. Type `hidden` into the search box
4. Editor panics immediately

## Environment
- OS: Linux (openSUSE Leap)
- Version: 0.3.0
- Terminal: Konsole

## Additional Context
The panic message indicates the string slicing logic at `crates/fresh-editor/src/view/settings/render.rs:2689` is using byte-based indexing on a string containing the `→` character (3 bytes in UTF-8), causing the panic. Switching to character-aware slicing (e.g., using `chars()` or `unicode-segmentation`) should fix this.

PR #1740, opened 2026-04-27
Fix UTF-8 panic in settings search description truncation

sinelaw:
## Summary
Fixes a panic that occurred when searching in the Settings dialog with certain terminal widths. The issue was caused by byte-based string slicing that could split multi-byte UTF-8 characters.

## Key Changes
- **Fixed unsafe string truncation in `render.rs`**: Replaced byte-based slicing with character-aware truncation using `floor_char_boundary()` to safely handle multi-byte UTF-8 sequences
  - Changed from `desc.len()` (byte count) to `desc.chars().count()` (character count) for length comparison
  - Used `floor_char_boundary()` to find the nearest valid UTF-8 boundary before the truncation point
  
- **Added comprehensive regression test**: Created `issue_1718_settings_search_utf8_panic.rs` that:
  - Tests the specific case from the issue (searching "hidden" which matches the `whitespace_show` setting)
  - Sweeps terminal widths from 40–250 to exercise all possible truncation cut-points
  - Verifies no panic occurs when rendering descriptions containing multi-byte characters like `·` (U+00B7, 2 bytes) and `→` (U+2192, 3 bytes)

## Implementation Details
The root cause was in the settings search result rendering code, which truncated setting descriptions to fit the terminal width. When `area.width - 5` fell inside a multi-byte UTF-8 character sequence, the byte slice would panic with "byte index X is not a char boundary". The fix ensures we only slice at valid UTF-8 character boundaries.

https://claude.ai/code/session_01YazhKuonyPoFfu4w321Rux


sinelaw:
Fixes #1718

PR #1764, opened 2026-04-28
feat: Add go to line with selection feature (absolute + relative)

PavelLoparev:
~This is `go to line with selection` feature for non vim mode (and a follow up for the `go to negative numbers for the relative line numbers mode` feature). We already have `ctrl + g` and `:` in command palette which moves cursor to the line. This PR adds `Got to line with selection` action and `::` (for consistency with `:`) for command palette which will move cursor + make selection from current position to target one.~

Should be useful for those users who won't use vim mode and also I believe in VScode there is such possibility too (probably with plugin, don't quite remember but i used it a lot there).

https://github.com/user-attachments/assets/de25a1bd-a090-42b5-a235-304538b98c63

### DIFF
--- a/crates/fresh-editor/plugins/goto_with_selection.i18n.json
+++ b/crates/fresh-editor/plugins/goto_with_selection.i18n.json
@@ -1,0 +1,58 @@
+{
+  "en": {
+    "cmd.goto_line_with_selection": "Go to Line with Selection",
+    "cmd.goto_line_with_selection_desc": "Select from current position to target line"
+  },
+  "cs": {
+    "cmd.goto_line_with_selection": "Go to Line with Selection",
+    "cmd.goto_line_with_selection_desc": "Select from current position to target line"
+  },
+  "de": {
+    "cmd.goto_line_with_selection": "Go to Line with Selection",
+    "cmd.goto_line_with_selection_desc": "Select from current position to target line"
+  },
+  "es": {
+    "cmd.goto_line_with_selection": "Go to Line with Selection",
+    "cmd.goto_line_with_selection_desc": "Select from current position to target line"
+  },
+  "fr": {
+    "cmd.goto_line_with_selection": "Go to Line with Selection",
+    "cmd.goto_line_with_selection_desc": "Select from current position to target line"
+  },
+  "it": {
+    "cmd.goto_line_with_selection": "Go to Line with Selection",
+    "cmd.goto_line_with_selection_desc": "Select from current position to target line"
+  },
+  "ja": {
+    "cmd.goto_line_with_selection": "Go to Line with Selection",
+    "cmd.goto_line_with_selection_desc": "Select from current position to target line"
+  },
+  "ko": {
+    "cmd.goto_line_with_selection": "Go to Line with Selection",
+    "cmd.goto_line_with_selection_desc": "Select from current position to target line"
+  },
+  "pt-BR": {
+    "cmd.goto_line_with_selection": "Go to Line with Selection",
+    "cmd.goto_line_with_selection_desc": "Select from current position to target line"
+  },
+  "ru": {
+    "cmd.goto_line_with_selection": "Go to Line with Selection",
+    "cmd.goto_line_with_selection_desc": "Select from current position to target line"
+  },
+  "th": {
+    "cmd.goto_line_with_selection": "Go to Line with Selection",
+    "cmd.goto_line_with_selection_desc": "Select from current position to target line"
+  },
+  "uk": {
+    "cmd.goto_line_with_selection": "Go to Line with Selection",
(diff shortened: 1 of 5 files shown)


PavelLoparev:
Hi @sinelaw, what is the process of updating docs to reflect new features? do you gather all features before release and update all the docs (site docs, inner fresh docs, fresh help txt file etc) or you expect docs to be incrementally updated in each PR?

sinelaw:
This feature is better implemented as a plugin or custom init.ts script. I had to fix gotoline mark handling to get it to work.

See https://github.com/sinelaw/fresh/pull/1768

PavelLoparev:
> This feature is better implemented as a plugin or custom init.ts script. I had to fix gotoline mark handling to get it to work.
> 
> See #1768

ok, will rework to plugin implementation later

sinelaw:
👍 there's an example in #1768

PavelLoparev:
I thought about plugin initially but didn't find how to register plugin action as `::` custom "command" in command palette (similar to `:` for go to line action).

PavelLoparev:
Hey @sinelaw, reimplemented with existing actions - much much easier!

sinelaw (changes requested):

- **Interactive test passed**: checked out PR branch, built binary, opened editor in tmux, used Ctrl+P → searched "Go to Line with Selection" → typed "5" → lines 1–4 were visually highlighted (selection background) and cursor moved to Ln 5, Col 1 with status "Jumped to line 5". The feature works correctly.
- **Plugin implementation is clean**: 17-line TypeScript plugin reusing existing `set_mark` + `goto_line` actions — simple, elegant, and exactly aligned with the owner's feedback.
- **CONTRIBUTING.md guideline #10 violated (locale keys)**: `goto_with_selection.i18n.json` uses English placeholder text verbatim for all 13 non-English locales (cs, de, es, fr, it, ja, ko, pt-BR, ru, th, uk, vi, zh-CN). The guideline states: *"Don't commit English placeholders."* Please provide real translations or coordinate with the project on how to handle this (e.g., if native speakers are needed, note that in the PR).
- **CONTRIBUTING.md guideline #2 violated (E2E tests observe, not inspect)**: `tests/e2e/plugins/goto_with_selection.rs` asserts on internal model state via `harness.editor().active_cursors().primary()`, `cursor.selection_range()`, `harness.has_selection()`, and `harness.editor_mut().active_state_mut().get_text_range()`. The guideline says: *"Do not call accessors that return model, view, or context state — if an invariant isn't visible on screen, cover it with a unit test on the component."* Tests should assert on rendered output instead (e.g., check that selected lines appear highlighted in the screen string, or that the status bar shows the expected line position).

---

PavelLoparev:
hey @sinelaw 

> if native speakers are needed, note that in the PR

do we really need to get 13 natives to close this PR? I could do just google mt or llm translations but just noticed a lot of existing plugins have default en strings in non en locales. Localisation is the whole another story I guess.

sinelaw:
I use only LLM for translations unless someone corrects me

sinelaw:
You can leave the testing comments, if you want, and I'll handle them with the new testing infra it should be easier

PavelLoparev:
okay.

about tests - do I understand correctly that assertion of selected text must be done trough reading the rendered data?

p.s. could you share you skill you use for the PR reviews? this will help to avoid these ping pongs with commits/reviews step by steps.

sinelaw:
It's basically checking that the PR follows CONTRIBUTING.md (https://github.com/sinelaw/fresh/blob/master/CONTRIBUTING.md)

The text selection specifically is more difficult to assert from the render. I would maybe check that using 'copy' after the selection and then paste results in the string appearing exactly two times on the screen

PavelLoparev:
Ok. added translations and reworked tests.

PavelLoparev:
test is basically selects text then copies then selects all then pastes - buffer contains only previously selected text

sinelaw:
That test will pass even if no selection, maybe? Copy and paste both do nothing in that case

PavelLoparev:
didn't get.

1. go to with selection - you get selected text.
2. copy it.
3. select all
4. paste - now the whole buffer contains only text from step 1
5. assert needed lines.

PavelLoparev:
test file contains 5 lines of text. test jumps around middle lines so it's not like "select all assert all"

sinelaw:
I was wondering if select doesn't happen, because someone breaks this feature: then it just copies nothing, selects all, and then paste does nothing because there's nothing to paste
Then the test might still pass

PavelLoparev:
ok, got it. then I guess one more step needed: delete after selecting all

PavelLoparev:
wait. test will still fall because it asserts that there must be only two lines but will be 5

Issue #1785, opened 2026-04-30
"Set Compose Width" suggestion popup overlaps prompt line when prompt is auto-hidden


sinelaw:
## Description

When invoking **Markdown: Set Compose Width** from the Command Palette, the suggestions popup (`None — Use viewport width`, `120 — Default compose width`, …) is drawn over the bottom rows of the screen, overlapping the area normally reserved for the prompt line. With the default `editor.show_prompt_line = false` (auto-hide) configuration, this means the user cannot see what they are typing in the `Compose width:` prompt while the popup is visible.

The prompt input only becomes visible once the user starts typing — at which point the popup auto-dismisses and the prompt re-appears.

## Steps to reproduce

1. Launch fresh with default config (`show_prompt_line` is false, the default).
2. Open the Command Palette (`Ctrl+P`), type `compose width`, select **Markdown: Set Compose Width**.
3. Observe: the suggestion popup is shown at the bottom of the screen. There is no prompt line visible above/below it where the input would be displayed.

## Expected

The popup should not occlude the prompt line. Either:
- reserve the prompt line row even when auto-hide is enabled while the prompt is active, and place the popup above it; or
- offset the popup vertically so the prompt line remains visible.

## Environment

- fresh built from source on `claude/summarize-github-issues-NwkCd`
- terminal: 200×50 (tmux)
- default config (no overrides)

sinelaw:
Closing: verified fixed on master (5c0d61c) by 33a3799 — `app/render.rs:202-216` now reserves the prompt row whenever a prompt is active (`prompt_line_visible || prompt.is_some()`), the issue's first suggested remedy. With default `show_prompt_line: false`, the "Set Compose Width" suggestion popup renders on rows above the prompt line and `Compose width: …` stays visible on the bottom row at all times, including while typing.

---
_This comment is autogenerated: an automated verification against a debug build._

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1800, opened 2026-04-30
fix(prompt): recompute layout after mid-render plugin command dispatch

sinelaw:
## Summary

Fixes a mid-render plugin command race in `Editor::render` that produced visible status-bar leakage on the prompt's row when the audit_mode plugin's "Review PR Branch" command was invoked from the command palette under auto-hide mode (`show_prompt_line = false`). Adds harness infrastructure (`PluginManager::test_inject_command`) and a deterministic regression test that reproduces the bug without booting the real plugin runtime.

This PR replaces my earlier #1785 attempt (a `render.rs` clamp on the prompt's render rect), which was reverted because the clamp turned out to be a no-op — the layout's `prompt_chunk` is only ever off-screen because of this race, not because the constraint solver places it there on consistent state.

## Root cause

`Editor::render` (`crates/fresh-editor/src/app/render.rs:402`) drains the plugin manager's command queue *after* the layout has been computed but *before* the prompt/status bar are rendered:

```rust
// Layout computed at line 164 with current self.prompt
let main_chunks = Layout::default()...
// Hooks fire (lines_changed, view_transform_request)
// Line 402: drain *all* plugin commands and dispatch
let commands = self.plugin_manager.process_commands();
for command in commands { self.handle_plugin_command(command); }  // may set self.prompt = Some(...)
// Status bar + prompt rendered using stale main_chunks
```

When the audit_mode plugin's `editor.prompt(label, initial)` queues a `StartPromptAsync` between renders, the next render's mid-render `process_commands()` dispatches it and sets `self.prompt = Some(AsyncPrompt)`. The cached `main_chunks` still reflects `prompt = None`, so:
- `prompt_chunk = (y: size.height, h: 0)` — off-screen, 0 height
- `status_chunk = (y: size.height-1, h: 1)` — gets the bottom row

Status bar paints to the bottom row; prompt paints to a 0-height slot. Confirmed against the real binary's debug log:

```
DBG_RENDER prompt_type=AsyncPrompt sugg_count=0 status_chunk=Rect{y:49,h:1} prompt_chunk=Rect{y:50,h:0}
```

The next frame self-corrects (`status_chunk={y:48,h:1} prompt_chunk={y:49,h:1}`), which is the "shifts up by one row" the user reported.

## Fix

After the mid-render command dispatch, if any commands were processed, recompute the layout-affecting flags (`has_suggestions`, `has_file_browser`, `prompt_is_overlay`, `show_search_options`) and re-split `main_chunks`. The bottom-row render code (status bar, prompt) then paints into the correct slots.

- **Conservative trigger**: "any commands processed" rather than "any layout-affecting command processed". Layout::split is cheap; this avoids a PluginCommand whitelist that would silently regress as new variants are added.
- **Bounded**: single drain + single recompute. No second `process_commands` call, no recursion. Hooks fired by the dispatch handlers above queue their responses for the next render/`editor_tick` — the same one-frame behaviour the existing comment already accepts as imperceptible at 60fps.
- **`main_content_area` intentionally not re-derived**: the file explorer / split rendering already painted to the original `main_content_area`. The bottom-row recompute may overwrite a single row of main content where the new status bar / prompt now sits; that brief overlap self-corrects on the next frame, where the layout is built consistently from the start.

## What this PR adds

1. **`PluginManager::test_inject_command`** (`crates/fresh-editor/src/services/plugins/manager.rs`) — always-present (zero overhead) `pending_injected_commands: Vec<PluginCommand>` queue. `process_commands()` drains it alongside the real plugin thread's commands, and `is_active()` returns true if the queue is non-empty so the in-render `process_commands` block (gated on `is_active`) actually runs in tests with no real plugin runtime.

2. **Regression test** — `test_mid_render_start_prompt_async_keeps_prompt_visible` in `tests/e2e/prompt.rs`:
   - Toggles auto-hide on, sets a status message containing a marker word, asserts the marker lands on the bottom row (sanity).
   - Injects a `StartPromptAsync` and renders once. The render fires the in-render command processing path, dispatches the injected command, and asserts the bottom row shows the prompt's label and contains no status-bar leftovers.
   - Without the fix this asserts-out cleanly with "Status-bar text from before the prompt opened leaked through the prompt's row".

3. **Fix** in `Editor::render` — recompute layout if mid-render command dispatch processed anything.

## Test plan

- [x] `cargo test -p fresh-editor --lib` — 2289 passing
- [x] `cargo test -p fresh-editor --test e2e_tests prompt_` — 75 passing
- [x] `cargo test -p fresh-editor --test e2e_tests popup` — 101 passing
- [x] `cargo test -p fresh-editor --test e2e_tests markdown_compose` — 62 passing
- [x] `cargo test -p fresh-editor --test e2e_tests test_mid_render_start_prompt_async_keeps_prompt_visible` — passes with fix; reverting just the fix portion makes it fail with the expected diagnostic
- [x] `cargo fmt`, `cargo check --all-targets` clean; no new clippy warnings on the changed files
- [x] Manual reproduction in tmux (200×50, default config, README.md, `Ctrl+P` → `Review PR` → Enter) — bottom row now correctly shows `Base ref to compare against (default: master): master` with the status bar one row above; prior to the fix the prompt row showed status-bar text leaking past the prompt's content

https://claude.ai/code/session_017NxQWM8Q3y1HDKoVUWbsfE

### DIFF
--- a/crates/fresh-editor/src/app/render.rs
+++ b/crates/fresh-editor/src/app/render.rs
@@ -102,8 +102,16 @@ impl Editor {
             }
         }
 
-        // Determine if we need to show search options bar
-        let show_search_options = self.prompt.as_ref().is_some_and(|p| {
+        // Determine if we need to show search options bar.
+        // (Held in mutable bindings because the in-render
+        // `process_commands` block below can dispatch commands —
+        // e.g. `StartPromptAsync`, `SetPromptSuggestions` — that
+        // mutate `self.prompt`. When that happens we recompute these
+        // flags and re-split `main_chunks` so the bottom-row
+        // rendering uses an up-to-date layout. See the
+        // "Recompute layout if mid-render commands changed state"
+        // block below.)
+        let mut show_search_options = self.prompt.as_ref().is_some_and(|p| {
             matches!(
                 p.prompt_type,
                 PromptType::Search
@@ -120,13 +128,13 @@ impl Editor {
         // wrong. Floating-overlay prompts (Live Grep, issue #1796)
         // are exempt because their suggestions live inside the
         // centred frame, not above the bottom row.
-        let prompt_is_overlay = self.prompt.as_ref().is_some_and(|p| p.overlay);
-        let has_suggestions = self
+        let mut prompt_is_overlay = self.prompt.as_ref().is_some_and(|p| p.overlay);
+        let mut has_suggestions = self
             .prompt
             .as_ref()
             .is_some_and(|p| !p.suggestions.is_empty())
             && !prompt_is_overlay;
-        let has_file_browser = self.prompt.as_ref().is_some_and(|p| {
+        let mut has_file_browser = self.prompt.as_ref().is_some_and(|p| {
             matches!(
                 p.prompt_type,
                 PromptType::OpenFile | PromptType::SwitchProject | PromptType::SaveFileAs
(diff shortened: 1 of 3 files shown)


sinelaw:
Not actually reproduced yet

sinelaw:
Also fixes #1847

PR #1803, opened 2026-04-30
feat(file-explorer): add follow_active_buffer config option

ko3n1g:
<details><summary>Claude summary</summary>

Adds `file_explorer.follow_active_buffer` (default: `false`) so users can opt in to automatic file explorer sync on tab switch.

When enabled, switching tabs expands the tree and highlights the file corresponding to the newly active buffer — without stealing keyboard focus from the editor.

**Usage** (`~/.config/fresh/config.json`):
```json
{
  "file_explorer": {
    "follow_active_buffer": true
  }
}
```

**How it works:**

`sync_file_explorer_to_active_file()` already existed and handled exactly this expand-and-select logic. It was just never called on tab switch. The only addition is calling it from `set_active_buffer` when the explorer is visible, the new setting is enabled, and the user isn't actively navigating inside the explorer panel (`key_context != FileExplorer`).

The `file_explorer_sync_in_progress` guard inside `sync_file_explorer_to_active_file()` prevents re-entrant async tasks on rapid tab switching.

**Changes:**
- `config.rs` — `follow_active_buffer: bool` field on `FileExplorerConfig` (default `false`)
- `partial_config.rs` — corresponding `Option<bool>` in `PartialFileExplorerConfig`, wired into `merge_from`, `From`, and `resolve`
- `active_focus.rs` — call `sync_file_explorer_to_active_file()` in `set_active_buffer` when the new setting is `true`

</details>

### DIFF
--- a/crates/fresh-editor/plugins/config-schema.json
+++ b/crates/fresh-editor/plugins/config-schema.json
@@ -133,7 +133,8 @@
         "width": "30%",
         "preview_tabs": true,
         "side": "left",
-        "auto_open_on_last_buffer_close": true
+        "auto_open_on_last_buffer_close": true,
+        "follow_active_buffer": false
       }
     },
     "file_browser": {
@@ -952,6 +953,11 @@
           "description": "Automatically focus the file explorer when the last buffer is\nclosed. Set to `false` for a \"blank workspace\" workflow where\nnothing opens automatically and the user explicitly invokes the\nfile explorer (e.g. via keybinding or command palette).\nDefault: true",
           "type": "boolean",
           "default": true
+        },
+        "follow_active_buffer": {
+          "description": "When the file explorer sidebar is open, automatically expand the\ntree and highlight the file that corresponds to the active buffer\nwhenever you switch tabs. Set to `true` to keep the explorer\nselection in sync with the active tab.\nDefault: false",
+          "type": "boolean",
+          "default": false
         }
       }
     },
--- a/crates/fresh-editor/src/app/active_focus.rs
+++ b/crates/fresh-editor/src/app/active_focus.rs
@@ -121,6 +121,7 @@ impl Editor {
         self.ensure_active_tab_visible(active_split, buffer_id, self.effective_tabs_width());
 
         if self.file_explorer_visible
+            && self.config.file_explorer.follow_active_buffer
             && self.key_context != crate::input::keybindings::KeyContext::FileExplorer
         {
             self.sync_file_explorer_to_active_file();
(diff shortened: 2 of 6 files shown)


ko3n1g:
hey @sinelaw, `fresh` is an awesome CLI in the age of agentic workflows! One feature I was deeply missing when switching from vscode to fresh is the option to follow files in the explorer. I get that the flickering might be disruptive for some folks, so this PR hides it behind a setting (default: false) (commit for this is in-flight).

sinelaw (changes requested):

**Interactive testing** (tmux session, `follow_active_buffer: true`, 3 files in a flat dir):
- opened `file1.rs`, `file2.rs`, `file3.rs`; toggled file explorer open
- cycled tabs with `Ctrl+PageDown` — explorer selection correctly followed each active buffer (bg highlight moved file3→file1→file2) ✓
- focus-steal guard (`key_context != FileExplorer`) and re-entrancy guard (`file_explorer_sync_in_progress`) both observed to work correctly ✓

**CONTRIBUTING.md violations that must be fixed before merge:**

- **Missing test** (guideline §Testing #1): every new feature must include a test that *fails without the change and passes with it*. No test is present in either commit. Please add an e2e test that opens multiple files, switches tabs, and asserts the explorer selection matches the active buffer (observe rendered output only, no model accessors).
- **Schema not regenerated** (guideline §Code Guidelines #6): `FileExplorerConfig` derives `JsonSchema`; after adding `follow_active_buffer` you must run `./scripts/gen_schema.sh` and commit the updated `plugins/config-schema.json` (and `config-schema.json` at repo root if it also encodes this type).

Code quality and architecture are otherwise clean — the implementation correctly reuses `sync_file_explorer_to_active_file()` with a minimal, well-guarded call site.

---

sinelaw:
Does it flicker when switching tabs now? That was the original reason for not doing this

ko3n1g:
> Does it flicker when switching tabs now? That was the original reason for not doing this

It does flicker a bit, yes. That's another reason why I proposed it as opt-in rather than opt-out. Personally, I'd be willing to bite the bullet for this feature, but I understand for some this might be a dealbreaker.

ko3n1g:
hey @sinelaw, could you re-launch CI another time when you have a moment?

ko3n1g:
hey @sinelaw , thanks for rebasing! Looks like CI is now passing. Let me know if other changes are required for getting it in

sinelaw:
Thanks! Merged

The flicker isn't as bad as it used to be, probably rendering changes since then. Maybe we can consider making it opt-out instead

PR #1819, opened 2026-04-30
fix(indent): don't push existing content right when Enter is at column 0

### LINKED ISSUE #1425: Auto Indent is too diligent
[ I am using ``fresh-editor-0.2.20-x86_64.AppImage`` ]

With ``Auto Indent``  switched on, the following happens:

```
····line1
····line2
········line3
········line4
```

i.e. line2 follows the indentation of line1,
and line4 follows the indentation of line3.
So far, so good.

But if an unindented line appears after some empty lines:

```
····line1
····line2
········line3
········line4

unindented line
```

and to insert another newline
I hit ``Enter`` at the beginning of the unindented line 
then the unindented line suddenly becomes indented, and follows line4:

```
····line1
····line2
········line3
········line4

········unindented line
```

To correct this behaviour manually is rather cumbersome.

sinelaw:
## Summary

Fixes #1425.

**Root cause.** Auto-indent (`calculate_indent_pattern`) checked whether the current line was empty by looking only at bytes *before* the cursor. When the cursor was at column 0 of a non-empty line, that slice was empty, so the algorithm walked back through blank lines to a previous indent and inserted that indent between the new `\n` and the existing line, displacing it rightward.

**Fix.** When the cursor is inside (or at the boundary of) a non-empty line's leading whitespace, return the cursor's column as the auto-indent. This preserves the existing content's column position for free across all column positions in the indent area:

- col 0 of `unindented line` → 0 (no displacement)
- col 0 of `    indented_target` → 0 (the 4 spaces ride over with the content untouched)
- col 2 of `    indented_target` → 2 (line A keeps 2 spaces; 2 more remain in front of `indented_target`)
- col 4 of `    foo()` (just before `f`) → 4

The rule is **language-agnostic** — it does not look at what character starts the content (a word, `}`, `end`, `</tag>`, `fi`, `done`, …) — and matches the behaviour of VS Code, Sublime Text, and similar editors. There is no hand-maintained list of closing tokens that would have failed to generalise to Lua/Ruby `end`, HTML `</tag>`, Bash `fi`/`done`, etc.

Smart-indent (tree-sitter / pattern matching) still fires when the cursor is past the leading whitespace, where it actually adds value (end of `fn main() {`, after `:` in Python, etc.).

The existing `test_indent_after_empty_line_uses_reference` test was repurposed to cover what its name describes — a cursor on a *truly* empty line between code lines — since its prior setup placed the cursor at col 0 of the `}` line and asserted the displacement behaviour this PR removes.

## Test plan

Unit tests in `crates/fresh-editor/src/primitives/indent.rs`:
- [x] `test_enter_at_start_of_unindented_line_after_blank_does_not_indent` — the exact #1425 fixture
- [x] `test_enter_at_start_of_indented_line_does_not_displace_content` — col 0 of an already-indented line
- [x] `test_enter_at_start_of_unindented_line_python` — same, on a tree-sitter language
- [x] `test_enter_in_middle_of_leading_ws_preserves_content_column` — col 2 of `    indented_target` returns 2
- [x] `test_enter_at_start_of_closing_brace_line_does_not_displace` — `}` stays at col 0 (matches VS Code)
- [x] `test_enter_inside_content_still_uses_smart_indent` — sanity check that smart-indent isn't short-circuited when past the leading ws

Suites:
- [x] All 26 `primitives::indent` tests pass
- [x] Full `fresh-editor` lib suite green (2235 tests)
- [x] Full `auto_indent` e2e suite green (56 tests)
- [x] `cargo fmt --check` clean

Manual validation in tmux (verified pane output + saved bytes via `cat -A`):
- [x] Plain text `unindented line` after a blank line → preserved at col 0
- [x] Plain text `    indented_target` cursor at col 0 → preserved at col 4
- [x] Plain text `    indented_target` cursor at col 2 (mid-indent) → preserved at col 4
- [x] Plain text mid-content split → smart-indent fires
- [x] Rust `}` at col 0 → preserved at col 0 (was being displaced)
- [x] Rust end of `fn main() {` → smart-indent fires (col 5)
- [x] Lua `end` at col 0 → preserved at col 0 (no `end` rule required)
- [x] HTML `</div>` at col 0 → preserved at col 0 (no tag rule required)
- [x] Python `unindented` after blank line → preserved at col 0
- [x] Python after `:` → smart-indent fires (col 5)

https://claude.ai/code/session_01Vp3JuxNnaPFzi9euUnCQz5

### DIFF
--- a/crates/fresh-editor/src/primitives/indent.rs
+++ b/crates/fresh-editor/src/primitives/indent.rs
@@ -198,6 +198,14 @@ impl IndentCalculator {
         language: &Language,
         tab_size: usize,
     ) -> Option<usize> {
+        // When the cursor is at column 0 of an existing non-empty line,
+        // pressing Enter splits before the line content. Adding any indent
+        // here would shift the existing content rightward, so the auto-indent
+        // for the new (empty) line above must be 0. See #1425.
+        if Self::is_at_start_of_nonempty_line(buffer, position) {
+            return Some(0);
+        }
+
         // Try tree-sitter-based indent
         if let Some(indent) =
             self.calculate_indent_tree_sitter(buffer, position, language, tab_size)
@@ -222,6 +230,11 @@ impl IndentCalculator {
         position: usize,
         tab_size: usize,
     ) -> usize {
+        // See `calculate_indent` for the rationale (#1425).
+        if Self::is_at_start_of_nonempty_line(buffer, position) {
+            return 0;
+        }
+
         // Pattern-based indent (for incomplete syntax)
         if let Some(indent) = Self::calculate_indent_pattern(buffer, position, tab_size) {
             return indent;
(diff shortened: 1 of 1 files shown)


crates/fresh-editor/src/primitives/indent.rs
```diff
@@ -231,6 +244,38 @@ impl IndentCalculator {
...
+            position == 0 || Self::byte_at(buffer, position.saturating_sub(1)) == Some(b'\n');
+        if !at_line_start {
+            return false;
+        }
+
+        // Look ahead on the same line for the first non-whitespace character.
+        let mut pos = position;
+        while pos < buffer.len() {
+            match Self::byte_at(buffer, pos) {
+                Some(b'\n') => return false,
```
sinelaw: can we avoid hard-coding delimiters?

PR #1843, opened 2026-05-02
feat: Status bar tokens registration API + git branch status bar token

PavelLoparev:
https://github.com/user-attachments/assets/bbe656aa-8c78-4b46-9a34-5ea377200020

Re-implemented as discussed:
1. Added new api for registering status bar elements from plugins
2. Added git_statusbar plugin that uses new API and introduces `{git_statusbar:branch}` element

### DIFF
--- a/crates/fresh-editor/src/config.rs
+++ b/crates/fresh-editor/src/config.rs
@@ -801,6 +801,13 @@ pub fn get_custom_status_bar_tokens() -> Vec<(String, String)> {
         .collect()
 }
 
+/// Clear all registered custom statusbar tokens.
+/// Used for test isolation between parallel test runs.
+pub fn clear_custom_status_bar_tokens() {
+    let mut tokens = CUSTOM_STATUS_BAR_TOKENS.lock().unwrap();
+    tokens.clear();
+}
+
 /// Get the value for a custom token by key.
 pub fn get_custom_status_bar_value(key: &str) -> Option<String> {
     let token = {
--- a/crates/fresh-editor/src/config.rs
+++ b/crates/fresh-editor/src/config.rs
@@ -749,18 +749,6 @@ impl Default for StatusBarConfig {
     }
 }
 
-
-
-
-
-
-
-
-
-
-
-
-
 /// Editor behavior configuration
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
 pub struct EditorConfig {
--- a/crates/fresh-editor/src/config.rs
+++ b/crates/fresh-editor/src/config.rs
@@ -730,13 +730,13 @@ pub struct StatusBarConfig {
     /// Elements shown on the left side of the status bar.
     /// Default: ["{filename}", "{cursor}", "{diagnostics}", "{cursor_count}", "{messages}"]
     #[serde(default = "default_status_bar_left")]
-    #[schemars(extend("x-section" = "Status Bar", "x-dual-list-sibling" = "/editor/status_bar/right"))]
+    #[schemars(extend("x-section" = "Status Bar", "x-dual-list-sibling" = "/editor/status_bar/right", "x-dynamicly-extendable-status-bar-elements" = true))]
     pub left: Vec<StatusBarElement>,
 
     /// Elements shown on the right side of the status bar.
     /// Default: ["{line_ending}", "{encoding}", "{language}", "{lsp}", "{warnings}", "{update}", "{palette}"]
     #[serde(default = "default_status_bar_right")]
-    #[schemars(extend("x-section" = "Status Bar", "x-dual-list-sibling" = "/editor/status_bar/left"))]
+    #[schemars(extend("x-section" = "Status Bar", "x-dual-list-sibling" = "/editor/status_bar/left", "x-dynamicly-extendable-status-bar-elements" = true))]
     pub right: Vec<StatusBarElement>,
 }
 
--- a/crates/fresh-editor/src/config.rs
+++ b/crates/fresh-editor/src/config.rs
@@ -7,7 +7,7 @@ use std::borrow::Cow;
 use std::collections::HashMap;
 use std::ops::Deref;
 use std::path::Path;
-use std::sync::{Arc, LazyLock, Mutex};
+
 /// Newtype for theme name that generates proper JSON Schema with enum options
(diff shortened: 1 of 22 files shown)


sinelaw:
As usual I prefer keeping the core small and doing these sort of featured as plugins, like we have for the git gutter. That means we need some new API for plugins to register status bar elements. 
For the branch checking some kind of polling like we have for git gutter is probably good enough, once every few seconds

PavelLoparev:
Hi @sinelaw, I introduced API for registering custom statusbar elements from plugins + added git statusbar plugin which adds token for current branch.

sinelaw (changes requested):

**Interactive testing:** built from branch `git-branch-status-bar`, launched the editor with `{git_statusbar:branch}` in the left status bar config pointing at a git repo. Status bar correctly displayed the branch name (`pr-1843`); showed `Not in git` in a repo without commits. The API and rendering pipeline work end-to-end.

**Requested changes (CONTRIBUTING.md violations):**

- **Sleep in test (`git_statusbar.rs` lines 54–57):** `std::thread::sleep(Duration::from_millis(50))` inside a loop violates the "No timeouts or time-sensitive tests" guideline. Replace with semantic waiting, e.g. `harness.wait_until(|h| fresh::config::get_custom_status_bar_tokens().iter().any(|(k, _)| k == "{git_statusbar:branch}")).unwrap()`.

- **Global static breaks test isolation (`config.rs` `CUSTOM_STATUS_BAR_TOKENS`):** The `LazyLock<Mutex<HashMap>>` is process-global; tokens registered in one parallel test persist into other tests, causing spurious passes or failures. CONTRIBUTING.md requires tests to be isolated and run in parallel. The registry needs to be either scoped to the editor instance (passed via context rather than a static) or cleared/reset between test runs.

**Minor:**
- `items.rs`: the `schema.path.contains("status_bar")` heuristic is fragile; any future field whose JSON path contains that substring will also get custom tokens injected. Consider a more targeted discriminant (e.g. a marker in the schema `x-` extension rather than a substring match).

---

PavelLoparev:
> this is an autogenerated comment
> 
> **Interactive testing:** built from branch `git-branch-status-bar`, launched the editor with `{git_statusbar:branch}` in the left status bar config pointing at a git repo. Status bar correctly displayed the branch name (`pr-1843`); showed `Not in git` in a repo without commits. The API and rendering pipeline work end-to-end.
> 
> **Requested changes (CONTRIBUTING.md violations):**
> 
> * **Sleep in test (`git_statusbar.rs` lines 54–57):** `std::thread::sleep(Duration::from_millis(50))` inside a loop violates the "No timeouts or time-sensitive tests" guideline. Replace with semantic waiting, e.g. `harness.wait_until(|h| fresh::config::get_custom_status_bar_tokens().iter().any(|(k, _)| k == "{git_statusbar:branch}")).unwrap()`.
> * **Global static breaks test isolation (`config.rs` `CUSTOM_STATUS_BAR_TOKENS`):** The `LazyLock<Mutex<HashMap>>` is process-global; tokens registered in one parallel test persist into other tests, causing spurious passes or failures. CONTRIBUTING.md requires tests to be isolated and run in parallel. The registry needs to be either scoped to the editor instance (passed via context rather than a static) or cleared/reset between test runs.
> 
> **Minor:**
> 
> * `items.rs`: the `schema.path.contains("status_bar")` heuristic is fragile; any future field whose JSON path contains that substring will also get custom tokens injected. Consider a more targeted discriminant (e.g. a marker in the schema `x-` extension rather than a substring match).
> 
> _Generated by [Claude Code](https://claude.ai/code)_

* Used semantic wait
* Introduced function for clearing up tokens
* Tuned `status_bar` check in config

crates/fresh-editor/src/view/settings/items.rs
```diff
@@ -240,10 +240,24 @@ fn build_dual_list_state(
...
+    // Start with static schema options (built-in tokens)
+    let mut all_options: Vec<(String, String)> = options
         .iter()
         .map(|o| (o.value.clone(), o.name.clone()))
         .collect();
+
+    // Add runtime options (custom tokens from plugins) if this is status bar config
+    // Check for exact status bar left/right field paths
+    if schema.path == "/editor/status_bar/left" || schema.path == "/editor/status_bar/right" {
+        let custom_tokens = crate::config::get_custom_status_bar_tokens();
```
sinelaw: This code is supposed to be generic and only depend on the schema structure. We shouldn't have custom reference to specific concepts like the status bar. Needs some design

crates/fresh-editor/src/config.rs
```diff
@@ -739,6 +749,88 @@ impl Default for StatusBarConfig {
...
 }
 
+/// Registry for custom statusbar tokens registered by plugins.
+/// Key format: "plugin_name:token_name" (e.g., "git:branch")
+pub struct CustomStatusBarToken {
+    pub title: String,
+    pub value: Mutex<Option<String>>,
+}
+
+static CUSTOM_STATUS_BAR_TOKENS: LazyLock<Mutex<HashMap<String, Arc<CustomStatusBarToken>>>> =
```
sinelaw: All state should be part of the context passed down through the call stack and editor state. Not global

crates/fresh-editor/src/view/settings/items.rs
PavelLoparev: we probably need to add new `dynamicly_extendable: true` property to the lef/right nodes and then check this property value instead of checking paths. Sounds good?

crates/fresh-editor/src/config.rs
PavelLoparev: Agree. Will check this.

crates/fresh-editor/src/view/settings/items.rs
PavelLoparev: Added `x-dynamically-extendable-status-bar-elements` to the setting item to not to rely on the specific paths.

crates/fresh-editor/src/config.rs
PavelLoparev: Moved state to editor.

PavelLoparev: Not sure this is the best way of doing this refactoring. But all tests pass.

PavelLoparev:
tested branch. works fine.

sinelaw:
🔥

PR #1849, opened 2026-05-03
feat: Add tab actions plugin (close all, close other, close to right, close to left, move tab left, move tab right)

### LINKED ISSUE #1844: Add buffer actions: close all, close to the right, move buffer tabs
It would be really handy to have these actions available:
* Close all buffers/tabs
* Close others buffers/tabs
* Close buffers/tabs to the right
* Move active buffer/tab to the left or right

PavelLoparev:
Closes https://github.com/sinelaw/fresh/issues/1844

# Summary
Added `tab_actions` plugin with next actions:
* Close all
* Close others
* Close to left
* Close to right
* Move tab left
* Move tab right

# Why
In VScode (and any many others text editor) there are these actions and it's very handy to clean up your bloated split at some point.

Also, sometimes you just need to re-order tabs. I find it useful to map `ctrl + alt + pageUp/pageDown` to drag tabs left and right accordingly (to be aligned with `alt + pageUp/pageDown` for switching tabs left and right)

# Details
There were already these methods in the rust backend I just exposed those to plugin API and wrote a simple wrapper plugin to register actions.

# Demo

https://github.com/user-attachments/assets/bf1e2f8d-5962-46ed-9d6e-df68666a22c9

https://github.com/user-attachments/assets/e42e5cec-256d-4104-8ed7-e5c43544e425


crates/fresh-editor/plugins/lib/fresh.d.ts
```diff
@@ -707,17 +707,6 @@ type AuthorityPayload = {
...
-	* Optional host↔remote workspace path mapping. The dev-container
-	* authority sets both roots (editor.getCwd() on host;
-	* remoteWorkspaceFolder on container) so LSP URIs translate at the
-	* host/container boundary. Local and SSH authorities omit it.
-	*/
-	path_translation?: PathTranslationSpec;
-};
-type PathTranslationSpec = {
-	host_root: string;
-	remote_root: string;
```
PavelLoparev: Regenerated with `cargo test -p fresh-plugin-runtime write_fresh_dts_file -- --ignored`. Not manual changes

sinelaw:

Tested interactively by building the PR branch, launching `fresh` in a tmux session with 3 files, and exercising the plugin via the command palette. Close actions (Close Other Tabs, Close All Tabs, Close Tabs To Left/Right) all registered and functioned correctly. Move Tab actions did **not** register at all — "Move Tab Left" and "Move Tab Right" are absent from the command palette.

Issues found:

- **[critical] Move Tab Left / Move Tab Right not functional** — the two commands never appear in the command palette. Likely cause: `move_tab_left` / `move_tab_right` in `quickjs_backend.rs` use the `#[plugin_api(js_name = ...)]` macro, but the four close methods that *do* work have no such macro. If `#[plugin_api]` only generates TypeScript type stubs rather than actual JS bindings, `editor.moveTabLeft()` would be `undefined` at runtime and registration would silently fail. Verify that the JS method is actually reachable via the QuickJS runtime.
- **[guideline 2 violation] E2E tests access internal model state** — `harness.editor().get_active_split()` and `harness.editor().get_split_tabs(split_id)` inspect model state directly. Per CONTRIBUTING.md, E2E tests must assert only on rendered output; invariants not visible on screen should be covered by unit tests.
- **[guideline 3 violation] E2E test uses `std::thread::sleep`** — `tab_actions_harness()` loops with `sleep(Duration::from_millis(50))`. Per CONTRIBUTING.md, use semantic waiting instead of fixed timers.
- **[wrong reference path] `tab_actions.ts`** — uses `/// <reference path="../../types/fresh.d.ts" />` but every other plugin in the same directory uses `/// <reference path="./lib/fresh.d.ts" />`. This breaks `check-types.sh`.
- **[unrelated change] `fresh.d.ts` removes `PathTranslationSpec`** — deletion of `PathTranslationSpec` type and `path_translation` field from `AuthorityPayload` is unrelated to tab actions and should be in a separate PR.
- **[minor] Unused import** — `use tempfile::TempDir;` in `tab_actions.rs` (compiler warns).
- **[minor] Missing doc comments** on new `PluginCommand` variants `CloseOtherBuffersInSplit`, `CloseAllBuffersInSplit`, `CloseBuffersToRightInSplit`, `CloseBuffersToLeftInSplit`.


---

sinelaw (changes requested):
Requesting changes: Move Tab Left/Right commands are non-functional (not registered at runtime), E2E tests violate CONTRIBUTING.md guidelines (sleep-based waiting and direct model state access), wrong TypeScript reference path, and an unrelated `PathTranslationSpec` removal. See inline comment for full details.

---

PavelLoparev:
@sinelaw fixed everything from above. not sure about missing move actions from UI - on videos above they are visible but then after syncing master and rebasing to this branch it stopped working. I just renamed action handlers so that do not match actions names from plugin api and it worked but I'm not sure about the root cause - probably something changed in the core and there was naming conflict somewhere. works fine now (at least on my local build).

sinelaw:
I'll rerun the review bot, let's see if it's still an issue

PavelLoparev:
more over, tests were passing too! it's interesting, in tests there are assertions that command palette DOES contain move action. so there is a difference between test env and prod env in runtime. I wasn't able to dig down to the root cause. just renamed.

PavelLoparev:
aah, let me fix `test_plugin_i18n_files_have_matching_keys`. never ending stuff...

sinelaw (changes requested):

Tested interactively on the PR branch (debug build, tmux session, 3 files open):

- **All 6 commands appear in the command palette** and execute correctly: Close Other Tabs, Close All Tabs, Close Tabs To Left, Close Tabs To Right, Move Tab To Left, Move Tab To Right — all listed under `tab_actions`
- **Close Other Tabs** ✓ — executed; only the active tab remained
- **Move Tab To Left** ✓ — executed; active tab shifted one position left
- **Move Tab To Right** ✓ — executed; active tab shifted one position right
- **Stale `fresh.d.ts` [blocking]** — `fresh.d.ts` declares `moveTabLeft()` / `moveTabRight()` but the Rust methods are now `move_tab_to_left` / `move_tab_to_right`, so the macro generates JS binding names `moveTabToLeft` / `moveTabToRight`. `tab_actions.ts` correctly calls `editor.moveTabToLeft()` at runtime, but `check-types.sh` fails with `TS2551: Property 'moveTabToLeft' does not exist on type 'EditorAPI'. Did you mean 'moveTabLeft'?`. Fix: re-run `cargo test -p fresh-plugin-runtime write_fresh_dts_file -- --ignored` after the rename so the `.d.ts` reflects the current Rust method names.

---

PavelLoparev:
@sinelaw fixed i18n keys issue + fresh.d.ts stale types.

PavelLoparev:
@sinelaw rebased from master to make it mergable again. let's merge?

PavelLoparev:
I think this plugin must be in this repo (core). These are basic tab actions.

sinelaw:
Thanks! As a follow up we should wire these commands to the tab context menu (don't remember if there's a plugin api for that yet)

PavelLoparev:
> Thanks! As a follow up we should wire these commands to the tab context menu (don't remember if there's a plugin api for that yet)

haha, I didn't even know there is such thing as tab context menu, I don't use mouse in fresh 😃

PR #1855, opened 2026-05-03
fix(prompt): re-derive bottom-row layout after in-render plugin drain (#1847)

### LINKED ISSUE #1847: "Save theme as:" text box doesn't show up if certain actions are done
While testing stuff for #1846, I noticed that "save theme as:" text box doesn't show up after if you have changed eg. color of some element.

See the bottom of the terminal window:

Image

It shows up if mouse cursor hovers over that area.

Terminal: xfce4-terminal 1.1.3.

sinelaw:
## Summary

Fixes #1847: when triggering "Save theme as" on a builtin theme, the prompt input row sometimes failed to render and the suggestions popup ate the bottom row.

## Root cause

`render()` already drains pending plugin commands once, after firing the `lines_changed` hooks, so the renderer can pick up overlay/conceal responses without a one-frame lag. That same drain, however, also processes any other commands the plugin happens to have in the queue — including layout-affecting ones like `StartPrompt` / `SetPromptSuggestions`.

The bottom-row chunks (`status_bar_idx`, `search_options_idx`, `prompt_line_idx`) are computed at the top of `render()` using the pre-drain `self.prompt`. When the drain flips `self.prompt` from `None` to `Some(prompt with suggestions)`, `main_chunks[prompt_line_idx]` keeps its pre-drain geometry: height 0 with y past the visible area. Two visible artifacts followed:
- The prompt input row tried to render to an off-screen Rect, so it was clipped entirely.
- The suggestions popup, anchored to `prompt_area.y` via `prompt_area.y - height`, landed one row too low and overpainted the row that should have hosted the prompt.

The `prompt` value at line 640 was already taken after the drain, so the inputs to the bottom-row renderers were inconsistent: prompt content was current, but layout was stale.

## Fix

Immediately after the in-render plugin-command drain, re-derive `prompt_is_overlay`, `has_suggestions`, `has_file_browser`, `show_search_options`, and `main_chunks`. Every later bottom-row renderer (status bar, search options, prompt input, suggestions popup) then sees chunk geometry that agrees with the post-drain `self.prompt`.

## Manual validation

Repeated `Ctrl+S` on a builtin theme after editing a color in a 120×40 xfce4-terminal:
- **Without fix:** ~25% of trials showed the bug (popup occupies bottom 3 rows, "Save theme as:" missing).
- **With fix:** 0/50 trials showed the bug.

## Test plan
- [x] `cargo test -p fresh-editor --test e2e_tests -- e2e::plugins::theme_editor` (32 tests pass)
- [x] `test_builtin_theme_requires_save_as` extended to assert the prompt input row is at the bottom of the screen while the suggestions popup is shown
- [x] Manual run: theme editor → edit color → Ctrl+S, repeated 50× without bug

https://claude.ai/code/session_01Qxf19fkKfe7FwLiTGRSR2H

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qxf19fkKfe7FwLiTGRSR2H)_

### DIFF
--- a/crates/fresh-editor/src/app/render.rs
+++ b/crates/fresh-editor/src/app/render.rs
@@ -415,6 +415,64 @@ impl Editor {
             self.flush_pending_grammars();
         }
 
+        // The plugin-command drain above can mutate `self.prompt`
+        // (e.g. a plugin's `lines_changed` handler kicks off a Save-As
+        // prompt via `editor.startPrompt` + `editor.setPromptSuggestions`).
+        // The bottom-row chunks (`status_bar_idx`, `search_options_idx`,
+        // `prompt_line_idx`) computed at the top of `render()` would then
+        // reflect the pre-drain state — typically `prompt` height 0 and
+        // y past the visible area — so the prompt input row wouldn't
+        // render and the suggestions popup, anchored to `prompt_area.y`,
+        // would land one row too low (clipping the prompt row entirely).
+        // Re-derive the layout-affecting flags and `main_chunks` here so
+        // every later renderer sees a chunk geometry consistent with the
+        // post-drain `self.prompt`. See issue #1847.
+        let prompt_is_overlay = self.prompt.as_ref().is_some_and(|p| p.overlay);
+        let has_suggestions = self
+            .prompt
+            .as_ref()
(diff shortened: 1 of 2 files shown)


sinelaw:
obsoleted by #1800

PR #1860, opened 2026-05-03
perf: populate profile.dist for smaller release binaries

PanAchy:
## Summary

The `[profile.dist]` section in `Cargo.toml` exists but is empty, just `inherits = "release"`. This populates it with profile flags that reduce the release binary from 62.4 MB to 40.7 MB (-34.7%), with the xz-compressed tarball going from ~12.4 MB to ~7.1 MB (-43%).

Dev builds (`--release`) stay unchanged for fast iteration.

Also, thanks for making Fresh! Been using it daily and really enjoying it. Especially since I didn't get around to learning NeoVim yet :laughing: 

## Changes

`Cargo.toml`: Populate `[profile.dist]` with `strip = true`, `lto = "fat"`, `codegen-units = 1`, `panic = "abort"`, `opt-level = "z"`, `force-frame-pointers = false`.

## Individual knob decomposition

Each knob was measured independently against the baseline (62,386 KB):

| Knob | Savings |
|------|---------|
| `strip = true` | -7,507 KB (12%) |
| `panic = "abort"` | -7,620 KB (12.2%) |
| `lto = "fat"` | -6,808 KB (10.9%) |
| `opt-level = "z"` | -1,787 KB (2.9%) |
| **All combined** | **-21,660 KB (34.7%)** |

No single knob dominates, the gains are roughly additive.

## Note on release workflow

The release workflows still use `--release`. Switching to `--profile dist` requires updating the build command and all artifact path references from `target/<triple>/release/` to `target/<triple>/dist/`, plus removing the now-redundant manual `strip` step in `linux-packages.yml`. I left that out of scope since you are better positioned to decide how to handle `cargo-deb`/`cargo-generate-rpm` path assumptions.

## Build time impact

Baseline `--release`: ~310s
`--profile dist` (full combo): ~370s (+19%)
`strip + panic` only (if build time is a concern): ~267s, actually faster than baseline, 21% smaller

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,6 +78,12 @@ force-frame-pointers = true
 # The profile that 'dist' will build with
 [profile.dist]
 inherits = "release"
+strip = true
+lto = "fat"
+codegen-units = 1
+force-frame-pointers = false
+panic = "abort"
+opt-level = "z"
 
 # Optimize the critical dependencies in the plugin-load path at -O3 even
 # in dev/test builds. These are the only deps where unoptimized code
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,11 +78,8 @@ force-frame-pointers = true
 # The profile that 'dist' will build with
 [profile.dist]
 inherits = "release"
-strip = true
 lto = "fat"
 codegen-units = 1
-force-frame-pointers = false
-panic = "abort"
 opt-level = "z"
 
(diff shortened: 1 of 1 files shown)


sinelaw:
Main downside (for me) is that user panics won't include backtrace information making it *much* harder to debug and fix any panics.

sinelaw:
Thanks! Merged with tweaks to keep call stack in panic

PanAchy:
Awesome!

PR #1867, opened 2026-05-04
fix(cursor): match secondary cursor shape to configured cursor style (#620)

### LINKED ISSUE #620: multi select not consist with cursor
I use ctrl + d to select, but they are block and blicking bar, not consist.

Image

sinelaw:
## Summary

Closes #620. Multi-cursor rendering always used `Modifier::REVERSED` for secondary cursor cells, so they appeared as block highlights even when the user configured a non-block primary cursor (bar or underline). The terminal's hardware cursor still drew the primary in the configured shape, so multi-cursor selections looked inconsistent.

This change plumbs a `secondary_cursor_modifier` through the rendering pipeline, derived once at the top level from `editor.cursor_style`:

- Block / Default → `REVERSED` (unchanged — matches the terminal's block shape)
- Bar / Underline → `UNDERLINED` (closest "thin marker" available in a TUI cell, so secondaries visually approximate the bar/underline shape the terminal draws for the primary)

A TUI cell can't render a true vertical bar without breaking layout, so bar-style cursors fall back to `UNDERLINED` (a thin horizontal marker) — visually closer to the user's configured shape than a full REVERSED block.

In software-cursor-only mode (GUI backend) the primary cell is already forced to `REVERSED` downstream, so secondaries also fall back to `REVERSED` to stay consistent with the primary.

## Reproduction

Launch with a non-block cursor style and add a secondary cursor:

```bash
cat > /tmp/cfg.json <<'EOF'
{ "editor": { "cursor_style": "blinking_bar" } }
EOF
fresh --config /tmp/cfg.json some-file.txt
# press Ctrl+D twice on a word with multiple matches
```

**Before:** primary appears as a thin bar (terminal hardware cursor), secondary appears as a full REVERSED block — inconsistent.
**After:** primary still bar, secondary uses UNDERLINED — both are thin markers.

## Test plan

- [x] New unit test `test_cursor_style_secondary_modifier_matches_shape` — verifies `CursorStyle::secondary_cursor_modifier()` returns `REVERSED` for block-like styles and `UNDERLINED` for bar/underline styles.
- [x] New e2e test `test_secondary_cursor_style_matches_configured_cursor_shape` — drives multi-cursor with `BlinkingBar` and `SteadyUnderline` config, asserts secondary cursor cells carry `UNDERLINED` rather than `REVERSED`.
- [x] All 35 existing multicursor e2e tests still pass.
- [x] Manual validation in tmux for `blinking_bar`, `steady_underline`, `blinking_block`, default — block-style behaviour unchanged, non-block now uses `UNDERLINED` for secondaries.
- [x] `cargo check --all-targets`, `cargo check --all-targets --features gui`, `cargo fmt`, `cargo clippy -p fresh-editor`.

https://claude.ai/code/session_012VjBvHX3roxbbSdBpBtSi7

---
_Generated by [Claude Code](https://claude.ai/code/session_012VjBvHX3roxbbSdBpBtSi7)_


sinelaw:
Doesn't really model the same cursor shape (it's not possible in normal terminals)

PR #1880, opened 2026-05-05
Feature: Global Panel UI that support anchor to edges

bluzky:
## Summary
  - Add showGlobalPanel / updateGlobalPanel / closeGlobalPanel to the plugin  API, allowing plugins to display a display-only overlay panel anchored to any  edge of the screen (Top, Bottom, Left, Right; default Bottom).
  - Panels are managed by a new GlobalPanelManager in editor state and rendered   as the topmost overlay layer, above content but below the menu bar.
  - Panel positioning respects existing chrome: the menu bar and tab bar at the   top, the status bar at the bottom. Left/Right panels span the full content  height.
  - Key handling is intentionally left to defineMode — panels carry no  interactive model of their own.

  **Test plan**

  - E2E tests in tests/e2e/global_panel.rs cover: open/close, all four anchor  positions, row rendering, and frame-clamp safety (zero-size guard).
  - Manual: open a panel from a plugin via editor.showGlobalPanel(...), verify  it appears at the correct edge and disappears on  editor.closeGlobalPanel(...).
  - Manual: resize the terminal while a panel is open; verify no ratatui panic  and correct reflow.

## Test UI from my WIP plugin for transient menu
- Stick to bottom
image

- Stick to righ edge
image

### DIFF
--- a/crates/fresh-editor/src/app/file_open_input.rs
+++ b/crates/fresh-editor/src/app/file_open_input.rs
@@ -800,6 +800,23 @@ impl Editor {
 
         // Double-click in file list opens/navigates
         if layout.is_in_list(x, y) {
+            // In Switch Project (folder-only) mode, double-clicking a directory
+            // entry should navigate INTO it like a file manager would, rather
+            // than immediately selecting it as the new project root. Selecting
+            // a project root remains available via Enter (or by typing a path)
+            // so keyboard users keep one-keypress confirmation.
+            if self.is_folder_open_mode() {
+                let selected_dir = self.active_window().file_open_state.as_ref().and_then(|s| {
+                    s.selected_index
+                        .and_then(|idx| s.entries.get(idx))
+                        .filter(|e| e.fs_entry.is_dir())
+                        .map(|e| e.fs_entry.path.clone())
+                });
+                if let Some(path) = selected_dir {
+                    self.file_open_navigate_to(path);
+                    return true;
+                }
+            }
             self.file_open_confirm();
             return true;
         }
--- a/crates/fresh-editor/tests/e2e/open_folder.rs
+++ b/crates/fresh-editor/tests/e2e/open_folder.rs
@@ -4,7 +4,7 @@
 //! the command palette or File menu.
 
 use crate::common::harness::EditorTestHarness;
-use crossterm::event::{KeyCode, KeyModifiers};
+use crossterm::event::{KeyCode, KeyModifiers, MouseButton, MouseEvent, MouseEventKind};
 use std::fs;
 use tempfile::TempDir;
 use tracing_subscriber::{fmt, prelude::*, EnvFilter};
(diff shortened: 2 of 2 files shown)


PavelLoparev:
Just curious what is this for? You can do exactly the same actions from the command palette. Is this just not to type command name and use single letter?

PavelLoparev:
For managing tabs in splits I introduced this PR: https://github.com/sinelaw/fresh/pull/1849. There are some useful tab actions like close, close others, close to left/right etc. I hope it will be included to the core.

bluzky:
> Just curious what is this for? You can do exactly the same actions from the command palette. Is this just not to type command name and use single letter?

It's which-key menu, which help to quick run command with your own keybinding instead of typing to much with command pallet.
And it's helpful because not you may not remember all your key short cut, except when you use it for too long

PavelLoparev:
> > Just curious what is this for? You can do exactly the same actions from the command palette. Is this just not to type command name and use single letter?
> 
> It's which-key menu, which help to quick run command with your own keybinding instead of typing to much with command pallet. And it's helpful because not you may not remember all your key short cut, except when you use it for too long

API for such panels might be useful indeed, but I still don't get your particular use case. There is a help page with all the shortcuts currently setup in the editor so you can find your needed one there too. In your case it's more like "one key" vs "multiple keys", but you still need to:
1. Open your panel always.
2. Remember your key for action.
Bu you already can just remember needed shortcuts and it will be faster than open global panel + hit one button.

Sorry. I'm not a maintainer and not judging/criticizing in any way. Just thinking outloud. You may ignore my comments :)

sinelaw:
Just to understand - why not use the split system for this? Because you want it to float above the existing UI without upsetting the buffer width/height?

bluzky:
> Just to understand - why not use the split system for this? Because you want it to float above the existing UI without upsetting the buffer width/height?

Let me try with split system. I never think of that use of split system. Thank you for your suggestion.

bluzky:
Hi @sinelaw, 
I've tried with split but I encountered a problem when I switch to new split, then cursor and focus is now in new split, is there anyway to interact with previous split?
And I think with a global panel, it'll not bother with current splits/editor state.

bluzky:
> > > Just curious what is this for? You can do exactly the same actions from the command palette. Is this just not to type command name and use single letter?
> > 
> > 
> > It's which-key menu, which help to quick run command with your own keybinding instead of typing to much with command pallet. And it's helpful because not you may not remember all your key short cut, except when you use it for too long
> 
> API for such panels might be useful indeed, but I still don't get your particular use case. There is a help page with all the shortcuts currently setup in the editor so you can find your needed one there too. In your case it's more like "one key" vs "multiple keys", but you still need to:
> 
>     1. Open your panel always.
> 
>     2. Remember your key for action.
>        Bu you already can just remember needed shortcuts and it will be faster than open global panel + hit one button.
> 
> 
> Sorry. I'm not a maintainer and not judging/criticizing in any way. Just thinking outloud. You may ignore my comments :)

I live in Emacs world so that is like a primitive :)

sinelaw:
> Hi @sinelaw, I've tried with split but I encountered a problem when I switch to new split, then cursor and focus is now in new split, is there anyway to interact with previous split? And I think with a global panel, it'll not bother with current splits/editor state.

What's the exact behavior you want?  *Why* do you need to switch to the new split, if you want focus to remain on the old one?

bluzky:
> > Hi @sinelaw, I've tried with split but I encountered a problem when I switch to new split, then cursor and focus is now in new split, is there anyway to interact with previous split? And I think with a global panel, it'll not bother with current splits/editor state.
> 
> What's the exact behavior you want? _Why_ do you need to switch to the new split, if you want focus to remain on the old one?

I don't know if my usecase is special. When the plugin open new panel it expect to receive key event from user, update UI based on user select, then decide which action to apply to editor, or current active split.

I'll close this request if it's not generic enough.

sinelaw:
Your use case makes perfect sense. I'm just trying to understand what feature is missing from the split system to get what you want. As a plug-in, you should be able to capture keys regardless of which buffer is focused. For example, see the VI mode plug-in

Issue #1845, opened 2026-05-02
Add symbols list LSP action


PavelLoparev:
How it works:
* Trigger action `Open LSP symbols list`
* List looks like:
```
[class] MyClass
  [method] myMethod
    [variable] myVariable
[class] MyAnotherClass
...
```
* When you navigate the list item then:
  * Cursor moves to the given symbol start position
  * The whole symbols block/line gets selected

Basically just like it's done in VScode (`ctrl + shift + o`)

I tried to implement it by myself but got lost pretty fast. It requires LSP knowledge but I don't have enough time to take a look at this protocol. I mean fresh already contains some code around running LSP commands but my first attempt just failed - I was not able to get list of symbols :(

PavelLoparev:
I figured it out. Will do a PR soon.

sinelaw:
Closing as completed: implemented by PR #1886 (merged 2026-05-31) — the `lsp_navigation` plugin provides a "Go to LSP Symbol" command (`goto_lsp_symbol` in `crates/fresh-editor/plugins/lsp_navigation.ts`) that lists document symbols with their kinds and navigates the cursor to the selected symbol, VS Code `Ctrl+Shift+O`-style.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1886, opened 2026-05-05
feat: LSP go to Symbol navigation

PavelLoparev:
# Summary
See https://github.com/sinelaw/fresh/issues/1845. Works just like in VScode.

No tests yet.

https://github.com/user-attachments/assets/41305406-96cf-4fbd-845c-6b4ff55ac467


crates/fresh-editor/plugins/lsp_navigation.ts
```diff
@@ -0,0 +1,261 @@
...
+
+  editor.setBufferCursor(bufferId, bytePos);
+
+  const lineCount = sym.endLine - sym.startLine + 1;
+
+  if (lineCount > 1) {
+    editor.executeActions([
+      { action: "select_line", count: 1 },
+      { action: "select_down", count: lineCount - 2 },
+      { action: "select_line_end", count: 1 },
```
PavelLoparev: Select symbol line start line
select down lines end - 1
select last line to the end (we do not want cursor to be after last line, just to be on the last line)

sinelaw:

Tested interactively (tmux + fresh binary built from this branch): the "Go to LSP Symbol" command appears correctly in the command palette and error handling fires as expected when no LSP server is running. The core navigation logic looks solid. However several issues need addressing before merge:

**Critical**
- `sendLspRequest("typescript", ...)` is hardcoded — the plugin only works on TypeScript files. Every other LSP-enabled language (Rust, Go, Python, …) silently fails. Fix: `editor.getBufferInfo(bufferId)?.language` gives the buffer's detected language dynamically.

**Architectural / plugin vs core**
- This plugin reimplements the prompt lifecycle (`startPrompt` / `setPromptSuggestions` / manual `prompt_changed` / `prompt_confirmed` / `prompt_selection_changed` handlers) that `lib/finder.ts` already encapsulates. `find_references.ts` and `live_grep.ts` both use `Finder` with `mode: "filter"` — this should too, for consistent UX (floating overlay, preview, utility dock, fuzzy filter).
- `textDocument/documentSymbol` is a language-agnostic LSP method, the same way go-to-definition is. Consider whether this belongs in core (like existing LSP navigation) rather than as a separate plugin. If it stays a plugin, it must be language-agnostic (see above).

**Integration / correctness**
- `lsp_navigation.ts` is missing from `tsconfig.json` → skipped by `check-types.sh`.
- Reference path `../../types/fresh.d.ts` is wrong for a bundled plugin; should be `./lib/fresh.d.ts` (the path used by every other plugin in this directory).
- `DocumentSymbol` response is hierarchical — children are not recursively traversed, so methods inside a class are silently dropped.
- `container` field is parsed but never surfaced in the UI — either display it or drop it.
- `editor.setStatus("LSP navigation plugin loaded")` overwrites other plugins' status messages on startup; use `editor.debug(...)` instead.

**Tests**
- CONTRIBUTING.md requires a test that fails without the change and passes with it — none provided.

---

sinelaw (changes requested):
Requesting changes — see inline comment for the full list of issues found during review and interactive testing.

---

PavelLoparev:
Don't know how I overlooked hardcoded language. all makes sense. will fix all above

PavelLoparev:
> hardcoded language in lsp request

Fixed. Language is fetched from the current buffer.

> lsp_navigation.ts is missing from tsconfig.json → skipped by check-types.sh.

Added.

> DocumentSymbol response is hierarchical — children are not recursively traversed, so methods inside a class are silently dropped.

This is false positive. Take a look:

https://github.com/user-attachments/assets/f657e99c-1f4c-4755-8b1a-32fbc49544ba

> container field is parsed but never surfaced in the UI — either display it or drop it.

Removed completely.

> editor.setStatus("LSP navigation plugin loaded") overwrites other plugins' status messages on startup; use editor.debug(...) instead.

Changed to debug message.

PavelLoparev:
Reimplemented using finder logic. added test. added i18n support

PavelLoparev:
Also, on videos you see a lag between selecting list item and selecting lines in editor - that was optimised too (cache current buffer id on prompt open and reuse instead of fetching it from editor object every time we need it). Now it realtimish.

sinelaw:

Re-reviewed after the latest 5 commits. Great progress — Finder abstraction, dynamic language, correct reference path, tsconfig, debug message, removed container field, and a real E2E test are all solid improvements. A few things still need fixing before merge:

**Must fix**
- `lsp_navigation.ts` lines 3–11: `class MyClass { constructor()... myMethod()... }` is leftover test/debug code sitting in the production plugin file, with an `import` statement below it. Must be removed.
- `loadSymbols` catches all exceptions and silently returns `[]` — when the LSP server is not running, the prompt opens showing "No matches" with no explanation. Verified interactively: the user has no idea what went wrong. Add an `editor.setStatus(...)` call in the catch block (matching the error handling pattern from the v1 code or from `openSymbolsListHandler` in `find_references.ts`).

**Should fix**
- `loadSymbols` navigates to the first symbol as a side-effect of loading (`await navigateToSymbol(cachedBufferId, symbols[0])`). This jumps the cursor before the user has selected anything.
- The `filter` callback also navigates as a side-effect (`navigateToSymbol(cachedBufferId, filtered[0])`). `FilterSource.filter` is documented as a pure transform; navigation on each filter call can double-fire alongside `onSelectionChanged` on every keystroke. Move navigation exclusively to `onSelectionChanged`.
- `cachedLanguage` is typed `string` but `editor.getBufferInfo(...)?.language` returns `string | undefined` — this is a type error that `check-types.sh` should catch.

**Minor**
- `lsp_navigation.i18n.json`: all non-English locales contain English strings. CONTRIBUTING.md says "Don't commit English placeholders."

---

sinelaw (changes requested):
Requesting changes — see comment for details. Main blockers: leftover MyClass test class in the plugin file, silent error swallowing in loadSymbols, and navigation side-effects in the filter callback.

---

PavelLoparev:
@sinelaw navigation side effect is needed. when user types for filtering and list is being updated finder doesn't fire any events. thus user doesn't see updated selection (suggestions list is updated, new "first" item is in the list but cursor is not moved to that symbol)

PavelLoparev:
@sinelaw i18n and cached language type issues BUT didn't fix others. Please review manualy - those "side effects" are WANTED.

PavelLoparev:
> loadSymbols navigates to the first symbol as a side-effect of loading (await navigateToSymbol(cachedBufferId, symbols[0])). This jumps the cursor before the user has selected anything.

Wanted behaviour.

> The filter callback also navigates as a side-effect (navigateToSymbol(cachedBufferId, filtered[0])). FilterSource.filter is documented as a pure transform; navigation on each filter call can double-fire alongside onSelectionChanged on every keystroke. Move navigation exclusively to onSelectionChanged.

Wanted behaviour - I really want it select lsp symbol as you type.

PavelLoparev:
@sinelaw you know what, I agree that selecting on load is redundant and might be not wanted by users (cursor was in the middle of a file and after prompt opening became in the beginning) + I can live without "live selection" on filtering.

additionally, fixed tests - assert rendered output instead of model's states.

PavelLoparev:
I think this one should be in the core too but if you want it to be a separate plugin - tell me, will do.

PavelLoparev:
manually tested after rebase. works as expected.

sinelaw:
A few issues from manual testing:
1. cursor position is lost if you cancel the popup: open the command,then move up/down to see the live preview move around, then press escape - it leaves the cursor in the last visited position
2. also the preview positions get added to the navigation history - after using the popup if I use "navigate back" command it goes through all the preview positions instead of skipping to the last actual cursor position
3. when pressing enter to visit, it jumps to the end of a line instead of the symbol's name
(this was tested in a python file with pylsp)

PavelLoparev:
> A few issues from manual testing:
> 
> 1. cursor position is lost if you cancel the popup: open the command,then move up/down to see the live preview move around, then press escape - it leaves the cursor in the last visited position
> 2. also the preview positions get added to the navigation history - after using the popup if I use "navigate back" command it goes through all the preview positions instead of skipping to the last actual cursor position
> 3. when pressing enter to visit, it jumps to the end of a line instead of the symbol's name
>    (this was tested in a python file with pylsp)

1. Added `onClose` invoke to `finder` - now cursor is returned after cancelling prompt.
2. We now do not call execute actions so actions history is not polluted with lsp symbols preview.
3. Now it jumps to the lsp symbol start position after selecting.

Trade off: we've lost lsp symbol visual selection on preview because this is what was causing actions history pollution. It's possible to add selection but it requires rust backend changes which I do not want to do. Simple plugin is good enough and it's not required to copy VScode functionality 1 to 1 (visual selection looks really clear and provides more visual context though). What do you think?

PavelLoparev:
just and an option - for visual "selection" preview we could just use native `preview: true` feature (patched to show `>` for lines range). Works just fine and also uses native approach for previewing things.

So, two options, both are free from issues you described earlier.

Option 1 (no native preview, just moving cursor around):

https://github.com/user-attachments/assets/ad5cf3cd-088e-4f56-a02f-b2c6d3361f3c

Option 2 (native preview split panel):

https://github.com/user-attachments/assets/17eb7062-b0a0-417d-ab56-ea3f2a9cbade

I personally like cursor jumps in the current split/tab though but it's up to you to decide :)

p.s. (latest 2 commits contains code for option 2)

PavelLoparev:
hey @sinelaw, I lived some time with both options and I think option 1 is better though. it's just visually more clear where exactly lsp symbol is

PavelLoparev:
Hi @sinelaw, sorry to ping you again. did you have a chance to take a look here? I know it could be a separate plugin but depending on which way you prefer it might require changes to the existing ts/rust core (option #2 as it's implemented now).

sinelaw:
Option 1 looks good, let's go with that

PavelLoparev:
> Option 1 looks good, let's go with that

Ok. Here we go.

1. Rebased with fresh master
2. Restored option 1 behaviour
3. Additionally, I added logic that preselects lsp symbol item from the prompt list (lsp symbol under current cursor position/line). Default to the first lsp symbol list item if current cursor position/line has no lsp symbol detected.

PavelLoparev:
Fixed failing tests.

sinelaw:
Thanks! Needs a bit more polish, will do it after merging

sinelaw:
This unrelated test is flaky, fixing separately: test_file_explorer_git_markers_clear_after_external_commit

PavelLoparev:
great, thank you!

yea, I saw this test fails on another PR too.

Issue #1915, opened 2026-05-09
Rust code actions return "No code actions available" — empty CodeActionClientCapabilities causes rust-analyzer to return null


sinelaw:
## Description

Reported by `cybin` on Discord:

> Code actions for Rust does not work for me. Whenever I try Alt+. I get a 'No code actions available' in the status bar. LSP is On and language is correctly detected as Rust. Tried by forcing a missing field error, which should offer an action to initialize the missing field.

Reproduced on `claude/reproduce-rust-issues-2k14m` (fresh 0.3.5, Linux, rust-analyzer 1.95.0).

## Steps to reproduce

1. `cargo init` a dummy project, paste this into `src/main.rs`:
   ```rust
   struct Point { x: i32, y: i32, z: i32 }
   fn main() {
       let p = Point { x: 1 };
       println!("{}", p.x);
   }
   ```
2. Open in fresh, enable rust-analyzer LSP for the buffer.
3. Wait for indexing to finish (status shows `LSP (on)` and the diagnostic `E:2`).
4. Place the cursor anywhere inside or adjacent to `Point { x: 1 }` (line 8, col 13 over `Point`, or col 25 inside the braces).
5. Press <kbd>Alt</kbd>+<kbd>.</kbd>.

**Result:** status bar shows `No code actions available`. Same outcome with selection covering the struct expression.

**Expected:** popup offering "Fill struct fields" (the rust-analyzer quickfix that inserts `y: todo!(), z: todo!()`).

## Comparison with neovim (same machine, same rust-analyzer)

Minimal `init.lua`:
```lua
vim.api.nvim_create_autocmd("FileType", {
  pattern = "rust",
  callback = function()
    vim.lsp.start({
      name = "rust-analyzer",
      cmd = {"rust-analyzer"},
      root_dir = "/tmp/rust-repro",
      capabilities = vim.lsp.protocol.make_client_capabilities(),
    })
  end,
})
```

`nvim -u init.lua src/main.rs` → `:lua vim.lsp.buf.code_action()` over the same expression → neovim immediately offers:

```
Code actions:
1: Fill struct fields
```

So rust-analyzer is producing the action; only fresh fails to receive it.

## Root cause

Captured the LSP traffic by wrapping rust-analyzer with `tee`. Fresh sends `textDocument/codeAction` correctly:

```json
{ "method": "textDocument/codeAction",
  "params": { "textDocument": {"uri": "..."},
[...]

PR #1918, opened 2026-05-09
Advertise codeActionLiteralSupport in LSP client capabilities

sinelaw:
## Summary
This PR adds support for advertising `codeActionLiteralSupport` in the LSP client capabilities during initialization. This fixes an issue where rust-analyzer and other LSP servers would return `null` for code action requests when the client didn't advertise this capability.

## Key Changes
- Added imports for `CodeActionKindLiteralSupport` and `CodeActionLiteralSupport` types
- Configured `code_action_literal_support` in the `CodeActionClientCapabilities` with support for common code action kinds:
  - Empty string (generic actions)
  - `quickfix`
  - `refactor` and its variants (`extract`, `inline`, `rewrite`)
  - `source` and `source.organizeImports`
- Added a test case `code_action_capability_advertises_literal_support()` to verify the capability is properly advertised with all required code action kinds

## Implementation Details
The issue occurred because rust-analyzer cannot represent `WorkspaceEdit`-based assists (like "Fill struct fields") as the `Command`-only fallback when `codeActionLiteralSupport` is not advertised. By explicitly advertising this capability with the supported code action kinds, the server can now properly return these assists instead of returning `null`.

This resolves the "No code actions available" issue users were experiencing for Rust quickfixes (sinelaw/fresh#1915).

https://claude.ai/code/session_014f8p77TuSzTfseyUrwrkLY

### DIFF
--- a/crates/fresh-editor/src/services/lsp/async_handler.rs
+++ b/crates/fresh-editor/src/services/lsp/async_handler.rs
@@ -261,8 +261,9 @@ impl LspClientState {
 /// Create common LSP client capabilities with workDoneProgress support
 fn create_client_capabilities() -> ClientCapabilities {
     use lsp_types::{
-        CodeActionClientCapabilities, CompletionClientCapabilities, DiagnosticClientCapabilities,
-        DiagnosticTag, DynamicRegistrationClientCapabilities, FoldingRangeCapability,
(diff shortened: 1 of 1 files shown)


sinelaw:
Fixes #1915

PR #1921, opened 2026-05-09
fix(render): keep cursor visible inside REVERSED selections (#1900)

### LINKED ISSUE #1900: Cursor + shift selection highlight is hard to see, reverse rendering effect rules itself out

sinelaw:
Fixes #1900.

## Problem

When the active theme uses `Modifier::REVERSED` for `selection_modifier` (the built-in `terminal` theme is the canonical example), the cursor visually disappears whenever it lands on a selected cell — e.g. after `Shift+Left` lands the caret on the first letter of a word.

Selection adds `REVERSED` to the cell, then the cursor logic adds `REVERSED` again. Because `Style::add_modifier` is bitwise OR, adding the same bit twice is a no-op, and the cursor cell ends up indistinguishable from any other selected cell. The "reverse rendering effect rules itself out", as the issue puts it.

## Fix

Toggle the `REVERSED` modifier on the cursor cell instead of unconditionally adding it. When the cell is already `REVERSED` (from a `REVERSED` selection), the cursor cell drops the bit and visibly punches a hole in the surrounding selection block — both for the software cursor (secondary cursors / `software_cursor_only` mode) and for the cell underneath the terminal's hardware cursor.

The change is local to `compute_char_style` and only affects cells where `is_cursor` is true. Themes whose `selection_modifier` is empty (dark/light/etc.) are unaffected — the cursor still gets `REVERSED` added.

## Tests

Three unit tests in `char_style.rs`:
- `cursor_on_reversed_selection_is_visually_distinct` (regression test for #1900)
- `cursor_without_selection_still_uses_reversed`
- `cursor_on_non_reversed_selection_keeps_reversed`

The regression test fails on `master` (cursor cell == selection cell, both `add: REVERSED, sub: NONE`) and passes after the fix.

## Manual validation

Built `target/debug/fresh`, configured `theme: "terminal"`, opened a file in tmux and exercised both `Shift+Left` (selects backwards, cursor lands inside selection) and `Shift+Right` (selects forwards, cursor lands past selection). With the fix:

- Selected non-cursor cells render with SGR `[7m` (REVERSED), as before.
- The cursor cell renders without `REVERSED` — e.g. `[1;4;7m curso [0;1;4m r` — so the underlying syntax/semantic style shows through and the hardware cursor block on top is unambiguous.

## Test plan
- [x] `cargo test -p fresh-editor --lib` (2430 passed, 4 ignored)
- [x] `cargo fmt --check`
- [x] `cargo clippy -p fresh-editor` — no new warnings on touched file
- [x] Manual tmux validation on `terminal` theme (selection forwards + backwards)

https://claude.ai/code/session_01HSe6y1vfNH9B6vSHX2d8JT

---
_Generated by [Claude Code](https://claude.ai/code/session_01HSe6y1vfNH9B6vSHX2d8JT)_

### DIFF
--- a/crates/fresh-editor/src/view/ui/split_rendering/char_style.rs
+++ b/crates/fresh-editor/src/view/ui/split_rendering/char_style.rs
@@ -224,12 +224,28 @@ pub(super) fn compute_char_style(ctx: &CharStyleContext) -> CharStyleOutput {
     let is_secondary_cursor = ctx.is_cursor && ctx.byte_pos != Some(ctx.primary_cursor_position);
     if ctx.is_active {
         if ctx.is_cursor {
-            if ctx.skip_primary_cursor_reverse {
-                if is_secondary_cursor {
+            // Toggle REVERSED rather than always adding it. Without this, a
+            // cursor that lands on a cell already styled with `REVERSED`
+            // (e.g. a selection on a theme like `terminal` whose
+            // `selection_modifier` is `Modifier::REVERSED`) becomes
+            // indistinguishable from its surroundings, since `add_modifier`
+            // is bitwise OR — see #1900.
+            let toggle_reverse = if ctx.skip_primary_cursor_reverse {
+                // Hardware cursor draws over the primary cursor cell; only
+                // secondary cursors need a software REVERSED marker. We still
+                // toggle on the primary cell so that, when the surrounding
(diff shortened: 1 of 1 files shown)


sinelaw:
Not working as expected

Issue #1554, opened 2026-04-12
Issue while setting scrollbar track color


klonuo:
Hi,

if I set theme setting `"scrollbar_track_fg": [255, 0, 0]`, it has no effect.

According Devin:

```
In src/view/ui/split_rendering.rs, the scrollbar track colors are hardcoded based on whether the split is active split_rendering.rs:849-859 :

- Active split: Color::DarkGray
- Inactive split: Color::Black

This code completely ignores the theme.scrollbar_track_fg value you set.
```

If so please allow setting the color for scrollbar tracker.

sinelaw:
Fixed in https://github.com/sinelaw/fresh/pull/1638

PR #1933, opened 2026-05-10
fix(theme): honor scrollbar_track_fg and hover overrides from user themes (#1554)

sinelaw:
resolve_theme_key{,_mut} were missing match arms for ui.scrollbar_track_fg,
ui.scrollbar_track_hover_fg, and ui.scrollbar_thumb_hover_fg. The override
walk in apply_theme_overrides routes every user-JSON leaf through
resolve_theme_key_mut, so user themes setting these three keys had their
overrides silently dropped — only the inherited base value (e.g. DarkGray)
reached the renderer. The split-rendering scrollbar code already reads
theme.scrollbar_track_fg, so the renderer side was fine; just the resolver
table needed the three missing arms.

Reproducer: create ~/.config/fresh/themes/redbar/theme.json (extending dark)
with ui.scrollbar_track_fg=[255,0,0] and ui.scrollbar_thumb_fg=[0,255,0].
Before the fix the track renders Color::DarkGray; after, it renders red.

https://claude.ai/code/session_01MuLjExrF5LUQNBoepKoMiq


sinelaw:
Fixed #1554 again

PR #1942, opened 2026-05-11
Remove slide-in animation from dashboard bringup

sinelaw:
## Summary
This PR removes the slide-in animation that previously played when the dashboard opened. The dashboard now appears immediately at its final position without any animation effect.

## Key Changes
- **Removed animation infrastructure**: Deleted `activeAnimationId` tracking variable and `resolveSlideFrom()` configuration resolver that controlled the slide direction
- **Removed structural change detection**: Eliminated `lastPaintedStructuralKey` and `CLOCK_RE` regex that were used to distinguish between clock ticks (no animation) and actual content changes (animate)
- **Simplified paint logic**: Removed the animation trigger condition that fired `editor.animateVirtualBuffer()` when structural content changed
- **Removed animation cancellation**: Deleted calls to `editor.cancelAnimation()` in buffer close and file open handlers
- **Added viewport padding**: Added bottom padding to the dashboard buffer to ensure it fills the full viewport height and prevents the `editor.after_eof_bg` color from showing below the content
- **Updated tests**: Simplified the dashboard bringup test to just verify the banner renders without waiting for animation completion
- **Updated documentation**: Modified config schema description to remove mention of "dashboard bringup" from the animations feature description

## Implementation Details
The removal simplifies the paint cycle by eliminating the need to track two separate content hashes. The `fullKey` hash (which includes clock stamps) is now the only content hash tracked, and it drives both the `setVirtualBufferContent` skip check and the overall change detection. This means clock ticks will still cause redraws (as intended) but without any animation side effects.

The bottom padding addition ensures the dashboard buffer visually extends to fill the viewport, preventing the end-of-file background color from creating a visual artifact at the bottom of the screen.

https://claude.ai/code/session_01SXrbJitTgnu3tyweY3dFQJ

### DIFF
--- a/crates/fresh-editor/plugins/config-schema.json
+++ b/crates/fresh-editor/plugins/config-schema.json
@@ -278,7 +278,7 @@
       "type": "object",
       "properties": {
         "animations": {
-          "description": "Enable frame-buffer animations (tab-switch slides,\nplugin-driven effects). When `false`, every animation\ncall is a no-op: the UI is fully static and each render lands\nthe final frame immediately. Useful on slow terminals, over\nSSH, or for users who prefer no motion.",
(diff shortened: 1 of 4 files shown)


PavelLoparev:
this animation also started when you open setting panel from being on dashboard and the whole settings panel was animated too :)

PR #1979, opened 2026-05-14
fix: paste with Ctrl+V into Search & Replace fields (#1960)

### LINKED ISSUE #1960: Unable to paste value for search in "Search and Replace in Project"
I need to do this especially when the value has UTF-8 characters that I cannot type directly. In the image, I used Ctrl+Shift+v to paste "jovial" into the field. Ctrl-v by itself seems to be disabled.

Image

sinelaw:
Fixes #1960.

## Problem

The project Search & Replace panel (Command Palette → "Search and Replace in Project") runs the focused buffer in a `search-replace-list` mode whose host-side text-input handler in `app/input.rs` swallows every Ctrl/Alt-modified key. As a result Ctrl+V never reached the Paste action and users could not paste into the search or replace fields — a real blocker for any non-typable input (UTF-8 characters, long values from outside the editor, regexes, etc.).

## Fix

Three small changes, all confined to host-side input/clipboard routing:

1. **`app/input.rs` text-input-mode handler** — when a Ctrl/Alt-modified character key reaches the text-input bypass, resolve it against the Normal context first and let only `Action::Paste` through. Other normal bindings (Ctrl+O open-file, Ctrl+S save, …) stay blocked, so the search field is not promoted to a global command bar.
2. **`app/input.rs` `Action::Paste` handler** — route paste through the focused widget input *before* the `editing_disabled` guard, since the Search & Replace results buffer is a read-only virtual buffer and the existing guard would otherwise short-circuit.
3. **`app/clipboard.rs::paste_text` + `app/plugin_dispatch.rs`** — when the active buffer hosts a widget panel with a focused text widget, paste into the widget via a new `paste_into_focused_widget` helper. Newlines are flattened to spaces for single-line widgets; multi-line widgets receive the clipboard text verbatim.

The fix preserves the original safety property of the text-input-mode bypass (random Ctrl-shortcuts cannot leak through) by allowing only the Paste action.

## Tests

Two new e2e tests in `tests/e2e/search_replace.rs`:
- `test_search_replace_paste_into_search_field` — set the internal clipboard to `"jovial"`, open the panel, press Ctrl+V, assert the search field contains `"jovial"`.
- `test_search_replace_paste_into_replace_field` — same flow but Tab to the replace field first.

Both tests hang on master (the action never fires, so `"jovial"` never appears on the screen) and pass with this change. Manual validation in tmux: copying `"hello"` from the buffer, opening Search & Replace, clearing the prefilled field, and pressing Ctrl+V successfully populates the search field and finds the expected match.

## Test plan
- [x] `cargo test -p fresh-editor --test e2e_tests search_replace` — 25 pass / 2 ignored
- [x] `cargo test -p fresh-editor --test e2e_tests paste` — 63 pass / 2 ignored
- [x] `cargo check --all-targets -p fresh-editor` — clean
- [x] `cargo fmt --check` — clean
- [x] Manual tmux validation against the release binary

---
_Generated by [Claude Code](https://claude.ai/code/session_0138njJzQEp5kzmFedEGkHHe)_

### DIFF
--- a/crates/fresh-editor/src/app/clipboard.rs
+++ b/crates/fresh-editor/src/app/clipboard.rs
@@ -557,6 +557,17 @@ impl Editor {
             return;
         }
 
+        // If the active buffer hosts a widget panel with a focused
+        // text input (e.g. the project Search & Replace panel — issue
+        // #1960), paste into that widget instead of the underlying
+        // virtual buffer's text.
+        let active_buf = self.active_buffer();
+        if let Some(panel_id) = self.panel_with_focused_text_for_buffer(active_buf) {
+            self.paste_into_focused_widget(panel_id, &normalized);
+            self.active_window_mut().status_message = Some(t!("clipboard.pasted").to_string());
+            return;
+        }
+
         // Collect cursor info sorted in reverse order by position
         let mut cursor_data: Vec<_> = self
             .active_cursors()
--- a/crates/fresh-editor/src/app/input.rs
+++ b/crates/fresh-editor/src/app/input.rs
@@ -693,6 +693,24 @@ impl Editor {
                             let action_name = format!("mode_text_input:{}", ch);
                             return self.handle_action(Action::PluginAction(action_name));
                         }
+                        // Ctrl/Alt + char isn't a printable input, but
+                        // users still expect Ctrl+V to paste into the
+                        // focused widget input (e.g. the search field
+                        // of the project Search & Replace panel —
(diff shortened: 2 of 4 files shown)


sinelaw:
Closing as superseded: master's widget-runtime refactor already implements paste (Ctrl+V) into the Search & Replace fields. Rebasing this PR re-introduces duplicate handler methods that now live in `widget_runtime.rs`, so the change is no longer needed.

---

PR #1989, opened 2026-05-15
test(e2e): fix flaky Switch Project double-click tests per CONTRIBUTING.md

sinelaw:
The `test_switch_project_double_click_parent_navigates_up` test was
flaking in CI. Audit against `CONTRIBUTING.md` § Testing turned up
two rule violations that together explain the flake:

* **Rule #2 (E2E Tests Observe, Not Inspect)**: the assertion
  `harness.editor().should_restart()` is a model accessor —
  `should_restart()` returns `self.restart_with_dir.is_some()`,
  pure model state with no on-screen signature. The CONTRIBUTING
  rule says: "if an invariant isn't visible on screen, cover it
  with a unit test on the component."

* **Rule #3 (No timeouts; use semantic waiting)**: three sync
  assertions fired immediately after `double_click_at()` —
  `should_restart()`, then a bare `screen_to_string()`, then an
  `assert!(contains("Navigation:"))`. None of these waited for
  the click event to flow through to a steady state, so the
  outcome depended on event-queue timing. Only the final
  `wait_until(parent_marker.txt)` was semantic.

The fix collapses the racy sync assertions into one
`wait_until` that observes both invariants on rendered output:
the Switch Project browser is still showing its "Navigation:"
header AND the parent directory's marker file is now listed.
That combination is strictly stronger than the original three
checks — if the bug had triggered, the browser would have closed
to commit the selection, so "Navigation:" would disappear, and
the wait would never succeed. The harness reaches a stable state
before assertion, eliminating the flake.

Same fix applied symmetrically to
`test_switch_project_double_click_navigates_into_folder`, which
has identical rule violations even if not yet flagged flaky —
the race window exists there too.

### DIFF
--- a/crates/fresh-editor/src/primitives/grammar/mod.rs
+++ b/crates/fresh-editor/src/primitives/grammar/mod.rs
@@ -29,6 +29,7 @@
 // Loader requires filesystem access - runtime only
 #[cfg(feature = "runtime")]
 mod loader;
+mod shebang;
 mod types;
 
 // Re-export all public items for backward compatibility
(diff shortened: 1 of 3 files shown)


sinelaw:
Closing: when rebased onto current `master`, this branch produces no diff — its changes are already present in master.

---

PR #2001, opened 2026-05-16
fix(input): match Shift+letter bindings when terminal omits SHIFT modifier

### LINKED ISSUE #1899: Keybindings utilizing SHIFT don't work + keybinding editor fails to record/assign hotkeys utilizing the SHIFT key unless caps lock is on
Simple bug (or something wrong with my setup, hopefully the former) - the keybinding editor fails to record/assign hotkeys utilizing the SHIFT key unless caps lock is on, and also I cannot invoke the hotkey combo unless caps lock is on.   There are some places in Fresh where the keys are case sensitive, but I'm not sure they should be for typical hotkeys.

sinelaw:
## Summary

Fixes #1899.

Without kitty keyboard protocol, terminals typically encode `Shift+letter` by sending the uppercase character alone — no `SHIFT` modifier accompanies it. The previous `normalize_key` lowercased the char but kept the empty modifier set, so a binding stored as `(Char('p'), SHIFT)` never matched the incoming `(Char('P'), NONE)` event.

As reported in #1899, this made `Shift+letter` shortcuts only fire while **Caps Lock was on** (which causes some terminals to invert the case and re-introduce the `SHIFT` modifier on lowercase letters). The same root cause meant the **keybinding editor's add-binding dialog** captured `Shift+P` as plain `p`, losing the shift information at save time.

## Approach

- `normalize_key` now infers `SHIFT` when an uppercase letter arrives with no other modifiers. The `CapsLock+Ctrl+letter` → `Ctrl+letter` collapse is preserved (uppercase + `CONTROL` is intentionally not given a `SHIFT` because that case is ambiguous between caps lock and a real shift, and the existing behavior is to fold caps lock away).
- The add-binding dialog (`view/keybinding_editor.rs`) and the record-key search (`app/keybinding_editor/editor.rs`) route the recorded event through `normalize_key` so they capture `Shift+P` consistently — no matter whether the terminal reported `SHIFT`.
- The fallback to `InsertChar` is unaffected: it uses the original event, so typing capital `Z` in a buffer still inserts `Z`.

## Tests

- `test_shift_letter_binding_works_without_terminal_shift_modifier` — the core regression: `Char('P')` with no modifier now matches a `Shift+P` binding.
- `test_capslock_ctrl_letter_still_matches_ctrl_letter_binding` — guards the existing CapsLock+Ctrl+A → Ctrl+A behavior.
- `test_uppercase_without_binding_falls_through_to_insert_char` — confirms typing capital letters still inserts them when no binding exists.
- `record_search_key_normalizes_uppercase_letter_to_shift` — the record-key search captures the `SHIFT` modifier even when the terminal omits it.
- `add_dialog_records_shift_when_terminal_omits_shift_modifier` / `add_dialog_records_shift_when_terminal_includes_shift_modifier` / `add_dialog_preserves_ctrl_when_capturing_upper_letter` — full end-to-end dialog drive: send the simulated event through `handle_keybinding_editor_input` and assert the captured `(key_code, modifiers)`.

Full lib test suite: **2468 passed, 0 failed**.

## Manual validation

Reproduced and verified the fix end-to-end:

1. Added a `Shift+P → save` binding in `~/.config/fresh/config.json`.
2. Opened a modified buffer (`Xinitial content`) in `fresh` under tmux.
3. Pressed `P`. The status bar transitioned to `Saved`, the unsaved-changes marker `[+]` and the buffer's `*` cleared, and the file on disk reflects the new content.

## Test plan

- [x] `cargo test -p fresh-editor --lib`
- [x] `cargo check --all-targets`
- [x] `cargo check --all-targets --features gui`
- [x] `cargo fmt --check`
- [x] Manual smoke test: `Shift+P` triggers `save` via custom binding.

https://claude.ai/code/session_01H6FpwGdLrCA7zquAbVrhX9

---
_Generated by [Claude Code](https://claude.ai/code/session_01H6FpwGdLrCA7zquAbVrhX9)_

### DIFF
--- a/crates/fresh-editor/src/app/event_debug.rs
+++ b/crates/fresh-editor/src/app/event_debug.rs
@@ -13,14 +13,30 @@ const MAX_HISTORY: usize = 10;
 pub struct RecordedEvent {
     /// The raw key event
     pub event: KeyEvent,
-    /// Human-readable description
+    /// Human-readable description of the raw event
     pub description: String,
+    /// Human-readable description of the key *after* `normalize_key` — i.e.
+    /// the form keybinding lookup actually matches against. `None` when
+    /// normalization leaves the event unchanged (no extra line to show).
+    pub normalized: Option<String>,
 }
 
 impl RecordedEvent {
     fn new(event: KeyEvent) -> Self {
         let description = format_key_event(&event);
-        Self { event, description }
+        let (norm_code, norm_mods) =
+            crate::input::keybindings::normalize_key(event.code, event.modifiers);
+        let normalized = if norm_code == event.code && norm_mods == event.modifiers {
+            None
+        } else {
+            let norm_event = KeyEvent::new(norm_code, norm_mods);
+            Some(format_key_event(&norm_event))
+        };
+        Self {
+            event,
+            description,
+            normalized,
+        }
     }
 }
 
--- a/crates/fresh-editor/src/app/keybinding_editor/editor.rs
+++ b/crates/fresh-editor/src/app/keybinding_editor/editor.rs
@@ -4,7 +4,9 @@ use super::helpers::{format_chord_keys, key_code_to_config_name, modifiers_to_co
 use super::types::*;
 use crate::config::{Config, Keybinding};
 use crate::input::command_registry::CommandRegistry;
-use crate::input::keybindings::{format_keybinding, Action, KeyContext, KeybindingResolver};
+use crate::input::keybindings::{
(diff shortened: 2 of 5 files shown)


sinelaw:
In my terminal (xfce terminal) when enabling caps lock the editor receives only modifier "shift", i.e. "shift"+"A"

RandomGHUser:
@sinelaw, this has been sitting a few weeks with the test failure on windows, any chance to revive this commit?

sinelaw:
Done, thanks for reporting!

PR #2002, opened 2026-05-16
live_diff: add live_diff.enabled_on_startup config option

### LINKED ISSUE #1950: Feature request: Config option to enable Live Diff on startup
Currently, Live Diff needs to be toggled on manually each time the editor is opened. For users who rely on it as part of their regular workflow, this adds a small but repetitive step every session. A config setting to enable Live Diff by default on startup would remove this friction.

sinelaw:
Closes #1950.

## Summary

Live Diff is opt-in: every session starts with the global toggle off, so users who rely on it have to invoke `Live Diff: Toggle (Global)` from the command palette before decorations appear. This adds a new `live_diff.enabled_on_startup` config field (default `false`, preserving current behavior). When `true`, the plugin force-enables the global toggle on startup — equivalent to the manual toggle, but automatic.

The user can still toggle Live Diff off mid-session via the command palette; the next startup re-enables it, which matches the requested "always on" semantics.

## Implementation

- New `LiveDiffConfig` struct in `config.rs` with a single `enabled_on_startup: bool` field, alongside the existing `terminal`, `clipboard`, etc. sections. Default impl matches the `WarningsConfig`/`TerminalConfig` pattern. Schema regenerated via `./scripts/gen_schema.sh`.
- Mirror `PartialLiveDiffConfig` in `partial_config.rs` (Merge + From + resolve) so layered config loading handles the new section.
- `live_diff.ts` reads `config.live_diff.enabled_on_startup` via `editor.getConfig()` and force-enables the global state. The check is gated by a `startupConfigApplied` boolean and fires lazily on the first buffer event (`after_file_open` / `buffer_activated` / plugin init with an active buffer) rather than at plugin load — `getConfig()` reads from a snapshot that the editor refreshes between ticks, so it can return `{}` if the plugin runs first.

## Test plan

- [x] New e2e test `test_live_diff_enabled_on_startup_config_renders_without_manual_toggle` builds a `Config` with `enabled_on_startup = true`, opens a modified file, and asserts the `+` gutter glyph appears with no command-palette invocation. Fails without the change (the existing test had to call `enable_live_diff_globally` for the same gutter to appear).
- [x] All 10 `live_diff` e2e tests pass (`cargo test --test e2e_tests -p fresh-editor live_diff`).
- [x] All 326 plugin e2e tests pass (`cargo test --test e2e_tests -p fresh-editor plugins::`).
- [x] All 60 config unit tests pass (`cargo test -p fresh-editor --lib config::`), including the roundtrip-to-partial test.
- [x] Manual validation in tmux:
  - With `live_diff.enabled_on_startup: true`, opening a modified file shows the `+` gutter glyph with no toggling.
  - With `live_diff.enabled_on_startup: false` (default), the gutter stays empty until the user runs the toggle command.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KqJtUdMofULezVB9ptgKw3)_

### DIFF
--- a/crates/fresh-editor/plugins/config-schema.json
+++ b/crates/fresh-editor/plugins/config-schema.json
@@ -181,6 +181,13 @@
         "mouse_forwarding": "requested"
       }
     },
+    "live_diff": {
+      "description": "Live Diff plugin settings.\nControls the inline-diff view rendered against a git reference.",
+      "$ref": "#/$defs/LiveDiffConfig",
+      "default": {
+        "enabled_on_startup": false
+      }
+    },
     "keybindings": {
       "description": "Custom keybindings (overrides for the active map)",
       "type": "array",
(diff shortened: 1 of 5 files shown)


sinelaw:
Plugins need their own dedicated config value registration mechanism so that we don't hard code them into rust or touch the config.rs when updating a plugin's config abilities

PR #2003, opened 2026-05-16
feat(config): per-language ruler overrides (closes #1881)

### LINKED ISSUE #1881: Language-specific ruler configuration
It doesn't seem to be possible to set language-specific ruler settings? At least it doesn't appear in the documentation. I think that in general, being able to edit display settings on a per-language basis could be useful for users to follow the writing conventions of each language.

sinelaw:
## Summary

Closes #1881.

Lets language configs set their own vertical-ruler columns. Previously `rulers` was global-only, so users couldn't follow per-language line-length conventions without retraining their eyes every time they switched files.

### Design

`LanguageConfig.rulers: Option<Vec<usize>>`:
- `None` (default) → fall back to global `editor.rulers` (existing behaviour, no regression)
- `Some(vec)` → override per language
- `Some([])` → opt the language out of globally-configured rulers

Resolution mirrors the existing `line_wrap` / `wrap_column` plumbing:

- `buffer_config_resolve::rulers(language, config)` is the pure helper
- `Window::resolve_rulers_for_buffer(buffer_id)` is the orchestrator-side entry point
- All call sites that initialise a buffer's view state now resolve rulers via the buffer's language (file open, split, tab drag, plugin-driven open, virtual buffers, settings update, etc.). The boot-time empty buffer and terminal buffers keep using global rulers since they have no language.
- Settings update now resolves rulers per buffer so language overrides survive a runtime settings save.

### Example

```json
{
  "editor": { "rulers": [80] },
  "languages": {
    "rust":     { "rulers": [100, 120] },
    "markdown": { "rulers": []         }
  }
}
```

Rust files get rulers at 100 and 120; markdown files have no rulers; everything else inherits the global ruler at 80.

## Test plan

- [x] 4 new unit tests in `buffer_config_resolve` (override wins, fallback when unset, fallback when `None`, empty-vec opts out)
- [x] 3 new E2E tests in `tests/e2e/vertical_rulers.rs` driving real key/file flows and asserting on rendered ruler-bg cells
- [x] All 16 existing + new ruler E2E tests pass
- [x] All 2465 fresh-editor lib unit tests pass
- [x] `cargo check --all-targets` and `cargo check --all-targets --features gui` clean
- [x] `cargo clippy --all-targets` introduces no new lints
- [x] Manual validation in tmux with the three-language config above: Rust shows rulers at cols 30 / 50, plain text shows the global ruler at 10, Markdown shows none

## Notes

- Regenerated `crates/fresh-editor/plugins/config-schema.json` (new `rulers` field on language config). Stale unrelated changes that the schema generator also produced (`fresh.d.ts`, `theme.schema.json`) were reverted; they belong in their own PRs.
- Updated `docs/configuration/index.md` to mention `rulers` alongside the other per-language overrides.
- The interactive Add/Remove Ruler commands continue to act on `editor.rulers` (the global list). Making them aware of the current buffer's language would be a separate UX decision and is out of scope.

---
_Generated by [Claude Code](https://claude.ai/code/session_01X5yAooZ1LUNyB3TpRZ1tmX)_


sinelaw:
Works: Setting the ruler for a language to overide the global ruler
Doesn't work: removing the per-language ruler using Settings UI, saving the settings -> ruler is gone and global ruler is not applied. Need to exit and reload the editor to see it work

sinelaw:
after latest change - when deleting a language's specific Ruler config, saving settings -> the global ruler config is NOT applied, just zero rulers shown

PR #2071, opened 2026-05-20
fix(orchestrator): scope sessions per project + redesign the Open picker

sinelaw:
## Summary

Fixes the reported orchestration bug where launching in one project pulled in **another day's / another project's** directories, files, and sessions, and reworks the Open picker around per-project scoping.

**Root cause.** In 0.3.7 orchestrator persistence moved from per-cwd files to a single global `windows.json`. Two surfaces then leaked across projects:
1. the Open picker listed *every* session from *every* project in one flat, unlabeled list; and
2. at boot the editor could drop you straight into a previous session (its worktree root / label / plugin state) rather than a clean editor for the project you launched in.

## Changes

**Boot (no cross-project bleed):**
- `pick_active_window_for_cwd` is now strictly cwd-scoped: reopen the session last used **in the launch cwd's project** (`env.active` if it belongs to the cwd, else the most-recent session for that cwd, else a clean base window). A different project's session is never auto-activated. Honors a directory passed on the CLI (`fresh /path`), not just the shell cwd.
- A clean base keys off the *picked* window, so a stale persisted id-1 window from another project can't lend its label/root/state to it.

**Open picker:**
- Scopes to the current project by default; an `Alt+P` toggle reveals all projects. Filtering is always global.
- Tabular rows `[id]  NAME …  PROJECT` with a column header; the PROJECT column labels cross-project rows; sessions sort current-project-first. Active session's `[id]` renders in the active-tab colour.
- Labelled `Project: [ <name> ▾  (Alt+P) ]` scope control and a labelled `Filter` field; `/` focuses the filter.
- The floating panel now fits the session count instead of a tall box padded with blank rows.

**Host fixes (found while testing):**
- Widget row-zip clamped byte-unit overlays to the pre-truncation length, so a multi-byte ellipsis from column truncation could land mid-char and panic the span splitter — now snaps overlays to a char boundary.
- The floating-panel key dispatcher fed plain chars straight to the focused text input, consulting mode bindings only for named/Ctrl/Alt keys — so a bare key like `/` bound in a picker mode was dead. Plain chars now defer to an explicit mode binding first.

**Docs:** design doc updated with the scoping principles, wireframes, and the shipped layout.

## Test plan

- [x] `cargo test -p fresh-editor --lib orchestrator_persistence::tests` (cwd-scoped boot pick: never-crosses-projects, last-used, most-recent fallback, clean-base)
- [x] `cargo test -p fresh-editor --lib widgets::render` (char-boundary regression test)
- [x] `cargo test -p fresh-editor --test orchestrator_persistence_paths`
- [x] `cargo test -p fresh-editor --test e2e_tests orchestrator` (scoping default + `Alt+P` reveal-all + `/`-focus live filtering; form typing unaffected)
- [x] Widget / search_replace text-input e2e (host key-dispatch change is non-regressive)
- [x] Manual tmux walkthrough: created git-worktree sessions across `webapp`/`api-service` + an in-place session in non-git `data`; verified scoped vs all-projects views, the Project control, `/`-focus + filtering, clean per-cwd boot, and compact panel sizing

https://claude.ai/code/session_01UhCtdAe4SYg6QUsXuYKc9y

### DIFF
--- a/crates/fresh-editor/plugins/orchestrator.ts
+++ b/crates/fresh-editor/plugins/orchestrator.ts
@@ -480,10 +480,12 @@ function renderListItem(id: number, activeId: number): TextPropertyEntry {
   }
   // Project tag: in the "all projects" view, label every row with
   // its project so a cross-project session is obvious at a glance
-  // rather than blending into the current project's sessions.
+  // rather than blending into the current project's sessions. Use
+  // the project's basename (not the full parent/base label) so it
+  // fits the sessions column without truncating.
   if (openDialog?.scope === "all" && projectKeyOf(s) !== currentProjectKey()) {
     entries.push({
-      text: `  · ${projectLabel(projectKeyOf(s))}`,
+      text: `  · ${editor.pathBasename(projectKeyOf(s))}`,
       style: { fg: "ui.menu_disabled_fg", italic: true },
     });
   }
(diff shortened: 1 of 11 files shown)


sinelaw:
Merged in a different PR

PR #2091, opened 2026-05-24
fix: Fixed cursor column calculation in statusbar elements

### LINKED ISSUE #2090: Cursor statusbar elements (full and compact) show invalid column numbers in some cases
`Cursor` statusbar element renders columns count as characters bytes length instead of chars count.

Example:
String 1: `0123456789` if your cursor at char `9` it will render `Col 10` correctly.
String 2: `╚═╝  ╚═╝╚═` - it will show `Col 24` incorrectly.

PavelLoparev:
Fixes https://github.com/sinelaw/fresh/issues/2090

`Cursor` and `Cursor (compact)` statusbar elements calculate column count incorrectly (characters byte length instead of characters position).

### DIFF
--- a/crates/fresh-editor/docs/visual-regression/screenshots/Comprehensive_UI_A_01_state_a.svg
+++ b/crates/fresh-editor/docs/visual-regression/screenshots/Comprehensive_UI_A_01_state_a.svg
@@ -460,12 +460,12 @@
   <rect x="378" y="54" width="9" height="18" fill="#000000"/>
   <text x="379" y="68" fill="#ffff00" class="terminal" style="">n</text>
   <rect x="387" y="54" width="9" height="18" fill="#000000"/>
-  <text x="388" y="68" fill="#d4d4d4" class="terminal" style="">(</text>
+  <text x="388" y="68" fill="#8be9fd" class="terminal" style="">(</text>
   <rect x="396" y="54" width="9" height="18" fill="#000000"/>
-  <text x="397" y="68" fill="#d4d4d4" class="terminal" style="">)</text>
+  <text x="397" y="68" fill="#8be9fd" class="terminal" style="">)</text>
   <rect x="405" y="54" width="9" height="18" fill="#000000"/>
   <rect x="414" y="54" width="9" height="18" fill="#000000"/>
-  <text x="415" y="68" fill="#d4d4d4" class="terminal" style="">{</text>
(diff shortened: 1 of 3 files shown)


sinelaw (changes requested):

- **Interactive testing passed**: manually navigated to multi-byte Unicode lines (`café résumé`, `日本語テスト`) and confirmed the status bar now shows correct character-based column numbers (Col 12 and Col 7 respectively) instead of the previous byte-count values — the fix is functionally correct.
- **Missing test (CONTRIBUTING.md violation)**: CONTRIBUTING.md requires _"Every behavioral claim in a commit must be backed by a test that fails without the change and passes with the fix."_ This PR has no test — please add a unit test to `status_bar.rs` (or the buffer test suite) that demonstrates the buggy byte-count behaviour and confirms the character-count fix.
- **Broad error recovery**: the `from_utf8(&line_bytes)` arm silently returns `0` on `Err(_)`, which violates guideline #9 ("trigger recovery on the specific error it was designed for, not on catch-all branches"). Consider using `from_utf8_lossy` and proceeding rather than swallowing the error silently, or at minimum add a debug-level log.

---

sinelaw:
Pushed some changes on top of @PavelLoparev's fix (and rebased the branch onto current `master`, which was ~3 minor versions behind):

- **Fixed a regression the original approach introduced.** Deriving the line via the cached `primary_cursor_line_number` + `line_start_offset` disagrees with the live cursor position in diff/split views, which broke the column display (e.g. `Col 1` in the side-by-side diff, inflated column in the comprehensive visual snapshot — both were red in CI). `cursor_column` now derives the line start from `cursor.position` itself, like the pre-fix code did.
- **Count grapheme clusters, not code points.** The editor moves the cursor by grapheme boundaries, so the column should too — otherwise a combining mark or ZWJ emoji would advance `Col` by more than one per keypress. Uses the existing `primitives::grapheme::grapheme_count`. (The composite buffer in the diff view doesn't expose readable line content, so it falls back to the byte distance there, which equals the grapheme count for the ASCII those views render.)
- **Added tests.** Unit tests for combining-accent / ZWJ-emoji / multi-byte columns, plus an e2e test that drives Right one keypress at a time over a multi-byte line and asserts the rendered `Col` advances by exactly one (observes rendered output per CONTRIBUTING.md).

All status-bar + diff-cursor tests pass locally; clippy/fmt clean.

sinelaw:
Thanks!

PR #2093, opened 2026-05-24
Add built‑in support for opening ePub files

### LINKED ISSUE #2087: $100 to upgrade Fresh IDE TUI for ePub reading
I donate to open source, and I'm loving the new Fresh IDE TUI for teaching students.

I'm offering $100 as an incentive for anyone here who can add epub reader capabilities to Fresh using Rust.

Spec: in the left-side file explorer, I select file `foo.epub`, then the right-side file viewer shows me the epub content as human-readable text not bytes.

Scope: The goal is to be good enough for a developer to be able to read epub documentation within Fresh in the TUI, without needing to call pandoc or launch a separate app or helper or plugin. Fresh runs locally on the developer's laptop, and the ePub is on a remote server and for security reasons cannot be downloaded to the local laptop.

Implementation: I can suggest trying the Rust crate `epub` and `bookokrat` because these work well for me in other projects. The implementation can't use JavaScript or TypeScript because of compliance aspects beyond me.

As an aside, I'm not affiliated with Fresh in any way, other than as a happy user and open source sponsor.

Update: I can keep this donation incentive open for two months, until August 1 2026; after that I'll need to move on.

servusdei2018:
Resolves #2087.

This change lets Fresh display EPUB documents directly in the editor:

- Selecting an EPUB in the left‑hand file explorer opens it on the right side as a read‑only text view.
- The file is parsed completely in memory with the `epub` crate, then each chapter’s HTML is turned into plain‑text using `html2text`.
- No temporary files are written, so the remote file never lands on the local disk.
- The buffer is marked read‑only and receives Markdown‑style syntax highlighting for a clean reading experience.
- New unit and end‑to‑end tests verify that an EPUB loads correctly, the content is shown, and editing is prevented.

image


sinelaw:

**Interactive testing**: Built and opened a real `.epub` file — EPUB content loaded correctly (title, author, chapter text visible), status bar showed `[RO]`, and typing was fully blocked. The core feature works.

**Issues to fix before merging:**

- **Missing locale translations (CONTRIBUTING.md #10 — blocker)**: Only `en.json` was updated. The two new keys (`buffer.epub_file`, `buffer.opened_epub`) must be added to all 13 remaining locale files: `cs.json`, `de.json`, `es.json`, `fr.json`, `it.json`, `ja.json`, `ko.json`, `pt-BR.json`, `ru.json`, `th.json`, `uk.json`, `vi.json`, `zh-CN.json`. Per guidelines: "Don't commit English placeholders."
- **E2E test uses model-state accessor (CONTRIBUTING.md testing rule #2)**: `harness.editor().active_window().is_editing_disabled()` inspects internal model state; replace with a rendered-output assertion (e.g. assert that `[RO]` or the read-only indicator appears on screen).
- **Duplicate chapter heading**: The extracted text renders both `## Chapter One` (from TOC label) and `# Chapter One` (from the HTML `<h1>` tag), producing a doubled heading for every chapter. Consider suppressing the `<h1>` tag conversion or the TOC-derived `##` header when both would be identical.

---

sinelaw (changes requested):
Requesting changes — see inline comment for details. Main blockers: missing locale keys in all non-English locale files, and an E2E test that inspects model state instead of rendered output.

---

sinelaw:
Thanks!
Aside for the technical notes produced above by the automatic code review, I'm trying to decide if this architectural solution is the best way to go. It kind of breaks the model of how buffers are loaded by introducing a bespoke path just for ePub files, which bypasses the entire lazy loading piece tree backed-by-a-file model. 
The original issue mentioned limitations about JavaScript. I'm trying to understand the context for that as well as much of fresh functionality is actually implemented as JavaScript plugins running in quick.js

joelparkerhenderson:
Thanks Noam. I'm the dev who posted the $100 feature incentive for epub, that is motivating this PR.

Your way of lazy loading is better & smarter. At the same time, I'm seeking any architectural solution that's soonish because I'm teaching students and this will make a large improvement in their immediate learning. I

If you're thinking broadly and long term, I have similar kinds of upcoming needs for PDF books, TSV data grids that can correctly render emoji (see ModernCSV for my favorite), PlantUML files that I currently dynamically render into PNG images, and DICOM files for x-rays, PET scans, and MRIs. All have the same need of remote storage and local render, because the MacBook Neo has minimal RAM & SSD.

My need for Rust in the core versus JavaScript plugins is because I work in a regulated industry. It's easy to proof and assure Rust code in the core; it's wildly hard to proof and assure JavaScript plugins, and I'm seeing lots of worldwide plugin-related attacks and extension-related attacks, including the massive Microsoft GitHub one just last week. Currently it takes me many months to get a JavaScript plugin approved for security, whereas Rust is just a few days. This gap isn't specific to Fresh; it's the same for VS Code, AstroNVim, etc. 

Happy to chat if you wish. I'm joel@joelparkerhenderson.com and you can see a bit of what I'm doing on my LinkedIn & GitHub etc. You're building great software and I've been wanting a Rust TUI IDE for five years, so I'm very encouraged to see you building it. <3

sinelaw:
I'm not convinced this belongs inside Fresh, running any external program to convert arbitrary formats to text should probably be the right approach here. I'm concerned about introducing bloat (binary size), dependency tree size, supply chain risk, security risk (parsing arbitrary binary formats), etc inside the editor itself.

Not entirely sure but I'm holding back for now

joelparkerhenderson:
Understood and I appreciate your update. Thank you for considering it.

Issue #2116, opened 2026-05-26
Cannot scroll Live Grep text window


mandolyte:
I am unable to scroll the content on the right side. Whether I use up/down arrow or scroll wheel, it is the match list that scrolls, not the content panel. Also, I can't really see what I have typed into the search field - a theme issue I suppose...

To reproduce, use attached xml file and search for `<em>`. Then try to scroll the three matches into view on the right. 

[chapter-16.txt](https://github.com/user-attachments/files/28273505/chapter-16.txt)
 

<img width="1388" height="769" alt="Image" src="https://github.com/user-attachments/assets/6067d532-e9de-4b78-8e5e-5fe08b7cb5d6" />

sinelaw:
Which theme is this?

sinelaw:
Reproduced in Dracula theme

As for the scrolling - I didn't think that would be needed, I'll consider adding that

mandolyte:
Yeah, you are correct. The real issue is that I can't see the match. These long lines are such a headache. 

So reproduce with a small sized terminal window. Then at least the second match will not be viewable

sinelaw:
Theme issue is fixed in https://github.com/sinelaw/fresh/pull/2120

sinelaw:
The scroll issue is fixed in https://github.com/sinelaw/fresh/pull/2121

sinelaw:
One more improvement is shipped in https://github.com/sinelaw/fresh/pull/2123 - match highlighting and better centering

PR #2120, opened 2026-05-26
Add suggestion_fg theme color for overlay prompt text

sinelaw:
## Summary
Fixes invisible text in overlay prompts (Live Grep) by introducing a dedicated `suggestion_fg` theme color for content drawn on `suggestion_bg`. Previously, the overlay prompt borrowed `prompt_fg` for its title and input text, which could match the editor background in some themes (e.g., Dracula), making the text invisible.

## Key Changes
- **Added `suggestion_fg` color field** to `UiColors` struct with fallback to `popup_text_fg` for backward compatibility with existing themes
- **Updated overlay prompt rendering** in `render.rs` to use `suggestion_fg` instead of `prompt_fg` for the title, and `editor_fg` for the input field
- **Updated footer rendering** in the overlay prompt to use `suggestion_fg` instead of `prompt_fg`
- **Added `popup_fg` serde alias** to `popup_text_fg` to allow themes to use the naming convention consistently
- **Added comprehensive documentation** to `UiColors` explaining the `*_bg`/`*_fg` pairing convention and which backgrounds intentionally lack matching foregrounds
- **Added regression test** (`test_suggestion_fg_falls_back_and_contrasts`) to ensure `suggestion_fg` contrasts with `suggestion_bg` and falls back correctly
- **Updated theme schema and i18n translations** across 13 languages to include the new `suggestion_fg` field

## Implementation Details
- `suggestion_fg` is optional in theme files and defaults to `None`, with the conversion logic falling back to `popup_text_fg` when unset
- The change maintains full backward compatibility—existing themes without `suggestion_fg` will automatically use `popup_text_fg`
- The fix addresses the root cause: renderers must pair foreground/background colors from the same surface to prevent text from becoming invisible when colors accidentally match

https://claude.ai/code/session_014U9BCU8j95NaV8tz7Q5UGK


sinelaw:
Fixes the theme issue mentioned in https://github.com/sinelaw/fresh/issues/2116 and also a similar issue with Light theme

PR #2139, opened 2026-05-27
fix(status-bar): count characters, not bytes, for cursor column

### LINKED ISSUE #2090: Cursor statusbar elements (full and compact) show invalid column numbers in some cases

sinelaw:
## Summary

Fixes #2090. The `Cursor` and `CursorCompact` status-bar elements computed the column as the **byte** distance between the line start and the cursor, so multi-byte characters (box-drawing glyphs, accents, …) inflated the reported column. For example, `╚═╝  ╚═╝╚═` (box-drawing chars are 3 bytes each) reported `Col 24` instead of the correct character count.

The fix introduces a `cursor_column_chars` helper that counts the number of Unicode scalar values in the line prefix up to the cursor (counting UTF-8 non-continuation bytes), and uses it for both the full and compact cursor indicators. Only the line prefix up to the cursor is read, keeping the operation viewport-localized.

## Test plan

- [x] New unit test `test_cursor_column_counts_chars_not_bytes` (fails with the old byte-based logic, passes with the fix) covering ASCII, box-drawing multi-byte glyphs, and per-line reset.
- [x] Full `view::ui::status_bar` test module passes (24 tests).
- [x] `cargo fmt` + `cargo clippy -p fresh-editor --lib` clean (no new warnings).
- [x] Manual validation in tmux: opened a file containing `╚═╝  ╚═╝╚═` and stepped the cursor right one character at a time — the column now increments by exactly 1 per character (Col 1→2→3→4→5) where the byte-based bug would have jumped by 3 per glyph.

https://claude.ai/code/session_01V1FhAZZbui9hxJ9EWMvBCW

---
_Generated by [Claude Code](https://claude.ai/code/session_01V1FhAZZbui9hxJ9EWMvBCW)_

### DIFF
--- a/crates/fresh-editor/src/view/ui/status_bar.rs
+++ b/crates/fresh-editor/src/view/ui/status_bar.rs
@@ -5,6 +5,7 @@ use std::path::Path;
 
 use crate::app::WarningLevel;
 use crate::config::{StatusBarConfig, StatusBarElement};
+use crate::model::buffer::Buffer;
 use crate::primitives::display_width::{char_width, str_width};
 use crate::state::EditorState;
 use crate::view::prompt::Prompt;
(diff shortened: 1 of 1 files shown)


sinelaw:
Closing as a duplicate of #2091, which already fixes #2090 with the same approach and is further along (reviewed). No need for two PRs on the same issue.

---

Issue #2077, opened 2026-05-21
pwsh.exe - Application Error


ket000:
Using Fresh 0.3.8On Windows 11 23H2, Trying to open view Terminal -> Open Terminal Results in below error. 
---------------------------
pwsh.exe - Application Error
---------------------------
The application was unable to start correctly (0xc0000142). Click OK to close the application. 
---------------------------
OK   
---------------------------

<img width="390" height="153" alt="Image" src="https://github.com/user-attachments/assets/67254b31-995c-4a1e-bfce-353bdb086a0a" />

Trying to manually open the pwsh.exe from command line works fine. 

pwsh.exe
PowerShell 7.6.1

sinelaw:
Labeling as `deferred` for now. The `0xc0000142` error indicates a Windows DLL/initialization failure when Fresh spawns `pwsh.exe`, even though launching it manually works — likely tied to the environment/working directory Fresh passes to the child process. This is Windows-specific and can't be reproduced in our Linux-based e2e/tmux test harness, so it doesn't yet meet the "easily reproducible" actionable criterion.

A couple of details would help narrow it down when someone picks this up:
- Does the same error occur if you set your terminal shell to `cmd.exe` instead of `pwsh.exe`?
- Any entries in the Fresh logs at `C:\Users\<user>\.local\state\fresh\logs` around the time the terminal is opened?

---
_Generated by [Claude Code](https://claude.ai/code)_

ket000:
The issue happens with both cmd.exe and windows terminal wt.exe

sinelaw:
Hey I just tried this on a brand new Windows 11 installation and couldn't reproduce it...

sinelaw:
@ket000 which installation method did you use for Fresh? winget? or something else

ket000:
cargo binstall fresh on windows 11

sinelaw:
can you try `winget install fresh-editor` instead?

ket000:
Ok will try. Not sure what is the difference. Will post here.

ket000:
The same error occurred even after installing via winget. I’m not sure if it’s related to having PowerShell version 7.6.2 installed from the Microsoft Store. On another machine running the default PowerShell 5 that comes with Windows 11, view terminal copmmand  worked without any issues.


winget install fresh-editor
Found fresh-editor [sinelaw.fresh-editor] Version 0.3.9
This application is licensed to you by its owner.
[...]

PR #2149, opened 2026-05-27
Skip Microsoft Store pwsh App Execution Alias when opening terminal

sinelaw:
The Store build of pwsh.exe is a zero-byte APPEXECLINK reparse stub under
WindowsApps. Spawning it through ConPTY crashes with 0xc0000142
(STATUS_DLL_INIT_FAILED), so detect_shell now resolves candidates to a
concrete path, prefers a real PowerShell 7 install, and falls back to
Windows PowerShell when only the alias stub is present.

https://claude.ai/code/session_013W1JsUysvPDVk89rnwPpTa


sinelaw:
Hopefully fixes https://github.com/sinelaw/fresh/issues/2077

PR #2150, opened 2026-05-27
feat(syntax): Add distinct highlight category for built-in variables

masmu:
This PR introduces the VariableBuiltin highlight category to allow precise styling of language-specific built-in variables (such as self, this, super, and arguments).

Previously, built-in variables were generally mapped to standard variables or constants, preventing theme authors from customizing them separately. This update parses the variable.language (TextMate) and equivalent Tree-Sitter scopes and assigns them to the new category.


sinelaw:

Reviewed and manually tested by opening JS and TS files with `this`/`super`/`arguments` in a tmux session. The TypeScript case works correctly (these identifiers render in RGB 86,156,214 as intended), but the PR has three issues that need to be addressed before merging:

- **Feature silently does not work for JavaScript (tree-sitter path):** `DEFAULT_HIGHLIGHT_CAPTURES` (used by JavaScript and all non-TypeScript languages) does not include `"variable.builtin"`, so the new category is never reached for `.js` files. Only TypeScript (which has its own `TYPESCRIPT_HIGHLIGHT_CAPTURES` that already lists `"variable.builtin"`) and TextMate-engine languages benefit from this change. Either add `"variable.builtin"` to `DEFAULT_HIGHLIGHT_CAPTURES` or create a JavaScript-specific captures list, and add the matching index → `VariableBuiltin` arm to `from_default_index` (or a new `from_javascript_index`).
- **No tests (CONTRIBUTING.md §Testing rule 1):** Every behavioral claim must be backed by a test that fails without the change and passes with it. At minimum, unit tests for `scope_to_category` in both `highlight_engine.rs` and `textmate_engine.rs` are needed (e.g., assert `"variable.language.this"` → `VariableBuiltin`). An e2e test that opens a TS/JS file and asserts the rendered color of `this` would be even stronger.
- **Schema not regenerated (CONTRIBUTING.md §Code Guidelines rule 6):** `SyntaxColors` was extended with `variable_builtin`, but `plugins/schemas/theme.schema.json` was not regenerated (`./scripts/gen_schema.sh`). The schema currently has no entry for `syntax.variable_builtin`, which breaks schema validation and theme authoring tooling.

---

sinelaw (changes requested):
Requesting changes — see inline comment for details on the three issues found during review.

---

masmu:
I can't run the entire test suite locally. There are always some tests that time out. Is this a known issue and are there any tips which help with that?

PR #2151, opened 2026-05-27
fix: classify storage.type as a keyword rather than a type in syntax highlighting

masmu:
### Summary
This PR adjusts the mapping of TextMate scopes to our internal highlight categories. Specifically, it maps `storage.type` to `HighlightCategory::Keyword` instead of `HighlightCategory::Type`.

### Details
Previously, storage types (representing tokens like `class`, `def`, `function`, `var`, `let`, and `const`) were incorrectly categorized as Types in both the generic highlight engine and the TextMate engine. By updating this explicitly, these structural keywords receive the correct keyword coloring, bringing syntax highlighting more in line with user expectations.


sinelaw (changes requested):

- **Fix is semantically correct**: `storage.type` scopes (tokens like `class`, `function`, `var`, `let`, `const`, `def`) are structural keywords, not type names — moving them from `HighlightCategory::Type` to `HighlightCategory::Keyword` is the right call. The change is consistent across both `highlight_engine.rs` and `textmate_engine.rs`.
- **Interactive testing passed**: Opened a JS file in the editor via tmux; confirmed `class`, `function`, `var`, `let`, `const`, `return`, `new` all render in keyword color (cyan). No regressions observed on strings, comments, or function-name highlights.
- **Missing required test (CONTRIBUTING.md §Testing rule 1)**: Every behavioral fix must include a test that fails without the change and passes with it. `textmate_engine.rs` already has `test_scope_to_category` and `highlight_engine.rs` has similar unit-test coverage for `scope_to_category`. Please add assertions such as `assert_eq!(scope_to_category("storage.type"), Some(HighlightCategory::Keyword))` (and a negative check against `Type`) to both test suites before this can be merged.

---

sinelaw:
Thanks!

masmu:
You are welcome! Thanks for the merge.

PR #2152, opened 2026-05-27
feat: Add ClearSearch action and expose active search state to plugins

masmu:
This PR adds a new `ClearSearch` action (`cmd.clear_search`). This command allows users to clear active search highlights explicitly without closing the find widget.

Additionally, it adds a `has_active_search` boolean property to the `EditorStateSnapshot` and exposes it via QuickJS (`hasActiveSearch()`), allowing plugins to query whether search highlights are currently visible.

### Changes
* **Input Actions:** Added `Action::ClearSearch` and mapped it to the new `cmd.clear_search` command.
* **API/Plugin Runtime:** Expose `has_active_search` to `EditorStateSnapshot` and implement `has_active_search()` in the QuickJS backend.
* **Locales:** Added new translations for `cmd.clear_search`, `cmd.clear_search_desc` and `action.clear_search` across all supported languages.


sinelaw (changes requested):

Tested interactively via tmux: opened the editor, performed a search for "Hello" (highlights appeared on both occurrences), then ran **Clear Search Highlights** from the command palette — highlights were correctly cleared. The command is discoverable in the palette and works as described. Architecture is clean and minimal, correctly reusing `clear_search_highlights()`.

Two CONTRIBUTING.md violations need to be fixed before merge:

- **Missing tests (CONTRIBUTING.md §1):** Every new feature must be backed by a test that fails without the change and passes with it. Neither the `ClearSearch` action nor the `has_active_search` plugin API have any tests.
- **English placeholders in locale files (CONTRIBUTING.md §10):** `cs.json` has `"action.clear_search": "Clear Search Highlights"` (English), and `uk.json` has `"cmd.clear_search"` / `"cmd.clear_search_desc"` left in English. Real translations are required for all keys in all locales.

---

sinelaw:
Thanks for the quick follow-up — the locale fixes look great! Please take another look at [CONTRIBUTING.md](https://github.com/sinelaw/fresh/blob/master/CONTRIBUTING.md) for the two remaining test issues:

- **`test_clear_search_action`** calls `harness.count_search_highlights()`, which is an internal model accessor. Per §2, e2e tests must assert only on rendered output — please replace this with a screen-content assertion (e.g. `assert_screen_contains`) or move the invariant to a unit test.
- **`test_plugin_has_active_search`** uses `harness.sleep(Duration::from_millis(200))` in two places. Per §3, fixed timers are not allowed — please replace both with `wait_until(...)` that waits for a specific observable state change.

---

sinelaw (changes requested):
Re-review. The feature itself is good — I reproduced it interactively (tmux): searching `hello` highlights both matches and shows `Found 2 matches for 'hello'`; running **Clear Search Highlights** from the palette removes the highlight overlays (verified via SGR capture: the `48;5;226m` background on the match disappears) and clears the status. Architecture is clean and minimal: it reuses `clear_search_highlights()` and follows the existing Action/Command/keybinding pattern. The earlier locale and §2/§3 follow-ups are mostly addressed, and `test_plugin_has_active_search` passes.

Two blocking issues remain:

**1. `test_clear_search_action` fails and doesn't actually test the fix (CONTRIBUTING §1 + §3).**
It fails deterministically for me (3/3 runs) at `assert_screen_contains("Found 2")`. `Found N matches` is a *transient* status that `editor_tick` clears — your sibling `test_search_f3_navigates_all_matches_after_scroll` only survives this because its large file changes the timing. With this 4-line file the message is already gone by assert time, and a `wait_until(... "Found 2")` hangs indefinitely because the transient never reappears.

Worse, the post-clear assertion `!screen.contains("Found 2")` is vacuous: `Found 2` is absent regardless of whether `ClearSearch` does anything, so it proves nothing. Please assert on something stable that actually reflects cleared highlights (the `test_plugin_has_active_search` flow — confirm `hasActiveSearch()` flips true→false around the command — is a good model, or assert on the match-highlight overlay rather than the transient status line).

**2. `fresh.d.ts` was not regenerated (CONTRIBUTING §6).**
The PR adds the `hasActiveSearch()` plugin API but `crates/fresh-editor/plugins/lib/fresh.d.ts` is stale — running `cargo test -p fresh-plugin-runtime write_fresh_dts_file -- --ignored` adds the missing `hasActiveSearch(): boolean;`. TS plugin authors (and `check-types.sh`) won't see the method without it. (CI's `schema` job only diffs `config-schema.json`, so it doesn't catch this.)

Minor (non-blocking): `has_active_search()` is the only method in the impl block without a `#[plugin_api]` attribute — it works (the macro auto-exports it, `bool`→`boolean`), but adding the attribute matches its neighbors.

---

sinelaw:
Thanks for the work on this — the feature itself is solid and I confirmed it end-to-end in tmux (search highlights appear, and **Clear Search Highlights** correctly removes the overlays). Just two small things left before merge, noted in the review above: `test_clear_search_action` currently fails because it asserts on the transient `Found N matches` status, and `fresh.d.ts` needs regenerating to pick up `hasActiveSearch()`. Happy to help if anything's unclear!

---

sinelaw:
:fire: :fire:

PR #2153, opened 2026-05-27
feat: Hide current-line highlight when text is selected

masmu:
## Description
This PR introduces a new editor configuration setting `hide_current_line_on_selection`.

By default, the `current_line_bg` highlight remains visible even when the user is selecting text, which can sometimes clutter the interface or reduce the visual contrast of the selection.

By enabling `hide_current_line_on_selection = true` in the configuration (defaults to `false`), the background highlighting of the active line is dynamically suppressed as soon as any cursor has a non-empty selection range. Once the selection is cleared, the current-line highlight reappears.

## Changes
- Added `hide_current_line_on_selection` to `EditorConfig` and `PartialEditorConfig` (under the `Display` section).
- Passed the configuration down into the split rendering orchestration.
- Suppressed `effective_highlight_current_line` conditionally by checking the selection state of the split's cursors.


sinelaw (changes requested):

Manually tested with `hide_current_line_on_selection = true`: the current-line background highlight (`#121212` dark gray) correctly disappears from both the gutter and unselected text when a selection is active, reverting to the normal line background. Feature works as described.

Two CONTRIBUTING.md violations need to be addressed before merge:

- **Missing test (§1):** Every new feature must be backed by a test that fails without the change and passes with it. `crates/fresh-editor/tests/e2e/rendering.rs` already contains a directly analogous test for `highlight_current_line` (line 668) — please add a similar e2e test for `hide_current_line_on_selection` that selects text and asserts the current-line bg is absent.
- **Missing schema regeneration (§6):** `plugins/config-schema.json` is missing the new `hide_current_line_on_selection` field. Please run `./scripts/gen_schema.sh` and commit the updated schema file.

---

sinelaw:
🥳

Issue #2155, opened 2026-05-28
[Bug] Pasting text freezes editor for a long time


EtienneMaire37:
When pasting any text (to anywhere in a file it seems) with Ctrl+V or Ctrl+Shift+V the editor freezes for up to 2-3 minutes.
I use fresh 0.3.9 on alacritty 0.15.1.
I am on debian Trixie and use the deb package version from the "Quick Install" command, though I've tested the flatpak package and it has the same bug.

sinelaw:
Thanks for reporting - I haven't seen this before and can't reproduce it.

1. does the same problem happen when using the Edit... Paste menu?
2. does it depend on the size of the paste (e.g. paste just a few words also gets stuck)?
3. can you paste directly into your terminal without opening Fresh and check the paste also?

I tried it in alacritty and other terminals and have not hit this problem...

EtienneMaire37:
After a bit more testing I found that it only happens after I run a command the terminal which opens another window (in my case qemu), but it is not everytime and it is quite inconsistent.
Sometimes the problem happens not after I run a command which opens another window but after I Alt+Tab...
I am using gnome 48. 
The problem seems to only happen with the LSP server running, though I did not test it thoroughly when it is disabled.
Quitting and starting fresh again fixes the problem until the bug starts occurring again.
After it happens once, there doesn't seem to be a way to get rid of it and it will happen on any paste until I restart fresh.

Here is my window configuration in case it has anything to do with it:
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/a32236e9-b3c1-4dd1-986a-c1ca9bdb0ea1" />

sinelaw:
Reproduced by creating a script that holds on to the clipboard:

```
cat > /tmp/dead_clip.py <<'EOF'
# /// script
# requires-python = ">=3.10"
# dependencies = ["python-xlib"]
# ///
import time
from Xlib import display, X
d = display.Display()
w = d.screen().root.create_window(0, 0, 1, 1, 0, d.screen().root_depth,
                                  X.InputOutput, X.CopyFromParent,
                                  event_mask=X.PropertyChangeMask)
w.set_selection_owner(d.intern_atom('CLIPBOARD'), X.CurrentTime)
d.flush()
print("Holding CLIPBOARD; ignoring all SelectionRequests", flush=True)
while True:
    if d.pending_events():
        d.next_event()  # drain, do NOT reply
    time.sleep(0.1)
EOF
[...]

PR #2159, opened 2026-05-28
Implement async system clipboard paste with visual anchors

sinelaw:
## Summary

This PR implements asynchronous system clipboard reading for paste operations, allowing the editor to remain fully interactive while waiting for clipboard data. Instead of blocking on the system clipboard, pastes now spawn a background thread that reads the clipboard and delivers results through the async message bridge. A visual placeholder ("▍") anchored to the buffer marks each paste location and floats with edits via the marker tree.

## Key Changes

- **Async paste infrastructure**: Added `PendingPaste` and `PasteAnchor` structures to track in-flight clipboard reads, with a 500ms deadline to prevent indefinite hangs
- **Placeholder anchors**: Each paste location gets a virtual text "▍" marker that automatically adjusts position as the user edits, ensuring pasted text lands at the correct location even if the buffer changes during the read
- **Deadline management**: Implemented `check_paste_deadline()` to cancel pastes that exceed the timeout window, with cleanup of placeholder markers
- **Input dispatch consolidation**: Moved all raw terminal event handling into `Editor::handle_input_event()` so input is dispatched immediately even while pastes are pending, rather than being queued
- **Buffer lifecycle integration**: Pastes anchored in a closed buffer are automatically discarded to prevent insertion into freed memory
- **Column-mode paste support**: Multi-cursor pastes distribute lines across cursors when the clipboard contains the same number of lines as active cursors
- **Line ending normalization**: Clipboard text is normalized to LF, then converted back to the buffer's configured line ending (LF/CRLF/CR) at insertion time

## Implementation Details

- Background clipboard reads use `arboard::Clipboard::new().get_text()` on dedicated threads, capturing the OS clipboard state at the moment the thread reaches the read call
- The `NEXT_PASTE_REQUEST_ID` atomic counter ensures each paste gets a unique identifier for matching results to pending entries
- Paste selection deletions are applied immediately (before the async read) so the insertion point is finalized, then the async result inserts at the anchor's current position
- The main loop's tick function checks paste deadlines before processing async messages, allowing a timeout fallback and a late result in the same tick to be handled correctly
- Defensive fast-paths skip async reads for prompt/terminal/file-explorer modes and when no async bridge is available
- Concurrent paste limit (64) prevents runaway macros from exhausting system resources

## Localization

Added "clipboard.pasting" message key across all supported language files to indicate paste-in-progress state.

https://claude.ai/code/session_01JyGCnVXRUyWDh8o6X9CC9g


sinelaw:
Fixes #2155

PR #2175, opened 2026-05-31
fix(scroll-panel): align focus offsets with render width when scrollbar is present

masmu:
### Description

This PR fixes a bug where jumping to a target item in a `ScrollablePanel` (such as navigating to a search result in the settings) would often leave the target item completely off-screen.

#### The Problem
When a scrollbar is needed, `ScrollablePanel::render()` reduces the available content width by 1 column (`area_width - 1`). However, `ensure_focused_visible` was performing its Y-offset calculations using the full `area_width`. 

For elements that wrap their text differently based on width (like long setting descriptions), the narrow width makes them taller. Because the offset calculator assumed the wider, "shorter" items, the cumulative Y position of items deep in the list drifted out of sync with the render engine, resulting in an incorrect viewport jump.

#### The Fix
- `update_content_height` now does a two-pass calculation: it first checks if the content overflows at the full width, and if so, it re-computes the layout at `area_width - 1`.
- A new `content_width()` helper exposes the effective render width.
- `ensure_focused_visible` now calls `update_content_height` to determine the scrollbar's presence and uses `content_width()` to calculate offsets. Now the offset calculator and the renderer agree perfectly.

#### Testing
- Added `test_ensure_focused_visible_uses_render_width_when_scrollbar_present` to prevent regression.
- Implemented a blanket `ScrollItem` trait for function pointers (`fn(u16) -> u16`) in the test module to allow concise, closure-based mocking of width-dependent items.


sinelaw:

- **`cargo fmt` fails**: trailing whitespace on line 251 in `update_content_height`, and `item_y_offset`'s signature needs multi-line formatting — CONTRIBUTING.md requires `cargo fmt` to pass per-commit; run `cargo fmt -p fresh-editor` to fix
- **Incomplete caller update**: `item_y_offset`'s parameter was renamed to `content_width` (docstring says "effective content width") but its two direct callers in `state.rs` (lines 532 and 635) still pass the raw `area_width` — they should call `self.scroll_panel.content_width(width)` instead, otherwise the "snap section to top" code paths exhibit the same root bug on narrow viewports with a scrollbar present
- **Redundant `update_content_height` calls**: `ensure_focused_visible` now calls `update_content_height` internally, but several call sites in `state.rs` (e.g. lines 521→537, 929→946, 2521→2524) pre-call `update_content_height` then immediately call `ensure_focused_visible` — the pre-calls are now redundant and can be removed

The core fix in `ensure_focused_visible` and the two-pass height calculation are correct. The regression test is well-designed and would fail on master. Please address the `cargo fmt` and the incomplete caller update before merge.

---

sinelaw (changes requested):
Requesting changes: `cargo fmt` fails, two `item_y_offset` callers in state.rs still pass area_width instead of content_width (leaving the snap-to-top paths with the same underlying bug), and several pre-calls to `update_content_height` before `ensure_focused_visible` are now redundant. Details in the issue comment.

---

sinelaw:
Needs fmt

sinelaw:
:fire:

PR #2241, opened 2026-06-04
feat(file-explorer): add non-breaking slot override API for icons, status, and name color

Agrejus:
## Motivation

The file explorer already had useful native behavior, but the rendering surface was still too rigid for the kind of plugin-driven customization I was hoping for. The goal of this PR is to make the explorer more flexible without breaking existing behavior: plugins should be able to change filename colors, add custom glyph/icon content, and override trailing status content/tooltips while the default explorer experience continues to work unchanged unless someone opts into the new API.

## Summary

- add an additive, non-breaking file explorer slot override API for leading content, trailing content, and filename color
- preserve the existing compatibility behavior by default so current users and plugins keep the same explorer rendering unless they opt in
- allow explicit suppression/override of default leading slot, trailing slot, and name color behavior on a per-path basis
- wire the new slot API through the editor, plugin runtime, generated typings, documentation, and targeted tests
- keep recent per-window file explorer async/update behavior intact while merging with upstream explorer changes

## Compatibility

- default file explorer behavior is unchanged unless a plugin calls the new slot API
- slot overrides are path-scoped and plugin-namespaced to avoid accidental bleed-through or cross-plugin collisions
- filename color overrides can coexist with custom badges/glyphs instead of forcing one rendering path to win globally

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo check -p fresh-editor`
- [x] `cargo test -p fresh-core --lib`
- [x] `cargo test -p fresh-plugin-runtime --lib`
- [x] `cargo test -p fresh-editor --lib`
- [x] `cargo test -p fresh-editor --test e2e_tests e2e::file_explorer::test_file_explorer_duplicate_refreshes_git_decorations -- --exact`
- [x] `cargo test -p fresh-editor --test e2e_tests e2e::file_explorer::test_file_explorer_git_status_async_update_stays_with_origin_window -- --exact`

## Demo

Demo using the new slot overrides. Icons and colors come from a plugin using the new overrides. They are not provided by default. Only extensibility is provided 

image


sinelaw (changes requested):

The slot override API concept is solid and the implementation is well-structured, but there are three issues that need to be addressed before this can merge:

- **CONTRIBUTING.md #5 violation (blocking):** `compute_file_explorer_git_status_cache` in `app/file_explorer.rs` uses `std::process::Command::new("git")` directly. The guideline explicitly requires using `ProcessSpawner` (see how `resolve_git_index` in `file_operations.rs` does it with `self.authority.process_spawner`). The current workaround (clear cache on remote connections and return early) avoids a crash but silently disables git status on SSH/remote sessions, while the ProcessSpawner pattern would make it work transparently on both local and remote.

- **Backward-compatibility claim is incorrect (blocking):** The PR description states "default file explorer behavior is unchanged unless a plugin calls the new slot API," but both `show_file_icons` and `color_git_status_names` default to `true`. Manually tested on the PR branch: 2-letter file type abbreviations (MD, RS, TS, {}, TL, PK…) appear in the leading slot for all users immediately on upgrade, and directory/file names are colored by git status. These are non-trivial visible changes that affect every existing user. These new defaults should either be `false`, or the compatibility claim needs to be removed and the behavior change documented as intentional.

- **Missing e2e test for the new API (blocking per CONTRIBUTING.md Testing #2):** There is no end-to-end test that calls `setFileExplorerSlots` from a plugin and asserts on the rendered output. The tests added cover internal git-status cache routing (window isolation) but not the user-visible slot override rendering. CONTRIBUTING.md requires: "Any new user flow must include an end-to-end test that drives keyboard/mouse events and asserts only on rendered output."

---

sinelaw:

*(supersedes the two previous autogenerated comments on this PR)*

Three issues to address before merging:

- **`ProcessSpawner` violation:** `compute_file_explorer_git_status_cache` calls `std::process::Command::new("git")` directly. CONTRIBUTING.md §5 requires using `ProcessSpawner` for all process spawning so features work on remote/SSH sessions. The existing `resolve_git_index` in `file_operations.rs` shows the correct pattern.

- **Two features belong in plugins, not Rust:** The embedded `git_explorer.ts` plugin already provides M/A/U/D badges and parent-directory status via `setFileExplorerDecorations`; `decorations.rs` on master contains no git logic. The new `git_status.rs` + background task duplicate that pipeline in Rust, and are shadowed by the plugin in normal use. Similarly, `file_icons.rs` is a static extension→glyph table with no performance requirement. Both belong as TypeScript plugins using the new `setFileExplorerSlots` API this PR introduces: `git_explorer.ts` can be extended to also call `setFileExplorerSlots` for `nameColor`; a new `file_icons.ts` plugin can handle leading-slot icons. This eliminates the `ProcessSpawner` violation and removes ~650 lines of new Rust.

- **Default behavior change:** `show_file_icons: true` and `color_git_status_names: true` add visible UI changes for all existing users on upgrade, contradicting the stated compatibility guarantee. If these features move to opt-in plugins the config keys and defaults can be removed entirely.

- **Missing e2e test:** CONTRIBUTING.md §Testing requires a rendered-output end-to-end test for any new user flow. No e2e test exists for `setFileExplorerSlots`.

---

Agrejus:
A few quick notes on what changed:

- The native Rust file explorer git pipeline is gone. `git_status.rs` and the background git cache work were removed, so the earlier `Command::new("git")` issue from the file explorer path should no longer apply.
- The native Rust file icon implementation is also gone. `file_icons.rs` was removed.
- Git status in the explorer is still plugin-driven. `git_explorer.ts` continues to provide the badges, and it now also uses `setFileExplorerSlots(...)` for optional filename coloring.
- There is no bundled default file-icons behavior anymore. The temporary `file_icons.ts` plugin was removed as well, so the slot API is there for user plugins to take advantage of, but Fresh itself does not ship default leading icons from this PR.
- The old config-driven visual changes were removed. `show_file_icons` and `color_git_status_names` are gone, so existing users should keep the same default explorer behavior unless a plugin opts into the new slot API.
- There is now an e2e test covering `setFileExplorerSlots` at `tests/e2e/plugins/file_explorer_slots.rs`.

So the intended end state here is: the API for leading content, trailing content, and name color exists, but default explorer behavior stays the same unless a plugin uses it.

Agrejus:
## CI fixes

**Formatting** — `cargo fmt` was failing because `tests/e2e/plugins/mod.rs` had modules out of alphabetical order (`file_explorer_slots` before `env_manager`). Reordered to satisfy the formatter.

**Schema** — `config-schema.json` was drifting between macOS and Linux CI because two defaults are platform-specific: `active_keybinding_map` (`"macos"` vs `"default"`) and `process_limits.enabled` (`false` vs `"true"`). Added explicit `#[schemars(default = ...)]` helpers so schema generation is consistent, then regenerated the schema.

**`test_file_explorer_git_markers_clear_after_external_commit`** — The poller detected `.git/index` mtime changes but never fired `after_file_explorer_change`, so `git_explorer` never rescanned and the "M" markers stayed forever. Fixed by calling `notify_file_explorer_change` when the index mtime changes.

**`test_file_explorer_slots_leading_override_renders`** — On Windows, plugin paths from `pathJoin` use forward slashes while the explorer tree uses canonical native paths (`C:/...` vs `C:\...`). Slots were silently dropped by `starts_with` / cache lookup, so `foo.txt` showed but `◆` never did and the test timed out. Fixed with `normalize_explorer_plugin_path()` for slot and decoration paths, plus a bit of test hardening (drain plugin commands after init, `wait_until_stable`).

sinelaw:
Looks good, needs rebase

Agrejus:
Rebased 😄

Agrejus:
I checked the CI failure, looks like a flaky test @sinelaw ?

sinelaw:
👍

Issue #1939, opened 2026-05-11
thread 'main' (506036) panicked at crates/fresh-editor/src/app/render.rs:841:58: called `Option::unwrap()` on a `None` value


zdooder:
Running under WSL2 on Windows 11.  I just did some wierd stuff to my git checkout, which might be related (I had rebased from the wrong branch, so renamed the "bad" branch, recreated a "good" version, cherry-picked the commits I wanted onto the new branch, and deleted the "bad" branch).  

I'll obviously try to figure out what is poison about my current environment, but thought you'd want to know about the crash so that it can hopefully be avoided.

```
$ fresh --version
fresh 0.3.5
$ fresh .

thread 'main' (506036) panicked at crates/fresh-editor/src/app/render.rs:841:58:
called `Option::unwrap()` on a `None` value
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
$ RUST_BACKTRACE=full fresh .

thread 'main' (506232) panicked at crates/fresh-editor/src/app/render.rs:841:58:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0:     0x5f3450cb9039 - <unknown>
   1:     0x5f345044eee4 - <unknown>
   2:     0x5f3450cb8491 - <unknown>
   3:     0x5f3450cb86da - <unknown>
   4:     0x5f3450cb81e3 - <unknown>
   5:     0x5f3450cde1b8 - <unknown>
   6:     0x5f3450cde139 - <unknown>
   7:     0x5f3450cde70c - <unknown>
   8:     0x5f345044ed7b - <unknown>
   9:     0x5f3450450561 - <unknown>
  10:     0x5f3450450548 - <unknown>
  11:     0x5f3450639795 - <unknown>
  12:     0x5f34503cffaa - <unknown>
  13:     0x5f34503e34c2 - <unknown>
  14:     0x5f34503d9fde - <unknown>
  15:     0x5f34503b2ddf - <unknown>
  16:     0x5f34503e85c9 - <unknown>
  17:     0x73309422a1ca - __libc_start_call_main
                               at ./csu/../sysdeps/nptl/libc_start_call_main.h:58:16
  18:     0x73309422a28b - __libc_start_main_impl
                               at ./csu/../csu/libc-start.c:360:3
  19:     0x5f34503ad62e - <unknown>
  20:                0x0 - <unknown>
$
```

I'm not sure what else might be helpful.

[...]

PR #2266, opened 2026-06-07
Fix orphaned split-leaf blank-pane bug on window restore

sinelaw:
## Summary

Fixes issue #1939 where restoring a window with an empty `[No Name]` active tab alongside resolvable non-file tabs (e.g., terminals) results in a blank pane, even though the status bar reports a live buffer.

## Root Cause

When restoring a window:
1. A fresh window is seeded with an empty `[No Name]` buffer
2. The empty buffer was never persisted (empty unnamed buffers get no recovery id)
3. Restoring non-file tabs (terminals, content-bearing unnamed buffers) creates new buffers without replacing the seed
4. The restore logic fails to resolve the saved active tab (since it wasn't persisted) and leaves the split-manager leaf pointing at the throwaway seed buffer
5. `clean_orphaned_buffers` removes the seed since it's not in any tab list, leaving the leaf dangling at a dead `BufferId`
6. The render path paints a blank pane while `effective_active_pair` falls back to another buffer for status queries

## Changes

**crates/fresh-editor/src/app/workspace.rs:**
- Added fallback logic in `restore_split_view_state` to resolve the active leaf to the first surviving tab when the saved active tab cannot be resolved
- This ensures the split-manager leaf, view state, and tab list all agree on which buffer is active
- Prevents the seed buffer from being orphaned and subsequently removed

**crates/fresh-editor/tests/e2e/workspace.rs:**
- Added regression test `test_restore_orphaned_active_unnamed_tab_renders_surviving_tab` that reproduces the bug using a content-bearing unnamed buffer as the resolvable background tab
- Verifies that the restored pane renders the surviving tab's content (not blank) and that the split tree remains consistent

**docs/internal/orphaned-leaf-investigation.md:**
- Added comprehensive investigation document explaining the symptom, root cause, fix rationale, and adjacent observations

## Implementation Details

The fix applies the fallback at the right layer — repairing the pane-buffer invariant at its source rather than masking it in the render or status-bar paths. This preserves existing behavior for edge cases (e.g., when all tabs are unresolvable) while ensuring the split tree remains consistent across save/restore cycles.

https://claude.ai/code/session_01Rz9XEuj24fycfgLood6QVT

### DIFF
--- a/crates/fresh-core/src/api.rs
+++ b/crates/fresh-core/src/api.rs
@@ -2235,6 +2235,9 @@ pub enum PluginCommand {
         cwd: Option<String>,
         command: Option<Vec<String>>,
         title: Option<String>,
+        /// Restore-time argv (agent resume); see
+        /// `CreateWindowWithTerminalOptions::resume`.
+        resume: Option<Vec<String>>,
         request_id: u64,
     },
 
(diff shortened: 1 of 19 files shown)


sinelaw:
Fixes #2267.

This repairs the "orphaned split-leaf" blank-pane bug at its source: when a session window is restored with an empty `[No Name]` buffer as the saved active tab (alongside a resolvable non-file tab such as a terminal), the active split leaf was left pointing at the throwaway seed buffer, which `clean_orphaned_buffers` then removed — leaving the leaf dangling at a dead `BufferId` and the pane rendering blank.

The fix makes `restore_split_view_state` fall back to the first surviving tab when the saved active tab can't be resolved, so the split tree, the view state, and the tab list all agree before orphan cleanup runs. Includes a PTY-free regression test (`test_restore_orphaned_active_unnamed_tab_renders_surviving_tab`) that fails without the fix and passes with it, plus a root-cause writeup in `docs/internal/orphaned-leaf-investigation.md`.

---

PR #2273, opened 2026-06-08
add smali highlight

### LINKED ISSUE #2265: smali highlight
https://github.com/tree-sitter-grammars/tree-sitter-smali

asukaminato0721:
fix #2265

I don't use https://github.com/tree-sitter-grammars/tree-sitter-smali/blob/master/Cargo.toml because this use 0.20.10 while tree-sitter-smali-valonsodev using 0.26


sinelaw:
Thanks! Same note as on #2274: we're avoiding new bundled tree-sitter grammars in favor of syntect highlighting + the indent-rules family tier — see [docs/development/adding-languages.md](https://github.com/sinelaw/fresh/blob/master/docs/development/adding-languages.md). (FYI the `indents.scm` here uses `@indent.begin/.end` captures, which Fresh's indenter doesn't read, so it has no effect.)

---

sinelaw:
👍

PR #2274, opened 2026-06-08
feat: add fish syntax highlighting and auto-indentation 

### LINKED ISSUE #2272: fish shell support
now .fish file will be highlighted as bash file.

https://github.com/ram02z/tree-sitter-fish

asukaminato0721:
fix #2272

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1481,7 +1481,6 @@ name = "fresh-languages"
 version = "0.4.0"
 dependencies = [
  "tree-sitter",
- "tree-sitter-fish",
  "tree-sitter-go",
  "tree-sitter-highlight",
  "tree-sitter-javascript",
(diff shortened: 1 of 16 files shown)


sinelaw:
Thanks! Heads-up that we're moving away from bundling tree-sitter grammars (they add large parse tables to the binary) toward syntect highlighting + the regex indent-rules family tier. Could you redo Fish along those lines? The how-to is here: [docs/development/adding-languages.md](https://github.com/sinelaw/fresh/blob/master/docs/development/adding-languages.md).

---

sinelaw:
(Automatic comment)
The highlighting mostly works, but indented control keywords (`if`/`for`/`end`) seem to render in the command color rather than as keywords — only column‑0 ones look right. I might be wrong, but maybe the `commands` rule's `(^|\s)` eats the leading space and out‑ranks `keywords` via syntect's leftmost‑match? e.g.:

```fish
function greet
    if test -n "$name"   # "if" looks like a command, not a keyword
    end
end
```

sinelaw:
:fire:

PR #2277, opened 2026-06-08
fix(win32): try to attach parent process's stdio, otherwise force GUI…

### LINKED ISSUE #2276: TUI mode won't work on `target_os = "windows"` `-gui` ships
On Windows, executables linked with the `WINDOWS` subsystem do not have standard I/O streams attached, so any error messages written to stdout/stderr are effectively invisible to the user.

At the moment, when `fresh` is started without `--gui`, it still attempts to initialize the TUI on Windows. If that initialization fails, the application may crash without displaying any useful error message, leaving users with no indication of what went wrong.

I would suggest either:

1. Ignoring the `--gui` flag on Windows and always using the GUI implementation; or
2. Finding a documented, non-tricky way to hide the console window only when `--gui` is explicitly specified, while keeping console output available in the default case.

This would improve the user experience significantly by ensuring that startup errors remain visible and debuggable.

mokurin000:
Fix #2276

Try to attach console from parent process, if that failed, force to use GUI mode.

Otherwise, if the `--gui` argument was not passed, TUI mode in entered.

## What works now

- `--help`
- GUI mode in explorer.exe by just double-clicking
- TUI mode in terminal, with `subsystem=windows`

## TODO

- [ ] write integration test runner on Windows, with `subsystem=windows` launcher and `subsystem=console` launcher, to test startup behavior

### DIFF
--- a/crates/fresh-editor/src/main.rs
+++ b/crates/fresh-editor/src/main.rs
@@ -3367,7 +3367,10 @@ fn dump_config_command(args: &Args) -> AnyhowResult<()> {
 /// Route non-interactive subcommands. Returns `Some(result)` if a subcommand
 /// was handled so the caller can propagate it; returns `None` for an
 /// interactive editor launch.
-fn run_if_subcommand(args: &Args, console_available: bool) -> Option<AnyhowResult<()>> {
+fn run_if_subcommand(
+    args: &Args,
+    #[cfg(feature = "gui")] console_available: bool,
+) -> Option<AnyhowResult<()>> {
     if args.show_paths {
         return Some(show_paths_command());
     }
@@ -3666,7 +3669,7 @@ fn real_main() -> AnyhowResult<()> {
     // Early check, otherwise terminal check would fail.
     #[cfg(all(windows, feature = "gui"))]
     let console_available = unsafe { AttachConsole(ATTACH_PARENT_PROCESS) != 0 };
-    #[cfg(not(all(windows, feature = "gui")))]
+    #[cfg(all(not(windows), feature = "gui"))]
     let console_available = true;
 
     // Enable backtraces for error reporting if not already set.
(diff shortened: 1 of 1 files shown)


mokurin000:
@sinelaw Would you please take a review when you have time? I did implemented like the planned method but slightly different, that forces to start GUI mode when `--gui` was not passed but no parent process console was available.

mokurin000:
Will quickly fix the cross platform issue, I coded it just for windows

update: force-pushed with latest fixes

sinelaw (changes requested):

The premise is correct — `windows_subsystem = "windows"` prevents console attachment, `AttachConsole(ATTACH_PARENT_PROCESS)` is the right Windows API, and the GUI fallback logic is sound. `windows-sys` with `Win32_System_Console` is already a project dependency so no new dep is introduced. The approach is clean and minimal.

Two issues need fixing before this can merge:

- **Compiler warning (blocks merge per CONTRIBUTING.md):** `console_available` is unused in non-`gui` feature builds because the `if !console_available || args.gui` block is inside `#[cfg(feature = "gui")]`. Running `cargo check --all-targets` (required by CONTRIBUTING.md) produces `warning: unused variable: console_available`. The project already uses the `#[cfg_attr(not(target_os = "linux"), allow(unused_variables))]` pattern for exactly this situation (see `main.rs:1742`). Apply the same here: annotate the `run_if_subcommand` parameter with `#[cfg_attr(not(feature = "gui"), allow(unused_variables))]`, or restructure so the parameter isn't passed at all in non-gui builds.

- **Missing test (required by CONTRIBUTING.md):** Every bug fix must include a test that fails without the change and passes with it. Even a minimal integration test that verifies the `console_available` detection path (e.g., mocking `AttachConsole` failure and asserting GUI mode is selected) would satisfy this.

---

mokurin000:
@sinelaw, thank you for your review. I’ve fixed the compiler warning by only introducing the argument when `gui` is enabled.

Regarding the integration test, using `EditorState` could be inefficient, since the code changes are limited to executable initialization. As mentioned in https://github.com/sinelaw/fresh/issues/2276#issuecomment-4664464371, I believe both the design approach and the use of executable-based integration testing (instead of session mocking) would benefit from further discussion among maintainers before moving forward.

sinelaw:
🔥

PR #2296, opened 2026-06-10
feat(languages): Assembly support via asm-lsp

### LINKED ISSUE #1964: Feature request: ASM (Assembly) file support via asm-lsp
### Discussed in https://github.com/sinelaw/fresh/discussions/1925

<div type='discussions-op-text'>

<sup>Originally posted by **viti95** May 10, 2026</sup>
Hi! It would be great if fresh-editor could add support for ASM (Assembly) files.
Maybe integrating with asm-lsp could be a good approach, since it already provides language server features for assembly languages. Thanks for the great project!</div>

sinelaw:
Closes #1964

Assembly files previously opened as plain text (`.asm`) or were mis-highlighted as **R** (`.s`/`.S` — syntect's built-in R grammar claims those extensions), and no language server was wired for assembly.

## What's in here

- **New `Assembly` sublime-syntax grammar** (hand-written, ~120 lines) covering the dialects [asm-lsp](https://github.com/bergercookie/asm-lsp) serves: GAS/AT&T and Intel-style NASM/MASM across x86, x86_64, ARM and RISC-V. Mnemonics are highlighted positionally (first word of an instruction line) rather than enumerating every ISA, keeping the grammar small and architecture-agnostic. Registered in `types.rs` and the build-time syntax packdump.
- **Two language entries sharing the one grammar**, mirroring how Verilog/SystemVerilog share svls:
  - `asm` (`.asm`/`.nasm`) — Intel/NASM, `;` comments
  - `gas` (`.s`/`.S`) — GAS/AT&T, `#` comments

  The split exists so comment toggling produces valid code in each dialect (`;` is a statement separator, not a comment, in x86 GAS).
- **asm-lsp wired as the default LSP** for both entries — opt-in per project (`auto_start: false`) like the other niche servers, root markers `.asm-lsp.toml` / `.git`.
- **`asm-lsp.ts` helper plugin** (modeled on `nim-lsp.ts`):
  - When the server binary is missing, the status-bar click shows install instructions (`cargo install asm-lsp` / `cargo binstall asm-lsp`) and a disable action.
  - **Dialect config offer**: asm-lsp ignores the LSP languageId and defaults to GAS/x86-64, so NASM/MASM/ARM/RISC-V files get bogus diagnostics without a config (true in every editor today). When an assembly buffer opens and no project or user-global `.asm-lsp.toml` exists, the plugin guesses the assembler and instruction set from the buffer contents and offers to generate the project config — written only after the user picks an option, with alternates, "not now", and a persisted per-project "don't ask again". Accepting restarts the Assembly LSP so the dialect applies immediately. For Intel-syntax dialects the generated config also disables asm-lsp's default diagnostics (they shell out to gcc/clang, which only assemble GAS syntax).

## Testing

- `test_highlight_asm_nasm` / `test_highlight_asm_gas` coverage tests with new `hello.asm` / `hello.s` fixtures — fail without the grammar.
- `test_asm_language_detection` pins `.asm`→`asm` and `.s`→`gas`, so the config override keeps beating syntect's R grammar on the `s` extension (this is the reproducer for the prior misdetection).
- `asm_lsp_config` e2e tests drive the config-offer popup against the real plugin: dialect detection for NASM and GAS fixtures, the generated config contents, and no offer when a config already exists.
- Manually validated in tmux with asm-lsp v0.10.1 installed: highlighting, comment toggling per dialect, server start, hover docs (`RDI`/`mov`), completions (`pu` → `push`...), and the config offer end-to-end — accepting it writes `.asm-lsp.toml`, restarts the server, and the bogus GAS diagnostics on NASM files disappear.
- `cargo fmt`, `cargo clippy --all-targets` (no new warnings), plugin type-check on `asm-lsp.ts` passes.

https://claude.ai/code/session_01A3AGNQNhKCT4489AWthWq6


sinelaw:
When using LSP hover on a nasm file (included hello.asm) I get

2026-06-14 18:32:05.209 WARN fresh::services::lsp::async_handler: LSP response error from 'asm-lsp' (asm): No information available (code -32803)

PR #2305, opened 2026-06-10
feat: add Smali syntax highlighting and auto-indentation

### LINKED ISSUE #2265: smali highlight

sinelaw:
## Summary

Closes #2265.

Adds full editing support for Smali — the human-readable assembly form of Dalvik/Android bytecode (`.smali` extension). Previously these files were detected as plain text (syntect ships no grammar) with no language-aware indentation.

## 1. Syntax highlighting

New embedded `smali.sublime-syntax` grammar, registered the same way as every other bundled-but-not-in-syntect language (TOML, Zig, Nim, C3, …):

- directives (`.class`, `.method`, `.end method`, `.field`, `.line`, `.local`, `.annotation`, …)
- instruction mnemonics with `-suffix`/`/variant` tails (`invoke-virtual/range`, `iget-object`, `add-int/2addr`, `const-string`, `if-eqz`, …)
- type descriptors (`Lcom/example/Foo;`, primitives `V Z B S C I J F D`, array-prefixed forms)
- registers (`v0`, `p1`), labels (`:cond_0`, `:goto_0`), access modifiers, strings, numbers, comments

Registered in both `types.rs` (`add_embedded_grammars`) and `build.rs` (the build-time packdump generator), which must stay in sync.

## 2. Auto-indentation

Smali has no tree-sitter grammar and previously had no entry in the regex indent-rules tier, so it fell through to the universal bracket heuristic — which only reacts to `{ } [ ] ( ) :`, none of which delimit Smali blocks. Pressing Enter after a `.method` header left the body at column 0.

Added a `SmaliLike` family to the regex indent-rules tier:
- **increase** after block-opening directives that always have a matching `.end …`: `.method`, `.annotation`, `.subannotation`, `.array-data`, `.packed-switch`, `.sparse-switch`
- **decrease** on any `.end …` line
- `.field` is deliberately excluded (usually a one-liner). Comment/string scope masking (already provided by the tier) prevents a directive inside a `#` comment from triggering.

## Testing

- `test_smali_grammar_loaded` — asserts `.smali` resolves to the `Smali` grammar (fails before the change, passes after).
- 6 new `indent_rules` tests covering method-body indent, other block directives, body maintenance, `.end` not opening a block, typed-`.end` dedent, and syntect-name resolution.
- `cargo test -p fresh-editor --lib primitives::grammar` → 41 passed; `… indent_rules` → 35 passed.
- `cargo fmt` / `cargo clippy` clean for touched code.
- **Manual validation (tmux):** opened `.smali` files in the editor — status bar reports the language as **Smali**, all token classes highlight correctly, and pressing Enter after `.method public foo()V` auto-indents the body to 4 spaces with subsequent body lines maintaining indent.

https://claude.ai/code/session_012ACTacoPpLGVZmUgWnTHsn

### DIFF
--- a/crates/fresh-editor/build.rs
+++ b/crates/fresh-editor/build.rs
@@ -368,6 +368,7 @@ fn generate_syntax_packdump() -> Result<(), Box<dyn std::error::Error>> {
         ("src/grammars/systemverilog.sublime-syntax", "SystemVerilog"),
         ("src/grammars/vhdl.sublime-syntax", "VHDL"),
         ("src/grammars/c3.sublime-syntax", "C3"),
+        ("src/grammars/smali.sublime-syntax", "Smali"),
     ];
 
     let mut loaded = 0;
--- a/crates/fresh-editor/src/grammars/smali.sublime-syntax
+++ b/crates/fresh-editor/src/grammars/smali.sublime-syntax
@@ -1,0 +1,89 @@
+%YAML 1.2
+---
+# Smali — the human-readable assembly format for Dalvik (Android) bytecode,
+# as produced by baksmali / used by smali. Files use the .smali extension.
+name: Smali
+file_extensions: [smali]
+scope: source.smali
+contexts:
+  main:
+    - include: comments
+    - include: directives
+    - include: access-modifiers
+    - include: labels
+    - include: opcodes
+    - include: types
+    - include: member-references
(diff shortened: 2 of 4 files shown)


sinelaw:
See #2273

Issue #2146, opened 2026-05-27
Markdown compose (preview) mode hides LSP diagnostic higlighting, gutters


sinelaw:
e.g. enabling harper-ls to find spelling mistakes
dosen't show up in the compose mode, but works without it

{
        "command": "harper-ls",
        "enabled": true,
        "name": "harper",
        "args": [
          "--stdio"
        ],
        "auto_start": false,
      }

sinelaw:
It works now, with a caveat that in compose/preview mode the highlighting can sometimes be hidden because of preview formatting (e.g. in tables it doesn't always show up, control characters, etc)

PR #2325, opened 2026-06-12
Inline image rendering via kitty graphics protocol (generic placeImage API)

Agrejus:
## Motivation

I wanted to open a README in Fresh and see it rendered roughly the way GitHub or Cursor would show it, instead of staring at raw markdown. Opening it in an external viewer works, but Fresh already has the rendering machinery to do this in place — so a command-palette action that previews markdown without ever leaving Fresh felt like a natural fit.

Rather than baking markdown support into the core, this PR exposes the underlying capability as a generic API: any plugin can place rendered images, whether that's markdown, diagrams, plots, or anything else it can rasterize. The markdown/mermaid preview is just the first consumer.

## Summary

- add `placeImage` / `clearImages` / `getGraphicsCapability` plugin APIs: plugins place raster images (PNG) anchored to buffer byte positions; the core reserves virtual placeholder rows and renders kitty Unicode placeholders over them, so images scroll with text and survive the cell-diff renderer
- detect graphics capability from the environment (kitty / WezTerm / Ghostty / Konsole; `FRESH_GRAPHICS` override), shared between core and plugin runtime so both always agree
- wrap transmit/delete escapes in DCS passthrough inside tmux (requires `allow-passthrough on`); placeholder cells ride through as normal text
- fix non-deterministic virtual-line ordering: equal-priority range-query results rendered in arbitrary order (an image's N reserved rows came back shuffled, displaying diagrams as torn bands); queries now tiebreak by insertion order
- add `clearConcealsInRangeForNamespace` so one plugin's per-line rebuild no longer destroys other plugins' conceals (same motivation as the namespaced overlay clear from #2146); markdown_compose switched to it
- make markdown_compose fence-aware: cache fenced-code-block ranges per buffer and skip markdown processing inside them (fixes `|`-leading code lines growing table borders); conceal fence/heading markers with cursor-line reveal
- render block quotes (`>` → dimmed `▌` bar, nesting included), list bullets (`•`), task checkboxes (☐ / ☑), horizontal rules (full-width `─`), and footnotes (`[^1]` → superscript ¹, with dimmed definition lines) in markdown_compose — these constructs were parsed for indentation but displayed as raw source

- render markdown tables as aligned, word-wrapped frames in markdown_compose: distribute column widths by water-filling (narrow label columns keep their natural width; only wide prose columns shrink), wrap long cells into virtual continuation rows so a row spans as many visual lines as its tallest cell — like a rendered README table — instead of truncating or splitting the source line, cap the frame at the compose width with a one-cell right margin so it neither stretches edge-to-edge nor overflows the final column, and count colored-circle/square status emoji (U+1F7E0..U+1F7EB, e.g. 🟡) and regional indicators as two cells so borders stay aligned; also harden the per-line compose loop (code-point slicing + try/catch) so one malformed line — e.g. an emoji sliced between its surrogate halves — can no longer throw and leave that line and every line after it rendered as raw markdown

## Compatibility

- the core stays content-agnostic — it doesn't know what an image shows; markdown/mermaid rendering lives in an external plugin
- no behavior change for existing plugins beyond deterministic same-priority virtual-text ordering and the namespaced conceal clear
- no new dependencies (base64 was already present); measured release-binary delta for the whole branch: **+44 KB**

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --all-features --all-targets`
- [x] `cargo nextest run --no-fail-fast --locked --all-features --all-targets` (CI: ubuntu / macos / windows)
- [x] `cargo test --test e2e_tests markdown_compose` (67 passed, incl. table wrapping/alignment, tall-table scroll, and emoji-row composition regression tests)
- [x] unit tests for placeholder encoding, transmit/delete sequences, tmux wrapping, namespace cleanup, conceal range removal (incl. property tests), virtual-text ordering

## Demo

https://www.loom.com/share/af49ee3975024c2884346282859fc9c3

Exercised end-to-end in Ghostty + tmux 3.6a with an external markdown-preview plugin: inline images, mermaid diagrams, and a fullscreen zoom view.


sinelaw:
This is a big PR doing several different things. I need to think about the graphics and kitty protocol stuff and allowing plugins directly to read files, but the PR does a few other things that are easier to merge.

Can we break into a separate PR doing only the "remove_in_range_for_namespace" and the changes in virtual_text.rs that are NOT related to the graphics stuff? It will be easier to know we aren't breaking anything? 

i.e. anything not related to graphics - if possible to break that into a separate PR we can merge first

crates/fresh-editor/plugins/markdown_compose.ts
```diff
@@ -983,40 +1234,186 @@ function effectiveComposeWidth(viewportWidth: number): number {
...
+  }
+  return widths;
+}
+
+/**
+ * Terminal cell width of a single code point. Tables are padded to align in
+ * *cells*, not code units — an emoji like ✅ is one UTF-16 unit but occupies
+ * two cells, so `.length`-based padding pushes the right border out of line.
+ */
+function charDisplayWidth(cp: number): number {
```
sinelaw, line 1320: This should be a replaced by a core function exposed to plugins, and since we already do unicode width in the rust core, the implementation should use the same underlying unicode logic we have in the rust core instead of duplicating it

Agrejus:
> This is a big PR doing several different things. I need to think about the graphics and kitty protocol stuff and allowing plugins directly to read files, but the PR does a few other things that are easier to merge.
> 
> Can we break into a separate PR doing only the "remove_in_range_for_namespace" and the changes in virtual_text.rs that are NOT related to the graphics stuff? It will be easier to know we aren't breaking anything?
> 
> i.e. anything not related to graphics - if possible to break that into a separate PR we can merge first

Yes, good call.  I'll start breaking it down now and close this one

Agrejus:
Closing this in favor of smaller, focused PRs, per the review above:

> *"This is a big PR doing several different things. I need to think about the graphics and kitty protocol stuff and allowing plugins directly to read files, but the PR does a few other things that are easier to merge. ... anything not related to graphics — if possible to break that into a separate PR we can merge first."*

Breaking it up so the easy, low-risk, non-graphics pieces can land first and the graphics/kitty + plugin-file-read work can get its own dedicated review.

**First split is up now:** #2399 — *Namespaced conceal range-clear + deterministic virtual-line ordering*. That's exactly the `remove_in_range_for_namespace` work and the non-graphics `virtual_text.rs` ordering change you called out, with nothing else attached.

Planned follow-ups (each its own PR):
- A core unicode-width plugin API to replace the duplicated `charDisplayWidth` table in `markdown_compose.ts` (using the same `unicode-width` logic the rust core already uses) — addresses your inline comment.
- The markdown-compose rendering itself (tables, wrapping, conceals, escaped pipes), once the two core PRs above land.
- The graphics / kitty inline-image rendering + plugin file-read APIs — held for the dedicated review you wanted.

Closing now; please direct review to #2399.

Agrejus:
@sinelaw — the non-graphics work from this PR is now split into three small, independent PRs, each branched from `master` and mergeable in any order:

- **#2399 — Namespaced conceal range-clear + deterministic virtual-line ordering**
  The `remove_in_range_for_namespace` work plus the non-graphics `virtual_text.rs` ordering change you specifically called out. Pure additions.

- **#2400 — Skip synchronized-update markers inside tmux**
  Bugfix: stray border glyphs left on screen when scrolling under tmux (DEC-2026 markers passed through tmux unwrapped, dropping frame-diff cell clears).

- **#2401 — Expose core unicode display-width to plugins (`charWidth`/`stringWidth`)**
  Addresses your inline review note — a shared `fresh-core` width function (same `unicode-width` logic the core already uses) exposed to plugins, with `fresh-editor` re-exporting it so there's one implementation, not a duplicated table.

Still held back for your separate review (not yet opened): the graphics / kitty inline-image rendering and the plugin file-read APIs. The markdown-compose rendering will follow as its own PR once #2399 and #2401 land (it depends on both, and is where `charDisplayWidth` actually gets deleted in favor of #2401's API).

Issue #2117, opened 2026-05-26
Review Diff: "Discard hunk" fails with "patch does not apply" even when patch is valid


sinelaw:
## Steps to reproduce

1. Open any git-tracked file (e.g. `README.md`)
2. Add 2 lines at the end of the file and save with `Ctrl+S`
3. Verify `git diff HEAD` shows the expected hunk (e.g. `+2/-0`)
4. Open the command palette (`Ctrl+P`) → `Review Diff` → `Enter`
5. Press `n` to navigate to the first hunk (status bar: "Review Diff: Hunk 1 of 1")
6. Press `d` to discard — a confirmation dialog appears: "Discard this hunk in 'README.md'? This cannot be undone."
7. Select "Discard hunk" (highlighted by default) → press `Enter`

**Observed behavior:**
Status bar immediately shows:
```
Patch failed: error: patch failed: README.md:275error: README.md: patch does not apply
```
The file is unchanged on disk. The diff still shows `+2/-0` after the failed operation.

## Expected behavior

The added lines are removed from the file on disk (the hunk is discarded). This is the behavior described in the docs: "Stage, unstage, or discard the hunk, file, or a line-level visual selection on the cursor row."

In comparison, the equivalent shell command (`git diff HEAD -- README.md | git apply --reverse`) succeeds without error.

## Actual behavior

The patch fails to apply with "patch does not apply" even though the reverse patch is valid and can be applied from the command line. The error message is emitted as a concatenated string (`error: patch failed: README.md:275error: README.md: patch does not apply`) suggesting Fresh captures git's stderr and surfaces it directly.

## Workaround

Apply the reverse patch manually from the shell:
```sh
git diff HEAD -- <file> | git apply --reverse
```

## Reproduction details

- **Reproduced:** 3 consecutive times in independent attempts during automated TUI testing
- **Branch:** `tui-automated-testing-state` (identical README.md content to `master`)
- **File tested:** `README.md` (last 2 lines modified, context lines match exactly)
- **tmux send-keys sequence used:**
  ```bash
  # Make change and save
  tmux send-keys -t fresh_test "C-End" ""
  tmux send-keys -t fresh_test "Enter" ""
  tmux send-keys -t fresh_test "# TEST LINE" ""
  tmux send-keys -t fresh_test "C-s" ""
  # Open Review Diff, navigate to hunk, discard
  tmux send-keys -t fresh_test "C-p" ""
  tmux send-keys -t fresh_test "Review Diff" ""
[...]

PR #2332, opened 2026-06-14
fix(review-diff): discard a fully-staged file back to HEAD (#2318)

### LINKED ISSUE #2318: Review Diff: file-level Discard (`D`) reports "Discarded" but leaves staged changes intact for a fully-staged file
**Version:** fresh 0.4.0 (built from `master` @ `1b5d7f8c8`)

In Review Diff, the file-level **Discard** action (`D`) only reverts a file's
**working-tree (unstaged)** changes. When a file's changes are entirely
**staged** (no working-tree diff), confirming the destructive discard dialog
("Discard changes in file — *Permanently lose changes* … *This cannot be
undone.*") does **nothing** to the staged content, yet the status bar reports
`Discarded: <file>`. The change remains in the index and on disk and is still
shown in the diff. A user is told an irreversible discard succeeded when in fact
their change is untouched.

## Steps to reproduce
1. Create a git repo and **stage** a one-line change so the file is fully staged
   with no working-tree diff:
   ```bash
   mkdir /tmp/rdiff && cd /tmp/rdiff && git init -q
   printf 'def add(a, b):\n    return a + b\n' > calc.py
   git add -A && git -c commit.gpgsign=false commit -qm initial
   sed -i 's/return a + b/return a + b  # CHANGED/' calc.py
   git add calc.py
   git status --short        # → "M  calc.py"  (staged, working tree clean)
   ```
2. Launch from that dir (`cd /tmp/rdiff` inside the shell first), `fresh --no-restore`.
3. Command Palette (`Ctrl+P`) → **Review Diff** → Enter.
4. In the `STAGED` group, move the cursor onto the file's changed hunk line
   (cursor must be on a hunk content row for `D` to fire).
5. Press `D` → a dialog appears: `Discard changes in file` / `Permanently lose
   changes` / `Discard "calc.py"? This cannot be undone.`
6. Select **Discard changes in file** → Enter.

## Expected behavior
Either:
- The staged change is reverted to HEAD (the file goes back to its committed
  state — the obvious meaning of a "this cannot be undone" discard), **or**
- If the action is intentionally scoped to working-tree changes only, the dialog
  and result should say so (e.g. "no unstaged changes to discard — unstage
  first"), and it must **not** report `Discarded` when nothing was discarded.

For reference, VS Code only offers "Discard Changes" on working-tree changes;
staged changes are unstaged first, and "Discard All Changes" reverts the file to
HEAD. Fresh presents a single generic "Discard changes in file" that implies a
full discard.

## Actual behavior
The status bar shows `Discarded: calc.py`, but:
- `git status --short` is still `M  calc.py` (change still staged),
- `git diff --cached` still shows the change,
- the file on disk is unchanged (still contains `# CHANGED`),
- the diff in Review Diff still lists the file with its hunk.

Nothing was discarded despite the success message and the irreversible-action
warning. (Verified by `git status` / `git diff --cached` before and after.)

**Contrast — works for unstaged changes (same session):** with the same file
changed but **not** staged (` M calc.py`), `D` correctly reverts the working
tree to HEAD (`Discarded: calc.py`, `git status` clean, file reverted). So the
discard path functions — it simply ignores the index, while the UI claims a full
file discard regardless of staged-vs-unstaged state.

## Workaround
Unstage the file first (`u`/`U`) so the change becomes a working-tree change,
then `D` to discard it; or run `git restore --staged --worktree <file>` from the
shell.

---
*Search queries used (only my own line-level #2317 and the closed hunk-level
#2117 found — both different code paths): `review diff discard staged file does
nothing`, `discard changes file staged not reverted`, `review diff D discard
reports discarded no change`.*

*Filed by tui-agent-auto-bug (automated TUI QA agent, Run #36, 2026-06-11).*

sinelaw:
## Summary

Fixes #2318.

In Review Diff, the file-level **Discard** action (`D`) ran `git checkout -- <path>`, which only rewrites the **working tree** from the index. When a file's change is entirely **staged** (clean working tree), that command is a no-op — yet the status bar still reported `Discarded: <file>`, telling the user an irreversible discard succeeded while the staged change remained in the index, on disk, and in the diff.

## Fix

- Reset the file all the way back to its committed state with `git checkout HEAD -- <path>`, which restores **both** the index entry and the working tree to HEAD, so staged and unstaged edits alike are dropped (the obvious meaning of the "this cannot be undone" discard dialog).
- Honor the command's exit code: only report `Discarded` when it actually succeeded; otherwise surface `Discard failed: …` instead of a false success.

The untracked-file branch (`rm`) is unchanged.

## Testing

- New e2e test `test_issue2318_discard_fully_staged_file_reverts_to_head`: stages a change so the working tree is clean, opens Review Diff, discards the file via `D`, and asserts the change disappears from the diff and that **neither** the working tree **nor** the index retains it. The test times out against the old `git checkout -- <path>` path (the change is never discarded) and passes with the fix.
- All 9 existing discard-related e2e tests pass, including the pre-existing hunk-discard regression test (#2117) — no regression.
- Plugin type-check (`tsc --strict`) passes.

## Manual validation

Built the `fresh` binary and reproduced the issue's exact scenario in tmux (commit a file, stage a one-line change, `Review Diff` → `D` → confirm):

- Before: staged change survived, status falsely said `Discarded`.
- After: UI shows **"No changes to review"**, status shows a truthful `Discarded: calc.py`, `git status --short` is empty, `git diff --cached --quiet` returns 0, and `calc.py` on disk is back to its committed contents.

https://claude.ai/code/session_01FqagsaYk1AERmbNY5GbnK2

---
_Generated by [Claude Code](https://claude.ai/code/session_01FqagsaYk1AERmbNY5GbnK2)_

### DIFF
--- a/crates/fresh-editor/plugins/audit_mode.ts
+++ b/crates/fresh-editor/plugins/audit_mode.ts
@@ -5110,13 +5110,28 @@ editor.on("prompt_confirmed", async (args) => {
         const f = pendingDiscardFile;
         if (f) {
             const cwd = gitCwd();
-            if (f.category === 'untracked') {
-                await editor.spawnProcess("rm", ["--", f.path], cwd);
-            } else {
-                await editor.spawnProcess("git", ["checkout", "--", f.path], cwd);
-            }
+            // "Discard changes in file" means lose the file's changes entirely
+            // and bring it back to its committed (HEAD) state. For tracked
+            // files this must reset BOTH the index and the working tree:
+            // `git checkout -- <path>` only rewrites the working tree from the
+            // index, so a fully-staged change (clean working tree) would slip
+            // through untouched while we still reported success. `git checkout
+            // HEAD -- <path>` restores the index entry and the working tree to
+            // HEAD, dropping staged and unstaged edits alike.
+            const result = f.category === 'untracked'
+                ? await editor.spawnProcess("rm", ["--", f.path], cwd)
+                : await editor.spawnProcess("git", ["checkout", "HEAD", "--", f.path], cwd);
             await refreshMagitData();
-            editor.setStatus(`Discarded: ${f.path}`);
+            // Only claim a discard happened when the command actually
+            // succeeded — never tell the user an irreversible discard worked
+            // when it didn't.
+            if (result.exit_code === 0) {
+                editor.setStatus(`Discarded: ${f.path}`);
+            } else {
+                editor.setStatus(
+                    `Discard failed: ${(result.stderr || "").trim() || f.path}`,
+                );
+            }
         }
     } else {
         editor.setStatus("Discard cancelled");
--- a/crates/fresh-editor/tests/e2e/plugins/audit_mode.rs
+++ b/crates/fresh-editor/tests/e2e/plugins/audit_mode.rs
@@ -4450,3 +4450,86 @@ fn test_review_branch_detail_enter_opens_file_at_commit() {
         .unwrap();
     harness.wait_until(|h| !file_view_active(h)).unwrap();
 }
+
+/// Issue #2318: file-level Discard (`D`) on a fully-*staged* file reported
+/// "Discarded" but left the staged change intact. The discard ran
+/// `git checkout -- <path>`, which only rewrites the working tree from the
+/// index — so a change that lives entirely in the index (clean working tree)
+/// survived untouched while the status bar still claimed success. The fix
+/// resets the file all the way back to HEAD (`git checkout HEAD -- <path>`),
+/// dropping both the staged and any unstaged edits.
+#[test]
+fn test_issue2318_discard_fully_staged_file_reverts_to_head() {
+    init_tracing_from_env();
+    let repo = GitTestRepo::new();
+    setup_audit_mode_plugin(&repo);
+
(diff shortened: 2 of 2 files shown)


sinelaw (commented):

Superseded. Suggest closing.

Issue #2318 was fixed by PR #3160, merged 2026-09-04 as e0022f57b, which rewrote the discard path in `audit_mode.ts`. That is why this branch no longer rebases cleanly.

This PR does two things, both now on master:

1. `git checkout --` -> `git checkout HEAD --`, so a fully-staged file goes back to HEAD. Master uses `git restore --source=HEAD --staged --worktree`, which covers the same case and also resets the index.
2. Report a non-zero exit instead of always printing "Discarded". Master reports failures and additionally re-asks git afterwards via `fileIsClean`, so a silent no-op is caught too.

Checked for anything here that master lacks and found nothing.

Note also that PR #2334 is the same fix against the same issue, and is a superset of this one — it additionally handles `status === 'A'`. Whichever way you go, these two should not both stay open.

sinelaw:
Superseded by #3160, merged as e0022f57b. Master's discard already covers this case. https://github.com/sinelaw/fresh/pull/3160

PR #2333, opened 2026-06-14
fix(status-bar): refresh cached cursor line on plugin cursor jump (#2301)

### LINKED ISSUE #2301: Go to LSP Symbol: status bar line number stays stale after jump (only column updates) until next keypress
## Steps to reproduce
Environment: `fresh` v0.3.12 (built from `origin/master` @ `f4ee3630`), clangd 18 LSP on a small C file, workspace Trusted, `LSP (on)`.

1. Open a C file with several top-level symbols and start clangd (e.g. a file whose `main` is on line 44).
2. Move the cursor to a known line — e.g. `Ctrl+G` → `30` → Enter. Status bar reads `Ln 30, Col 1`.
3. Open the command palette (`Ctrl+P`), run **Go to LSP Symbol**.
4. Filter to a symbol on a *different* line (type `main`), press **Enter** to jump.
5. Read the status bar immediately, then look at where the cursor actually is.

## Expected behavior
After jumping to the symbol, the status bar should report the symbol's real location — `Ln 44, Col 5`. Both VS Code and Sublime update the status-bar line/column immediately after "Go to Symbol". Fresh's own `Go to Definition` (F12) and `Go to Line` (Ctrl+G) both update the status bar correctly and immediately (F12 even prints `Jumped to definition at …:44`).

## Actual behavior
The cursor *does* move to the correct line (verified: the terminal cursor and the highlighted line are on line 44, `int main(void) {`), but the status bar shows the **previous** line number: `Ln 30, Col 5`.
- The **column** is updated correctly (`Col 5` = the new position).
- The **line** is stale — it retains the value from before the jump.
- It self-corrects to `Ln 44, …` the instant any cursor movement happens (e.g. pressing `Right`, `End`).

Reproduced 3/3 times with different start/target lines (start Ln 1 → `main`/`print_rectangle`; start Ln 30 → `main`). In every case `Col` updated but `Ln` retained the pre-jump value. This is specific to **Go to LSP Symbol** — `Go to Definition` (F12) and `Go to Line` (Ctrl+G) do not exhibit it.

Severity is low (cursor and editing are correct; it's a status-bar display glitch that clears on the next keystroke), but the line readout briefly misreports the user's location right after the command whose entire purpose is to navigate there.

## Workaround
Press any cursor-movement key (e.g. `→`) after jumping; the status bar then shows the correct line.

---
Searched before filing: `status bar line number symbol`, `Go to LSP Symbol`, `status bar cursor position not updating`, `in:title symbol navigation line` — no existing issue. Filed by automated black-box TUI QA agent.

sinelaw:
Fixes #2301.

## Problem

After **Go to LSP Symbol** jumps the cursor, the status bar shows the **previous** line number (`Ln 30`) with the **new** column (`Col 5`) until the next keypress, which self-corrects it. The cursor itself lands correctly — only the status-bar line readout is stale, right after the command whose whole purpose is to navigate there.

## Root cause

The "Go to LSP Symbol" plugin moves the cursor through the `setBufferCursor` plugin op, handled by `Window::set_buffer_cursor_in_splits`. That function moved the cursor position but never refreshed `state.primary_cursor_line_number`.

The status bar (`status_bar.rs`) reads the **line** from that cached `primary_cursor_line_number`, while it computes the **column** live from the cursor's byte position. That asymmetry is exactly the reported symptom: column updates, line is stale until any later cursor movement recomputes the cache.

This is the same class of bug noted in the issue comment ("status bar not refreshed after a programmatic cursor jump") for any path that bypasses the canonical jump entry point.

## Fix

Refresh the cached line number inside `set_buffer_cursor_in_splits` after moving the cursor, mirroring the canonical `Window::jump_active_cursor_to`. This covers every plugin-driven cursor jump that uses `setBufferCursor`, not just the LSP symbol case.

## Testing

- Added `test_lsp_navigation_status_bar_line_updates_after_jump`: jumps to `myMethod` (file line 5) and asserts the status bar reads `Ln 6, Col 3` **immediately**, with no intervening cursor movement. The pre-existing `test_lsp_navigation_jumps_to_precise_column` only asserted the column, which is why the stale line went unnoticed.
- Verified the new test **times out / fails without** the fix (the status bar never shows the new line) and **passes with** it.
- All 8 `lsp_navigation` e2e tests pass; `cargo fmt --check` and `cargo clippy` are clean.

https://claude.ai/code/session_015Cs1411QRfGHJzz2cjsczW

---
_Generated by [Claude Code](https://claude.ai/code/session_015Cs1411QRfGHJzz2cjsczW)_

### DIFF
--- a/crates/fresh-editor/src/app/window/mod.rs
+++ b/crates/fresh-editor/src/app/window/mod.rs
@@ -1536,6 +1536,17 @@ impl Window {
                     state.primary_cursor_line_number =
                         crate::model::buffer::LineNumber::Absolute(line);
                 }
+                // Refresh the cached primary-cursor line number. The status bar
+                // reads the line from this cache (the column is computed live),
+                // so a programmatic jump that skips it leaves the status bar
+                // reporting the pre-jump line until the next cursor movement
+                // recomputes it (#2301). Mirror the canonical jump path in
+                // `Window::jump_active_cursor_to`.
+                let clamped = position.min(state.buffer.len());
+                if let Some(pos) = state.buffer.offset_to_position(clamped) {
+                    state.primary_cursor_line_number =
+                        crate::model::buffer::LineNumber::Absolute(pos.line);
+                }
             });
     }
 
--- a/crates/fresh-editor/tests/e2e/plugins/lsp_navigation.rs
+++ b/crates/fresh-editor/tests/e2e/plugins/lsp_navigation.rs
@@ -272,6 +272,48 @@ fn test_lsp_navigation_jumps_to_precise_column() -> anyhow::Result<()> {
     Ok(())
 }
 
+/// Regression test for #2301: after confirming a "Go to LSP Symbol" jump,
+/// the status bar must report the symbol's *line* immediately — not just its
+/// column. The jump goes through the plugin `setBufferCursor` op, which moved
+/// the cursor but left the cached primary-cursor line number stale, so the
+/// status bar showed the pre-jump line (`Ln 1`) until the next cursor move.
+/// We assert the full `Ln 6, Col 3` (myMethod is on file line 5 / 1-indexed
+/// line 6) right after the jump, with no intervening cursor movement.
+#[test]
+#[cfg_attr(windows, ignore)]
+fn test_lsp_navigation_status_bar_line_updates_after_jump() -> anyhow::Result<()> {
+    let (mut harness, _temp_dir) = setup_lsp_test()?;
+
+    // Confirm the starting cursor is on line 1 — the line the bug would leave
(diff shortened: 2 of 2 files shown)


sinelaw:
Closing as obsolete. `master`'s `Window::set_buffer_cursor_in_splits` (`window/mod.rs:1522-1537`) already refreshes `state.primary_cursor_line_number` after a plugin cursor jump — the comment there even cites the same off-by-one symptom this PR fixes. After rebasing, the PR's code hunk is a duplicate refresh; only its regression test adds anything. Closing the fix as already-present; if you'd like to keep the regression test, I can open a small test-only PR for it.

---

PR #2346, opened 2026-06-14
fix(vi-mode): make % jump to matching bracket

ianyepan:
## TLDR: The bug is a single underscore typo.

The vi plugin called executeAction("go_to_matching_bracket"), but the action is registered as "goto_matching_bracket". Action::from_str returned None, so % (and the operator-pending %) was a silent no-op.

Corrects the action string at both call sites. Adds a passing e2e test for the % motion, and an #[ignore]'d reproducer documenting that the operator form (d%/c%/y%) is still unimplemented (goto_matching_bracket has no motionToSelection entry).

### DIFF
--- a/crates/fresh-editor/plugins/vi_mode.ts
+++ b/crates/fresh-editor/plugins/vi_mode.ts
@@ -456,7 +456,7 @@ function vi_page_up() : void {
 registerHandler("vi_page_up", vi_page_up);
 
 function vi_matching_bracket() : void {
-  editor.executeAction("go_to_matching_bracket");
+  editor.executeAction("goto_matching_bracket");
 }
 registerHandler("vi_matching_bracket", vi_matching_bracket);
 
@@ -1657,8 +1657,13 @@ function vi_op_doc_end(): void {
 }
 registerHandler("vi_op_doc_end", vi_op_doc_end);
 
+// NOTE: operator + `%` (d%/c%/y%) is currently a no-op. `applyOperatorWithMotion`
+// resolves a motion via `atomicOperatorActions` or `motionToSelection`, and
+// `goto_matching_bracket` is in neither map, so it bails without deleting. Making
+// this work needs a selection-extending action (e.g. `select_to_matching_bracket`)
+// plus a `motionToSelection` entry. See test_vi_bug_d_percent_ignored.
 function vi_op_matching_bracket(): void {
-  handleMotionWithOperator("go_to_matching_bracket");
+  handleMotionWithOperator("goto_matching_bracket");
 }
 registerHandler("vi_op_matching_bracket", vi_op_matching_bracket);
 
--- a/crates/fresh-editor/tests/e2e/vi_mode.rs
+++ b/crates/fresh-editor/tests/e2e/vi_mode.rs
@@ -1153,3 +1153,43 @@ fn test_vi_colon_split() {
         })
         .unwrap();
 }
+
+// =============================================================================
+// Matching Bracket Tests
+// =============================================================================
+
+/// Test '%' jumps between matching brackets (both directions).
+#[test]
+fn test_vi_percent_matching_bracket() {
+    let (mut harness, _temp_dir) = vi_mode_harness(80, 24);
+
+    // foo(bar)\n  ->  '(' at byte 3, ')' at byte 7
+    let fixture = TestFixture::new("test.txt", "foo(bar)\n").unwrap();
+    harness.open_file(&fixture.path).unwrap();
+    harness.render().unwrap();
+
+    enable_vi_mode(&mut harness);
+
+    // Move cursor onto the '(' at byte 3.
+    for _ in 0..3 {
+        harness
+            .send_key(KeyCode::Char('l'), KeyModifiers::NONE)
+            .unwrap();
+        harness.render().unwrap();
+    }
+    harness.wait_until(|h| h.cursor_position() == 3).unwrap();
+
+    // '%' jumps forward to the matching ')'.
+    harness
+        .send_key(KeyCode::Char('%'), KeyModifiers::NONE)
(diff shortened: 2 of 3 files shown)


ianyepan:
## Test Notes — vi mode `%` (goto matching bracket) fix

## What changed

**Bug:** In vi mode, `%` (jump to matching bracket) did nothing. The plugin called
`editor.executeAction("go_to_matching_bracket")`, but the action is registered as
`goto_matching_bracket` (no underscore between "go" and "to"). `Action::from_str`
returned `None`, so the keystroke was a silent no-op.

**Fix** — `crates/fresh-editor/plugins/vi_mode.ts`:
- `vi_matching_bracket` (line 459): `go_to_matching_bracket` → `goto_matching_bracket`
- `vi_op_matching_bracket` (line 1661): same correction, plus a comment noting that
  `d%`/`c%`/`y%` is still a no-op (separate, unimplemented case).

**Tests:**
- `crates/fresh-editor/tests/e2e/vi_mode.rs::test_vi_percent_matching_bracket` —
  covers the fixed `%` motion (passes).
- `crates/fresh-editor/tests/e2e/vi_mode_bugs.rs::test_vi_bug_d_percent_ignored` —
  `#[ignore]`'d reproducer documenting the still-unimplemented operator+`%` case.

---

## Root-cause confirmation

The plugin's `executeAction` string is resolved by `Action::from_str` in
`crates/fresh-editor/src/input/keybindings.rs`:

```rust
"goto_matching_bracket" => GoToMatchingBracket,   // line 975
...
_ => None,                                         // line 1286 (no alias/normalization)
```

`handle_execute_action` (`crates/fresh-editor/src/app/plugin_dispatch.rs:1969`) turns a
`None` into a dropped action:

```rust
if let Some(action) = Action::from_str(&action_name, &HashMap::new()) {
    ... self.handle_action(action) ...
} else {
    tracing::warn!("Unknown action: {}", action_name);   // old "go_to_matching_bracket" hit this
}
```

---

## Commands run & output

### 1. `%` motion test passes (with the fix)

```sh
RUST_LOG=fresh=debug cargo test -p fresh-editor --test e2e_tests \
  test_vi_percent_matching_bracket -- --nocapture --test-threads=1
```

Output (trimmed):

```
running 1 test
test e2e::vi_mode::test_vi_percent_matching_bracket ... 
  DEBUG fresh::app::plugin_dispatch: Executed action: goto_matching_bracket
  DEBUG fresh::app::plugin_dispatch: Executed action: goto_matching_bracket
ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2895 filtered out; finished in 0.12s
```

The two `Executed action: goto_matching_bracket` lines correspond to the two `%`
presses in the test (jump to the matching `)`, then back to `(`), confirming the
action now resolves and runs.

### 2. Ignored `d%` reproducer compiles and is skipped (won't run in CI)

```sh
cargo test -p fresh-editor --test e2e_tests d_percent_ignored
```

Output:

```
running 1 test
test e2e::vi_mode_bugs::test_vi_bug_d_percent_ignored ... ignored

test result: ok. 0 passed; 0 failed; 1 ignored; 0 measured; 2895 filtered out; finished in 0.00s
```

`cargo nextest run --all-targets` (the CI invocation) skips `#[ignore]` tests by
default, so this reproducer documents the gap without hanging CI.

### 3. The `d%` reproducer genuinely fails (hangs) until implemented

Run the ignored test explicitly, bounded by a hard timeout:

```sh
timeout --signal=KILL 45 cargo test -p fresh-editor --test e2e_tests \
  d_percent_ignored -- --ignored --nocapture
echo "exit=$?"
```

Output:

```
exit=137
```

Exit `137` = killed by the timeout, i.e. the test hangs: `d%` deletes nothing, so
`wait_for_buffer_content("foobaz\n")` never resolves. This is the expected
"fails/times out until the underlying bug is fixed" behavior for a
`vi_mode_bugs.rs` reproducer. It will pass once a `select_to_matching_bracket`
action and a `motionToSelection["goto_matching_bracket"]` entry are added.

---

## Notes / scope

- Only the normal-mode `%` motion is fixed and tested as passing. The operator
  forms (`d%`/`c%`/`y%`) remain unimplemented — `goto_matching_bracket` has no entry
  in `atomicOperatorActions` or `motionToSelection`, so `applyOperatorWithMotion`
  bails without deleting. That is captured by the ignored test, not fixed here.
- Pre-commit gates not yet run in this session: `crates/fresh-editor/plugins/check-types.sh`
  (plugin type-check), `cargo fmt`, `cargo clippy --all-targets`, `cargo check --all-targets`.

ianyepan:
This PR fixes https://github.com/sinelaw/fresh/issues/2347

sinelaw (approved):
This is an autogenerated comment.

**Approved** — root cause confirmed: the registered action is `goto_matching_bracket` (keybindings.rs:975); `go_to_matching_bracket` has no alias, so `%` was a silent no-op. Fix corrects both call sites.

Validated interactively in tmux, side-by-side against `vim` on `foo(bar)`:
- **`%` on `(`** → jumps to `)` (Col 4→8); **`%` on `)`** → back to `(` (Col 8→4). vim cursor_x 3↔7. Match.
- **`d%`** confirmed still a no-op (cursor jumps, buffer unchanged) — accurately documented by the `#[ignore]`'d reproducer + comment; not overclaimed.

Code quality (per CONTRIBUTING.md): minimal one-char fix, reuses the core `goto_matching_bracket` action, passing e2e test follows file conventions, honest scoping. No duplication or architectural concern.

Minor (non-blocking, out of scope): `%` from a non-bracket cell doesn't scan forward to the next bracket like vim — that's pre-existing behavior of the core `goto_matching_bracket` action, not this plugin. LGTM.

---

sinelaw:
🥇

PR #2361, opened 2026-06-16
Non-terminal (web) UI: every editor surface as native HTML from one shared core model

sinelaw:
## Summary

Adds a **non-terminal (web) UI** for Fresh that renders the editor in the browser as **native HTML chrome over the real, syntax-highlighted buffer**, driven end-to-end by a single shared core — no mocks. The TUI is unchanged and stays byte-identical; both frontends are fed by the same `Editor` and differ only in the final drawing (cells for the terminal, HTML/SVG for the web) and the input bridge.

The guiding principle: **the TUI and the web must not re-implement the same logic.** Everything semantic — what menus/items exist, what's enabled/checked, accelerators, the status segments, the settings tree, etc. — is derived **once** in the core; each frontend only renders it. Design + phased plan: `docs/internal/UNIFIED_SCENE_DESIGN.md` (original research in `docs/internal/NON_TERMINAL_UI_RESEARCH.md`).

```
cargo run -p fresh-editor --example webui_server -- 127.0.0.1:8137
# then open http://127.0.0.1:8137
```

> ⚠️ The bridge binds plain localhost HTTP with `Access-Control-Allow-Origin: *` and hosts a live editor with filesystem access — fine for local development, but a prototype, **not for exposure** on a shared interface.

## Architecture

- **Bridge** (`crates/fresh-editor/src/webui`): a dependency-free single-threaded `TcpListener` HTTP server hosting a real `Editor` (plugin runtime on, `init.ts` loaded). Routes: `/state`, `/key`, `/mouse`, `/action`, `/resize`, `/widget`, `/settings`, `/kbedit`, plus `/reset` + `/step` test hooks. Browser key/mouse map straight onto the real `handle_key`/`handle_mouse`.
- **Frame parity with the TUI loop**: each poll runs `editor_tick` (drains async LSP/plugin/file events, steps animations), so frames advance without user input; the frontend polls adaptively and diffs scenes to avoid churn.
- **Shared semantic model** (`crates/fresh-editor/src/view/scene.rs`): `Editor::*_view()` projections are the single derivation of menu / tabs / status / command-palette / popups / file-explorer / trust-dialog / plugin-widgets / context-menu / keybinding-editor / settings. The bridge only serializes them.
- **Layout-vs-draw seam** (`Editor::suppress_chrome_cells`): the pipeline always computes chrome *layout/geometry/semantics* but skips *drawing* chrome cells when suppressed. The web gets pane-only cells (zero chrome bleed) and renders all chrome as native HTML; for the TUI the flag is always off, so its rendering is unchanged. The status bar was refactored from a cell-scraper to a real semantic `StatusBarLayout.segments` model — the last cell-scraper is gone.
- **Shared dispatch**: hit→action logic is extracted so the TUI mouse handler and the web bridge call the *same* methods (`deliver_widget_hit`, `dispatch_settings_hit`, `entry_dialog_*`, `kbedit_select_display_row`, …). The web routes by the semantic identity it rendered, not fragile pixel hit-testing.

## Surfaces rendered as native HTML (all from the shared model)

- Menu bar + dropdowns (submenus, accelerators, checkmarks, enabled/disabled), keyboard nav not fought by a stationary mouse
- Tab bar, status bar (labeled segments), command palette / pickers, completion / hover / action popups
- File-explorer sidebar, workspace-trust dialog, right-click / new-tab context menus
- Plugin widgets (floating + dock surfaces, e.g. the Git/orchestrator dock) from their real `WidgetSpec`, including card lists
- Full keybinding editor (header / search / filters / table / **edit dialog**)
- Full Settings modal (category tree, every control type incl. DualList + Map, search, and the add/edit **entry dialog**)
- Buffer interiors stay the pipeline's **real syntax-highlighted cells** (SVG)

## Interaction + polish in this PR

- **Theme-accurate chrome**: scene carries the resolved editor `Theme`; the frontend seeds its CSS variables from it (matches the active theme instead of a fixed palette).
- **Settings DualList + entry-dialog mouse wiring**: native two-column Available⇄Included with add/remove/reorder; the add/edit sub-dialog is clickable (rows + Save/Cancel/Delete).
- **Live-grep / quick-open preview pane**: surfaced on the web as real buffer cells, laid out as a single unified overlay (full-width header + results | preview body) matching the TUI; the palette/overlay card now has a visible themed border.
- **Menu position + dropdown alignment**: the menu bar stays full-width at x=0 when a left dock opens (no shift); dropdowns open directly under their title.
- **Resizable dock / file-explorer borders** via explicit drag grips that drive the real editor resize.
- **Scrollable** Settings content pane and dock/widget lists (native scroll, preserved across re-renders).
- **Cell-grid text rendering**: vertical box-drawing (gutter separator, tree guides) rendered as crisp vector rules instead of dashed glyphs; the **line-number gutter is emitted as its own block** (real gutter width from the split's viewport), kept out of the buffer-text flow so a future native text selection covers only code; and **every glyph is pinned to its exact cell column** (per-character x), so toggling occurrence / current-line highlights no longer nudges text.

## Testing

- **Headless Chromium / Playwright** end-to-end suite (`web-ui/test/drive.mjs`) drives the real bridge: **50/50 green**, asserting buffer interiors are real highlighted cells while all chrome is native HTML (no cell-drawn chrome), and that key/mouse/menu/settings/widget interactions run through the real `Editor`.
- **Web/TUI scene-parity test** (`crates/fresh-editor/tests/scene_parity.rs`): one `Editor`, two renderers must agree — the chrome the web scene reports also appears in the TUI's cell rendering (tabs, status language, menu + dropdown item, file-explorer rows, settings categories).
- Glyph-grid measurement confirms every cell glyph sits at exactly `col*CW` under occurrence/current-line highlighting (max error 0px).
- Full `fresh-editor` lib suite passes; `cargo fmt` + `clippy` clean.

https://claude.ai/code/session_01EwQbLpJk16oyJkNHdsa5N8

---
_Generated by [Claude Code](https://claude.ai/code/session_01EwQbLpJk16oyJkNHdsa5N8)_


crates/fresh-editor/src/view/chrome_layout.rs
```diff
@@ -0,0 +1,665 @@
+//! Chrome layout engine (PoC): turn a [`ChromeSnapshot`] + window size into
```
sinelaw: **Blocking (architecture): this module — plus `chrome_html.rs` and `chrome_snapshot.rs` — is dead relative to the shipped web UI (~1,300 lines).**

I grepped for production usage: `chrome_layout` / `chrome_snapshot` / `chrome_html` are referenced only by each other and by the `chrome_web_demo` example (which renders a *static* HTML mockup from a *sample* `ChromeSnapshot`). The live bridge (`webui/mod.rs`) and the actual TUI/web rendering go entirely through `view/scene.rs` (`menu_view`/`status_view`/… → `RectView`), never through `ChromeSnapshot`/`ChromeLayout`/`PxRect`.

So this PR introduces a *second, competing* "chrome model" vocabulary (with placeholder `buffer#{b}` / `"(group)"` tab labels that don't match reality) sitting next to the real one — which is precisely the divergence the PR's single-source-of-truth thesis exists to eliminate.

Recommend pulling `chrome_layout.rs` + `chrome_html.rs` + `chrome_snapshot.rs` + the `chrome_web_demo` example/HTML into a separate "GUI/Tauri PoC" PR (or deleting until Phase 4 actually consumes them). It would shrink this diff by ~1.3k lines and remove the conceptual fork.

---

crates/fresh-editor/src/webui/mod.rs
```diff
@@ -0,0 +1,794 @@
...
+            let s = scene_json(editor, *cols, *rows).to_string();
+            respond(stream, "200 OK", "application/json", s.as_bytes())
+        }
+        _ => respond(stream, "404 Not Found", "text/plain", b"not found"),
+    }
+}
+
+fn respond(stream: &mut TcpStream, status: &str, ctype: &str, body: &[u8]) -> Result<()> {
+    let header = format!(
+        "HTTP/1.1 {status}\r\nContent-Type: {ctype}\r\nContent-Length: {}\r\nAccess-Control-Allow-Origin: *\r\nConnection: close\r\n\r\n",
```
sinelaw: **Blocking (security): `Access-Control-Allow-Origin: *` on a live editor with filesystem access is more than a prototype caveat.**

`/state` returns real buffer contents, and `*` lets *any* website the user visits while this is running read that response cross-origin, and POST to `/key` / `/action` / `/mouse` (simple requests → no preflight) to drive the editor. That's a drive-by read/control vector, not just "exposed on a shared interface."

The frontend is served from the same origin, so it doesn't need CORS at all. Recommend dropping the `Access-Control-Allow-Origin: *` header entirely (zero functional cost — same-origin policy then re-protects `/state`), and ideally gating the bridge behind a per-process token in the URL. Cheap hardening that closes the actual hole.

---

crates/fresh-editor/src/webui/mod.rs
```diff
@@ -0,0 +1,794 @@
...
+        }
+    } else if step.get("key").is_some() {
+        apply_key(editor, step);
+    } else if step.get("kind").is_some() {
+        apply_mouse(editor, step);
+    } else if let Some(name) = step.get("action").and_then(|a| a.as_str()) {
+        if let Some(act) =
+            crate::input::keybindings::Action::from_str(name, &std::collections::HashMap::new())
+        {
+            editor.dispatch_action_for_tests(act);
```
sinelaw: **Nit: production routes call a `#[doc(hidden)]` "test-only shim."**

`apply_step` here (and the `/action` route below at ~L233) dispatch via `editor.dispatch_action_for_tests(act)`, which is documented as a test-only shim that just forwards to `handle_action`. Functionally fine, but the name lies about its use in a shipping bridge. Either call `handle_action` directly here, or rename the shim to something neutral (e.g. `dispatch_action`) and drop the `_for_tests`/`#[doc(hidden)]`.

---

web-ui/README.md
```diff
@@ -0,0 +1,70 @@
+# Fresh — web UI (wired to the real editor, no mocks)
+
+A non-terminal UI for Fresh, driven **end-to-end by the real Rust `Editor`** — no
+mock model. See `docs/internal/NON_TERMINAL_UI_RESEARCH.md` (Direction A).
+
+- **Chrome is DOM/CSS** — menu bar, tabs, split panes via CSS grid + dividers,
+  status bar — rendered from the editor's real `ChromeSnapshot`.
```
sinelaw: **Should-fix (docs): this README documents the wrong (unused) architecture.**

It says chrome is "rendered from the editor's real `ChromeSnapshot`" (line 7) — but `ChromeSnapshot` is the unused PoC; the shipped UI renders from `view/scene.rs` projections. The "Status … Next:" section (L46–54) also lists mouse, tab/split/palette clicks, and real highlighting as not-yet-done, which contradicts both the actual code and this PR's description (they're implemented). A contributor reading this will be misled about how the bridge works. Please rewrite around the `scene.rs` projection path and refresh the status/test-count section.

---

sinelaw (commented):
## Staff review — strong architecture, a few things to tidy

Nice work. The core design is the right one: a single semantic projection in `view/scene.rs` that both frontends consume, a clean `suppress_chrome_cells` layout-vs-draw seam (geometry/hit-caches always computed, only cell *painting* gated), a real `Editor` + plugins behind the bridge (no mocks), and genuine de-duplication via the shared `deliver_widget_hit`. The regression safety is real, not asserted — `scene_parity` proves the two renderers agree, and `menu_render_golden` pins the TUI menu byte-for-byte across the Phase-3 refactor.

**Verified locally:** `cargo build -p fresh-editor --example webui_server` → clean (only pre-existing `animation.rs` dead-code warnings); `scene_parity` 1/1; `menu_render_golden` 2/2.

### Blocking / should-fix (inline)
1. **~1,300 lines of dead, parallel chrome code** (`chrome_layout.rs` / `chrome_html.rs` / `chrome_snapshot.rs` + `chrome_web_demo`) — not wired into the shipped UI and a competing "chrome model" vocabulary. Split out or drop.
2. **`Access-Control-Allow-Origin: *`** on a live editor with FS access — drop it (same-origin frontend doesn't need it) and ideally add a bind token.
3. **`web-ui/README.md`** documents the unused `ChromeSnapshot` path and lists done features as "Next."

### Non-blocking nits
- `dispatch_action_for_tests` used in the `/step` and `/action` production routes (flagged inline).
- **Per-frame cost on the TUI hot path:** `StatusBarRenderer::render_status` now always builds `layout.segments` (a `Vec<StatusSegmentInfo>` with `String` clones) every frame, even when `draw=true` and the TUI never reads them. Consider only populating it when a consumer needs it.
- **Frontend (acknowledged prototype):** full `app.innerHTML=""` rebuild per changed poll won't scale (the README's own xi-editor "diff the visible window" TODO); poll `/state` and input POSTs can interleave on the single-threaded server (safe, but can jitter). `esc()`/SVG escaping looked correct at the injection points I checked.
- **Size:** 6.9k additions / 57 commits / 36 files in one PR. The design doc's own phase boundaries (1 → 2 → 3 → 5) would make this far more reviewable as a series; the parity/golden tests are currently carrying most of the review confidence.

None of these block the *approach* — they're about trimming scope and one real CORS hardening. Happy to re-review once the dead-code split and the ACAO change land.

---

crates/fresh-editor/src/view/chrome_layout.rs
sinelaw: Confirmed and removed in `8176741`. Verified the shipped UI never used this PoC: the live path renders from `view/scene.rs` projections and the separate per-frame `ChromeLayout` in `app/types/layout.rs`, never from `ChromeSnapshot`/`PxRect`/`view/chrome_layout.rs`. Deleted `chrome_layout.rs` + `chrome_snapshot.rs` + `chrome_html.rs` + the `chrome_web_demo` example and its generated HTML (~1.3k lines), dropped the `view/mod.rs` decls, and marked the `NON_TERMINAL_UI_RESEARCH.md` section superseded. Lib + examples build clean; scene-parity + e2e (50/50) green.

---

crates/fresh-editor/src/webui/mod.rs
sinelaw: Fixed in `9199f2f` — dropped `Access-Control-Allow-Origin: *` from every response. The frontend is served same-origin so it needs no CORS, and same-origin policy now re-protects `/state`. e2e suite still 50/50.

---

crates/fresh-editor/src/webui/mod.rs
sinelaw: Fixed in `9199f2f` — `/step` and `/action` now call the real `handle_action` directly (surfacing errors via `eprintln`) instead of the `#[doc(hidden)]` `dispatch_action_for_tests` shim.

---

web-ui/README.md
sinelaw: Fixed in `9199f2f` — rewrote `web-ui/README.md` around the real `scene.rs` projection path (it had described the unused `ChromeSnapshot` PoC and listed already-shipped features as "next"), and refreshed the test count to 50 plus the example file/port.

---

crates/fresh-editor/src/view/ui/status_bar.rs
```diff
@@ -1698,6 +1739,13 @@ impl StatusBarRenderer {
                     area.x + start_col as u16,
                     area.x + (start_col + width) as u16,
                 );
+                layout.segments.push(StatusSegmentInfo {
```
sinelaw: **Nit (TUI hot-path allocation):** `layout.segments` is built on every status-bar render (here and the two sibling pushes at ~L1787/L1857) — a `Vec<StatusSegmentInfo>` with a `String` per segment, allocated each frame. But it's only consumed by `status_view()` on the web; the TUI (`draw == true`) never reads it. So this is pure per-frame allocation on the terminal's render path.

Cheap to gate: only populate `segments` when the semantic model will actually be read (e.g. behind the same condition that suppresses cell drawing, or a `want_segments` flag passed in). Not blocking, just avoidable churn on the hot path.

---

web-ui/index.html
```diff
@@ -0,0 +1,1233 @@
...
+document.addEventListener("keydown",e=>{
+  // forward single characters (with or without modifiers, so Ctrl+P / Ctrl+S
+  // shortcuts reach the editor) and the named navigation/edit keys.
+  const single=e.key.length===1;
+  if(single||NAMED.has(e.key)){ e.preventDefault(); sendKey({key:e.key,ctrl:e.ctrlKey,meta:e.metaKey,shift:e.shiftKey,alt:e.altKey}); }
+});
+// Mouse + wheel: forward to the REAL Editor::handle_mouse at cell coordinates.
+const cellAt = e => ({ col: Math.max(0, Math.floor(e.clientX/CW)), row: Math.max(0, Math.floor(e.clientY/CH)) });
+const btn = e => (e.button===2?"right":e.button===1?"middle":"left");
+async function sendMouse(o){ try{ scene=await (await fetch("/mouse",{method:"POST",headers:{"content-type":"application/json"},body:JSON.stringify(o)})).json(); render(); }catch(err){ showErr("mouse failed: "+err); } }
```
sinelaw: **Nit (frontend consistency):** `sendMouse` bypasses the scene-dedup that every other sender uses. `sendKey` / `sendWidget` / `sendSettings` / `sendKbedit` all route through `applyScene(...)`, which skips the re-render when the new scene is byte-identical to the last (`str === sceneStr`). `sendMouse` instead does `scene = …; render();` unconditionally, so every mouse event — including pure `moved`/hover and no-op drags — forces a full `app.innerHTML` rebuild even when nothing changed.

Routing it through `applyScene` like the others would remove that churn for free:

```js
async function sendMouse(o){ try{ applyScene(await (await fetch("/mouse",{method:"POST",headers:{"content-type":"application/json"},body:JSON.stringify(o)})).json()); }catch(err){ showErr("mouse failed: "+err); } }
```

---

web-ui/index.html
```diff
@@ -0,0 +1,1233 @@
...
+    window.addEventListener("mousemove",move); window.addEventListener("mouseup",up);
+  };
+  return grip;
+}
+
+// --- bridge to the real editor ---
+// Apply a scene from any endpoint, re-rendering only when it actually changed
+// so the idle poll loop doesn't churn the DOM (or restart the caret blink).
+let sceneStr=null;
+function applyScene(s){ const str=JSON.stringify(s); if(str===sceneStr) return; sceneStr=str; scene=s; render(); }
```
sinelaw, line 1172: **Nit (scale, prototype-acceptable):** the poll loop re-serializes the entire scene to JSON on both ends every tick and diffs it as a string (`JSON.stringify` here, compared against `sceneStr`). The scene includes every visible buffer cell, so on large files / big terminals this is a non-trivial stringify+compare every 40ms while anything is animating. Fine for a local prototype, but it's the first thing that'll bite when this grows — the README's own "ship only the visible-window cell diff" TODO is the right eventual fix.

Also note (no action needed, just flagging the model): because the server is single-threaded and the frontend fires overlapping fetches, a `/state` poll can land between two input POSTs and run an extra `editor_tick`. Harmless given each request renders unconditionally, but worth keeping in mind if timing-sensitive behavior is ever added.

---

sinelaw (commented):
## Re-review of `8176741` — blocking items resolved ✅

Thanks for the quick turnaround. All three items from my first pass are addressed correctly:

- **Dead PoC removed** (`chrome_layout`/`chrome_html`/`chrome_snapshot` + example + HTML, ~1.3k lines) — verified no dangling refs (remaining `ChromeLayout` hits are the unrelated per-window layout cache), and the research doc carries an honest "superseded/historical" note.
- **CORS wildcard dropped** — closes the cross-origin `/state` read vector; same-origin policy is the protection, as the new comment notes.
- **`handle_action` in the prod routes**; the README is rewritten to describe the real `scene.rs` projection path (no more `ChromeSnapshot`).

I also did a deeper pass on the web→core hit-routing this round and confirmed the three trickiest paths route by consistent semantic identity, not pixels: **Settings** item/category indices match `selected_item`/`selected_category`; **DualList** row cursors line up with the `included_items()`/`available_items()` enumerations that `add_selected`/`remove_selected` index; and **context menus** invert cleanly (`item_idx = row - menu_y - 1` ↔ the frontend's `(x+1, y+1+i)`, with clamped-vs-raw position consistent on both sides).

Verified locally on this SHA: `webui_server` build clean (exit 0), `scene_parity` green. CI shows fmt/clippy/doc/schema/check already passing; only the OS test matrix is still in-flight (the sole reason for "unstable").

The three inline comments here are all **minor / non-blocking** polish — a TUI hot-path allocation, a frontend dedup inconsistency, and a scale note on the poll loop. No objection to merging once the test matrix is green.

---

crates/fresh-editor/src/view/ui/status_bar.rs
sinelaw: Fixed in `9b2a5cb`. Gated all three `segments` pushes — and their text allocations — behind `!draw`, so the TUI render path no longer builds the `Vec<StatusSegmentInfo>` / per-segment `String`s it never reads (the doc comment at L718 already stated that intent; the gate was just missing). The web path (`draw == false`) still builds them, so `status_view()` is unaffected — scene-parity test green, e2e 50/50.

---

web-ui/index.html
sinelaw: Fixed in `9b2a5cb` — `sendMouse` now routes through `applyScene` like the other senders, so a byte-identical scene (pure `moved`/hover, no-op drags) skips the `app.innerHTML` rebuild. e2e still 50/50.

---

web-ui/index.html
sinelaw: Agreed — leaving this one open as tracked future work rather than papering over it. Both points are accurate: the poll loop full-scene stringify+diff every ~40ms is the right thing to replace with a visible-window cell diff (the established xi-editor lesson), and the single-threaded server means a `/state` poll can interleave an extra `editor_tick` between input POSTs (harmless today since every request renders unconditionally). Neither is a correctness bug for the local prototype, so I'd rather address the cell-diff properly as its own change than rush it here. (Note: my README rewrite dropped the old "ship only the visible-window cell diff" TODO line — I'll make sure that future-work item is captured in `UNIFIED_SCENE_DESIGN.md` so it isn't lost.)

---

crates/fresh-editor/src/view/scene.rs
```diff
@@ -0,0 +1,1489 @@
...
+            .map(|m| m.menu_areas.iter().cloned().collect())
+            .unwrap_or_default();
+
+        // Same expanded menu list the TUI renderer uses (config + plugin menus),
+        // so the two frontends never diverge on which menus/items exist.
+        let menus: Vec<MenuEntry> = self
+            .all_menus_expanded()
+            .iter()
+            .enumerate()
+            .map(|(i, m)| MenuEntry {
```
sinelaw, line 192: `menu_view()` projects a `MenuEntry` for every expanded menu, including `when`-hidden ones (they just get `x: None`), and leaves it to the frontend to skip them. The TUI instead filters via `menu_visible` in `MenuRenderer`. So menu visibility is computed independently in each frontend — exactly the kind of per-frontend divergence this design is meant to eliminate. Suggest deriving visibility here (e.g. add a `visible: bool` to `MenuEntry` from the same `when` evaluation) so both renderers consume one answer.

---

crates/fresh-editor/src/view/scene.rs
```diff
@@ -0,0 +1,1489 @@
...
+            p.scroll_offset,
+            p.suggestions.len(),
+            p.suggestions.len(),
+        ));
+        Some(PaletteView {
+            query: p.input.clone(),
+            message: p.message.clone(),
+            prompt_type: prompt_type_tag(&p.prompt_type),
+            overlay: p.overlay,
+            title: p.title.iter().map(|t| t.text.as_str()).collect(),
```
sinelaw, line 424: The prompt title is flattened to a plain `String` here (`p.title.iter().map(|t| t.text...).collect()`), discarding the per-span styling that `StyledText` carries and that the TUI renders. The web title will lose any plugin-set colors/emphasis. If title styling matters for parity, project the styled spans instead of collapsing to text.

---

crates/fresh-editor/src/view/scene.rs
```diff
@@ -0,0 +1,1489 @@
...
+
+    /// Semantic status bar: the whole bar tiled into labeled indicator segments
+    /// plus the untracked text runs between them (file name / Ln,Col). The
+    /// segment *text* is lifted from the rendered `buf` for now. Single
+    /// derivation shared by both frontends.
+    pub fn status_view(&self) -> Option<StatusView> {
+        let chrome = self.active_chrome();
+        let (sy, sx, sw) = chrome.status_bar_area?;
+        let mid = sx.saturating_add(sw / 2);
+        let side = |x: u16| if x < mid { "left" } else { "right" };
```
sinelaw: `side` is decided by a raw midpoint heuristic (`x < mid`) rather than the renderer's actual left/right tiling. A segment that straddles the center (or the boundary one when widths are odd) can be assigned to the wrong side, so the native status bar can order/justify segments differently from the TUI. Since the renderer already knows which run is left- vs right-aligned, carry that on `StatusSegmentInfo` instead of re-deriving it from x.

---

crates/fresh-editor/src/webui/mod.rs
```diff
@@ -0,0 +1,802 @@
...
+            eprintln!("handle_mouse error: {e}");
+            break;
+        }
+    }
+}
+
+/// ratatui Color → CSS hex (or None for the terminal default).
+fn color_css(c: Color) -> Option<String> {
+    let hex = |r: u8, g: u8, b: u8| format!("#{r:02x}{g:02x}{b:02x}");
+    Some(match c {
```
sinelaw, line 777: `color_css` and `indexed_css` (below) hardcode the 16-color basic palette twice, and the two tables disagree at the endpoints: here `White`→`#e5e5e5` and `Black`→`#000000`, while `indexed_css` maps index 15→`#ffffff` and index 0→`#000000`. So the same logical color resolves to a different hex depending on whether it arrives as a named `Color` or an `Indexed`. Factor the 16-entry table into one shared constant and have both paths read from it.

---

crates/fresh-editor/src/webui/mod.rs
```diff
@@ -0,0 +1,802 @@
...
+        for x in 0..buf.area.width {
+            out.push_str(buf.cell((x, y)).map(|c| c.symbol()).unwrap_or(" "));
+        }
+        out.push('\n');
+    }
+    out
+}
+
+pub fn run(addr: &str, files: &[PathBuf]) -> Result<()> {
+    let (mut cols, mut rows) = (140u16, 44u16);
```
sinelaw: The initial `140x44` default is hardcoded here and then re-hardcoded independently in the `/reset` route (`*cols = 140; *rows = 44;`). They can silently drift apart. Pull the default size into a shared `const (u16, u16)` and reference it from both.

---

crates/fresh-editor/src/app/mod.rs
```diff
@@ -1377,6 +1386,28 @@ impl Editor {
             .viewport
     }
 
+    /// Width (in cells) of the line-number gutter for a given split leaf, or 0
+    /// when that split hides line numbers. The same value the renderer uses, so
+    /// a frontend can peel the gutter off the buffer text at the exact column.
+    pub fn leaf_gutter_width(&self, leaf: fresh_core::LeafId, buffer_id: BufferId) -> u16 {
```
sinelaw, line 1392: `leaf_gutter_width` recomputes the gutter width via `vs.viewport.gutter_width(&buffer)` outside the render pass, and the bridge relies on it matching the column where the renderer actually split the gutter from the buffer text. There's no shared helper, so a future change to how the renderer sizes the gutter would silently desync the web's gutter/text slice (off-by-N column) until both sites are updated. Consider having the renderer record the gutter width it used (alongside the other per-leaf layout it already caches) and reading that here instead of recomputing.

---

sinelaw (commented):
Non-test, non-security review notes — six items on shared-model fidelity and duplicated derivations. None blocking; the layout-vs-draw seam and shared-dispatch extraction are clean.

---

crates/fresh-editor/src/view/scene.rs
sinelaw: Fixed in `33d4f17`. Added a shared `is_menu_visible(menu, context)` in `view/ui/menu.rs`; the TUI `MenuRenderer::is_menu_visible` now delegates to it, and `menu_view` calls the same function to set a derived `visible` on each `MenuEntry`. The frontend skips `visible === false` menus instead of re-deriving from `x == null`. So `when`-visibility is computed once. Parity + e2e (50/50) green.

---

crates/fresh-editor/src/view/scene.rs
sinelaw: Fixed in `33d4f17`. `StatusSegmentInfo` now carries a `side` ("left"/"right") set by the renderer in its actual left/right tiling passes, and `status_view` reads `s.side` instead of the `x < mid` midpoint guess — so a straddling/odd-width segment can't be mis-assigned and the web orders/justifies exactly as the TUI.

---

crates/fresh-editor/src/webui/mod.rs
sinelaw: Fixed in `33d4f17`. Factored the 16 ANSI colors into one shared `ANSI16` table; both `color_css` (named-color arm, each mapped to its ANSI index) and `indexed_css` read from it. The endpoint disagreement is gone — a logical color now resolves to one hex whether it arrives named or as `Color::Indexed` (e.g. white is `#e5e5e5` at index 7, bright white `#ffffff` at index 15 in both paths).

---

crates/fresh-editor/src/webui/mod.rs
sinelaw: Fixed in `33d4f17` — pulled the default into `const DEFAULT_SIZE: (u16, u16) = (140, 44)`, referenced by both `run()` and the `/reset` route.

---

crates/fresh-editor/src/view/scene.rs
sinelaw: Valid parity gap, but deferring as tracked follow-up rather than fixing in this PR. Projecting styled title spans means adding a styled-span structure to `PaletteView` and teaching the frontend to render per-span colors/emphasis. The prompt title is plain text in the overwhelming majority of cases (plugin-set title styling is rare), so the cost/benefit favors doing it as its own focused change. Leaving this thread open so it isn't lost; I'll capture it in `UNIFIED_SCENE_DESIGN.md` alongside the other web-bridge follow-ups.

---

crates/fresh-editor/src/app/mod.rs
sinelaw: Fair coupling concern — leaving open as tracked follow-up. Worth noting there's no *actual* desync today: `leaf_gutter_width` calls the exact same `viewport.gutter_width(&buffer)` the renderer uses to size the gutter, so the column always matches. The risk is purely future — if the renderer's gutter sizing changes without updating this call. The clean fix is to have the render pass record the gutter width it used into the per-leaf layout cache and read that here; that touches the render pass, so I'd rather do it deliberately than rush it into this PR. I'll capture it in `UNIFIED_SCENE_DESIGN.md` with the other follow-ups.

---

PR #2372, opened 2026-06-16
Theme audit for correct APCA contrast ++

dcog989:
All themes audited against APCA for acceptable contrast, adjusted accordingly. Minor tweaks for e.g. selected text bg being almost the same as current line bg; refined high-contrast to be better for protan/deutan color blindness; added new theme one-dark-x, a derivation of 'One Dark Pro'.

All code eyeballed, and themes viewed against JS + CSS + Python.


sinelaw:
Some of the changes actually lower the contrast between BG and FG, can you please confirm?

dcog989:
I've just quickly scanned again, and a couple were reduced contrast slightly but still ok for APCA.

I also missed a few, although some were intentional to maintain theme aesthetic - e.g. nord is very muted. If I make everything hard APCA-compliant, it will effectively 'break' some themes - and people will complain!

I'll go through them again and tweak the script I used for the first run.

Note: I realised theme.rs needs to reflect the new colors, so will update that as well - sorry for scope creep.

dcog989:
I've gone down a rabbit hole on this - my 'clever' script was not that clever. Need to rethink.

There are some real issues with current themes that need fixing, but I might just address those by hand.

I'm closing this as I need to start fresh. Sorry for the noise. v2 will be good!

PR #2377, opened 2026-06-17
docs(ssh): document persistent sessions over SSH from a single jump box

Monear:
## What

Documents the "jump box" workflow for SSH remote editing: launching a **persistent session bound to a remote target** from a single machine that has Fresh installed — `fresh -a <name> user@host:path`. Adds a short subsection to the existing *Alternative: SSH + Session Persistence* part of `docs/features/ssh.md`.

## Why

Remote editing and session persistence are documented separately. The current "SSH + Session Persistence" section only shows running `fresh -a` **on the remote host**, which assumes Fresh is installed there. But the CLI already supports passing a remote target straight to a persistent session, so you can keep Fresh on **one** jump/bastion box (targets only need Python 3) and still get detach/reattach persistence. That's a genuinely useful workflow that wasn't written down anywhere.

## This is existing, intended behavior — not a new feature

The combination is already implemented and guarded in `crates/fresh-editor/src/main.rs`:

- `run_attach()` calls `extract_ssh_url_from_files()` and forwards the result to `spawn_server_detached(session_name, ssh_url)` — the comment notes the remote URL "becomes the daemon's startup authority," and that it's consumed only when *starting* a server (reattach ignores it).
- `extract_ssh_url_from_files()` enforces the single-backend rule with explicit errors: "Cannot open files from multiple remote hosts" and "Cannot mix local and remote files."

The doc wording mirrors those semantics (one session = one backend; reattach with bare `fresh -a <name>`; unsaved buffers preserved via hot-exit).

## Notes

- Docs-only change; no code touched.
- Happy to adjust wording/placement, or fold it into the main flow instead of the alternatives section, if you'd prefer.

### DIFF
--- a/docs/features/ssh.md
+++ b/docs/features/ssh.md
@@ -62,3 +62,21 @@ fresh -a
 ```
 
 You can also pair SSH with `tmux` for a similar effect—run `tmux` on the remote host and launch Fresh inside it. Daemon mode has the advantage of being built into Fresh, so editor state (open files, terminals, undo history) is preserved without an external multiplexer.
+
+### Single-install "jump box" pattern
+
+You don't need Fresh on every machine you edit. Pass a remote target directly to a daemon: the remote URL becomes the daemon's startup authority, so the daemon runs on the machine you launched it from (e.g. a jump/bastion host) and edits the remote over SSH. The target still only needs Python 3; Fresh is installed solely on the box you launch from.
+
+```bash
+# On the one box where Fresh is installed:
+fresh -a webserver deploy@server.example.com:/etc/nginx/nginx.conf
+
+# Detach (Command Palette → "Detach"), then reattach later from the same box:
+fresh -a webserver
+```
+
+Notes:
+
+- The remote target is only consumed when the daemon is *started*; reattaching to an existing daemon ignores it, so `fresh -a webserver` is enough to reconnect.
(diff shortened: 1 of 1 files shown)


sinelaw:
Actually, it is possible to connect both to a remote and the local machine in the same fresh process using the orchestrator new session feature. 

add a short note about that?


By the way I'm working on changing the terminology dropping the word session as it's too confusing and moving to
- "workspace" for the thing that's managed by the orchestrator and 
- the word "daemon" for the persistent process that you can reattach to.

Monear:
Thanks! Added a note: clarified that the single-backend rule is *per session*, and that **Orchestrator: New Session** opens an additional session with its own authority, so a local session and a remote session (or different hosts) can run side by side in one Fresh process.

I kept the current "session" wording so it matches what ships today (the `-a` flag, the "Orchestrator: New Session" palette label, and the linked Session Persistence page). Happy to switch the doc over to the workspace/daemon terminology once that rename lands.

sinelaw:
The terminology rename is merged now, so you can update the wording (also you need to rebase anyway)

Monear:
Done on both counts:

- **Rebased** onto current `master` (now a single clean commit, no conflicts).
- **Updated the wording** to the merged terminology: the persistent process is now a *daemon*, and the per-backend unit is a *workspace*. The note now points at the **Orchestrator: New Workspace** command and explains that a local workspace and a remote workspace can run side by side in one Fresh daemon.

PR #2379, opened 2026-06-17
fix(popup): drive completion popup colors from the active theme

### LINKED ISSUE #2378: Popup colors don't follow the active theme
Popups — including LSP completion suggestions, plugin action popups, and others created via `Event::ShowPopup` — ignore the theme's `popup_*` color keys. The popup background renders as a hardcoded near-black rectangle, and item text inherits the terminal's default foreground instead of `popup_text_fg` / `popup_selection_fg`.

This is most visible on themes that customize popup colors (especially those relying on ANSI named colors via a custom terminal palette).

peanball:
Popups (LSP completion, plugin actions, anything pushed via
Event::ShowPopup) ignored the theme's popup_* keys: the background
was hardcoded to Color::Rgb(30, 30, 30), and PopupContent::List
rows never stamped popup_selection_fg / popup_text_fg, so item
text inherited the terminal's default foreground.

- Thread popup_bg / popup_border_fg through convert_popup_data_to_popup.
  Editor-level callers pass the live theme; the replay path falls
  back to default_popup_bg / default_popup_border_fg.
- PopupContent::List row rendering now stamps popup_selection_fg/bg on
  selected rows and popup_text_fg/popup_bg on unselected rows.
- Stamp ui.popup_text_fg on unselected rows in the prompt-toolbar
  completion dropdown (widgets/render.rs).

Existing issue_2_show_popup_ignores_theme_popup_bg test is reframed
from a known-fail bug doc into a passing regression guard.

Before:
image

After:
image

Closes #2378

### DIFF
--- a/crates/fresh-editor/src/app/lsp_requests.rs
+++ b/crates/fresh-editor/src/app/lsp_requests.rs
@@ -209,6 +209,10 @@ impl Editor {
             crate::app::popup_actions::build_completion_popup_from_items(all_popup_items, 0);
         let accept_hint = self.completion_accept_key_hint();
         let focus_hint = self.popup_focus_key_hint();
+        let (popup_bg, popup_border_fg) = {
+            let theme = self.theme();
+            (theme.popup_bg, theme.popup_border_fg)
+        };
 
         {
             let buffer_id = self.active_buffer();
@@ -220,7 +224,8 @@ impl Editor {
                 .get_mut(&buffer_id)
                 .unwrap();
             // Convert PopupData to Popup and use show_or_replace to avoid stacking
-            let mut popup_obj = crate::state::convert_popup_data_to_popup(&popup_data);
+            let mut popup_obj =
+                crate::state::convert_popup_data_to_popup(&popup_data, popup_bg, popup_border_fg);
             popup_obj.accept_key_hint = accept_hint;
             popup_obj.resolver = crate::view::popup::PopupResolver::Completion;
             popup_obj.focus_key_hint = focus_hint;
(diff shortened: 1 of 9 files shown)


peanball:
The comments in the code are a bit extensive. Let me know if I should reduce them.

sinelaw (approved):

- **Fix is genuine and non-duplicated**: The previous partial workaround in `plugin_dispatch.rs` only re-stamped colors after construction; this PR properly threads `popup_bg`/`popup_border_fg` as parameters into `convert_popup_data_to_popup`, removing the inconsistency and covering the additional cases in `lsp_requests.rs` and `popup_actions.rs` that were never patched.
- **Architecture is sound**: The two-path design (direct callers pass live theme; the event-replay path in `EditorState::apply` uses defaults, then `show_popup` re-stamps with live theme) is correct and the PR handles both paths consistently.
- **`popup_text_fg` added to unselected rows**: The additional fix in `popup.rs` (stamping `popup_selection_fg` on selected rows, `popup_text_fg`/`popup_bg` on unselected) and in `widgets/render.rs` (unselected completion items in the toolbar dropdown) correctly prevents text from inheriting the terminal's default foreground.
- **Test correctly promoted**: `issue_2_show_popup_ignores_theme_popup_bg` transitions from a known-fail bug doc to a passing regression guard; assertions unchanged, error messages updated.
- **Interactive validation**: Tested with the built binary using the built-in `light` theme (`popup_bg: [232,238,245]`). Plugin popup (direnv) and command palette both rendered with the correct light background (`48;5;254m`) and medium-gray border (`38;5;244m`). With the dark theme the same elements used black background and cyan border — confirming the fix activates the theme-defined colors instead of the hardcoded defaults.
- **Minor nit (non-blocking)**: The inline comment in `state.rs` says "Editor-level callers … intercept `Event::ShowPopup` *before* it reaches `state.apply`" but the event does reach `state.apply` (the re-stamp happens *after* via `show_popup`). Consider softening to "after `state.apply` pushes the popup with defaults, `show_popup` re-stamps with the live theme." Comment verbosity in general is a bit high (as you noted) — trimming the longest explanatory blocks would help readability, but it is not a blocker.

---

sinelaw:
:fire:

PR #2398, opened 2026-06-19
feat(vi_mode): add compatibility motions and word search

NihilDigit:
## Summary

This PR fills in several common vi/Vim navigation and search behaviors in the `vi_mode` plugin:

- Enables arrow-key navigation in vi mode behind a default-on `arrowKeys` setting.
- Adds `W`, `B`, and `E` WORD motions in normal, operator-pending, and visual mode.
- Adds `{` and `}` paragraph motions in normal, operator-pending, and visual mode.
- Adds `*` and `#` word-under-cursor search behind a default-on `searchWordUnderCursor` setting.
- Supports count, wraparound, and `n` / `N` repeat behavior for `*` / `#` searches.
- Preserves counted WORD operator motions across `.` repeat, while allowing an explicit repeat count such as `2.` to override the original count.
- Covers byte-offset edge cases for multibyte text and astral Unicode in WORD operator motions.

## Notes

This PR adds two plugin settings:

- `arrowKeys` (default: `true`): binds the arrow keys in vi modes. When disabled, vi mode no longer handles arrow-key movement.
- `searchWordUnderCursor` (default: `true`): binds `*` and `#` in vi normal mode. When enabled, `*` searches forward for the word under the cursor and `#` searches backward; both support count, wraparound, and `n` / `N` repeat.

`W` / `B` / `E` and `{` / `}` are treated as baseline vi motion coverage, so they are always enabled rather than hidden behind settings.

The optional settings are limited to behaviors that are more editor-flavor dependent: arrow-key handling and `*` / `#` word-under-cursor search.

The repeat-count handling matters for counted operator motions: after `d3W`, `.` repeats the three-WORD delete; a prefix such as `2.` repeats the same change with count 2.

## Testing

- `./check-types.sh vi_mode.ts`
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p fresh-editor --test e2e_tests vi_vim_compat -- --nocapture`
- `cargo test -p fresh-editor --test e2e_tests vi_mode -- --nocapture`
- `cargo check --all-targets`
- `cargo clippy --all-targets`

### DIFF
--- a/crates/fresh-editor/plugins/check-types.sh
+++ b/crates/fresh-editor/plugins/check-types.sh
@@ -28,6 +28,8 @@ for file in "${FILES[@]}"; do
     --lib esnext,dom \
     --skipLibCheck \
     --allowImportingTsExtensions \
+    --ignoreConfig \
+    --ignoreDeprecations 6.0 \
     "$file" 2>&1; then
     ERRORS=$((ERRORS + 1))
   fi
--- a/crates/fresh-editor/plugins/vi_mode.ts
+++ b/crates/fresh-editor/plugins/vi_mode.ts
@@ -25,6 +25,7 @@ type TextObjectType = "inner" | "around" | null;
 type ModeBinding = [string, string];
 type WORDMotionKind = "forward" | "backward" | "end";
 type WordSearchDirection = "forward" | "backward";
+type WordSearchTarget = { start: number; end: number; text: string; wholeWord: boolean };
 
 // Types for tracking repeatable changes
 type ChangeType = "simple" | "operator-motion" | "operator-textobj" | "insert" | "line-op";
@@ -49,9 +50,11 @@ interface ViState {
   lastChange: LastChange | null; // For '.' repeat
   lastYankWasLinewise: boolean; // Track if last yank was line-wise for proper paste
   visualAnchor: number | null; // Starting position for visual mode selection
+  visualHead: number | null; // Active end of computed visual selections
+  visualRange: { start: number; end: number } | null; // Characterwise visual range for computed motions
   insertStartPos: number | null; // Cursor position when entering insert mode
   visualBlockAnchor: { line: number; col: number } | null; // For visual block mode
-  lastWordSearch: { word: string; direction: WordSearchDirection } | null; // For n/N after * or #
+  lastWordSearch: { text: string; direction: WordSearchDirection; wholeWord: boolean } | null; // For n/N after * or #
 }
 
 const state: ViState = {
@@ -64,6 +67,8 @@ const state: ViState = {
   lastChange: null,
   lastYankWasLinewise: false,
   visualAnchor: null,
+  visualHead: null,
+  visualRange: null,
   insertStartPos: null,
   visualBlockAnchor: null,
   lastWordSearch: null,
@@ -149,6 +154,8 @@ function switchMode(newMode: ViMode): void {
   // Clear visual anchor when leaving visual modes
   if (newMode !== "visual" && newMode !== "visual-line" && newMode !== "visual-block") {
     state.visualAnchor = null;
+    state.visualHead = null;
+    state.visualRange = null;
     state.visualBlockAnchor = null;
     // Clear any selection when leaving visual mode by moving cursor
     // (any non-select movement clears selection in Fresh)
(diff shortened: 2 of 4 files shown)


sinelaw (commented):
I built this branch and ran it side-by-side against Vim 9.1 in tmux, driving identical keystrokes into both and using **yank+paste / delete as ground truth** for selections (Fresh renders the visual caret on the trailing edge, so comparing the on-screen cursor mid-selection is misleading). The normal-mode and operator features are solid — but there's one class of **newly broken behavior** that the test suite doesn't cover.

### 🔴 Visual-mode `vW` / `vB` / `vE` are broken (new)

The new visual WORD motions select the wrong range — confirmed deterministically (5/5) and via both yank and delete:

| keys (cursor at start of `elephant zebra giraffe`) | Vim selects | Fresh selects |
|---|---|---|
| `vE` | `elephant` | **`el`** (2 chars) |
| `vW` | `elephant z` | **`el`** (2 chars) |
| `vB` (from a later word) | `beta g` | **nothing** (empty) |

Repro: `vEy$p` appends `el` in Fresh but `elephant` in Vim. The pre-existing lowercase `ve` / `vw` work correctly (`vey$p` → `elephant`), so this is specific to the new handlers.

Likely cause: `vi_vis_WORD` / `vi_vis_WORD_back` / `vi_vis_WORD_end` route through the new `selectToPosition()` helper, which extends the selection by looping `executeAction("select_right"/"select_left")` and breaking when `getCursorPosition()` "doesn't change". In Fresh's visual mode that read doesn't track the moving selection head the way the loop expects, so it bails after ~1 step forward and 0 steps backward. The existing visual word handlers (`vi_vis_word_end` etc.) instead drive native `select_word_right`/`select_left` directly by count, which works — the new motions probably want the same approach rather than the position-chasing loop.

Worth adding e2e coverage that asserts the *selected text* (e.g. `vE` then yank) for visual W/B/E — the current tests exercise normal- and operator-mode WORD motions but never visual-mode selection results.

### 🟠 `*` / `#` when the cursor is not on a word (new)

Vim scans **forward** on the current line to the next keyword and searches for it. Fresh instead steps **backward** to the preceding word, or no-ops on leading indentation:

- On the space after `cat` in `cat dog cat` → Vim searches `dog`; Fresh searches `cat`.
- On the indentation of `␣␣␣cat tail` → Vim jumps to `cat`; Fresh does nothing.

The common case (cursor on a word) matches Vim exactly, including whole-word, case-sensitivity, `count`, wrap, and `n`/`N`.

### 🟡 `}` at end-of-file (minor)

With a trailing newline, Vim's `}` lands on the last char of the last line; Fresh jumps to the empty virtual line after the final `\n`. Comes from the reused `move_to_paragraph_down` host action.

---

For completeness, these diverge from Vim too but are **pre-existing** (the lowercase equivalents behave identically), so they're not regressions from this PR: `cW` eats the trailing space (like `cw`); `dW` on the last WORD of a line joins the next line (like `dw`); and the status-bar `Ln` indicator goes stale after any `setBufferCursor`-based motion until the next keystroke (also affects `nG` on master — the cursor itself is correct, it's just the readout).

Everything else I tested matches Vim: `W`/`B`/`E` cursor motions across lines, `dW`/`d3W`, `.`-repeat with `2.` override, `*`/`#`/`n`/`N` on a word with counts/wrap, `{`/`}` interior paragraph motions, visual `}`, and arrow keys. Nice work on the `.`-repeat count preservation in particular.

---

sinelaw (changes requested):

NihilDigit:
I’ve addressed the regression cases from the review, and I also checked the affected behavior against real Vim.

While working through these cases, I noticed that most of the fixes are still local patches around specific edge cases. That is probably fine for this PR, but I’m not sure it is the best direction long term.

At the moment, `vi_mode` mixes plugin state, keybinding handlers, and direct editor actions. That keeps the implementation lightweight, but it also means the code does not have a shared model for motions, operators, visual selection, cursor position, and inclusive/exclusive ranges like Vim does.

I’m wondering whether it would make sense to gradually move `vi_mode` toward that kind of model. My guess is that it would make some of the current compatibility fixes simpler, and make future testing and debugging easier.

I would keep that separate from this PR. If that sounds like a direction you’d be open to, I can spend some time sketching out what the model might look like.

sinelaw:
Yeah that sounds reasonable, as I don't use vim I don't have the mental capacity to do that properly so would be happy if you want to take a shot at that. I would look into how other non-vim editors implement vi mode, eg evil mode in emacs, etc

sinelaw:
👍

PR #2399, opened 2026-06-19
Namespaced conceal range-clear (ConcealManager::remove_in_range_for_namespace)

Agrejus:
## What this is

One small, **additive** core primitive, carved out of #2325 so it can be reviewed and merged on its own. No graphics / kitty-protocol code is in here.

### `ConcealManager::remove_in_range_for_namespace()`
Clears only **one namespace's** conceal ranges that overlap a byte range, leaving other plugins' conceals in that range untouched. This mirrors the existing `OverlayManager::remove_in_range_for_namespace` (issue #2146) — conceals just never had the namespaced variant.

It's exposed to plugins as `editor.clearConcealsInRangeForNamespace(bufferId, namespace, start, end)` via a new `PluginCommand::ClearConcealsInRangeForNamespace`. This lets a plugin rebuild *its own* conceals for a line (the common per-line `lines_changed` rebuild pattern) without stomping conceals another plugin placed on the same line.

## Why it's split out like this

This is one of several PRs breaking up #2325, **which I'm closing in favor of these smaller PRs.**

This PR was originally scoped to also include a deterministic virtual-line ordering tiebreak in `virtual_text.rs`. Per @sinelaw's request, **that change has been removed** so the ConcealManager primitive can be merged on its own. The underlying scroll/flicker issue it addressed isn't reproducible on the maintainer's setup, so it'll be revisited separately rather than holding up this merge.

## Risk

Low. The change is purely additive — the conceal method is brand new; nothing previously called it.

## Tests
- `view::conceal::tests` — namespace isolation (only the target namespace's in-range conceals are removed; other namespaces and out-of-range entries survive) plus range/version no-op semantics.

Full local CI passes (`fmt`, `clippy`, `doc`, schema, `check --no-default-features`, and the touched test modules). The `ts_export` API-surface tests (incl. the generated `.d.ts` validation) pass with the new method.

---
Supersedes #2325 (being closed). Follow-up PRs will cover the virtual-line ordering, the markdown-compose rendering, and, separately, the graphics/kitty work for review.

### DIFF
--- a/crates/fresh-core/src/api.rs
+++ b/crates/fresh-core/src/api.rs
@@ -2627,6 +2627,17 @@ pub enum PluginCommand {
         end: usize,
     },
 
+    /// Remove conceal ranges overlapping a byte range, restricted to a single
+    /// namespace. Lets one plugin rebuild its conceals for a line without
+    /// destroying another plugin's ranges there (cf.
+    /// `ClearOverlaysInRangeForNamespace`, issue #2146).
+    ClearConcealsInRangeForNamespace {
+        buffer_id: BufferId,
+        namespace: OverlayNamespace,
+        start: usize,
+        end: usize,
+    },
+
     /// Add a collapsed fold range. Hides the byte range
     /// `[start, end)` from rendering — the line containing `start - 1`
     /// (the fold's "header") stays visible while the lines covered by
--- a/crates/fresh-editor/plugins/lib/fresh.d.ts
+++ b/crates/fresh-editor/plugins/lib/fresh.d.ts
@@ -2559,6 +2559,11 @@ interface EditorAPI {
 	*/
 	stringWidth(text: string): number;
 	/**
+	* Clear conceal ranges overlapping a byte range, restricted to one
+	* namespace — other plugins' conceals in the range are untouched.
+	*/
+	clearConcealsInRangeForNamespace(bufferId: number, namespace: string, start: number, end: number): boolean;
+	/**
 	* Add a collapsed fold range. Hides bytes [start, end) from
 	* rendering — the line containing `start - 1` (the fold "header")
 	* stays visible, while subsequent lines covered by the range are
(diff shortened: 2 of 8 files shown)


sinelaw:
What's the observable bug that is fixed by "Deterministic virtual-line ordering" ? Did you encounter an issue while building the graphics pr?

Agrejus:
> What's the observable bug that is fixed by "Deterministic virtual-line ordering" ? Did you encounter an issue while building the graphics pr?

@sinelaw when I was viewing the markdown preview, scrolling vertically caused the pipe table borders in continuation rows to jump out of position.  They would snap back into the correct position as I scrolled farther.

Turns out the continuation lines for a wrapped row are virtual texts sharing the same anchor and priority. They live in a `HashMap`, and the old sort only went position → priority, so equal-priority items at the same anchor had no real tie-break and just fell back to whatever order the `HashMap` felt like that frame. Every re-query on scroll could shuffle them, hence the jumping pipes.

Fix is just adding `VirtualTextId` (insertion order) as the final tie-break, so same position + same priority always resolves consistently and the rows stop reordering.

sinelaw:
Same here as in the other PR - I can't reproduce the scroll issue causing table borders to move.

Can we change this PR to only add the ConcealManager stuff so I can merge it separately?

sinelaw:
By the way which OS / Platform / Terminal app are you using?

Agrejus:
> Same here as in the other PR - I can't reproduce the scroll issue causing table borders to move.
> 
> Can we change this PR to only add the ConcealManager stuff so I can merge it separately?

I wonder if it is specific to my setup, I would think so

Agrejus:
> By the way which OS / Platform / Terminal app are you using?

I am using:

PC: Mac Book Pro M1
OS: Tahoe 26.3
Platform: Apple Silicon arm64
Terminal App: Ghostty running tmux

Agrejus:
> Same here as in the other PR - I can't reproduce the scroll issue causing table borders to move.
> 
> Can we change this PR to only add the ConcealManager stuff so I can merge it separately?

Yes, completed

sinelaw:
Needs rebase

PR #2400, opened 2026-06-19
Skip synchronized-update markers inside tmux (fixes stray glyphs on scroll)

Agrejus:
## The bug

When scrolling inside **tmux**, stray box-drawing glyphs (e.g. a leftover `│`) get left behind on otherwise-blank lines and stay there until a full redraw (`Ctrl-L`).

## Root cause

Every frame is wrapped in DEC private mode 2026 synchronized-update markers (`Begin/EndSynchronizedUpdate`) in the render loop, sent straight to stdout:

```rust
stdout().execute(crossterm::terminal::BeginSynchronizedUpdate)?;
terminal.draw(|frame| editor.render(frame))?;
stdout().execute(crossterm::terminal::EndSynchronizedUpdate)?;
```

Inside tmux those markers pass through to the outer terminal **unwrapped**, and tmux's handling can drop individual cell-clear updates from a frame diff. The composed frame buffer is correct, but the cell that should have been cleared never reaches the screen — so the old glyph persists until the whole screen is repainted.

## How I narrowed it down

- Reproduced live in tmux; characterized it: stray cell sits at a decoration's column, on blank lines, and clears on `Ctrl-L`.
- Proved the **composed buffer is always correct** — a headless ratatui-backend test (which reflects real terminal state via diffs, vt100-cross-checked) stayed clean across 60 scroll steps. So it isn't a logic/layout bug.
- Disabled the synchronized-update markers, rebuilt, scrolled again → strays gone. That isolates the markers as the cause.

## The fix

Gate the markers on **not running inside tmux** (`TMUX` env unset):

```rust
let synchronized_update = std::env::var_os("TMUX").is_none();
// ...
if synchronized_update { stdout().execute(BeginSynchronizedUpdate)?; }
terminal.draw(...)?;
if synchronized_update { stdout().execute(EndSynchronizedUpdate)?; }
```

tmux already batches its own pane refreshes per display cycle, so synchronized update buys nothing there. Outside tmux, the tear-free behavior is unchanged.

## Risk / testing

Low — one boolean gate around two escape writes; the actual frame draw is untouched, and non-tmux behavior is identical.

No automated test: this is terminal-output (escape-sequence) behavior under tmux specifically. The headless backend used in tests composes a fresh, correct buffer every frame and so cannot reproduce a tmux passthrough drop (that's exactly why the bug was invisible to CI). Verified manually: built a release binary, ran inside tmux, scrolled a decoration-heavy buffer, and confirmed zero stray cells across many scroll passes (vs. reliably reproducible before).

---
Split out of #2325 (closed); this is one of the non-graphics pieces broken out for easier review. Independent of the other splits — branches from `master` and can merge in any order.

### DIFF
--- a/crates/fresh-editor/src/main.rs
+++ b/crates/fresh-editor/src/main.rs
@@ -4351,6 +4351,14 @@ where
     let mut last_render = Instant::now();
     let mut needs_render = true;
     let mut pending_event: Option<CrosstermEvent> = None;
+    // Synchronized-update markers (DEC private mode 2026) are emitted around
+    // each frame to avoid tearing. Inside tmux they are passed through to the
+    // outer terminal unwrapped, and tmux's handling can drop individual
+    // cell-clear updates from a frame diff — leaving stale cells (e.g. a
+    // leftover box-drawing border glyph) on otherwise-blank lines until a
+    // full redraw. tmux already batches its own pane refreshes, so the markers
+    // buy nothing there. Skip them when running inside tmux.
+    let synchronized_update = std::env::var_os("TMUX").is_none();
     // Time of the last real input event, used to start the wave-animation
     // screensaver after the configured idle period. Read from the editor's
     // injected time source so tests can drive idle time deterministically.
@@ -4447,9 +4455,13 @@ where
             {
                 let _span = tracing::info_span!("terminal_draw").entered();
                 use crossterm::ExecutableCommand;
-                stdout().execute(crossterm::terminal::BeginSynchronizedUpdate)?;
+                if synchronized_update {
+                    stdout().execute(crossterm::terminal::BeginSynchronizedUpdate)?;
+                }
                 terminal.draw(|frame| editor.render(frame))?;
-                stdout().execute(crossterm::terminal::EndSynchronizedUpdate)?;
+                if synchronized_update {
+                    stdout().execute(crossterm::terminal::EndSynchronizedUpdate)?;
+                }
             }
             tracing::info!(target: "paste_timing", "render: {}ms (paste_pending={})", r0.elapsed().as_millis(), was_paste_pending);
             last_render = Instant::now();


sinelaw:
I use tmux all the time and haven't seen an issue. 
Can you explain step by step how to reproduce this?

Agrejus:
Here is a good video of it @sinelaw.  I only see this when in markdown preview mode, otherwise I have never observed this before.

https://www.loom.com/share/af293f91148741569545fc9563832850

sinelaw:
Thanks for sharing that video. I've been trying to reproduce the same locally and found a few issues but not this specific one, it just keeps the same border width. Will keep at it

sinelaw:
The only issue I can reproduce here is that the borders are not aligned correctly when there's an emoji (multibyte unicode) character. I used your PR https://github.com/sinelaw/fresh/pull/2401 which exposes stringWidth to fix that, in https://github.com/sinelaw/fresh/pull/2477

But I don't see the same thing as your video shows by just scrolling - scrolling doesn't change anything :\

sinelaw:
does the problem reproduce on latest master (which includes the unicode width fix)?

Agrejus:
> does the problem reproduce on latest master (which includes the unicode width fix)?

Nope, I cannot reproduce with the unicode fix.  I use a video from a previous commit

Agrejus:
> The only issue I can reproduce here is that the borders are not aligned correctly when there's an emoji (multibyte unicode) character. I used your PR #2401 which exposes stringWidth to fix that, in #2477
> 
> But I don't see the same thing as your video shows by just scrolling - scrolling doesn't change anything :\

Hmm, has to be something with my setup.  I wonder if there is an agnostic way to solve this.  If not, I can forgo the readme rendering in fresh.  It is super nice when I am presenting to enter preview mode and see well formatted markdown with mermaid diargams (if any)

sinelaw:
> Nope, I cannot reproduce with the unicode fix. I use a video from a previous commit

Does that mean we don't need another fix? Or it still reproduces if you include that + some more changes?

Agrejus:
> > Nope, I cannot reproduce with the unicode fix. I use a video from a previous commit
> 
> Does that mean we don't need another fix? Or it still reproduces if you include that + some more changes?

Appologies for the confusion.  I went to an old commit to show you the problem I was solving.  We do not need another fix because this PR fixes the stray glyph issue that I demonstrated in the video.

Agrejus:
This PR fixes a different root cause than the unicode width issue.  It's the tmux synchronized-update/stale-cell problem, which is environment-specific (tmux passthrough) and won't show on every setup. The code fix is still valid and additive.

Agrejus:
It feels a bit odd to check for tmux in `main.rs`, but I was not sure how else to solve this issue.  Happy to go back to the drawing board and close this PR if this way forward isn't ideal

sinelaw:
I want to reproduce it, and then we can decide how to fix. I'll try with a similar setup hopefully tomorrow

Agrejus:
## Detailed steps to reproduce the stray glyphs / floating pipe borders on scroll

Here's a reliable recipe from my side. The key thing I think was missing before: it only shows up when a **pipe table soft-wraps**, and it's most visible in **compose/preview mode** while scrolling.

### Environment
- Apple Silicon (M1 Pro), macOS
- Ghostty running **tmux**
- **Stock `fresh` — no custom plugins needed.** The table/border rendering is the bundled `markdown_compose` plugin (`crates/fresh-editor/plugins/markdown_compose.ts`), so this reproduces on a plain build.

### Necessary condition
The terminal/pane must be **narrow enough that the table wraps** onto continuation rows. If the window is wide enough that the table fits on one line per row, you won't see it — make the pane narrower (or the table wider). The continuation rows are what misbehave.

### Sample file
A table engineered to wrap at any normal width (6 wide columns):

<details>
<summary>floating-pipe-table.md</summary>

```markdown
| Service name and owner            | Primary responsibility (long description)                          | Upstream dependencies                          | Downstream consumers                           | On-call rotation / escalation path                | Notes, caveats, and known limitations                          |
| --------------------------------- | ------------------------------------------------------------------ | ---------------------------------------------- | ---------------------------------------------- | ------------------------------------------------- | -------------------------------------------------------------- |
| ingest-gateway (team-foundations) | Accepts inbound webhooks, validates signatures, enqueues for fanout | edge-lb, secrets-broker, schema-registry-cache | fanout-worker, audit-log-sink, metrics-relay   | foundations-primary → foundations-secondary → SRE | Drops payloads >8MB; retries are at-least-once so consumers must dedupe |
| fanout-worker (team-foundations)  | Reads the queue and fans messages out to per-tenant partitions     | ingest-gateway, partition-router, quota-store  | tenant-projector, billing-meter, search-feeder | foundations-primary → foundations-secondary → SRE | Backpressure trips at 50k depth; cold partitions add ~200ms first-write latency |
| tenant-projector (team-records)   | Materializes per-tenant read models from the fanout stream         | fanout-worker, schema-registry, blob-store     | api-edge, export-service, reporting-rollups    | records-primary → records-secondary → SRE         | Rebuilds are O(events); large tenants can take minutes to catch up after a reset |
| api-edge (team-surface)           | Public GraphQL + REST surface, auth, rate limiting, response cache | tenant-projector, auth-service, cache-cluster  | web-app, mobile-app, partner-integrations      | surface-primary → surface-secondary → SRE         | Cache TTL is 30s; stale reads possible right after a write on the same tenant |
| export-service (team-records)     | Streams bulk exports (CSV/Parquet) from the read models to blob    | tenant-projector, blob-store, quota-store      | partner-integrations, data-warehouse-loader    | records-primary → records-secondary → SRE         | Exports >2GB are chunked; partial-failure resume relies on idempotent chunk keys |
| billing-meter (team-commerce)     | Aggregates usage events into metered line items for invoicing      | fanout-worker, pricing-catalog, ledger-service | invoice-builder, finance-export, alerting      | commerce-primary → commerce-secondary → SRE       | Late events (>24h) are bucketed to the current period, not backdated |
| search-feeder (team-discovery)    | Transforms records into the search index document shape and ships  | fanout-worker, schema-registry, blob-store     | search-cluster, suggest-service, ranking-eval  | discovery-primary → discovery-secondary → SRE     | Reindex is throttled to protect the cluster; full reindex ~6h at current volume |
| audit-log-sink (team-foundations) | Durable append-only sink for every inbound and admin action        | ingest-gateway, secrets-broker, blob-store     | compliance-export, security-siem, forensics    | foundations-primary → foundations-secondary → SRE | Write-once; corrections are new compensating entries, never edits |
```
</details>

### Steps
1. Inside **tmux**, open the file in fresh: `fresh floating-pipe-table.md`.
2. Size the pane **narrow enough that each table row wraps** onto 2+ continuation lines.
3. Open the **Command Palette** (Cmd+Shift+P, or your configured binding) and run **`Markdown: Toggle Compose/Preview`**. The table now renders with drawn borders.
4. **Scroll up and down through the table repeatedly.** Half-page scrolls that re-enter the table from the top are the most effective.
5. Watch the vertical `|` borders on the **wrapped** rows.

### Expected
Borders stay aligned in the same columns as you scroll.

### Actual
Border glyphs on wrapped rows jump to the wrong column (a "floating pipe") / stray box-drawing glyphs appear on otherwise-blank cells, then snap back to the correct position on a full redraw.

### Notes on reliability
It's intermittent — give it a few scroll passes. Knobs that make it more reliable:
- **Narrower pane** → more wrapping → more visible shuffling.
- **More rows** → duplicate the rows above to 20+.
- **Faster repeated scrolling** → more re-renders per second.

### Loom

https://www.loom.com/share/b03002c03a35402b987f96df7129b87f

Agrejus:
Were you able to reproduce this @sinelaw?  Anything I can do to help?

sinelaw:
Sorry not yet (I tried a few times but not on the macOS). I'll try soon (TM). if you see it on other platforms/setups let me know

Agrejus:
No worries, sounds good.  I'll try on my Windows and Linux (Fedora) PC's and see if I can reproduce it

sinelaw (changes requested):
- Code itself is small and clean (single boolean gate around 2 escape writes), no duplication/complexity concerns.
- Violates CONTRIBUTING.md's "Reproduce Before Claiming" rule: no automated test pins the claimed tmux passthrough behavior (PR description admits this).
- Root cause is still unconfirmed after weeks of back-and-forth in this thread — the maintainer has been unable to reproduce it, and the original repro video predates an unrelated unicode-width fix.
- Interactive testing: built this branch, ran `fresh` inside tmux (`TMUX` env correctly detected, sync-update markers disabled per this fix) on a soft-wrapping markdown table in Compose/Preview mode. A table-border rendering glitch (missing `│` borders on a wrapped row) reproduced deterministically on the very first paint, with **zero scrolling**. This indicates the border glitches are a static width/layout computation bug in `markdown_compose.ts`, not the tmux synchronized-update passthrough issue this PR targets — so the fix doesn't actually resolve the visible defect.
- Recommend per CONTRIBUTING.md #16 ("Be wary of fixes you can't reproduce"): hold off until the root cause is confirmed with a concrete, automatable reproducer before merging a behavior change that disables tear protection for all tmux users.

---

sinelaw:
That comment was autogenerated, I still haven't gotten around to trying to repro. It's still on my list though

Agrejus:
@sinelaw I cannot reproduce on Linux (Fedora SilverBlue) or Windows.  I can go ahead and close this, seems like it may be a one off issue.  Will try to investigate further on my Mac and see if I can find a root cause.

sinelaw:
Thanks
I noticed there's still another bug in markdown table layout and borders, but it isn't a rendering artifact

PR #2404, opened 2026-06-20
fix(explorer): apply configured custom_ignore_patterns to the file tree

sfjohansson:
## Why
  
  Working in a Unity project, every asset has a sibling `.meta` file — so the
  file explorer shows two entries for everything (`Player.cs` + `Player.cs.meta`,
  `Scene.unity` + `Scene.unity.meta`, …), which roughly doubles the tree and
  buries the files you actually edit.

  Fresh already has a `custom_ignore_patterns` config field meant for exactly
  this. Setting it to `["*.meta"]` *looked* like it should hide them — but it had
  no effect.
  
  ## What Claude found
  
  The feature was only half-built upstream. The config field
  (`custom_ignore_patterns`) and the matcher in `view/file_tree/ignore.rs`
  (`add_custom_pattern` / `clear_custom_patterns` / `matches_custom_pattern`)
  both already exist — but nothing ever fed the config into the matcher. The
  field was parsed and then dropped on the floor.
  
  ## Approach

Claude hooked up the one that's already there.
  The config-to-matcher wiring is added at the two points where the explorer's
  ignore state is set:
  
  - explorer install — `FileExplorerViewDefaults` (initial open)
  - live settings re-apply — `settings_actions.rs` (config change without restart)

  Both clear and repopulate the matcher's custom patterns from
  `config.file_explorer.custom_ignore_patterns`. No behaviour change for anyone
  who leaves the field empty.
  
  ## Testing
  
  Adds an e2e test: sets `custom_ignore_patterns = ["*.log"]`, creates `keep.rs`
  and `debug.log` in a temp project, opens the tree, and asserts on rendered
  output — `keep.rs` is listed, `debug.log` is not. Fails without the wiring
  (`debug.log` shows up).

### DIFF
--- a/crates/fresh-editor/src/app/async_messages.rs
+++ b/crates/fresh-editor/src/app/async_messages.rs
@@ -1420,6 +1420,7 @@ impl Editor {
             show_hidden: self.config.file_explorer.show_hidden,
             show_gitignored: self.config.file_explorer.show_gitignored,
             compact_directories: self.config.file_explorer.compact_directories,
+            custom_ignore_patterns: self.config.file_explorer.custom_ignore_patterns.clone(),
         };
         let is_active = window == self.active_window_id();
         // Route the result back to the window that asked for it. If that window
--- a/crates/fresh-editor/src/app/file_explorer.rs
+++ b/crates/fresh-editor/src/app/file_explorer.rs
@@ -15,11 +15,12 @@ pub struct FileExplorerClipboard {
 /// Config-derived defaults handed to `Window::install_initialized_file_explorer`
 /// so the apply logic lives on `Window` without it borrowing the editor's
 /// `Config` (which would clash with the `&mut Window` borrow).
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone)]
 pub(crate) struct FileExplorerViewDefaults {
     pub show_hidden: bool,
     pub show_gitignored: bool,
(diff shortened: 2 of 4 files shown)


sinelaw:

- **Interactive validation**: checked out the PR branch, built locally, created a temp project with `config.json` containing `"custom_ignore_patterns": ["*.log"]`, and launched fresh — the file explorer showed only `keep.rs` and `config.json`; `data.log` and `debug.log` were correctly hidden.
- **Bug fix confirmed**: `custom_ignore_patterns` was parsed from config but never fed into the ignore matcher; the PR wires it up at both the initial-open path (`FileExplorerViewDefaults` → `install_initialized_file_explorer`) and the live settings-reload path (`settings_actions.rs`). Minimal, correct, and idiomatic.
- **Code quality**: removing `Copy` from the derive is necessary and correct; clear-then-repopulate pattern avoids stale state on reload.
- **Test**: e2e test observes only rendered output, uses `wait_until_stable` (no fixed timers), runs in an isolated temp project — all per CONTRIBUTING guidelines. Fails without the wiring, passes with it.
- **No duplicate**: the field existed in config but the feature was never functional; this completes it.

---

sinelaw (approved):
Approving after successful interactive validation — feature works as described.

---

sinelaw:
:fire:

PR #2479, opened 2026-06-23
Fix markdown table corruption under edit storms (per-line decorations + epoch-mapped coordinates)

### LINKED ISSUE #2484: Markdown compose: table borders corrupt when inserting lines above a table
## Summary

In Markdown compose/preview mode the `markdown_compose` plugin draws box-drawing
frames around tables (`┌─┬─┐` top, `├─┼─┤` separators, `└─┴─┘` bottom). When text
is inserted **above** a table — e.g. pressing Enter at the top of the document —
the table's frame becomes corrupted: the `┌─┬─┐` top border above the header is
lost (often replaced by a `├─┼─┤` separator), and duplicated / orphaned border
fragments and a stray blank line appear inside or around the frame.

The corruption is cumulative — the table renders fine for the first one or two
inserts and only breaks once several lines have been inserted — so rapidly
hammering Enter at the top of the file is the most reliable trigger.

## Steps to reproduce

1. Save the file below as `table.md`.
2. Open it: `fresh table.md`.
3. `Ctrl+P` → run **"Markdown: Toggle Compose/Preview"**. The table renders with a
   box-drawing frame (`┌─┬─┐ │ ├─┼─┤ └─┴─┘`).
4. `Ctrl+Home` to move to the very start of the file.
5. Press **Enter** several times in a row (e.g. 8×) to insert blank lines above
   the table.

### Example file (`table.md`)

```markdown
# Table Border Regression

Intro paragraph one.

| Task   | Owner | Status |
|--------|-------|--------|
| Build  | Alice | Done   |
| Deploy | Dave  | Done   |
| Docs   | Frank | Done   |

Tail paragraph one.
Tail paragraph two.
```

## Observed behavior

After inserting the blank lines above the table, the frame is no longer a single
well-formed box. Typical breakage:

- The `┌─┬─┐` top border is missing directly above the header row (the header
  renders with a separator or no top border at all).
- More than one top-left/bottom-left corner (`┌` / `└`) appears, or separators
  are duplicated / misplaced, sometimes with a blank line between the top border
  and the header.

```
┌────────┬───────┬────────┐
                              ← stray blank line
│ Task   │ Owner │ Status │
├────────┼───────┼────────┤
├────────┼───────┼────────┤   ← duplicated separator
│ Build  │ Alice │ Done   │
...
```

## Expected behavior

Inserting (or deleting) lines above a table should simply move the whole table —
and its borders — downward unchanged. The table should always remain a single,
well-formed frame: one `┌─┬─┐` top border directly above the header, `├─┼─┤`
separators between rows, and one `└─┴─┘` bottom border.

## Environment

- `fresh 0.4.1`
- View mode: Markdown compose/preview (`markdown_compose` plugin)
- Terminal: tmux, `TERM=xterm-256color`

sinelaw:
Fixes #2484

In compose/preview mode the `markdown_compose` plugin draws box-drawing frames
around tables (`┌─┬─┐` top, `├─┼─┤` separators, `└─┴─┘` bottom) and conceals the
raw markdown (`|` cell separators, `**` emphasis). Editing around a table
corrupted it. Three distinct symptoms, all fixed here:

- **Editing above/around a table** left a stray `┌`, **doubled separators**
  (`├─┼─┤` with no row between them), or a misaligned row.
- **Rapid Enter/Backspace storms** above a table left a frozen, mangled frame
  that only a cursor move would heal.
- **Rapid forward-delete of lines above a *wide* table** transiently scattered
  the cells — raw `|` and `**` leaked through, rows wrapped at the wrong
  columns — for as long as the deletes kept coming.

The root cause is the same throughout: `markdown_compose` decorates **per line**,
re-running on every `lines_changed`, and `lines_changed` is **fire-and-forget**
across the plugin-thread boundary. The byte coordinates a handler computes are
valid at the buffer version the event fired at, but the resulting decoration
commands apply *later*, after more edits have landed.

## Final architecture

Four pieces, layered:

**1. Per-line decorations, no stored positions.** Borders are emitted per line
exactly like conceals: for each line in a `lines_changed` batch the plugin
clears that row's border range and re-adds its frame, role determined locally
from the row and its batch neighbours. Border virtual lines are marker-backed
and auto-shift with edits — the same machinery conceals and overlays use — so
the plugin **stores no table byte positions at all**. This needed one missing
editor primitive, **`clearVirtualLinesInRange`** (the virtual-line analogue of
the existing `clearConcealsInRange`); the plugin's entire block/marker model is
deleted.

**2. Convergence on structure-changing edits.** The whole-viewport re-fire that
keeps per-line decorations consistent (`handle_refresh_lines` → clear
`seen_byte_ranges`) previously ran *only* for an inter-line `MoveCursor`. An
edit storm emits **no** `MoveCursor` events, so it never converged — the
per-edit `seen_byte_ranges` invalidation only drops the edited line, leaving the
table's rows re-fired at scattered versions with stale-role borders that froze
until the user moved the cursor. A newline insert/delete now triggers the same
convergence (`event_apply.rs`).

**3. Epoch-mapped coordinate repair.** A version-keyed edit-delta ring
(`model::coord_map::CoordMap`) records every edit, fed at the single
marker-adjustment chokepoint, so the editor can map a byte coordinate captured
at an older buffer version forward to the current version — reusing the exact
shift math the marker tree applies. `lines_changed` now carries its buffer
version as `epoch`.

**4. The runtime auto-stamps the epoch.** Rather than have the plugin thread the
epoch through every call by hand (fragile, easy to forget — which is exactly why
the conceal path lagged the border path), the plugin runtime stamps the active
hook's epoch onto **every** coordinate-bearing command (`addConceal`,
`clearConcealsInRange[ForNamespace]`, `add/clearSoftBreak`,
`add/clearVirtualLine`). The editor remaps the stale coordinate on apply via
`map_plugin_coord`. The plugin passes nothing and can't forget.

### Why both #2 and #3/#4 are required (validated empirically)

- **Without convergence (#2):** the narrow Enter/Backspace storm corrupts —
  incomplete re-fire leaves stale-role borders.
- **With convergence but without the epoch remap (#3/#4):** the wide delete
  storm corrupts — convergence forces a rebuild whose coordinate-based conceal /
  soft-break commands land at stale byte ranges under lag, leaking `|`/`**`.

Neither alone fixes both; together they do. (The coordinate desync is small —
≤1 version in light use — but under a sustained storm it grows enough to scatter
a wide table.)

## Per-split compose scoping (split view)

A separate but related bug, found while testing and reproduced interactively in
tmux: **compose decorations leaking across splits.** `markdown_compose` emits
its decorations onto the **buffer** whenever *any* split composes it
(`isComposingInAnySplit`), but `view_mode` is **per-split**
(`handle_set_view_mode` writes only the active split's buffer state). So a split
toggled back to source — while a sibling split kept composing the same buffer —
still rendered the boxed/concealed table instead of the raw `| Name | Role |`
markdown (the source split kept its line-number gutter yet drew the framed
table).

Fixed by gating the two still-ungated compose-decoration namespaces by the
split's `view_mode` at render time, mirroring the existing `md-emphasis` overlay
gate (`overlays.rs`):

- `md-syntax` conceals (cell separators / emphasis) are excluded from Source-mode
  rendering — new `ConcealManager::query_viewport_excluding` (`conceal.rs`),
  applied in `view_data.rs`.
- `md-tb` border virtual lines are dropped in Source mode (`inject_virtual_lines`,
  `transforms.rs`).

Soft breaks were already gated; `flash`'s source-mode conceals (`flash-label`)
and git-blame/diff virtual lines are unaffected.

## Solutions attempted, and why each was superseded

1. **Line-number bookkeeping** (original) — keyed namespaces, width cache, and
   first/last by line number. Inserting above a table renumbers every row → a
   separator drawn where the top border belongs. *Line numbers aren't stable
   under edits.*
2. **Interval markers, trust their coordinates** — track each table as a core
   interval marker and trust the shifted coordinates. *The plugin reads the
   marker off a shared snapshot at a different time than the event positions, so
   under load it sat a byte off and the merge doubled separators persistently.*
3. **Shift markers on every edit path** — added the missing
   `shift_plugin_markers_for_edit` to `insertAtCursor` / bulk edits. *Reduced but
   couldn't eliminate the race: the desync is temporal, not a forgotten shift.*
4. **Event-authoritative rebuild** — rebuild each table's rows from the live
   event every frame instead of merging marker coordinates. *Fixed the doubling,
   but the plugin still tracked byte positions — a parallel store that can
   desync.*
5. **Per-line decorations** (#1 above) — once it was clear border virtual lines
   already auto-shift and only lacked a per-range clear, the stored model became
   unnecessary. Shipped, and fixed the *editing-around-a-table* class.
6. **Convergence fix** (#2 above) — fixed the narrow Enter/Backspace storm. But
   instrumenting the *wide* delete storm showed the convergence-forced rebuild
   was itself leaking stale coordinates — so:
7. **Per-command epoch remap** — extended #3 to the border commands. Worked, but
   was opt-in per call and easy to miss (conceals were still broken). Superseded
   by **#4, runtime auto-stamp**, which makes it forget-proof and uniform, and
   let the manual epoch plumbing be deleted.

## Tests

- `coord_map` unit tests — the version-keyed remap: insert/delete/bulk,
  eviction, barriers, undo-as-forward-delta, replacement-as-net-delta.
- `state::coord_map_remaps_through_real_edits` — the marker-adjust chokepoint
  feeds the ring end-to-end through real `apply_insert`/`apply_delete`.
- `bulk_edit_records_into_coord_map_ring` — the forward bulk-edit path
  (paste / multi-cursor / query-replace / code actions) records into the ring.
- `structure_changing_edit_clears_seen_byte_ranges` — a newline edit clears
  `seen_byte_ranges` (forcing convergence); an intra-line edit does not (cheap
  path preserved).
- `test_table_border_no_doubled_separators_on_marker_event_desync`,
  `test_table_border_survives_insert_above`,
  `test_table_columns_uniform_width_under_partial_scroll`,
  `test_table_at_buffer_start_renders`, `test_table_at_buffer_end_has_bottom_border`,
  `test_clear_lines_in_range_tolerates_offset_anchor` — the per-line border
  regressions.
- `split_view_markdown_compose::test_disable_compose_in_one_split_restores_source_in_that_split`
  — with one split composing and one in source, the source split must show raw
  `|` while the composing split keeps its `┌` frame. Fails without the per-split
  decoration gate.
- `markdown_compose` e2e pass; `live_diff`/`git_blame`/`flash` (which share the
  now-auto-stamped commands) pass; `fmt` and the no-plugins build are clean.
  Verified interactively in tmux: rapid Enter/Backspace storms, wide-table
  forward-delete storms, and per-split compose toggling all stay clean.

## Known follow-ups (out of scope here)

- **Wrapped-last-row bottom border** — when a table's last row wraps, the `└─┘`
  border renders after the first wrapped visual row, spilling the continuation
  below the frame. Separate bug in `inject_virtual_lines`
  (`split_rendering/transforms.rs`); guarded by an `#[ignore]`d repro test
  (`test_table_bottom_border_below_wrapped_last_row`).
- **End-to-end rapid-delete repro** (`test_wide_table_not_corrupted_by_rapid_delete_above`)
  is `#[ignore]`d: faithful but races the real plugin thread, so it flakes under
  nextest's parallel load (a correct-but-not-yet-applied conceal reads the same
  as a leak). The fix's logic is covered deterministically by the `coord_map`
  tests above; run this repro locally with `--ignored`.


crates/fresh-editor/src/app/plugin_dispatch.rs
```diff
@@ -123,6 +123,61 @@ impl Editor {
...
+        };
+        for m in markers.values_mut() {
+            if removed == 0 {
+                if m.start >= pos {
+                    m.start += inserted;
+                }
+                if m.end > pos {
+                    m.end += inserted;
+                }
+            } else {
```
sinelaw, line 178: **Confirmed bug:** this branch drops the insert length on replacement edits. When `removed > 0` it only applies the deletion clamp (`x - removed`) and never adds `inserted`. The bulk path merges a same-position delete+insert into a single tuple with *both* lengths nonzero (type-over-selection, paste-over-selection, query-replace, LSP code actions), and that merged tuple reaches this function. `marker_list` (event_apply.rs:365) and `coord_map.record_replace` both shift by the net delta `inserted - removed`; this does not, so every plugin marker past the edit lands off by `inserted` bytes (wrong direction entirely when `ins > del`). The doc comment claims it mirrors the marker_list adjustments — it doesn't. Fix: shift by the net delta, ideally via the same `apply_delta` math the ring uses.

crates/fresh-editor/src/app/event_apply.rs
```diff
@@ -526,6 +526,15 @@ impl Editor {
             .highlighter
             .notify_edits(&edit_lengths);
 
+        // Bulk edits (multi-cursor typing, paste, code-action rewrites) apply
+        // here instead of through the Insert/Delete hook arms, so shift plugin
+        // interval markers for each edit too — mirroring the `marker_list`
+        // adjustments above — or plugin-tracked decorations keep stale coords.
+        #[cfg(feature = "plugins")]
+        for &(pos, del_len, ins_len) in &edit_lengths {
```
sinelaw, line 521: **Confirmed bug:** the forward bulk-edit path never records into the coord_map ring. Only `apply_insert`, `apply_delete`, and `replay_bulk_marker_adjustments` (undo/redo) call `coord_map.record_*`. This path bumps the buffer version and adjusts markers inline but produces version V+1 with **no ring delta**. A coordinate stamped at epoch V, remapped via `map_plugin_coord(coord, V)`, then finds no delta and returns an under-shifted (stale) byte — exactly the corruption coord_map exists to prevent — for paste / multi-cursor / query-replace / code actions. This violates the module's stated "ring coverage == marker coverage" invariant. Route the bulk path through `coord_map.record_*` too.

crates/fresh-editor/plugins/markdown_compose.ts
```diff
@@ -1491,159 +1523,103 @@ function processLineSoftBreaks(
...
-    // When mouse-scrolling, lines_changed only delivers NEW lines (not
-    // previously seen), so the group may not overlap with earlier cached
-    // rows of the same table. Scanning adjacently bridges the gap.
-    const firstLine = group[0].line_number;
-    const lastLine = group[group.length - 1].line_number;
-    for (let ln = firstLine - 1; widthMap.has(ln); ln--) {
-      mergeWith(widthMap.get(ln)!.maxW);
+    // Find this table's width memo (overlap query; write-through means two
+    // groups of one table in the same batch share it). Consolidate duplicates.
+    const near = (editor.queryMarkers(bufferId, gStart - 1, gEnd + 1) as Array<{
```
sinelaw: **Confirmed bug (crash):** when the table's first row is at byte 0 of the buffer, `gStart - 1 === -1`. `queryMarkers`'s Rust binding takes a `u32`, and rquickjs's `FromJs<u32>` returns `Err("Underflow")` for negatives → a thrown JS exception. The `(... as Array) || []` guard doesn't catch a *throw*, and there's no try/catch, so it propagates out of the `lines_changed` handler every render. A markdown file whose very first line is a table row never gets widths/conceals/borders and renders raw `|` pipes. Fix: `Math.max(0, gStart - 1)`.

crates/fresh-plugin-runtime/src/backend/quickjs_backend.rs
```diff
@@ -4993,6 +5032,147 @@ impl JsEditorApi {
...
+        let qs = start as usize;
+        let qe = end as usize;
+        let arr = rquickjs::Array::new(ctx.clone())?;
+        if let Ok(snapshot) = self.state_snapshot.read() {
+            if let Some(map) = snapshot.plugin_markers.get(&bid) {
+                let mut idx = 0usize;
+                for (key, m) in map.iter() {
+                    // overlap test: not (m.end <= qs || m.start >= qe), with
+                    // inclusive end so a zero-length query at a boundary hits.
+                    if m.end < qs || m.start > qe {
```
sinelaw, line 5161: **Confirmed bug:** the overlap skip test `m.end < qs || m.start > qe` is inclusive on both ends, but ranges are half-open `[start, end)`. Two markdown tables separated by one blank line: table A's memo ends at byte `e` (start of the blank line), B's header starts at `e+1`; B's padded query `[gStart-1, gEnd+1]` = `[e, …]` returns A's memo because `e < e` is false. The caller filters only by namespace prefix, so B adopts A's widths and extends A's span — the two tables fuse into one shared width memo and never separate. Use `m.end <= qs || m.start >= qe` (or drop the caller's ±1 padding).

crates/fresh-editor/plugins/markdown_compose.ts
```diff
@@ -1689,38 +1665,70 @@ editor.on("lines_changed", (data) => {
...
+      const prev = byLineNum.get(line.line_number - 1);
+      const next = byLineNum.get(line.line_number + 1);
+      // First/last is local. A row is first if it's the buffer's line 0 or its
+      // previous line is present in this batch and is NOT a table row; last if
+      // its next line is present and not a table row. When a neighbour is
+      // off-screen (absent from the batch) we conservatively treat the row as
+      // mid-table, so a tall table scrolled past its top/bottom never draws a
+      // spurious frame edge — it redraws when that neighbour re-enters a batch.
+      const isFirst = line.line_number === 0 || (prev !== undefined && !isTableRowContent(prev.content));
+      const isLast = next !== undefined && !isTableRowContent(next.content);
```
sinelaw: **Plausible bug (EOF):** `isLast` requires a following line in the batch. When the table's last row is the final line of the buffer, `next` is always undefined → `isLast` false → the `└─┴─┘` bottom border is never drawn, leaving an open-bottomed table. The old `!nextIsTable` semantics were true at end-of-buffer. Tests miss it because their tables always have a trailing line. Add a last-line-of-buffer fallback.

crates/fresh-editor/plugins/markdown_compose.ts
```diff
@@ -58,29 +46,113 @@ function isComposingInAnySplit(bufferId: number): boolean {
...
+function tableCells(content: string): string[] {
+  let inner = content.trim();
+  if (inner.startsWith("|")) inner = inner.slice(1);
+  if (inner.endsWith("|")) inner = inner.slice(0, -1);
+  return inner.split("|");
+}
+
+// Viewport-constrained per-column widths from accumulated max raw widths.
+function allocatedFor(maxW: number[]): number[] {
+  const viewport = editor.getViewport();
```
sinelaw, line 152: **Plausible bug (splits):** `allocatedFor` reads `editor.getViewport()` (the active split) and the width memo is keyed by BufferId alone. The same markdown buffer open in two side-by-side splits of different widths will lay out the inactive split's table using the active split's compose width — wrong column widths in the inactive split. The width memo key should include the split/view id.

crates/fresh-editor/src/model/coord_map.rs
```diff
@@ -0,0 +1,292 @@
...
+        self.deltas.len()
+    }
+}
+
+/// Shift a single coordinate through one edit that replaced `[pos, pos+removed)`
+/// with `inserted` bytes. Right-gravity at the boundary: a coordinate sitting
+/// exactly at an insertion point moves forward with the inserted text, matching
+/// how the marker tree shifts a right-gravity marker. A coordinate strictly
+/// inside a deleted range collapses to the deletion start.
+fn apply_delta(coord: usize, pos: usize, removed: usize, inserted: usize) -> usize {
```
sinelaw, line 179: **Plausible (niche):** `apply_delta` models a replacement as delete-then-insert, but the marker tree adjusts a same-position replacement by the *net* delta only. For a coordinate strictly inside the replaced span (`pos < c < pos+removed`), this collapses `c` to `pos+inserted` while the marker net-shift gives `c+(ins-del)` — the two disagree. So a stale plugin coordinate landing inside a replaced span (on undo/redo via `replay_bulk_marker_adjustments`, state.rs:990) is remapped to a different byte than its own auto-shifted marker, contradicting the "exact same shift math" guarantee.

crates/fresh-plugin-runtime/src/backend/quickjs_backend.rs
```diff
@@ -7665,6 +7850,16 @@ impl QuickJsBackend {
         self.services
             .set_js_execution_state(format!("hook '{}'", event_name));
 
+        // Stamp the hook's epoch (carried by `lines_changed` etc.) so every
+        // coordinate-bearing command the handlers emit is auto-tagged with the
+        // buffer version it was computed against — the editor remaps stale
+        // coordinates on apply. Saved/restored to nest correctly if a handler
+        // emits another event. Events without an `epoch` field clear it (None),
+        // which is correct: their commands carry no stale coordinate.
+        let prev_epoch = self.current_hook_epoch.get();
```
sinelaw, line 7887: **Plausible (niche):** `current_hook_epoch` is a single per-backend cell holding the firing hook's epoch (a per-buffer version), but it's stamped onto every coordinate-bearing command regardless of which buffer the command targets. A handler firing for buffer A that emits a command for buffer B stamps B's command with A's version; versions are per-buffer, so `map_plugin_coord` on B either replays unrelated deltas or returns None and silently drops the command. Stamp the epoch per target buffer.

crates/fresh-editor/src/app/plugin_commands.rs
```diff
@@ -434,6 +435,15 @@ impl Editor {
             .expect("active window present")
             .buffer_state_mut(buffer_id)
         {
+            // Repair a stale anchor: the plugin computed `position` against the
+            // `lines_changed` epoch, but later edits may have shifted it. Map it
+            // forward; if the epoch is too old to map, skip — convergence
+            // (cursor_moved → refreshLines) re-fires with fresh coordinates.
+            let position = match state.map_plugin_coord(position, epoch) {
```
sinelaw, line 442: **Altitude/duplication:** this `match (map_plugin_coord(start, epoch), map_plugin_coord(end, epoch)) { (Some, Some) => …, _ => return }` remap block is copy-pasted across ~8 handlers (conceals x3, soft-breaks x2, virtual-lines x2, add_conceal). Every new coordinate-bearing command must remember to re-add it or it silently regresses to stale-coordinate corruption. Extract a single `map_plugin_range(start, end, epoch) -> Option<(usize, usize)>` helper, or do the remap once at the dispatch chokepoint.

sinelaw (commented):
Automated code review (high-effort, multi-agent). Four confirmed correctness bugs + several plausible issues are posted as inline comments below.

**Theme:** the new coordinate-remap ring is meant to prevent stale-decoration corruption, but two holes (comments on `plugin_dispatch.rs` and `event_apply.rs`) leave that corruption fully present for the most common multi-byte edits (paste, multi-cursor, query-replace, LSP code actions). I'd treat those two as blocking and fix them at the shared edit chokepoint rather than per-branch.

Lower-priority cleanup/conventions not inlined: the `map_plugin_coord(start/end, epoch) else return` block is copy-pasted across ~8 handlers in `plugin_commands.rs` (and `shift_plugin_markers_for_edit` re-implements `coord_map::apply_delta` math — the duplication is *why* the inserted-length bug exists); `isSepRowContent()`/`tableCells()` helpers are added but raw copies remain at markdown_compose.ts:1114/1124/1420/1427 (one local shadows the module helper); war-story/incident comments at markdown_compose.ts:53-87, coord_map.rs:1-40, the test header in markdown_compose_table_border.rs, and `(issue #2515)` references in settings/{input,state}.rs.

crates/fresh-editor/src/app/plugin_dispatch.rs
sinelaw: Fixed in 6f7c46d9f. `shift_plugin_markers_for_edit` now reduces a same-position replacement to its net delta before shifting (`eff_removed`/`eff_inserted`), so it matches `marker_list` and `coord_map.record_replace`. Markers past the edit shift by `inserted - removed` in the correct direction.

---

crates/fresh-editor/src/app/event_apply.rs
sinelaw: Fixed in 6f7c46d9f. The forward bulk path now routes through `state.replay_bulk_marker_adjustments(&edit_lengths)` (event_apply.rs:370) instead of an inline marker loop, so paste / multi-cursor / query-replace / code actions all record into the ring. Ring coverage == marker coverage is restored, and a new test asserts `map_plugin_coord` shifts correctly after `apply_events_as_bulk_edit`.

---

crates/fresh-editor/plugins/markdown_compose.ts
sinelaw: Fixed. `queryMarkers` is now called with `Math.max(0, gStart - 1)` (markdown_compose.ts:1591), so a table whose first row is at byte 0 no longer underflows the `u32` binding. Covered by `test_table_at_buffer_start_renders`.

---

crates/fresh-editor/plugins/markdown_compose.ts
sinelaw: Fixed. `isLast` now has an end-of-buffer fallback: `bufEnd = getBufferLength(...)` is hoisted before the row loop and `isLast` is true when `next === undefined && line.byte_end >= bufEnd` (markdown_compose.ts:1713). A table that is the last line of the buffer now gets its `└─┴─┘` border. Covered by `test_table_at_buffer_end_has_bottom_border`.

---

crates/fresh-editor/src/model/coord_map.rs
sinelaw: Fixed in 6f7c46d9f. `apply_delta` now reduces a same-position replacement to its net delta first (mirroring how `replay_bulk_marker_adjustments` collapses the merged tuple before calling `marker_list.adjust_*`), so a coordinate strictly inside a replaced span maps to the same byte as its auto-shifted marker. New test `replacement_shifts_by_net_delta` covers both net-negative and net-positive cases, including a coord inside the span.

---

crates/fresh-plugin-runtime/src/backend/quickjs_backend.rs
sinelaw: Fixed in 6f7c46d9f. `current_hook_epoch` now stores `(buffer_id, epoch)` and `hook_epoch_for(buffer_id)` returns the epoch only when the firing hook's buffer matches the command's target buffer (quickjs_backend.rs:1030). A handler firing for buffer A that emits a command for buffer B now stamps no epoch on B's command (so it isn't remapped with A's unrelated version) rather than silently dropping it.

---

crates/fresh-editor/src/app/plugin_commands.rs
sinelaw: Fixed in 6f7c46d9f. Extracted `EditorState::map_plugin_range(start, end, epoch) -> Option<(usize, usize)>` (state.rs:459) and the range handlers (add_conceal, the two clear_conceals_in_range variants, clear_soft_breaks_in_range, clear_virtual_lines_in_range) now share it via a single `match … { Some(r) => r, None => return }` block.

---

crates/fresh-plugin-runtime/src/backend/quickjs_backend.rs
sinelaw: After investigating I'm leaving this one as-is — I believe the inclusive end is load-bearing rather than a bug. The same-table scroll-batch path relies on a row query `[gStart, gEnd]` re-finding the memo whose stored `end` equals `gStart` of the next batch within one table; making the test half-open (`m.end <= qs`) breaks memo sharing across scroll batches of a single table (`test_table_columns_uniform_width_under_partial_scroll`). The two-adjacent-tables fusion you describe would need a stale memo to survive the convergence re-fire, which now clears ephemeral decorations on structural edits. I added a comment at the call site documenting why the boundary is inclusive. Happy to revisit if you can point me at a repro where two adjacent tables actually fuse — I couldn't reproduce it in a test.

---

crates/fresh-editor/plugins/markdown_compose.ts
sinelaw: Acknowledged as a real but pre-existing issue, deferred out of this PR. It's orthogonal to the table-border corruption this PR targets (it predates these changes and needs the width-memo key to include the split/view id, which touches the layout cache more broadly). Left unresolved so it isn't lost — can spin it into a follow-up if you'd like.

---

PR #2505, opened 2026-06-27
fix(vi): make `cw` change to end of word like Vim, preserving trailing space

### LINKED ISSUE #2437: vi mode: `cw` deletes the trailing whitespace instead of acting like `ce`, unlike Vim
Tested on `fresh 0.4.1` (built from `origin/master` @ `eb3a349e6`), after the recent "feat: add Vim compatibility options to vi mode" / "fix: align vi compatibility motions with Vim" commits. Vi mode enabled via command palette → "Toggle Vi mode".

In Vim, `cw` is a documented special case: when the cursor is on a non-blank character, `cw` (and `cW`) behave like `ce` (and `cE`) — they change only up to the **end of the word** and do **not** consume the trailing whitespace. This is why `cw` is convenient: the space between words is preserved so you don't have to retype it. See `:help cw` / `:help word-motions`:

> Special case: When the cursor is in a word, "cw" and "cW" do not include the white space after a word, they only change up to the end of the word.

Fresh's vi mode instead makes `cw` behave like `dw` + insert: it deletes the word **and** its trailing space.

## Steps to reproduce
1. Create a file containing the single line `hello world foo bar` and open it: `fresh --no-restore motions.txt`.
2. Enable vi mode: `Ctrl+P` → "Toggle Vi mode" → Enter (status bar shows `NORMAL`).
3. Press `gg` then `0` to put the cursor on the first character (`h`, column 1).
4. Type `cw`, then type `X`, then press `Esc`.
5. Observe the resulting line.
6. For contrast, undo back to the original line, repeat with `ce` + `X` + `Esc` instead of `cw`.

## Expected behavior
Per Vim (the reference this feature is aligning to), `cw` and `ce` are equivalent on a non-blank character. Both should produce:

```
X world foo bar
```

(the space between `X` and `world` is preserved).

## Actual behavior
- `cw` + `X` produces `Xworld foo bar` — the trailing space is consumed (it changed `hello ` like `dw` would).
- `ce` + `X` produces `X world foo bar` — correct, space preserved.

So `cw` and `ce` diverge in Fresh, whereas in Vim they are identical. Reproduced 2/2 at the start of a word and also mid-word: with the cursor on column 3 (`l` of `hello`), `cw` + `X` yields `heXworld foo bar` (Vim: `heX world foo bar`).

For comparison, the operator/motion plumbing is otherwise working: `dw` from column 1 correctly yields `world foo bar` (deleting `hello ` *with* its trailing space, which is correct for `dw`). The bug is specifically that `cw` is not getting Vim's `cw`→`ce` special-case treatment.

## Workaround
Use `ce` instead of `cw` to change a word without eating the trailing space (or `cw` then re-type the space).

---
Search performed before filing (no existing issue): `cw vi mode change word`, `vi mode cw trailing whitespace`, `vim compatibility motion` — 0 results each.

sinelaw:
## Summary

Fixes #2437.

Vim documents `cw`/`cW` as a special case: when the cursor is on a non-blank character they behave like `ce`/`cE` — they change only up to the **end of the word** and do **not** consume the trailing whitespace (`:help cw`). This is what makes `cw` convenient: the space between words is preserved.

Fresh's lowercase `cw` instead behaved like `dw` + insert, deleting the word **and** its trailing space:

- Before: `cw` + `X` on `hello world` (cursor on `h`) → `Xworld` ❌
- After:  `cw` + `X` on `hello world` (cursor on `h`) → `X world` ✅ (matches `ce`)

The uppercase `cW` already implemented this (via `computeWORDOperatorRange`'s change-forward branch), but lowercase `cw` went through the native `move_word_right` motion, which has no such special case.

## Changes

- **`crates/fresh-editor/plugins/vi_mode.ts`**
  - `vi_op_word`: when the pending operator is `c` and the cursor is on a non-blank, compute a word-class-aware change range that stops at the end of the current word instead of advancing to the start of the next word. Plain `dw`/`yw`, and `cw` on a blank, keep the regular word-forward semantics.
  - New helpers `tokenRunEnd` / `viWordEndAdvance` / `computeWordChangeRange`. Unlike the whitespace-delimited `cW`, these respect Vim's word classes, so `cw` on `one.two` changes only `one` (stops at the punctuation).
  - The first word stops at the end of the current word with **no forward advance** (so the cursor on a word's last character changes only that character); each additional count behaves like Vim's `e`.
  - Dot-repeat (`.`) recomputes the range at the new cursor position, mirroring the existing `cW` repeat path.

## Testing

Four new e2e tests in `crates/fresh-editor/tests/e2e/vi_mode.rs` drive real keystrokes and assert on rendered output:

- `cw` keeps the following space (the issue's reproduction)
- `cw` stops at punctuation (`one.two` → `X.two`)
- `c2w` honors the count and still keeps the trailing space
- `.` repeat of `cw` replays the change (keep-space) semantics, not `dw`

Verified the 3 behavioral tests fail against the unfixed plugin and pass with the fix; the full 80-test vi-mode e2e suite still passes. Also manually validated in a tmux session: `cw` + `X` on `hello world foo bar` → `X world foo bar`.

https://claude.ai/code/session_012hiMwSvTcgCZwXL3F8b1Sv

---
_Generated by [Claude Code](https://claude.ai/code/session_012hiMwSvTcgCZwXL3F8b1Sv)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,7 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
 * **Regex search**: `^` / `$` now anchor per line in multi-line `Ctrl+F` search (#2495).
 * **Splits**: closing a buffer shown in two splits no longer desyncs the surviving cursor (#2496).
 * **Terminal**: a terminal restores its live/scrollback mode on refocus, and the last split is no longer left read-only after closing a second terminal (#2485).
+* **Vi mode**: `cw` on a non-blank now changes up to the end of the word like Vim (matching `ce`) instead of swallowing the trailing whitespace (#2437).
 * **Per-language indentation rules** are now actually applied — previously custom patterns were ignored (#2314).
 * **Brackets** are no longer highlighted inside comments and strings (#2405, reported by @TakemiSora).
 * **Quick-open `#` buffer switcher** now lists virtual buffers (#2373).
--- a/crates/fresh-editor/plugins/vi_mode.ts
+++ b/crates/fresh-editor/plugins/vi_mode.ts
@@ -751,6 +751,43 @@ function endWORDIndex(text: string, startIndex: number): number {
   return index;
 }
 
+// Word-class helpers for the `cw` special case (lowercase `w`, which — unlike
+// the whitespace-delimited WORD motions above — treats a run of word characters
+// and a run of punctuation as separate words).
+
+// Return the string index of the last character of the same-class (word or
+// punctuation) run that `index` is in. `index` must point at a non-whitespace
+// character.
+function tokenRunEnd(text: string, index: number): number {
+  const startIsWord = isWordChar(charAtStringIndex(text, index));
+  while (true) {
+    const next = nextStringIndex(text, index);
+    if (next >= text.length || isWhitespaceAt(text, next)) {
+      break;
+    }
+    if (isWordChar(charAtStringIndex(text, next)) !== startIsWord) {
+      break;
+    }
+    index = next;
+  }
+  return index;
+}
+
+// Vim `e`-style advance: from `index`, move forward one character, skip any
+// whitespace, then return the last character of the next word. Used for the
+// trailing words of a `cNw` change. Returns `index` unchanged if there is no
+// further word.
+function viWordEndAdvance(text: string, index: number): number {
+  let next = nextStringIndex(text, index);
+  while (next < text.length && isWhitespaceAt(text, next)) {
+    next = nextStringIndex(text, next);
+  }
+  if (next >= text.length) {
+    return index;
+  }
+  return tokenRunEnd(text, next);
+}
+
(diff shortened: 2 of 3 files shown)


sinelaw:
Manually validated in tmux: fresh (debug build) side-by-side with Vim 9.1 (`vim --clean`), identical keystrokes sent to both panes, buffers diffed byte-for-byte after `:w`.

- `cwX` + Esc on `hello world foo bar` (cursor col 1) → `X world foo bar` — matches Vim (trailing space preserved)
- `.` repeat on `one.two three` (col 1) → `X.two three` — matches (dot-repeat works and stops at punctuation)
- `c2wX` + Esc on `alpha beta gamma delta` → `X gamma delta` — matches (count honored, space preserved)
- `e` then `cwZ` + Esc (cursor on the word's last char) → `repeaZ one two` — matches (changes only that char)

All four scenarios ended byte-identical with Vim. (Validation binary was a local integration build with the five open vi-mode PRs cherry-picked on the shared base; this PR's changes merged cleanly.)

---

PR #2506, opened 2026-06-27
fix(indent): indent braceless control-flow bodies in TS/JS (#2492)

### LINKED ISSUE #2492: Auto-indent: braceless control-flow body not indented in TypeScript/JavaScript
## Summary

When auto-indent inserts a new line after a **braceless** control-flow statement (`if (...)`, `for (...)`, `while (...)` with no `{`), the body line is **not indented** in **TypeScript** and **JavaScript** files. The same edit in a C (or other C-style) file correctly indents the body one level.

## Steps to reproduce

1. Open a new file `test.ts` in fresh.
2. Type `if (a)` and press <kbd>Enter</kbd>.
3. Type `body`.

### Actual

The body stays at column 0:

```
if (a)
body
```

### Expected

The body should be indented one level under the braceless `if` (the standard behavior of VS Code / Prettier / clang-format):

```
if (a)
    body
```

## Comparison: same steps in a C file

Open `test.c`, type `if (a)`, <kbd>Enter</kbd>, `body`:

```
if (a)
    body
```

The C file indents correctly. TypeScript and JavaScript do not.

## Affected languages

| Language | Braceless body indented? |
|---|---|
| **TypeScript** | ❌ No (bug) |
| **JavaScript** | ❌ No (bug) |
| C | ✅ Yes |
| C++ | ✅ Yes |
| C# | ✅ Yes |
| Java | ✅ Yes |
| PHP | ✅ Yes |
| Kotlin | ✅ Yes |
| Dart | ✅ Yes |
| Scala | ✅ Yes |
| Swift | ✅ Yes |

Only **TypeScript** and **JavaScript** exhibit the problem. (Go and Templ require braces on control flow, so the braceless case does not arise there; JSON has no control flow.)

## Notes

- Reproduced with a debug build (`cargo build --bin fresh --features runtime`).
- Reproduced with `--no-plugins` and LSP off, so it is independent of plugins/LSP.
- Braced blocks (`if (a) { … }`), nested blocks, and dedent-on-`}` all work correctly in TypeScript/JavaScript — only the braceless control-flow body is affected.

sinelaw:
## Summary

Fixes #2492. Pressing Enter after a **braceless** control-flow head (`if (a)`, `for (…)`, `while (…)` with no `{`) in **TypeScript** or **JavaScript** left the body at column 0, while the same edit in a C-family file indented one level correctly.

```
if (a)
body        ← was: column 0 (bug)
    body    ← now: indented one level (matches C / VS Code / Prettier)
```

## Root cause

TS and JS have **bundled tree-sitter grammars**, so they use the AST indent path (`calculate_indent_tree_sitter`); the C-family languages without a bundled grammar use the regex rules tier, which already handles braceless heads via its `indent_next_line` pattern.

In the AST path, a braceless head's closing `)` is captured as `@dedent`. The "line ends with a closing token → maintain indent" short-circuit fired before the braceless-body indent could apply, so it returned the head's own indent (0). Mid-edit the enclosing block is still unclosed (its `}` is past the ~2000-byte parse window), so `if (a)` frequently parses as an `ERROR` node with no `if_statement` — which makes a purely structural fix unreliable.

## Fix

A line ending in `)` is genuinely ambiguous to the indent counter: it may close a braceless control head (indent one level), a call/group on its own line (`foo(a)` — no indent), or a multi-line argument list. The per-language regex rules tier already encodes exactly which keywords open a braceless body, so the AST path now **declines on a trailing `)` and defers to that tier** — restoring parity with the non-bundled C-family languages. Other closers (`}`, `]`, …) are unambiguous block closers and keep the maintain-indent behavior.

## Testing

New unit tests in `indent.rs`:
- `test_typescript_braceless_control_head_indents` / `test_javascript_braceless_control_head_indents` — `if (a)`, `for (x of y)`, `while (a)` → indent 4.
- `test_typescript_braceless_head_indents_when_nested` — body indents relative to the head's own column (→ 8).
- `test_typescript_complete_one_line_constructs_do_not_over_indent` — `function foo()`, `if (a) {}`, `const a = [1, 2]` must **not** gain a level (→ 0).

Full `fresh-editor` lib suite (2904 tests) and the auto-indent e2e suite pass. Manually validated in a real editor session (tmux):
- top-level `if (a)` + Enter → body at column 4
- `foo(a)` + Enter → next line at column 0 (no over-indent)
- nested `if (a)` inside a function + Enter → body at column 8

> Note: the `auto_indent` e2e suite has a pre-existing parallel-test race on the global `USER_RULES` (the config-driven tests call `clear_user_rules()`); it flakes on `master` as well and is unrelated to this change. The suite passes single-threaded.

---
_Generated by [Claude Code](https://claude.ai/code/session_014w7ADVZ9P7Ejq37gLGsM5C)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -112,6 +112,7 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
 
 ### Bug Fixes
 
+* **Auto-indent**: a braceless control-flow body (`if (…)`, `for (…)`, `while (…)` with no `{`) now indents one level in TypeScript and JavaScript, matching the C-family behavior (#2492).
 * **Regex search**: `^` / `$` now anchor per line in multi-line `Ctrl+F` search (#2495).
 * **Splits**: closing a buffer shown in two splits no longer desyncs the surviving cursor (#2496).
 * **Terminal**: a terminal restores its live/scrollback mode on refocus, and the last split is no longer left read-only after closing a second terminal (#2485).
--- a/crates/fresh-editor/src/primitives/indent.rs
+++ b/crates/fresh-editor/src/primitives/indent.rs
@@ -1093,11 +1093,32 @@ impl IndentCalculator {
             return Some(line_indent.saturating_sub(tab_size));
         }
 
-        // When the current line ends with a token captured as @dedent (`}`,
-        // `end`, `fi`, `done`, …), keep the new line at the same indent as
-        // the closing token. The grammar has already placed that token at the
-        // correct column matching its opener; the next line continues at the
-        // enclosing scope's indent.
+        // A line ending in `)` is ambiguous to the indent counter — and doubly
+        // so mid-edit, when the enclosing block is still unclosed and the parse
+        // window yields only ERROR nodes. The trailing `)` may close a braceless
+        // control head's condition (`if (a)`, `for (…)`, `while (…)` — the body
+        // must indent one level), a call/group on its own line (`foo(a)` — no
+        // indent), or a multi-line argument list. The per-language regex rules
+        // tier resolves this deterministically from the line text (its
+        // `indent_next_line` pattern encodes exactly which keywords open a
+        // braceless body), so decline here and let the public `calculate_indent`
+        // defer to it. This also restores parity with the C-family languages
+        // that have no bundled grammar and already use that tier (issue #2492).
+        let last_nonws_is_close_paren = last_nonws_offset
+            .and_then(|p| source.get(p))
+            .is_some_and(|&b| b == b')');
+        if last_nonws_is_dedent_capture && last_nonws_is_close_paren {
+            tracing::debug!(
+                "Cursor line ends with ')', deferring to the regex rules tier (issue #2492)"
+            );
+            return None;
+        }
+
+        // When the current line ends with another token captured as @dedent
+        // (`}`, `]`, `end`, `fi`, `done`, …), keep the new line at the same
+        // indent as the closing token. These are unambiguous block closers: the
+        // grammar has already placed the token at the correct column matching
+        // its opener; the next line continues at the enclosing scope's indent.
         if last_nonws_is_dedent_capture {
             let line_indent = Self::get_current_line_indent(buffer, position, tab_size);
             tracing::debug!(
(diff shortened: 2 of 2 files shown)


sinelaw:
Needs discussion

crates/fresh-editor-core/src/primitives/indent.rs
```diff
@@ -1216,11 +1216,32 @@ impl IndentCalculator {
...
+        // indent), or a multi-line argument list. The per-language regex rules
+        // tier resolves this deterministically from the line text (its
+        // `indent_next_line` pattern encodes exactly which keywords open a
+        // braceless body), so decline here and let the public `calculate_indent`
+        // defer to it. This also restores parity with the C-family languages
+        // that have no bundled grammar and already use that tier (issue #2492).
+        let last_nonws_is_close_paren = last_nonws_offset
+            .and_then(|p| source.get(p))
+            .is_some_and(|&b| b == b')');
+        if last_nonws_is_dedent_capture && last_nonws_is_close_paren {
```
sinelaw, line 1233: This is not TS/JS-only. `calculate_indent_tree_sitter` contains no `Language::` gate, so every bundled grammar reaches this line.

After this change, any line ending in `)` skips the AST tier and falls through to the regex rules — Rust `let x = foo(1)`, the closing paren of a multi-line signature, and so on, in every bundled language.

Tests cover TS/JS only, and the one non-TS-shaped guard is `function foo()`. Either gate on language, or add coverage for the other bundled grammars.

sinelaw (commented):

The fix works for the reported case. The concern is scope: the change is not TS/JS-specific. Details inline.

sinelaw (commented):

Follow-up to the previous comment. Tried to reproduce #2492 against the installed 0.5.1 binary and could not. It does not reproduce, because master already fixed it.

Repro attempt, 80x24 terminal, empty `.ts` buffer, typed `if (a)` then Enter then `body`:

```
1 | if (a)
2 |   body
```

Cursor lands at 2:7, i.e. the body is indented one level. Expected column 0 if the bug were present.

The fix is commit 88942ac0d (2026-08-10), "fix(indent): indent braceless control-flow bodies in TS/JS", at indent.rs:192-205. It keeps the tree-sitter result and asks the regex tier for `braceless_head_body_indent`, preferring it only when it is deeper. Its comment notes that languages whose rules have no `indent_next_line` pattern are unaffected — which is exactly the over-broad behavior flagged in the previous comment.

After the rebase this branch carries both fixes. They did not conflict because they sit in different functions:

- indent.rs:201 — master's `braceless_head_body_indent`
- indent.rs:1233 — this PR's `return None` for any line ending in `)`

The `return None` fires inside `calculate_indent_tree_sitter`, so it lands before master's "only if deeper" comparison can run, and discards the AST result for every bundled grammar. That is a regression against current master, not a fix. The CHANGELOG entry would also duplicate the one already there.

Suggest closing.

sinelaw (commented):

Checked the tests here against master's, to see if anything is worth keeping before closing.

Mostly not. Master already guards the case this PR's blanket `return None` would put at risk: `test_ts_plain_statement_keeps_indent` uses `if (a) {\n    foo(b)` — a line ending in `)` that must keep its indent rather than gain a level. Master also has the braceless if/while/for cases and the nested case.

Two cases here are not on master:

- `function foo()` on its own must not indent. Master has `function foo() {` (braced) and `function bar()\n    do_thing(` (multi-line call), but not the bare signature.
- JavaScript `for` and `while`. Master tests JS for `if` only.

Both are cheap to add to master's existing tests without any production change. Recommendation is unchanged: close this, port those two cases.

sinelaw:
Fixed on master by #2949, merged as 88942ac0d. https://github.com/sinelaw/fresh/pull/2949

PR #2507, opened 2026-06-27
fix(vi): make i"/a" quote text-objects search forward on the line (#2439)

### LINKED ISSUE #2439: vi mode: quote text-objects `i"`/`a"` don't search forward on the line (only work when cursor is already inside the quotes)
Found while testing the new vi "Vim compatibility options" / "align vi compatibility motions with Vim" work on master (v0.4.1 @ `8ee2baf31`). Reference: Vim (`:help v_iquote` / `:help a"`).

## Steps to reproduce
1. Open a file containing a quoted string on a line, e.g. `the "quick" brown fox` (`fresh --no-restore file.txt`).
2. Enable vi mode: `Ctrl+P` → "Toggle Vi mode" → Enter.
3. Put the cursor at the **start** of that line (`0`), i.e. **before** the opening quote, in NORMAL mode.
4. Press `d` `i` `"` (`di"`).

## Expected behavior
As in Vim: when the cursor is on a line that contains a quoted string but is *before* the quotes, `i"`/`a"` search forward on the **current line** and operate on that quoted string. So `di"` from column 1 of `the "quick" brown fox` yields `the "" brown fox`, and `ci"` does the same and enters INSERT mode.

## Actual behavior
`di"`/`ci"` are silent no-ops when the cursor is before the quote — the line is unchanged (reproduced 2/2 from column 1).

The operation itself works **only** when the cursor is already *inside* the quotes:
- Cursor on `q` of `"quick"` → `di"` → `the "" brown fox` ✓
- Cursor on `q` → `ci"` → enters INSERT, `the "WXYZ" brown fox` ✓

So the text object is implemented but is missing Vim's "search forward on the line for the quoted string" rule, which makes `ci"`/`di"` from the start of a line (the common case) fail. The `"` key transmits fine (verified by typing it in INSERT mode), so this is not an input issue.

## Workaround
Move the cursor inside the quotes first (e.g. `f"` then `l`, or `t"`) before pressing `ci"`/`di"`.

sinelaw:
## Summary

Fixes #2439. In vi mode, the quote text objects `i"` / `a"` (and `'` / `` ` ``) only matched the quote pair the cursor was already **inside**. Vim's actual rule is to *search forward on the current line* when the cursor is before the quotes, so `ci"` / `di"` from the start of a line (the common case, e.g. right after `0`) were silent no-ops.

## Root cause

In `applyTextObject` (`crates/fresh-editor/plugins/vi_mode.ts`), the quote case scanned the line for a pair containing the cursor and bailed out with no selection when the cursor was before the quotes.

A second, related issue surfaced once forward search worked: the `c`/`d` text-object operators relied on the cursor already being inside the deleted range, so after deleting they left the cursor at its original column. With forward search the cursor is *before* the range, so `ci"` from column 0 inserted text at the start of the line instead of between the quotes.

## Fix

- Pick the first complete quote pair whose closing quote is at or after the cursor — this covers both "inside" and "before" the quotes (Vim's forward-search rule), and still resolves to the containing pair when the cursor is inside one.
- After the change/delete text-object operators, pin the cursor to the deletion point (`selectStart`), so `ci"`/`di"` land correctly regardless of where the cursor started on the line.

## Testing

Added two e2e reproductions in `vi_mode_bugs.rs` that drive keys and assert on buffer content, both starting with the cursor at column 0:

- `di"` on `the "quick" brown fox` → `the "" brown fox`
- `ci"` then typing → `the "WXYZ" brown fox`

Both fail (time out / wrong output) without the fix and pass with it.

Validation run in a tmux session:

- New tests: `2 passed; 0 failed`
- Full vi suite (`e2e_tests vi_`): `103 passed; 0 failed; 5 ignored` — no regressions
- `cargo fmt -- --check` clean; `cargo clippy --tests` clean (only pre-existing, unrelated warnings)

https://claude.ai/code/session_01TdMpos2dBVKwAyqzJoc2fg

---
_Generated by [Claude Code](https://claude.ai/code/session_01TdMpos2dBVKwAyqzJoc2fg)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,7 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
 * **Regex search**: `^` / `$` now anchor per line in multi-line `Ctrl+F` search (#2495).
 * **Splits**: closing a buffer shown in two splits no longer desyncs the surviving cursor (#2496).
 * **Terminal**: a terminal restores its live/scrollback mode on refocus, and the last split is no longer left read-only after closing a second terminal (#2485).
+* **Vi mode**: `cw` on a non-blank now changes up to the end of the word like Vim (matching `ce`) instead of swallowing the trailing whitespace (#2437).
 * **Per-language indentation rules** are now actually applied — previously custom patterns were ignored (#2314).
 * **Brackets** are no longer highlighted inside comments and strings (#2405, reported by @TakemiSora).
 * **Quick-open `#` buffer switcher** now lists virtual buffers (#2373).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
 
 ### Bug Fixes
 
+* **Vi mode**: quote text objects `i"`/`a"` (and `'`/`` ` ``) now search forward on the line, so `ci"`/`di"` work from before the quotes — e.g. from the start of the line — instead of only when the cursor is already inside them (#2439).
 * **Cursor column** is preserved on vertical moves that used to drift it - indented soft-wrapped lines, off-screen up/down over short lines (#2565), and blank lines with indentation guides (#2564).
 * **LSP args** - an empty `args` list now overrides a built-in server's defaults, so you can use one that takes no arguments (e.g. markdown-oxide for marksman) (#2549, reported by @alex-ball).
 * **Settings icons** now render on all terminals by default - standard Unicode symbols replace the Nerd Font glyphs that showed as `?`; opt back in with `editor.nerd_font_icons` (#2032).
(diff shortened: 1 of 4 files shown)


sinelaw:
Manually validated in tmux: fresh (debug build) side-by-side with Vim 9.1 (`vim --clean`), identical keystrokes sent to both panes, buffers diffed byte-for-byte after `:w`.

- `di"` from col 1 on `the "quick" brown fox` → `the "" brown fox` — matches Vim (forward search works)
- `ci"` + `WXYZ` from col 1 on `say "hello world" now` → `say "WXYZ" now` — matches (cursor correctly pinned to the deletion point, insert lands between the quotes)
- `di"` with the cursor *inside* the quotes (`fp` then `di"` on `mid "point" here`) → `mid "" here` — containing pair still resolves

Out-of-scope observation (pre-existing, `a"` range code is untouched by this PR): `da"` from before the quotes on `one "two" three` gives `one  three` (two spaces) in fresh vs `one three` in Vim — Vim's `a"` also consumes the trailing whitespace after the closing quote.

---

PR #2510, opened 2026-06-28
fix(vi): support find-char motions (f/t/F/T) with operators (#2441)

### LINKED ISSUE #2441: vi mode: find-char motions don't work with operators (`df`/`dt`/`cf`/`ct` hang or mis-delete) and `;`/`,` repeat is a no-op
Found while continuing the vi "Vim compatibility motions" sweep on master **v0.4.1** (built from `origin/master` @ `3b8c2eca1`). Reference: Vim (`:help f`, `:help d`, `:help ;`).

The `f`/`t` find-char motions work correctly as **pure cursor motions**, but are broken in two ways: they are not recognized when an **operator** is pending, and the **repeat** commands `;`/`,` do nothing.

## Steps to reproduce
1. Create a file with the line `hello world foo bar baz` and open it: `fresh --no-restore motions.txt`.
2. Enable vi mode: `Ctrl+P` → "Toggle Vi mode" → Enter (status shows `-- NORMAL --`).
3. Put the cursor at column 1 (`0`).

**A — operator + find-char hangs:**
4. Press `d`, then `f`, then `r` (i.e. `dfr` — "delete up to and including the first `r`").

**B — operator + find-char silently mis-deletes (when the target char is itself a motion key):**
5. From a fresh `hello world foo bar baz` at column 1, press `d`, then `f`, then `w` (`dfw`).

**C — `;` repeat is a no-op:**
6. From `hello world foo bar baz` at column 1, press `f` then `o` (cursor lands on the `o` of `hello`, col 5 — correct). Then press `;` (repeat find).

## Expected behavior (Vim)
- **A:** `dfr` deletes from the cursor up to and **including** the first `r` (`hello wor`), leaving `ld foo bar baz`.
- **B:** `dfw` deletes up to and including the first `w` (`hello w`), leaving `orld foo bar baz`.
- **C:** `;` repeats the last `f`/`t`, advancing the cursor to the **next** `o` (col 8, the `o` of `world`).

## Actual behavior
- **A:** Nothing is deleted and the editor is **stuck in operator-pending** — the status bar shows `-- OPERATOR (d) --` indefinitely. Subsequent keys (including the find target `r`, and even unrelated keys like `Z`) are swallowed; the only way out is `Esc`. Reproduced with 0.4–0.6s and with deliberate 1.2s gaps between keys (so it is not a key-timing race). Same hang with `cf`/`ct` (`-- OPERATOR (c) --`) and with `t` (`dt,` on `find the comma, and stop here` hangs identically).
- **B:** `dfw` produces `world foo bar baz` — i.e. it deletes exactly `hello ` (a plain `dw`), **not** `hello w`. The `f` is dropped and the following `w` is consumed as a standalone word-motion, so `dfw` silently degrades to `dw` and deletes the wrong span with no error. (Targets that are not motion keys, like `r`/`o`, instead trigger the hang in case A.)
- **C:** `;` does nothing — the cursor stays on col 5. Verified by pressing `x` afterward: it deletes the `o` at col 5 (`hell world foo bar baz`), proving the cursor never advanced. `,` (reverse repeat) is likewise a no-op.

The find-char keys transmit fine and the feature is half-implemented: **pure** `f`/`t` cursor motions land correctly (e.g. `fr` then `x` deletes the `r` of `world`; `t,` then `x` deletes the char before the comma). Only the operator-pending and repeat paths are broken.

## Workaround
For deletion/change up to a character, fall back to a non-find motion (`dw`, `de`, visual select + `d`) or use search (`/char` then `d` ... ), and re-issue `f`/`t` manually instead of `;`/`,`.

---
Search queries used (no existing issue): `vi mode find char operator`, `vi df dt delete find`, `vi df dt`. 

<sub>Filed by the automated black-box TUI QA agent (Run #44).</sub>

sinelaw:
Fixes #2441.

## Problem

In vi mode the find-char motions (`f`/`t`/`F`/`T`) worked as pure cursor motions but were **unbound in operator-pending mode**. The reported symptoms:

- **A — hang:** `dfr` (and `cf`/`ct`/`dt`, backward `dF`/`dT`) left the editor stuck in `-- OPERATOR (d) --` indefinitely, swallowing all keys until `Esc`. The `f` key did nothing while pending, and the target char was not a motion, so nothing resolved the operator.
- **B — silent mis-delete:** when the find target was itself a motion key, e.g. `dfw`, the dropped `f` left `w` to run as a plain word motion, so `dfw` degraded to `dw` and deleted the wrong span.

(The issue also listed `;`/`,` repeat as a no-op — that part was already fixed on `master` since v0.4.1; this PR adds regression coverage for it.)

## Fix

Bind `f`/`t`/`F`/`T` in operator-pending mode to a new operator-aware find-char path that, after capturing the target character, converts the cursor→target span into a byte range and applies the pending operator via the existing `applyOperatorWithRange`:

- Find-char operators are **inclusive** motions in vim: forward `f`/`t` include the landing character; backward `F`/`T` delete from the landing column up to (but not including) the cursor.
- A **missing target** on the line cancels the operator and leaves the buffer unchanged, instead of being consumed as a different motion.
- Wired into `.` repeat (new `operator-find-char` change type) and into `;`/`,` repeat, with a count so `d2fx` targets the Nth occurrence.
- The line-context lookup and target search are factored into shared helpers reused by the existing pure-motion path, so cursor-only `f`/`t`/`F`/`T` behavior is unchanged.

## Testing

- New e2e reproductions in `vi_mode_bugs.rs` — `dfr`, `dfw`, `dtr`, `cfr` — which **hang without this change** and pass with it; plus `;`/`,` regression tests.
- Full `vi_mode` e2e suite passes (107 passed, 0 failed).
- `cargo fmt` clean; the TypeScript plugin introduces no new `tsc` errors (the 4 pre-existing `ActionSpec` errors are unrelated).
- Manually validated in a real terminal (tmux): `dfr` → `ld foo bar baz`, `dtr` → `rld foo bar baz`, and `cfr` then typing `XYZ` → `XYZld foo bar baz`, each returning cleanly to NORMAL/INSERT instead of hanging.

https://claude.ai/code/session_01P4ofi1eeqzb96bkpgc9tri

---
_Generated by [Claude Code](https://claude.ai/code/session_01P4ofi1eeqzb96bkpgc9tri)_

### DIFF
--- a/crates/fresh-editor/plugins/vi_mode.ts
+++ b/crates/fresh-editor/plugins/vi_mode.ts
@@ -28,7 +28,7 @@ type WordSearchDirection = "forward" | "backward";
 type WordSearchTarget = { start: number; end: number; text: string; wholeWord: boolean };
 
 // Types for tracking repeatable changes
-type ChangeType = "simple" | "operator-motion" | "operator-textobj" | "insert" | "line-op";
+type ChangeType = "simple" | "operator-motion" | "operator-textobj" | "operator-find-char" | "insert" | "line-op";
 
 interface LastChange {
   type: ChangeType;
@@ -36,6 +36,8 @@ interface LastChange {
   operator?: string;         // For operator+motion/textobj: "d", "c", "y"
   motion?: string;           // For operator+motion: the motion action
   textObject?: { modifier: TextObjectType; object: string }; // For operator+textobj
+  findType?: FindCharType;   // For operator+find-char (df/dt/cf/ct): the f/t/F/T variant
+  findCharTarget?: string;   // For operator+find-char: the target character
   count?: number;            // Count used with the command
   insertedText?: string;     // Text inserted during insert mode
 }
@@ -202,7 +204,8 @@ async function captureInsertedText(): Promise<void> {
         insertedText: text,
       };
     } else if (state.lastChange.type === "simple" || state.lastChange.type === "operator-motion" ||
-               state.lastChange.type === "operator-textobj" || state.lastChange.type === "line-op") {
+               state.lastChange.type === "operator-textobj" || state.lastChange.type === "operator-find-char" ||
+               state.lastChange.type === "line-op") {
       // A change command (c, s, etc.) was used - append the inserted text
       state.lastChange.insertedText = text;
     }
@@ -2044,6 +2047,22 @@ async function vi_repeat() : Promise<void> {
       break;
     }
 
+    case "operator-find-char": {
+      // Operator + find-char like dfx, dtx, cfx
+      if (change.operator && change.findType && change.findCharTarget) {
+        if (change.operator === "c") {
+          // Replay change as delete-then-insert so '.' doesn't block on input.
+          await executeFindCharOperator("d", change.findType, change.findCharTarget, count);
+          if (change.insertedText) {
+            editor.insertAtCursor(change.insertedText);
+          }
+        } else {
+          await executeFindCharOperator(change.operator, change.findType, change.findCharTarget, count);
+        }
+      }
+      break;
+    }
+
     case "insert": {
       // Pure insert (i, a, o, O)
       if (change.insertedText) {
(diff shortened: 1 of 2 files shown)


sinelaw:
Manually validated in tmux: fresh (debug build) side-by-side with Vim 9.1 (`vim --clean`), identical keystrokes sent to both panes (each line of `hello world foo bar baz`), buffers diffed byte-for-byte after `:w`.

All byte-identical with Vim, and no operator-pending hangs:
- `dfr` → `ld foo bar baz`
- `dtr` → `rld foo bar baz`
- `cfr` + `XYZ` → `XYZld foo bar baz` (enters insert correctly)
- `dfw` (target char that is also a motion key) → `orld foo bar baz` — no longer degrades to `dw`
- `$` then `dFb` (backward find with operator) → `orld foo bar z`
- `dfq` (target missing on line) → no-op, buffer unchanged, returns to NORMAL instead of hanging; a subsequent `d2fo` (count) → `rld foo bar baz`

---

PR #2512, opened 2026-06-28
fix(vi): clamp j/k onto a shorter line to its last character (#2442)

### LINKED ISSUE #2442: vi mode: `j`/`k` onto a shorter line park the cursor one past the last char, so `x` deletes the newline and joins lines
Found while continuing the vi "Vim compatibility motions" sweep on master **v0.4.1** (built from `origin/master` @ `3b8c2eca1`). Reference: Vim (cursor cannot rest past the last character in NORMAL mode; vertical motion clamps the cursor to the last char of a shorter line and remembers the desired column).

When you move down/up (`j`/`k`) from a long line onto a **shorter** line whose length is less than the current/desired column, Fresh's vi mode parks the cursor **one column past the last character** (the insert position), instead of clamping it to the last character. Because the cursor is then sitting on the line-ending boundary, pressing `x` deletes the **newline** and joins the following line — silent corruption from an ordinary navigate-then-delete.

## Steps to reproduce
1. Create a file whose first lines are a long line followed by a short line, e.g.:
   ```
   hello world foo bar baz
   short
   a longer line with many words here
   ```
   Open it: `fresh --no-restore motions.txt`.
2. Enable vi mode: `Ctrl+P` → "Toggle Vi mode" → Enter.
3. On line 1, press `$` (cursor to the last char `z`, col 23 — correct).
4. Press `j` to move down onto line 2 (`short`, 5 chars).
5. Press `x`.

(Also reproduces without `$`: from line 1 col 10 — e.g. nine `l` presses — then `j` onto `short`; and upward with `k` from line 3 onto `short`.)

## Expected behavior (Vim)
After `$` then `j`, the cursor lands on the **last char** of `short` (the `t`, col 5) — Vim clamps a too-large desired column to the last character and never lets the cursor rest past it in NORMAL mode. `x` then deletes the `t`, giving `shor`. `x` in NORMAL can never delete a newline / join lines.

## Actual behavior
After `$`/`j` (or any `j`/`k` onto the shorter line), the status bar shows **`Ln 2, Col 6`** — i.e. column `len+1` on a 5-character line, past the last char. Pressing `x` then deletes the line-ending newline and **joins line 3 onto line 2**:
```
short  +  a longer line with many words here   →   shorta longer line with many words here
```
Reproduced 3/3: via `$`+`j`, via `l`-to-col-10 + `j`, and via `k` from line 3 — all land on `Col 6` and all join lines on `x`.

This is specific to **vertical** motion. Horizontal clamping is correct: pressing `$` directly on `short` lands on `Col 5` (the `t`) and `x` gives `shor` (no join); `l` also stops at the last char. So the column-clamp logic for `j`/`k` is off by one relative to `$`/`l`.

## Workaround
After moving vertically onto a shorter line, press `$` (or `0`) to re-anchor the cursor onto a real character before using `x`; or use `dl`/`X` carefully. To merge lines deliberately, use `J`.

---
Search queries used (no existing issue): `vi cursor past end of line j k shorter`, `vi mode cursor end of line clamp`, `vi mode cursor end of line`.

<sub>Filed by the automated black-box TUI QA agent (Run #44).</sub>

sinelaw:
Fixes #2442.

## Problem

In vi mode, moving with `j` / `k` from a long line onto a **shorter** line parked the cursor **one column past the last character** (the insert position) instead of clamping it to the last character like Vim. Because the cursor then sat on the line-ending boundary, pressing `x` deleted the **newline** and joined the following line — silent corruption from an ordinary navigate-then-delete.

```
$ on line 1 → j onto "short" → status showed "Col 6" on a 5-char line
x → deleted the newline, joining the next line onto "short"
```

Horizontal clamping (`l`, `$`) was already correct (they use the vi-aware `move_*_in_line` actions); only the **vertical** motions went through the generic `move_down`/`move_up`, which legitimately land at end-of-line for a normal editor but not for vi NORMAL mode.

## Fix

Add vi-aware native vertical motions `ViMoveUp` / `ViMoveDown` (alongside the existing `ViMoveWordEnd`). They reuse the same vertical-motion logic — so the **goal column is still remembered** for the next vertical move (a long→short→long traversal returns to the original column, as in Vim) — but clamp the final position to the destination line's last character via a small `clamp_to_last_char_on_line` helper. The `vi_mode` plugin routes normal-mode `j`/`k` (and the Down/Up arrows) to these actions.

`MoveUp` / `MoveDown` / `SelectUp` / `SelectDown` are byte-for-byte unchanged (they pass `clamp = false`), so non-vi navigation and selection are unaffected.

## Why not clamp in the plugin

A plugin-side corrective move (e.g. `move_left` after `move_down`) would work for the join bug but resets `sticky_column` to 0, wiping the remembered goal column and regressing long→short→long navigation. Clamping inside the native vertical motion keeps the goal column intact, so the fix is done there.

## Tests

- New e2e reproductions in `vi_mode_bugs.rs` (drive real keystrokes, assert on buffer content / cursor position; fail/time-out without the fix):
  - `j` onto a shorter line then `x` → deletes a char, **does not** join lines
  - same for upward `k`
  - goal column survives a long→short→long traversal
- Updated `test_vi_vim_compat_search_count_does_not_leak_to_next_motion`, which had recorded the buggy off-by-one column (`Col 4` on a 3-char line) — now asserts the Vim-correct `Col 3`. Its actual intent (the search count not leaking into the next `j`) is unchanged.

All 109 vi-mode e2e tests pass; `cargo fmt`, `cargo clippy`, and `cargo check --all-targets` are clean. Manually validated in a tmux session: `$` → `j` onto `short` lands on `Col 5` and `x` yields `shor` without joining the line below.

https://claude.ai/code/session_01UCqVXtGtd7VyaLFFFVq9tC

---
_Generated by [Claude Code](https://claude.ai/code/session_01UCqVXtGtd7VyaLFFFVq9tC)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
 
 ### Bug Fixes
 
+* **Vi mode**: moving with `j` / `k` onto a shorter line now clamps the cursor to that line's last character (like Vim) instead of parking it one past the end, so a following `x` deletes a character rather than the newline and joins lines. The desired column is still remembered for the next vertical move (#2442).
 * **Regex search**: `^` / `$` now anchor per line in multi-line `Ctrl+F` search (#2495).
 * **Splits**: closing a buffer shown in two splits no longer desyncs the surviving cursor (#2496).
 * **Terminal**: a terminal restores its live/scrollback mode on refocus, and the last split is no longer left read-only after closing a second terminal (#2485).
--- a/crates/fresh-editor/plugins/vi_mode.ts
+++ b/crates/fresh-editor/plugins/vi_mode.ts
@@ -1081,12 +1081,15 @@ function vi_left() : void {
 registerHandler("vi_left", vi_left);
 
 function vi_down() : void {
-  executeWithCount("move_down");
+  // vi_move_down clamps the caret to the destination line's last char (Vim
+  // never lets the cursor rest past it in NORMAL mode), while still
+  // remembering the goal column for the next vertical move.
+  executeWithCount("vi_move_down");
 }
 registerHandler("vi_down", vi_down);
 
 function vi_up() : void {
-  executeWithCount("move_up");
+  executeWithCount("vi_move_up");
 }
 registerHandler("vi_up", vi_up);
 
--- a/crates/fresh-editor/src/app/composite_buffer_actions.rs
+++ b/crates/fresh-editor/src/app/composite_buffer_actions.rs
@@ -854,10 +854,10 @@ impl Editor {
             }
 
             // Cursor movement (without selection)
-            Action::MoveDown => {
+            Action::MoveDown | Action::ViMoveDown => {
                 self.handle_cursor_movement_action(split_id, buffer_id, CursorMovement::Down, false)
             }
-            Action::MoveUp => {
+            Action::MoveUp | Action::ViMoveUp => {
                 self.handle_cursor_movement_action(split_id, buffer_id, CursorMovement::Up, false)
             }
             Action::MoveLeft => {
--- a/crates/fresh-editor/src/input/actions.rs
+++ b/crates/fresh-editor/src/input/actions.rs
@@ -56,6 +56,23 @@ fn content_len_without_line_ending(content: &str) -> usize {
     content.trim_end_matches(LINE_ENDING_CHARS).len()
 }
 
+/// In vi NORMAL mode the caret may not rest past a line's last character.
+///
+/// Given a destination line (its start offset and full content, line ending
+/// included) and a tentative byte position on that line, clamp the position to
+/// the start of the line's last character — or to the line start for an empty
+/// line. Used by the vi-aware vertical motions so `j`/`k` onto a shorter line
+/// land on the last char (like Vim) instead of one past it.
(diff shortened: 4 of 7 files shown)


sinelaw:
Manually validated in tmux: fresh (debug build) side-by-side with Vim 9.1 (`vim --clean`), identical keystrokes sent to both panes, cursor positions compared via status bar/ruler, buffers diffed byte-for-byte after `:w`. Test file: `a much longer first line` / `short` / `another long line here ok`.

- `$` then `j` onto `short`: cursor lands at **2,5** in both editors (last char, no longer past EOL); `x` → `shor`, **no line join** — matches Vim
- `$` then `k` upward onto `short`: same, matches
- Goal column: `10l` (col 11) → `j` clamps to col 5 on `short` → `j` returns to col 11 on the next long line, in both editors; `x` deletes the same char — matches Vim
- Bug confirmed on the base commit (`f7912fd`): same keystrokes there put fresh at Col 6 on the 5-char line and `x` produced `shortanother long line here ok` (joined lines)

One adjacent gap, verified **pre-existing** on the base commit (unchanged by this PR): Vim treats `$` as a sticky "end of line" goal (`$jj` lands on the *last char* of each line, col 25 on line 3), while fresh remembers the numeric column (col 24). Only affects the `$`-then-vertical-motion case; plain column clamping and restore match Vim.

---

Issue #2434, opened 2026-06-22
[Enhancement] Next match key for Project Search and Replace


mandolyte:
In my use case, I am looking for all occurrences of a string. I have to review each match and its context to determine whether to make a change. At this moment, I have almost 150 matches. To review each one, I have to hit the down arrow in the S&R panel, press return, view the match, optionally modify it. Then use `Alt+]` to return the S&R panel, and repeat. 

It would be nice to have a key that would go to the next match for me, saving three keystrokes (which after enough iterations become error prone). 

Another thought, since I cannot use the replacement feature, perhaps a simpler "Project Search" would also be nice to have.

Thanks!

sinelaw:
Closing as completed: implemented by PR #2516 (merged in 0.4.2) — "Next Search Match" / "Previous Search Match" commands step through Project Search &amp; Replace results with a single keystroke from anywhere. Follow-up fixes also landed: panel selection/click-to-open (1a6c93910) and match positions anchored to buffer edits (PR #2654).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2516, opened 2026-06-28
Add next/previous match navigation to Project Search & Replace (#2434)

sinelaw:
## Summary

Implements #2434 (requested by @mandolyte). Reviewing a large Project Search & Replace result set match-by-match used to cost **three** keystrokes per match — `Down` to select the next row in the panel, `Enter` to open it (which moves focus to the source split), then `Alt+]` to get back to the panel before repeating. This adds a **single** keystroke that does the whole step, from wherever focus happens to be.

Two new commands — **Next Search Match** / **Previous Search Match** — advance the panel's selection to the next/previous match row (skipping file headers, wrapping around), open it in the source split so the cursor lands right on the match ready to edit, and report `Match n/total` in the status bar. Because they act on the last search's results and `openFileInSplit` keeps focus in the editor, you can edit a match and press the key again to move on **without ever returning to the panel**. The pattern mirrors the next/prev idiom already used by `diff_nav.ts`.

## Keybindings

Bound to `Ctrl+Alt+→` / `Ctrl+Alt+←` in the default keymap (normal/editor context) **and** in the Search & Replace panel mode, so the same key works whether focus is in the panel or back in the source buffer after an edit. The commands are also available from the palette (`Ctrl+P`).

`Alt+n`/`Alt+p` and `F3` were already taken; `Ctrl+Alt+Arrow` was chosen because arrow keys avoid the `Ctrl+[` (≡ ESC) and shifted-punctuation transmission pitfalls that the existing keymap comments warn about.

## Changes

- **`plugins/search_replace.ts`** — navigation helpers (`adjacentMatchIndex`, `jumpToMatch`, `matchOrdinal`, `navigateMatch`), the two action handlers, command registration, and panel-mode bindings.
- **`keymaps/default.json`** — `Ctrl+Alt+Right`/`Left` → next/prev match (normal context).
- **`input/keybindings.rs`** — allow the two plugin actions in built-in keymaps (`ALLOWED_PLUGIN_ACTIONS_IN_DEFAULTS`).
- **`plugins/search_replace.i18n.json`** — new command + status strings across all 14 locales.
- **`tests/e2e/search_replace.rs`** — e2e test driving both the panel-mode binding and the normal-context keymap binding.
- **`CHANGELOG.md`** — Features entry.

## Testing

- New e2e test `test_search_replace_next_match_navigation` passes; it presses the key from the panel (mode binding) and then from the editor (default keymap binding), asserting the ordinal moves `1 → 2 → 1` on rendered output.
- Full `search_replace` e2e suite green (35 passed), plugin i18n tests green (16 + 3), keymap-validation unit test green.
- `cargo fmt --check`, `cargo clippy -p fresh-editor`, and plugin type-check all clean (no new findings).
- Manually validated in a real terminal (tmux): opened the panel, searched, and stepped forward/back through all matches with `Ctrl+Alt+→`/`←` — the editor jumped to each match's line, wraparound worked both directions, and the palette showed both commands with their bindings.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VJ79Mtp4DJXbs6iE8PJZp2)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
 * **JSONC config files** — config files now accept comments and trailing commas, a stray syntax slip no longer silently resets you to defaults, and saves never clobber an unparseable file (the error is surfaced instead) (#2497).
 * **Per-buffer view toggles** — *Toggle Line Numbers (Current Buffer)* and *Toggle Line Wrap (Current Buffer)* scope to the active buffer and persist across restarts.
 * **LSP Go to Implementation** (`textDocument/implementation`) (#2384, by @Crocmagnon).
+* **Step through Search & Replace matches from the editor** — `Ctrl+Alt+→` / `Ctrl+Alt+←` (also available as palette commands and inside the panel) jump to the next/previous match, opening each in the source split so you can review or edit it without the Down → Enter → `Alt+]` round-trip back to the panel (#2434, requested by @mandolyte).
 
 ### Improvements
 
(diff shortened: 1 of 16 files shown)


sinelaw:
Requires fixes before merging

PR #2534, opened 2026-06-30
Enable rquickjs bindgen feature on NetBSD

ci4ic4:
## What

Add a `target_os = "netbsd"` dependency block that enables the `bindgen`
feature for `rquickjs` in the two crates that depend on it
(`fresh-core` and `fresh-plugin-runtime`), mirroring the existing
FreeBSD workaround.

## Why

`rquickjs-sys` ships pre-generated C bindings only for a handful of
platforms; NetBSD is not among them. Without the `bindgen` feature the
build fails on NetBSD because no bindings are available. Enabling
`bindgen` makes `rquickjs-sys` generate them at compile time, exactly as
already done for FreeBSD.

This lets NetBSD build `fresh` from the upstream source unmodified. It
is currently carried as a local patch in pkgsrc-wip; upstreaming it
removes the need to maintain that patch.

## Scope

Build configuration only — no behavior change on existing platforms,
since the new block is gated on `cfg(target_os = "netbsd")`.

### DIFF
--- a/crates/fresh-core/Cargo.toml
+++ b/crates/fresh-core/Cargo.toml
@@ -23,6 +23,10 @@ rquickjs-serde = { workspace = true, optional = true }
 [target.'cfg(target_os = "freebsd")'.dependencies]
 rquickjs = { workspace = true, optional = true, features = ["bindgen"] }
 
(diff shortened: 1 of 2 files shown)


sinelaw:
Thanks!

Issue #2032, opened 2026-05-18
Settings editor renders Unicode glyphs as `?` on some macOS terminals


sinelaw:
On macOS terminals whose primary font lacks the relevant glyphs (and fallback isn't kicking in), the Settings editor shows literal `?` characters instead of its decorative symbols.

Glyphs used by the settings UI (from `crates/fresh-editor/src/view/settings/render.rs`):

| Char | Where | Line |
|---|---|---|
| `▶`, `▼` | category expand/collapse chevrons | 522, 524 |
| `→` | inline arrow in entries | 2009 |
| `↑`, `↓`, `↑↓` | scroll indicators | 2622–2624 |
| `←/→/Tab` | help hint text | 3019, 3136 |
| `←→` | category-switch hint | 440 |
| `╭ ╮ ╰ ╯ ─ │` | dialog/panel borders | throughout |
| `▎` | left bar indicator | dashboard / list rows |
| `✓` | active-toggle marker | various |

These are all standard Unicode, so the fix isn't "drop them" — but right now there's no ASCII fallback when a font can't render them. Two reasonable directions:

1. Add a runtime fallback set (`>` / `v` for chevrons, `^v` for scroll indicators, `+--+ | -` for borders, `[x]` for the checkmark, …) gated on an `editor.ascii_ui` config flag, or auto-detected.
2. Document which macOS terminals/fonts are known good (Menlo, SF Mono, etc.) so users can switch font instead.

Same glyphs are also used in the rest of the UI (file explorer, popups, prompts), so worth solving once and applying across the board.

sinelaw:
automatic comment: not straightforward — the fix spans the whole UI (30+ glyph sites in `settings/render.rs` alone plus file explorer/popups/prompts), needs a design decision (config flag like `editor.ascii_ui` vs runtime auto-detection, which isn't feasible from a TUI) and plumbing of the flag through many render paths, and the macOS-terminal-font-specific failure mode can't be reliably reproduced or validated on a Linux dev environment.

---
_Generated by [Claude Code](https://claude.ai/code)_

comesuccingfuccsloot:
i believe this is (at least partly?) affecting me too, and i'm running fresh inside wsl2 debian running in windows terminal, not mac.
just installed v0.4.2 from .deb

<img width="182" height="318" alt="Image" src="https://github.com/user-attachments/assets/2fef383f-c926-426a-8332-ca4c5c857938" />

[...]

PR #2551, opened 2026-07-03
Add editor.ascii_ui fallback for chrome glyphs (#2032)

sinelaw:
On terminals/fonts that lack the decorative Unicode and Nerd-Font glyphs
the UI chrome uses, the settings editor, file tree, popups, and dialog
borders render symbols as literal `?` or blank boxes. The worst offender
is the settings category icons, which use Nerd-Font private-use-area
codepoints and break on any non-patched font.

Introduce a central chrome glyph set (view/glyphs.rs) with a Unicode set
(current appearance, unchanged) and an ASCII fallback set, selected per
frame from a new `editor.ascii_ui` config flag. Route the chrome glyph
sites through it: settings editor (chevrons, category icons, arrows,
bullets, separators, rounded/dialog borders, hints), file tree status
dot, file browser (sort arrow, separators, popup border), the
dropdown/dual-list/map-input controls, and the popup/keybinding-editor/
calibration-wizard/workspace-trust dialog borders.

Buffer text, whitespace indicators, indentation guides, and the
already-configurable file-explorer tree indicators are left untouched —
they have their own dedicated config. Regenerate config-schema.json so
the flag appears in the Settings UI, and document the option in
troubleshooting.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01QtFiKHF9rWPgFQUcPwHrkv

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
 * **Settings**:
   * language-entry field edits keep their committed value (including **Tab Size**), and `Esc` now cancels an in-progress edit while `Enter`/`Tab` commit it (#2537);
   * committing a text field with `Tab` persists the typed value on Save, matching `Enter` — previously the row showed as modified but Save wrote an empty value (#2515).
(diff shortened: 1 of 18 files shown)


sinelaw:
different solution implemented to just disable nerd fonts by default

PR #2576, opened 2026-07-06
Virtual space: cursor beyond end-of-line (off/block/on)

sinelaw:
Implements **virtual space** — the cursor can move beyond the end of a line, like Visual Studio, JetBrains IDEs, and Vim's `virtualedit` (and VSCode's long-stalled [microsoft/vscode#228680](https://github.com/microsoft/vscode/pull/228680)). Scoping analysis included at `docs/internal/virtual-space-scoping.md`.

## The setting

`editor.virtual_space` (Editing section of the Settings UI, default `off`):

- **`off`** — traditional behavior, cursor clamps to line content (unchanged).
- **`block`** — only block (rectangular) selections extend past line ends: rectangles keep their column over short lines, paint their full extent, copy space-padded true rectangles, and block-typing pads short lines so characters land in a straight column.
- **`on`** — everything in `block`, plus: Right at EOL walks into empty space instead of wrapping; Up/Down hold their column over short lines; clicks past EOL place the cursor at the clicked column; **the void below the last line is reachable too** — click anywhere down there, or press ArrowDown at the bottom, and the cursor floats on a virtual line (ArrowUp steps back through them); typing anywhere in the void materializes the missing newlines + column padding. The cursor renders at its virtual position (secondary cursors and inactive splits get padded indicator cells; horizontal scroll follows). Typing/Tab/paste materialize the gap as a **single undo step**; Backspace steps back through virtual columns without deleting; Enter inserts at the real content end, so no phantom trailing whitespace ever appears from movement alone.

A command-palette command — **Toggle Virtual Space (Current Buffer)** — switches virtual space off ↔ on for the active buffer only (localized in all 14 locales); the per-buffer override survives global settings changes and is persisted in the workspace session, so a restored workspace restores it.

## Design

Cursors stay byte offsets and the buffer is never mutated by movement. Instead of adding a `virtual_column` field to `Cursor` (which would touch all 210 `MoveCursor` construction sites), the implementation *derives* the horizontal virtual position from the existing `sticky_column` (the goal column): a cursor is virtual iff it is collapsed, sits exactly at its line's content end, and its sticky column exceeds the line's visual width. `model::virtual_space` is the single source of truth, with the mode gate in the derivation's signature so a disabled config can never leak a virtual position. Byte positions handed to LSP, plugins, and selections are always the clipped position.

The vertical dimension has nothing to derive from, so `Cursor` gains a transient `virtual_lines_below` field — meaningful only while the cursor sits at the buffer end, cleared by any move to a different position or any edit, and deliberately not carried by `MoveCursor` events (the same category of direct-mutation view state as `block_anchor`). Vertical *movement* into the void queues the count in `EditorState::pending_virtual_lines`, consumed when the move event applies (handlers emit events; events can't carry the count).

The sticky column already flows through every `MoveCursor` event, undo, and session persistence — but making it trustworthy surfaced real pre-existing bugs, fixed first with reproducers:

- **PageUp/PageDown stored byte columns as the goal**, and could land the cursor *mid-character* on wide-char (CJK) lines — corrupting the buffer on the next insert.
- **Mouse click/drag stored byte columns**, misplacing the next vertical move after wide chars.
- **Bracket expansion stored a byte *position*** as the goal column, sending the next vertical move far right.
- **Edits didn't reset the goal column** (and the bulk-edit path silently dropped every event's sticky update), so Up/Down after typing jumped to a stale column. Edits now reset it, matching mainstream editors.

## Scoped-out limits (documented in the scoping doc)

- Linear (Shift+arrow) selections stay byte-clamped; extending a selection from a virtual cursor collapses to the content end.
- Soft-wrapped lines don't take virtual columns past the wrap point.
- Block-selection geometry remains byte-column based, exact for the spaces padding materializes.
- Vertical scroll doesn't follow virtual lines past the viewport bottom (the floating cursor clamps to the last visible row until real lines exist).

## Testing

- E2E suite `tests/e2e/virtual_space.rs` (26 tests) drives keys/mouse and asserts rendered cursor position, screen cells, buffer content, and clipboard — covering `on`, `block`, and `off` counterparts for each behavior, the per-buffer toggle's scoping, and its workspace save/restore round-trip.
- Reproducers for the goal-column bugs in `tests/e2e/sticky_column_units.rs` and unit tests in `input/actions.rs`, `state.rs`, and `model/virtual_space.rs` (verified to fail without the fixes).
- **Manually validated in tmux against the real binary** (real key and SGR mouse sequences). This caught two bugs the fixtures had masked — a below-the-last-line click at a column within the last line's width landed mid-line instead of on a virtual line, and the floating cursor didn't render for end-of-buffer cursors resolved by the render fallback path — both fixed with regression tests.
- Fixed a pre-existing palette e2e test (`lsp_no_config`) whose fuzzy query ("ToggleLspBuffer") now matched the new enabled toggle command instead of the disabled LSP one; the query now uniquely matches the intended command.
- `cargo check --all-targets` (default and `--features gui`), fmt, clippy clean. Full suite left to CI per request.

https://claude.ai/code/session_01FhDoYywCqXyYNcDp7MRBiw


regs01:
> * **`off`** — traditional behavior, cursor clamps to line content (unchanged).
> 
> * **`block`** — only block (rectangular) selections extend past line ends: rectangles keep their column over short lines, paint their full extent, copy space-padded true rectangles, and block-typing pads short lines so characters land in a straight column.
> 
> * **`on`** — everything in `block`, plus: Right at EOL walks into empty space instead of wrapping;

Needs an opposing option - no block selection, but go beyond EOL. Similar to Delphi/Lazarus.

PR #2638, opened 2026-07-11
feat(theme): text attributes (bold/italic/…) for syntax highlighting

### LINKED ISSUE #2633: Implementing bold font weight for syntax highlighting
## Proposal: Implementing bold font weight for syntax highlighting
ANSI escape sequences afford the possibility of applying text attributes and font variants to text.
Visual Studio Code consistently applies bold formatting to its syntax highlighting.
Upon comparing the legibility of Rust code between Fresh Editor and VSCode, VSCode demonstrates notably superior readability, primarily due to the bolder character rendering.

Accordingly, we would respectfully suggest that Fresh Editor incorporate a Theme Editor feature that would permit users to selectively configure text attributes such as bold, underline, italic, and similar variants for each individual syntax element type.
This functionality would offer considerable utility and demonstrate meaningful value creation, as it would enable users to customize typographic attributes precisely according to their preferences and requirements.

## Note
* This translation was prepared with the assistance of Claude 4.5 Haiku.
* Should this translation contain any discourteous or offensive language, we sincerely apologize and respectfully request your understanding.

asukaminato0721:
Closes #2633.

Lets a theme render a syntax category with text attributes — `bold`, `italic`, `underlined`, `dim`, `reversed` — not just a foreground color.

## Theme format

A syntax category is either a bare color (as before) or an object bundling that color with the attributes to render it with:

```json
"syntax": {
  "keyword": { "color": [86, 156, 214], "modifier": ["bold"] },
  "string": [206, 145, 120],
  "comment": { "color": [106, 153, 85], "modifier": ["italic"] }
}
```

Existing themes are unaffected: a bare color carries no attributes, so every theme on disk today keeps parsing and rendering identically.

Bundling the pair into one value is deliberate. The alternative — a parallel `keyword_modifier` sibling key — forces the core to hard-code which modifier key belongs to which color key, in a hand-maintained table that has to stay in lock-step with the colors. With the bundle there is no such relation to declare: the color and its attributes travel together.

## Implementation

- **`StyledColorDef`** — untagged enum, `Plain(ColorDef)` or `Styled { color, modifier }`, with `color()` / `modifier()` / `from_parts()` to split and rejoin the pair. `SyntaxColors` fields are now `StyledColorDef`.
- **Runtime `Theme`** keeps its existing `syntax_*: Color` fields (read in ~a dozen places — highlighter, LSP semantic tokens, file explorer, …) alongside `syntax_*_modifier`, populated by splitting the bundle at parse time. No call site of an existing color field changes.
- **The color → modifier lookup is table-driven.** `resolve_modifier_key` is generated from the same `theme_color_keys!` row that declares the color, so adding a syntax category is still a one-line table edit and the two cannot drift apart.
- **Overrides**: the value defines the whole style, so overriding a bold keyword with a bare color clears the bold.
- **Renderer**: `char_style` ORs the category's modifier into the span style, keeping injected plugin tokens authoritative while preserving the attribute when an LSP semantic token refines the color.

## Theme editor

Each syntax color gets an `Attributes` row accepting a comma-separated attribute list. The editor keeps color and attributes flat internally and bundles them at the disk boundary, so the schema stays free of `_modifier` properties.

`theme.schema.json` is regenerated from the Rust types; docs updated.


sinelaw (approved):

- Reviewed against CONTRIBUTING.md: reuses the existing `ModifierDef`/`Modifier` pattern (already used for `selection_modifier`) instead of inventing a new mechanism — no duplication, no architectural break.
- i18n keys added for all locales (guideline 10), unit tests cover parse/round-trip and render application (guideline "Reproduce Before Claiming"), CI green on all platforms.
- Built the branch locally and manually tested in tmux: opened Theme Editor → Syntax Highlighting → `keyword_modifier`, set it to `bold`, confirmed live picker preview showed bold `fn`/`let`, saved the theme, and confirmed the real editor buffer rendered keywords in bold (SGR bold escape present in the rendered pane).
- No pre-existing equivalent feature; this closes a real gap (issue #2633 requested exactly this) without scope creep beyond bold/italic/underline/dim/reversed modifiers already defined by `VALID_MODIFIERS`.

---

sinelaw:
We should find way to do this without hard-coding the relation between the specific color and modifier keys in the rust core.

sinelaw:
:fire: 

Changed the format so that we don't need to duplicate all the theme keys

PR #2644, opened 2026-07-11
fix(indent): re-indent dedent keywords as they are typed (#2582)

### LINKED ISSUE #2582: Typing a dedent trigger does not re-indent the line: Python `else:` and custom `decrease_indent_pattern` tokens stay over-indented (only `}` dedents live)
**Version:** fresh 0.4.3 (built from `origin/master` @ `9f6135001`)

Follow-up to #2314 (closed as completed in 0.4.2). The core fix is confirmed working — `increase_indent_pattern`, `dedent_next_line_pattern`, and the Enter-time application of `decrease_indent_pattern` all fire correctly for custom languages now. But one row of #2314's own expected-behavior table still does not work:

> | `decrease_indent_pattern` | `^\s*CLOSE\b` | type `CLOSE` on an indented line | line dedents 1 level | stays indented |

Dedent rules are applied **only when Enter moves text down to a new line**. Typing a line whose content matches a dedent trigger never re-indents that line. This affects Fresh's **built-in Python rules** the same way, so it is visible without any configuration — only `}` (closing bracket) gets live "electric" dedent treatment.

## Steps to reproduce

Canonical case, built-in Python (no config needed):

1. `fresh --no-restore test.py`
2. Type `if a:` and press Enter → new line auto-indents to column 5 (correct).
3. Type `x = 1` and press Enter → new line keeps indent 4 (correct).
4. Type `else:` on that line.

Custom-language variant (same behavior): a `.fresh/config.json` language with `"decrease_indent_pattern": "^\\s*CLOSE\\b"`, an indented line, type `CLOSE`.

Contrast/control, built-in C: `fresh --no-restore test.c`, type `void f() {`, Enter (cursor lands at indent 4), type `}` → the line **immediately dedents to column 1**. So live dedent-on-type exists, but only for closing brackets.

## Expected behavior

When the typed line matches a dedent trigger, the line re-indents to one level shallower as you type it:

```
if a:
    x = 1
else:      ← dedents to column 1 when "else:" is typed
```

This is what VS Code (built-in Python language configuration), Sublime Text, and Vim all do for `else:`/`elif`/`except`, and what VS Code's `indentationRules.decreaseIndentPattern` (the model named in Fresh's docs) does for custom rules. It is also the exact behavior described in #2314's expected-results table. And Fresh itself already behaves this way for `}` — users will reasonably expect keyword triggers to work like the bracket ones.

## Actual behavior

The line stays at the inherited indent:

```
if a:
    x = 1
    else:      ← stays at column 5, never corrected
```

The error then **compounds**: pressing Enter after the mis-indented `else:` indents the next line to 8 (one level deeper than the wrong baseline), so the whole block drifts right:

```
    else:
        z = 3
```

The mis-indent is never corrected — not on later keystrokes and not at the following Enter. Same result for a custom language's `decrease_indent_pattern` token (`CLOSE` typed on an indent-4 line stays at 4; verified with `capture-pane | cat -A`). Note that the doc's own `begin`/`end` example (`"decrease_indent_pattern": "^\\s*end\\b"`) is affected in the way users would hit hardest: you always *type* `end` at the end of a block — you never Enter-split before it — so the rule effectively never dedents anything in real use.

What **does** work (verified, for scope): Enter-split — placing the cursor before `CLOSE tail` on an indented line and pressing Enter drops the moved-down `CLOSE tail` line one level shallower, exactly as `docs/configuration/index.md` describes.

## Workaround

Manually unindent each dedent-keyword line (Shift+Tab / Backspace) after typing it.

---
*Searches performed (no existing issue): `else dedent`, `indent typing keyword`, `auto-indent else elif`, `decrease_indent_pattern` (only #2314, closed).*

sinelaw:
Fixes #2582.

## Problem

Live "electric" dedent only fired for closing brackets (`}`). Typing a keyword that matches a language's `decrease_indent_pattern` — Python `else:`/`elif`/`except`/`finally`/`case`, Ruby/Lua/Fish/Pascal `end`, Bash `fi`/`done`/`esac`, or a user's custom token — left the line stuck at its inherited indent:

```
if a:
    x = 1
    else:      ← stays over-indented
```

The error then compounded: pressing Enter after the mis-indented `else:` indented the next line another level deeper, so the whole block drifted right. Dedent rules were applied only when Enter split text onto a new line, never when the keyword was typed — even though Fresh already gave `}` this treatment, and the docs' own `decrease_indent_pattern` examples (`end`, `CLOSE`) are tokens you *type* at the end of a block, never Enter-split.

## Fix

- Add `IndentRules::calculate_dedent_for_typed_char` in `indent_rules.rs` — the keyword counterpart of the existing `calculate_dedent_for_delimiter` (which handles brackets). It matches the line's leading content (after the just-typed char) against the language's `decrease` rule and returns the target indent, aligned to the reference line one level up.
- Hook it into the character-insert path (`insert_char_events`) so a completed dedent keyword snaps the line one level shallower, mirroring the existing closing-bracket `handle_auto_dedent`.

The hook is deliberately conservative:
- only when there's **no active selection** and the **cursor is at the end of the line's content** (the way these keywords are actually typed), so it never fires while editing mid-line;
- **idempotent** — it no-ops when the line is already at the correct indent, so it adds no redundant undo step on subsequent keystrokes.

Only the per-language regex rules tier is touched; curly-brace languages keep their bracket-only decrease rule (their `}` is handled by the existing path), so nothing changes for them.

## Testing

- **Unit tests** (`indent_rules.rs`): `else`/`except`/`finally`/`case` (Python, incl. nested), `end` (Ruby/Lua), `fi`/`done` (Bash); plus negatives — an identifier like `elsewhere`, an ordinary statement, a curly-brace language, and a first line with no reference above.
- **End-to-end typing tests** (`actions.rs`): drive the production `Action::InsertChar` pipeline key-by-key; verify `else:` dedents to column 0 and that the following Enter indents the body to column 4 (not the old drifted 8).
- **Manual validation**: ran the built `fresh` binary in tmux on a Python file, typed `else:`, and confirmed via `capture-pane` that the line dedents to column 0 and its body lands at column 4.
- `cargo fmt`, `cargo clippy`, and `cargo check --all-targets` all clean; existing indent / auto-pair / auto-indent scenario suites pass.

https://claude.ai/code/session_01DvjnQtGxsMn3UKywougFDd

---
_Generated by [Claude Code](https://claude.ai/code/session_01DvjnQtGxsMn3UKywougFDd)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
 
 ### Bug Fixes
 
+* **Auto-indent**: typing a dedent keyword now re-indents the line as you type it, matching the "electric" dedent already applied to closing brackets. Python `else:`/`elif`/`except`/`finally`/`case`, Ruby/Lua/Fish/Pascal `end`, Bash `fi`/`done`/`esac`, and any language's custom `decrease_indent_pattern` token snap one level shallower instead of staying over-indented until the next save (#2582).
 * **Cursor column** is preserved on vertical moves that used to drift it - indented soft-wrapped lines, off-screen up/down over short lines (#2565), and blank lines with indentation guides (#2564).
 * **LSP args** - an empty `args` list now overrides a built-in server's defaults, so you can use one that takes no arguments (e.g. markdown-oxide for marksman) (#2549, reported by @alex-ball).
 * **Settings icons** now render on all terminals by default - standard Unicode symbols replace the Nerd Font glyphs that showed as `?`; opt back in with `editor.nerd_font_icons` (#2032).
--- a/crates/fresh-editor/src/input/actions.rs
+++ b/crates/fresh-editor/src/input/actions.rs
@@ -731,6 +731,23 @@ fn rules_dedent(state: &EditorState, position: usize, ch: char, tab_size: usize)
     })
 }
 
+/// Keyword variant of [`rules_dedent`]: when typing `ch` at `cursor` completes
+/// a dedent keyword (Python `else:`, Ruby/Lua `end`, Bash `fi`/`done`, …) on
+/// the line starting at `line_start`, returns the visual indent the line
+/// should snap to. Returns `None` when the language has no matching rule.
+fn rules_typed_dedent(
+    state: &EditorState,
+    ch: char,
+    cursor: usize,
+    line_start: usize,
+    tab_size: usize,
+) -> Option<usize> {
+    let rules = buffer_indent_rules(state)?;
+    rules.calculate_dedent_for_typed_char(&state.buffer, line_start, cursor, ch, tab_size, |b| {
+        byte_is_code(state, b)
+    })
+}
+
 /// Calculate the correct indent for a closing delimiter.
 ///
 /// Tiering mirrors the Enter-indent path: a language with a *bundled*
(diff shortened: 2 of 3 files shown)


sinelaw (commented):

Superseded on the production side, but the tests are worth keeping.

Issue #2582 was fixed by PR #2949 ("Auto-indent engine gaps: braceless TS/JS bodies and live keyword dedent"), merged 2026-08-10. Its body says "Fixes #2492" and "Fixes #2582". Master now has `handle_keyword_dedent` in `input/actions.rs:1128` and the trigger data in `indent_rules.rs:281`.

Master's version is the more careful one. It also takes `char_after`, and documents a three-part trigger policy: cursor at end of line only, fires only on the keystroke where the trigger first consumes the whole line, and only ever dedents. This PR's `handle_typed_keyword_dedent` has no `char_after`. It also calls `calculate_dedent_for_typed_char`, which no longer exists on master.

What this PR has that master does not is test coverage. Master only tests Python, `else:` and `elif`. This PR covers:

- ruby `end` and ruby `else`
- lua `end`
- bash `fi`
- python `except` and `finally`
- a rust negative case, typing `els` inside `fn f() {` must not dedent

None of those appear anywhere in master's tests. Suggest closing the PR but lifting those cases onto master's `handle_keyword_dedent`. They will need re-pointing at the current entry point, since the helper they call is gone.

sinelaw:
Fixed on master by #2949, merged as 88942ac0d. https://github.com/sinelaw/fresh/pull/2949

PR #2656, opened 2026-07-12
fix(lsp): anchor diagnostics to the editor's CoordMap so they track every edit kind (#2602)

### LINKED ISSUE #2602: LSP diagnostics don't track buffer edits: gutter markers and F8 point at stale lines after lines are inserted above — even for editor-applied code actions
Found during an LSP-diagnostics deep-dive on v0.4.3 (built from `origin/master` @ `11dccfad5`), rust-analyzer 1.94.1 (Full mode).

## Steps to reproduce
1. In a small cargo crate, put an error on a line in the lower half of the file, e.g. line 8: `let bad = p.zzz;` (no field `zzz`). Save so check-on-save publishes — red `●` on line 8, `E:` count in the status bar.
2. Insert several lines **above** the error without saving. The easiest fully-editor-driven repro: put the cursor on the struct name, press `Alt+.`, and accept the "Generate `new`" assist — Fresh itself inserts a 5-line `impl` block above the error (`Applied: Generate `new` (1 change(s))`). Manually typing new lines above shows the same behavior.
3. Look at the gutter and press `F8`.

## Expected behavior
Diagnostic ranges shift with buffer edits, as in VS Code: after 6 lines are inserted above, the squiggle/gutter marker moves from line 8 to line 14 together with the text, and `F8` jumps to the real (shifted) position. This is especially expected when the edit was applied **by the editor itself** (a code-action `WorkspaceEdit`), where the exact deltas are known.

## Actual behavior
Positions are frozen at publish time:
- The red `●` stays on line 8 — which is now the middle of the freshly generated `fn new(...)` (perfectly valid code), while the real error line (now 14) shows no marker.
- `F8` jumps to `8:17` — inside the generated code — and shows the `zzz` error message there; the actual `p.zzz` line is not navigable.
- The Diagnostics panel likewise lists `[E] 8:17`.

Everything corrects on the next save (check-on-save re-publishes with fresh positions). Reproduced consistently; the "stale positions until re-check" behavior also matches what #2583 found for Search & Replace match stepping, but this is the diagnostics subsystem.

Note: the inline-diagnostics feature advertises "version-aware staleness dimming", so staleness is a known concept — but the gutter marker, `F8` navigation, and the panel neither shift nor indicate staleness; they confidently point at the wrong code.

## Workaround
Save the file to trigger a re-check.

---
Search performed before filing (0 results each): `diagnostics stale position edit`, `diagnostic position shift insert line`, `diagnostics markers wrong line after edit`.

sinelaw:
## Summary

Fixes #2602. LSP diagnostics stayed frozen at their publish-time positions after editing above an error. Inserting or deleting lines above a diagnostic left the gutter `●`, `F8` navigation, inline underline, hover, and the Diagnostics panel pointing at the pre-edit line — the wrong code — until the next save re-published fresh positions.

## Root cause

LSP servers publish diagnostics at fixed positions and only re-publish after a re-check (for rust-analyzer's cargo `checkOnSave`, on save). Between publishes:

- The diagnostic **overlays** ride the buffer's marker tree, so they briefly track edits.
- But the stored diagnostics (`stored_push` / `stored_pull` / merged) keep their original LSP positions. The next `merge_and_apply` **rebuilds** overlays from those stale positions, snapping the gutter/`F8` target back; hover and the panel read the stale positions directly.

## Approach

Rather than hand-shift stored positions, **anchor each diagnostic to the buffer and let the editor's own coordinate mapper carry it forward** — the same `CoordMap` that already remaps virtual-line anchors and mirrors the marker tree.

- `AnchoredDiagnostic { diagnostic, anchor: Option<(start_byte, end_byte, epoch)> }` — each received diagnostic is stamped with a byte range and the buffer version it was seen at.
- `at_current_version` maps the anchor forward via `EditorState::coord_map` → current byte → current LSP position; falls back to the published range when there's no anchor (closed file) or the epoch was evicted/barriered.
- `Window::recompute_merged_diagnostics` is the **single materialisation point** for the merged view (still `Vec<Diagnostic>`, so no reader changed). It runs on every publish/pull **and** on every edit, mapping the anchored push/pull stores to the buffer's current version.

`CoordMap` is fed at the marker-adjustment chokepoint (`record_insert`/`delete`/`replace`, with barriers for opaque snapshot restores), so the raw readers use the exact same shift as the overlays and can never disagree with them.

## Why this replaces the first cut

The initial version derived the shift from the outgoing LSP `didChange` payload. That silently dropped whole classes of edits: a `BulkEdit` (undo, redo, replace-all, multi-cursor, code actions) is sent as a full-document replacement with `range: None`, so the shift skipped it and the diagnostic snapped back. It was also a fourth, independent copy of shift math already living in the marker tree. Anchoring to `CoordMap` covers every edit kind by construction and reuses the one canonical implementation.

Removed with no leftovers: `shift_diagnostics_for_changes`, `shift_position`, `edit_new_end`, and `Window::shift_stored_diagnostics_for_changes`. The push/pull stores now hold `AnchoredDiagnostic`; the merged `stored_diagnostics` stays `Vec<Diagnostic>`.

## Files

- `crates/fresh-editor/src/services/lsp/diagnostics.rs` — `AnchoredDiagnostic` (`capture` / `at_current_version`) + unit tests; removed the hand-rolled shift.
- `crates/fresh-editor/src/app/window/mod.rs` — anchored store types; `recompute_merged_diagnostics` + `buffer_id_for_uri`; edit-path call site.
- `crates/fresh-editor/src/app/async_messages.rs` — anchor diagnostics on receipt (push + pull); `merge_and_apply` delegates to `recompute_merged_diagnostics`.
- `CHANGELOG.md` — entry under 0.4.3 Bug Fixes.

## Testing

- New unit tests exercise the real `EditorState` + `coord_map` edit path: insert-above (rides down), delete-above (rides up), edit-below (untouched), and the closed-file no-op. All pass.
- Full `fresh-editor` lib suite: **2978 passed, 0 failed** (4 ignored).
- `cargo fmt` and `cargo clippy -p fresh-editor` clean (the only clippy warnings are pre-existing `type_complexity` in unrelated files).

## Known limitation

Like the overlays, anchoring happens when diagnostics are **received**, interpreting the server's positions against the buffer at that moment. If the server was behind when it published, that initial placement is as imprecise as it is today — corrected on the next publish. What's now robust is tracking every edit *after* a publish, across all edit kinds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
 
 ### Bug Fixes
 
+* **LSP diagnostics** now track buffer edits: inserting or deleting lines above an error shifts its gutter marker, `F8`/Diagnostics-panel target, and inline text down/up with the text, instead of leaving them frozen at the pre-edit line until the next save re-publishes. Stored diagnostic positions are remapped by the same incremental changes sent to the server, matching VS Code (#2602).
 * **LSP Rename**: a cross-file rename (F2) invoked from a use site no longer steals the active tab, jumping you to the definition's file at its stale cursor position; focus stays on the buffer you were editing, matching VS Code / Sublime / IntelliJ (#2599).
 * **Vi mode**: the indent operators `>>` / `<<` (with counts and motions like `>j`) and visual, visual-line, and visual-block `>` / `<` now shift lines by one level instead of doing nothing; `.` repeats them (#2438, #2606).
 * **Vi mode**: quote text objects `i"`/`a"` (and `'`/`` ` ``) now search forward on the line, so `ci"`/`di"` work from before the quotes — e.g. from the start of the line — instead of only when the cursor is already inside them (#2439).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
 
 ### Bug Fixes
 
-* **LSP diagnostics** now track buffer edits: inserting or deleting lines above an error shifts its gutter marker, `F8`/Diagnostics-panel target, and inline text down/up with the text, instead of leaving them frozen at the pre-edit line until the next save re-publishes. Stored diagnostic positions are remapped by the same incremental changes sent to the server, matching VS Code (#2602).
+* **LSP diagnostics** now track buffer edits: inserting or deleting lines above an error shifts its gutter marker, `F8` target, inline underline, hover, and the diagnostics panel down/up with the text, instead of leaving them frozen at the pre-edit line until the next save re-publishes. Each diagnostic is anchored to the buffer and carried forward by the same edit-tracking the editor uses for its own markers, so it stays put through ordinary typing, multi-cursor edits, code actions, and undo/redo alike, matching VS Code (#2602).
 * **LSP Rename**: a cross-file rename (F2) invoked from a use site no longer steals the active tab, jumping you to the definition's file at its stale cursor position; focus stays on the buffer you were editing, matching VS Code / Sublime / IntelliJ (#2599).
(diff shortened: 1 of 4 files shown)


sinelaw:
Need to reconsider how changes are tracked here, it should probably be using the marker overlay system (or similar) instead of manual tracking

PR #2659, opened 2026-07-12
fix(lsp): re-pull diagnostics from every server, not just the first (#2615)

### LINKED ISSUE #2615: incremental type checking not working Bug
I use **ty** uv tool for type checking but it doesn't change types until  restart.

> fresh config.json  is like this what changes need to be done

```
{
  "theme": "builtin://dark",
     
  "lsp": {
    "python": [
      {
        "command": "ruff",
        "enabled": true,
        "args": ["server"],
        "auto_start": true,
        "process_limits": {
          "max_memory_percent": 50,
          "max_cpu_percent": 90,
          "enabled": true
        }
      },
      {
        "command": "ty",
        "enabled": true,
        "args": ["server"],
        "auto_start": true,
        "process_limits": {
          "max_memory_percent": 50,
          "max_cpu_percent": 90,
          "enabled": true
        }
      }
    ]
  },
  "version": 2,
  "plugins": {
    "vi_mode": {
      "enabled": true,
      "settings": {
        "autoStart": false,
        "arrowKeys": true,
        "searchWordUnderCursor": true
      }
    }
  },
  "editor": {
    "auto_save_enabled": true,
    "auto_save_interval_secs": 1,
    "confirm_quit": true,
    "nerd_font_icons": false,
    "quick_suggestions_delay_ms": 65,
    "suggest_on_trigger_characters": false
  }
}

```

sinelaw:
Fixes #2615.

## Problem

The reporter runs two Python LSP servers — `ruff` and `ty` (type checker) — both configured with `args: ["server"]`, and both serve diagnostics via the **pull model** (`textDocument/diagnostic`). Symptom: `ty`'s type errors never refresh after an edit; they only update on a restart.

## Root cause

`Diagnostics` is a **merged** feature — results are meant to be combined from *all* eligible servers. But every pull-diagnostic request site resolved a *single* handle via `handle_for_feature_mut(Diagnostics)`, i.e. the first-listed diagnostic-capable server. So only `ruff` (first in the config) was ever re-pulled after an edit; `ty` was never re-pulled and its diagnostics went stale until a restart re-opened the document.

A secondary latent bug: pull diagnostics and their `result_id`s were keyed by URI only, so two pull servers on the same file would have clobbered each other's results and cross-sent incremental cursors.

## Fix

- Every pull site now fans the `textDocument/diagnostic` request out to **all** diagnostic-capable handles:
  - the per-edit debounce (`check_diagnostic_pull_timer`)
  - the didOpen initial pull (`file_operations`)
  - `workspace/diagnostic/refresh` and post-quiescent re-pull (`pull_diagnostics_for_language`)
  - dynamic capability registration
  - the settings confirm-open path (`popup_overlay_actions`)
- Introduced `Window::pull_diagnostics_for_uri`, the single fan-out helper the pull sites share.
- Pull diagnostics are now stored **per-server** (`URI → server → diagnostics`), mirroring `stored_push_diagnostics`, and `result_id`s are tracked per-server (`URI → server → result_id`) so servers keep independent incremental cursors.
- `LspPulledDiagnostics` carries the originating `server_name`; the handler stores per-server and drops reports from servers that have shut down (parity with the push path).
- `clear_diagnostics_for_server` now also clears the server's pull diagnostics and result_ids.

Servers that never advertised `diagnosticProvider` drop the request on the async side, so fanning out to every eligible handle is safe.

## Tests

- New e2e regression test `lsp_multi_server_diagnostic_pull.rs`: spawns two fake pull-diagnostic servers for one language and asserts **both** are re-pulled after an edit. Verified it fails against the pre-fix behavior (second server never reaches its post-edit pull) and passes with the fix.
- Full `lsp` e2e suite green (235 passed), diagnostics subset green (41 passed).

## Manual validation

Ran the regression test in a tmux session; captured output shows both servers re-pulled after edits:

```
[TEST] linter pulls=3, typechecker pulls=2
test ... ok
```

https://claude.ai/code/session_01ExQjfkPm1voC4XXP6BN1HG

---
_Generated by [Claude Code](https://claude.ai/code/session_01ExQjfkPm1voC4XXP6BN1HG)_

### DIFF
--- a/crates/fresh-editor/src/app/async_dispatch.rs
+++ b/crates/fresh-editor/src/app/async_dispatch.rs
@@ -248,11 +248,18 @@ impl Editor {
                 AsyncMessage::LspPulledDiagnostics {
                     request_id: _,
                     uri,
+                    server_name,
                     result_id,
                     diagnostics,
                     unchanged,
                 } => {
-                    self.handle_lsp_pulled_diagnostics(uri, result_id, diagnostics, unchanged);
+                    self.handle_lsp_pulled_diagnostics(
+                        uri,
+                        server_name,
+                        result_id,
+                        diagnostics,
+                        unchanged,
+                    );
                 }
                 AsyncMessage::LspInlayHints {
                     request_id,
--- a/crates/fresh-editor/src/app/async_messages.rs
+++ b/crates/fresh-editor/src/app/async_messages.rs
@@ -221,42 +221,61 @@ impl Editor {
     pub(super) fn handle_lsp_pulled_diagnostics(
         &mut self,
         uri: String,
+        server_name: String,
         result_id: Option<String>,
         diagnostics: Vec<Diagnostic>,
         unchanged: bool,
     ) {
+        // Drop reports from servers that have since been shut down, matching
+        // the push path — queued messages can outlive the server.
+        if let Some(lsp) = self.lsp() {
+            if !lsp.has_server_named(&server_name) {
+                tracing::debug!(
+                    "Dropping pulled diagnostics from stopped server '{}' for {}",
+                    server_name,
+                    uri
+                );
+                return;
+            }
+        }
+
+        // Store result_id (per-server) even on an unchanged report so the
+        // next pull for this server keeps sending the correct cursor.
+        if let Some(result_id) = result_id {
+            self.active_window_mut()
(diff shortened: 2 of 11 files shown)


ak24watch:
when will the  next realease?

PR #2702, opened 2026-07-15
fix(explorer): let a new item's name create its missing parent directories (#2640)

### LINKED ISSUE #2640: Add options "Automatic Directory Creation with Slash"
## Proposal: Add the options "Automatic Directory Creation with Slash"
Currently, the Fresh Editor throws an error when a forward slash is included in the  filename during file creation, which causes the operation to fail.
However, VSCode provides a feature that automatically creates and places directories when slashes are included in the filename.
This is an extremely useful feature that significantly enhances user experience.

I believe this would be beneficial to implement, and I consider the implementation to be relatively straightforward.

To be more precise, rather than simply creating directories,
the feature interprets a relative path specification and automatically creates any missing directories along the path as needed.

## Considerations
* Permissions
    - An important aspect to consider in this implementation is permission handling.
    - Under no circumstances should unwrap or similar error-handling methods be used 
      in case of failure.
* Directory Traversal
    - Access attempts using "~", "$HOME", or "/" should be prevented.
    - These should be interpreted as literal strings and should not be expanded.
* Relative Path Access
    - Paths containing "." and ".." should be handled as-is.
    - These are extremely useful to users when creating files and directories.

## Note
* I used Claude Haiku 4.5 to translate this issue
* If this sentence including rude sentences, I really apologize to you

asukaminato0721:
## Summary

**Motivation.** Creating a file somewhere other than the selected directory meant leaving the explorer: make the directory first, then create the file inside it, or drop to a shell. `Ctrl+N` accepted a bare filename only — a `/` in the name was refused outright — so the natural thing to type, `src/components/app.rs`, simply didn't work.

**Goal.** Let a new item's name carry a path, create whatever directories are missing along the way, and refuse the names that would put the file somewhere the user didn't mean.

**Before → after**

| You type at `Ctrl+N` | Before | After |
|---|---|---|
| `src/components/app.rs` | `Name cannot contain '/'` | file created, both missing directories created with it |
| `src/main.rs`, and that file exists | *(unreachable — `/` was refused)* | `'src/main.rs' already exists`; the existing file is untouched |
| `../notes.md`, staying inside the project | `Name cannot contain '/'` | created in the parent directory |
| `../../../outside.rs` | `Name cannot contain '/'` | `New item path must stay within the project directory`; nothing is written |
| `/etc/passwd`, `C:\…`, `\\server\share\…` | `Name cannot contain '/'` | rejected as not relative |
| `~/notes.md` | `Name cannot contain '/'` | `~` stays literal — no home expansion |

The existing-file check matters because this PR is what makes it reachable: once a name can address an existing path, `rename` would otherwise replace that file silently, with no prompt and no undo. Rejecting is deliberate — every other name that fails validation here reports why and leaves the temporary item in place, and an "overwrite" answer would also have to invalidate the clobbered file's open buffer.

Symlink escapes are checked by resolving the deepest existing ancestor, not lexically, so a directory that links out of the project can't be used as a way around the containment check. Only a genuinely missing path is treated as "not created yet" — permission and loop errors are reported rather than being read as absence.

**Verified by hand**, driving the real UI and comparing against the released 0.4.10 binary: nested creation, the collision refusal with the target file's bytes intact, and the escape refusal with nothing written outside the project.

---

fix #2640 

New files can use nested paths such as src/components/app.rs; missing directories are created automatically.
Relative . and .. components remain supported.
Absolute, drive-prefixed, and UNC paths are rejected.
~ and $HOME remain literal—no expansion.
Filesystem and permission errors are handled without panics.
Added localized validation messages and regression coverage.

### DIFF
--- a/crates/fresh-editor/locales/bg.json
+++ b/crates/fresh-editor/locales/bg.json
@@ -1006,6 +1006,8 @@
   "explorer.moved_to_trash": "Преместено в кошчето: %{name}",
   "explorer.new_directory_prompt": "Име на новата папка: ",
   "explorer.new_file_prompt": "Име на новия файл: ",
+  "explorer.new_item_path_must_be_relative": "Пътят на новия елемент трябва да бъде относителен",
+  "explorer.new_item_path_outside_project": "Пътят на новия елемент трябва да остане в директорията на проекта",
   "explorer.opened": "Файловият браузър е отворен",
   "explorer.opened_file": "Отворен: %{name}",
   "explorer.paste_cancelled": "Поставянето е отменено",
--- a/crates/fresh-editor/locales/bg.json
+++ b/crates/fresh-editor/locales/bg.json
@@ -1006,6 +1006,7 @@
   "explorer.moved_to_trash": "Преместено в кошчето: %{name}",
   "explorer.new_directory_prompt": "Име на новата папка: ",
   "explorer.new_file_prompt": "Име на новия файл: ",
+  "explorer.new_item_path_exists": "'%{name}' вече съществува",
   "explorer.new_item_path_must_be_relative": "Пътят на новия елемент трябва да бъде относителен",
   "explorer.new_item_path_outside_project": "Пътят на новия елемент трябва да остане в директорията на проекта",
   "explorer.opened": "Файловият браузър е отворен",
--- a/crates/fresh-editor/locales/cs.json
+++ b/crates/fresh-editor/locales/cs.json
@@ -973,6 +973,7 @@
   "explorer.refreshing": "Obnovuji %{name}...",
   "explorer.rename_cancelled": "Přejmenování zrušeno",
   "explorer.new_item_path_must_be_relative": "Cesta nové položky musí být relativní",
+  "explorer.new_item_path_outside_project": "Cesta nové položky musí zůstat v adresáři projektu",
   "explorer.rename_invalid_dot": "Název nemůže být '.' ani '..'",
   "explorer.rename_invalid_separator": "Název nemůže obsahovat oddělovač cesty",
   "explorer.rename_prompt": "Přejmenovat na: ",
--- a/crates/fresh-editor/locales/cs.json
+++ b/crates/fresh-editor/locales/cs.json
@@ -972,6 +972,7 @@
   "explorer.refreshed_default": "Obnoveno",
   "explorer.refreshing": "Obnovuji %{name}...",
   "explorer.rename_cancelled": "Přejmenování zrušeno",
+  "explorer.new_item_path_exists": "'%{name}' již existuje",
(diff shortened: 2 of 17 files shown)


sinelaw (approved):

- Confirms bug: master's `perform_file_explorer_rename` rejects any `/` in new-file names, matching issue #2640.
- Fix reuses the existing `FileSystem::create_dir_all` trait method (no raw `std::fs`), keeps the pre-existing rename/validation flow, and adds the check only for `is_new_file` — minimal, non-duplicative, no architectural changes.
- All 14 locale files got the new `explorer.new_item_path_must_be_relative` key; unit + e2e tests added (fail-without-fix pattern).
- Interactively verified in tmux against a debug build on the PR branch:
  - `Ctrl+N` → typed `src/components/app.rs` → dirs auto-created, temp `untitled_*` file moved into place, buffer focused on the new file.
  - Typed an absolute path (`/etc/passwd_evil`) → rejected with "New item path must be relative", nothing created outside the project.
  - Typed `../outside_project/evil.txt` → allowed and escaped the project dir, matching the issue's explicit spec that `.`/`..` remain literal relative components (not a regression).
- No duplicate/adjacent existing feature found.

---

sinelaw:
Escaping the project dir shouldn't be allowed, e.g. we can evaluate the absolute canonical target path and check that its under the project dir

-

sinelaw:
Manually verified in a real terminal, not just via the test suite.

I built this branch stacked on current `master` and drove it by hand in a scratch project, then ran the identical steps against the released 0.4.10 binary as a baseline.

**This branch**

| Step | Result |
|---|---|
| `Ctrl+N` → `deep/nested/dir/newfile.rs` → Enter | file created, all three directories created |
| `Ctrl+N` → `src/main.rs` (file already exists) → Enter | `'src/main.rs' already exists`; the existing file's contents were **byte-for-byte intact** and the temporary item was left in place |
| `Ctrl+N` → `../../../REALESCAPE.rs` from the project root → Enter | `New item path must stay within the project directory`; nothing created anywhere outside |
| `Ctrl+N` → `../ESCAPED.rs` with a nested node selected | created at `deep/nested/ESCAPED.rs`, i.e. parent-relative is allowed as long as it stays inside — matches the intent |

**Released 0.4.10 baseline**: `Ctrl+N` → `src/main.rs` → `Name cannot contain '/'`. So nested paths are genuinely new here, which also means the overwrite hazard is one this PR opens — that's why the existence check belongs in this branch rather than as a follow-up.

One behaviour worth knowing: after the "already exists" rejection, pressing `F2` retries as a *rename*, which rejects `/` outright, so the nested retry has to go through a fresh `Ctrl+N`.

sinelaw:
:fire:

PR #2710, opened 2026-07-16
feat(tabs): extract a tab into a co-tenant workspace over the same project root

sinelaw:
## Extract a tab into its own workspace over the same project

Adds an "Extract Tab to New Workspace" command palette entry and an "Extract to New Workspace" tab right-click item. Both **peel the focused tab into a new, independent window (`Window`) co-tenanting the same project root** — a second workspace over the same project, with its own split layout / session, distinguished by a numbered label (`repo`, `repo (2)`, …). File tabs and terminal tabs (the live PTY moves along, the running process untouched) both root the new window at the source window's root.

The live `EditorState` moves — unsaved edits and undo history travel with the tab rather than being re-read from disk. Source-window leaves that displayed the buffer are re-pointed at another of their tabs, closed, or reseeded with a scratch buffer so the split tree never dangles. The new workspace **inherits the source's authority configuration** (its backend spec): a local source rebuilds a fresh local backend, a remote/container source's spec is carried so reconnect-on-activate rebuilds the same backend — extraction never silently downgrades a remote session to local.

> Root selection went through a couple of iterations in review. It landed on "the source window's root" — the only choice that makes the feature coherent (peel a tab into its own window over the *same* project), which in turn required supporting multiple workspaces per root, below.

## Multiple workspaces per root

The persistence layer stops treating "one file per directory" as the invariant and keys by durable identity (`stable_id`) instead, so a root can host several co-tenant workspaces:

- Storage is `workspaces/{encoded-root}.{stable-id}.json`; `discover_sessions` dedups by `stable_id` (not canonical root), so co-tenants each surface as their own session, while id-less legacy files still map per-root and are suppressed once any id-keyed file claims that root.
- `Workspace::load_by_id` restores one exact identity; each window restores its **own** `stable_id` rather than the freshest file for the root.
- `save` retires only the legacy root-keyed file it re-keys away from — never an id-keyed peer; `delete_by_id` closes a single co-tenant without taking its peers down, while whole-root `delete` stays the project-teardown path.
- `handle_open_file_in_inactive_session` migrates the buffer's metadata and undo/redo event log to the target window (a pre-existing latent bug: both used to stay behind, severing undo history for any buffer moved across windows).

### Co-tenants survive a cold reboot

Persisting co-tenants wasn't enough — restoring them exposed two bugs (both caught by a new deterministic round-trip test and fixed here):

- Bring-up dropped **every** persisted window whose root matched the launch cwd — the leftover "one session per directory" guard in `build_persisted_window_shells`. Given `pick_active_window_for_cwd`'s semantics it only ever fired on genuine co-tenants, so the second window over a root silently vanished on relaunch. The guard is gone: every persisted entry other than the active pick becomes its own window.
- The active (picked) window adopted its session's label/root/state but **not** its `stable_id`, minting a fresh one — so `load_by_id` missed its own file and fell back to the freshest sibling, restoring the wrong file. The active window now continues the picked session's identity, mirroring the background-shell path.

A dedicated integration binary (`orchestrator_co_tenant_restore`, isolated because it sets the process-global `XDG_DATA_HOME`) covers the full extract → save → cold-reboot round-trip: two windows over one root each reopen exactly their own file.

## Review fixes folded in

From the automated review: best-effort `delete`, corrupt-file diagnostic, shared `workspace_freshness_rank`, canonicalization raw-fallback, the tab-context-menu clamp dimension mismatch, and the retarget-leaf selection harmonized with the close path (LRU / hidden-from-tabs / last-buffer opt-out). The Windows `test_search_replace_file_glob` separator deadlock and a rebased test's legacy-filename assumption were also fixed.

## Deliberately out of scope (follow-up)

Opening the **same file in two windows** yields two independent buffers today (any two windows — local or remote — not just co-tenants; `BufferId` is global, buffers are per-window, `open_file` dedups only the active window). This is a pre-existing, general problem whose correct fix is shared buffers keyed by *(backend, path)* — a buffer-ownership re-architecture. It's tracked as a separate effort rather than bolted onto this PR; multiple-per-root doesn't add a new hazard class, only makes the existing one more likely.

https://claude.ai/code/session_01GQYut8xrWsYyQffc4D84dz


crates/fresh-editor/src/app/window_actions.rs
```diff
@@ -806,6 +806,378 @@ impl crate::app::Editor {
...
+        // the move while this window can still resolve it.
+        let name = self.get_buffer_display_name(buffer_id);
+
+        self.retarget_leaves_off_buffer(buffer_id);
+
+        let label = root
+            .file_name()
+            .map(|n| n.to_string_lossy().into_owned())
+            .unwrap_or_else(|| root.to_string_lossy().into_owned());
+        let target = self.create_window_at(root, label);
```
sinelaw: **Extracting a terminal tab into a dormant persisted workspace orphans the live PTY.**

`create_window_at` dedups by root against *all* `self.windows`, including windows still in `materialize_pending`. The guard above only rejects `root` matching the **active** window, not a dormant one. So if the terminal's cwd equals a backgrounded persisted workspace's root, `target` is that dormant shell.

`move_terminal_machinery_to_window` then `adopt`s the live `TerminalHandle` and re-registers the pid into it, but the subsequent `set_active_window(target)` materializes the shell → `apply_workspace_layout` rebuilds the split tree from the snapshot → `clean_orphaned_buffers` drops the terminal buffer, because it guards on `file_path().is_none() && !is_modified()` — which a terminal buffer (no file path) always satisfies. Result: the buffer and tab vanish, but the PTY, its pid registration, and the `terminal_buffers` binding remain — a running shell with no UI handle and no shutdown path for the session's life.

The file-tab sibling path is safe from this precisely because its buffer *is* file-backed and thus protected by that same guard. Suggest making the extract path aware of the durable-storage layer (adopt/materialize an existing on-disk workspace's identity for the root rather than reusing a dormant window and then rebuilding over it), or at minimum reap the terminal machinery when its anchoring buffer is cleaned.

---

crates/fresh-editor/src/workspace.rs
```diff
@@ -994,11 +1186,13 @@ impl Workspace {
...
+    /// Delete workspace state for a working directory. Removes *every*
+    /// file claiming the directory — the id-keyed snapshot plus any legacy
+    /// root-keyed file that survived migration — so a killed workspace
+    /// can't resurrect from a stale duplicate.
     pub fn delete(working_dir: &Path) -> Result<(), WorkspaceError> {
-        let path = get_workspace_path(working_dir)?;
-        if path.exists() {
-            std::fs::remove_file(path)?;
+        for ident in candidate_files_for_root(working_dir)? {
+            std::fs::remove_file(&ident.path)?;
```
sinelaw: **`delete` aborts on the first `remove_file` error, letting a killed workspace resurrect.**

The `?` propagates the first per-file error, so the loop stops before removing the remaining variants — contradicting the doc comment's "removes *every* file claiming the directory." A realistic TOCTOU: the checkpoint-save path retires legacy files best-effort (`remove_file(&legacy).ok()` at ~1116) from another process. If that legacy file is retired between this function's directory scan and its `remove_file`, the call returns `ErrorKind::NotFound`, the loop aborts, and the id-keyed snapshot survives — resurrecting on the next boot's `discover_sessions`, the exact outcome this delete was meant to prevent.

Make deletion best-effort per file: ignore `NotFound`, continue past errors, and optionally return an aggregate error only after attempting all candidates.

---

crates/fresh-editor/src/workspace.rs
```diff
@@ -926,6 +1098,26 @@ impl Workspace {
...
+            if let Ok(legacy) = get_workspace_path(&self.working_dir) {
+                if legacy != path && legacy.exists() {
+                    tracing::info!(
+                        "Retiring legacy workspace file {:?} (re-keyed to {:?})",
+                        legacy,
+                        path
+                    );
+                    // Best-effort: a failed delete just leaves a stale
+                    // duplicate that lookup arbitration already outranks.
+                    let _ = std::fs::remove_file(&legacy).ok();
```
sinelaw: **`save` never retires a stale *id-keyed* sibling, so duplicate workspace files accumulate.**

This retirement only removes the legacy root-keyed name (`get_workspace_path`). It never deletes a rival `<root>.<otherid>.json` claiming the same directory with a different `stable_id`. The load/discover arbitration merely *skips* the loser without deleting it, so once two files claim one directory (e.g. two processes each adopting the legacy id-less file and minting their own id, or a fresh window created at a root that already had a persisted `<root>.<oldid>.json`), both coexist forever and the shadowed session's state persists on disk indefinitely, re-accumulating on every such event.

The PR text's "duplicate files … resolve to the freshest snapshot" holds for reads, but nothing garbage-collects the losers. Consider having `save` (or the arbitration chokepoint) delete non-winning duplicates for the root, mirroring what `delete` already enumerates.

---

crates/fresh-editor/src/app/window_actions.rs
```diff
@@ -806,6 +806,378 @@ impl crate::app::Editor {
...
+        // one being extracted. The rect is a probe — only ids matter here.
+        let probe = ratatui::layout::Rect::new(0, 0, 1, 1);
+        let showing: Vec<(crate::model::event::LeafId, Option<fresh_core::BufferId>)> = mgr
+            .root()
+            .get_leaves_with_rects(probe)
+            .into_iter()
+            .filter(|(_, displayed, _)| *displayed == buffer_id)
+            .map(|(leaf_id, _, _)| {
+                let replacement = view_states.get(&leaf_id).and_then(|vs| {
+                    vs.open_buffers.iter().find_map(|t| match t {
```
sinelaw: **Source-window leaf re-pointing diverges from the real close-tab path in three ways.**

`retarget_leaves_off_buffer` re-implements "what tab replaces this one when it leaves a leaf," but differs from `resolve_close_replacement` / `close_tab_in_split`:

1. Picks the replacement as the **first** entry of `open_buffers` (tab order) instead of the focus-history LRU the close path uses (`focus_history.iter().rev()`).
2. Doesn't consult `hidden_from_tabs`, so a leaf can be re-pointed onto a hidden/helper buffer (the close path excludes those in both its LRU branch and its fallback).
3. The last-leaf branch calls `self.new_buffer()` directly (line ~1176), bypassing the `auto_create_empty_buffer_on_last_buffer_close` opt-out and `synthetic_placeholder` marking that the close path applies — forcing a visible `[No Name]` even for users who disabled that.

Two copies of this selection logic will drift as tab rules evolve. Prefer driving the source-window side through the existing silent close-tab path.

---

crates/fresh-editor/src/workspace.rs
```diff
@@ -720,6 +753,131 @@ pub fn get_workspace_path(working_dir: &Path) -> io::Result<PathBuf> {
...
+    for entry in entries.flatten() {
+        let name = entry.file_name();
+        let Some(name) = name.to_str() else {
+            continue;
+        };
+        if name != legacy_name && !(name.starts_with(&id_prefix) && name.ends_with(".json")) {
+            continue;
+        }
+        let path = entry.path();
+        let Ok(content) = std::fs::read_to_string(&path) else {
```
sinelaw, line 830: **`candidate_files_for_root` silently skips corrupt/unparseable workspace files, changing failure semantics.**

Any file that fails the `serde_json` value-parse (or lacks a matching `working_dir`) is dropped from the candidate set. Two consequences vs the pre-PR direct-deserialize path:

- `Workspace::load` now returns `Ok(None)` — indistinguishable from "no saved session" — for a corrupt **sole** workspace file, where base returned `Err(WorkspaceError::Json)`. On the startup restore path this downgrades a `tracing::warn!` diagnostic to a `debug!` "no session," and the editor starts fresh (next save overwrites).
- `delete` (which iterates the same candidates) never removes such a corrupt file, so it lingers on disk and can resurrect.

Buffer *contents* are safe (separate hot-exit recovery files), so this is layout/diagnostic loss rather than text loss — but the silent `Ok(None)` on a damaged file is worth surfacing as at least a logged warning, and `delete` should still unlink files it can't parse.

---

crates/fresh-editor/src/app/render.rs
sinelaw, line 1: **(This finding is about `render_new_tab_menu` / `NewTabMenu`, which is unchanged by the PR, hence a file-level comment.)**

The PR introduced the shared `clamped_position`/`height` helper and routed `TabContextMenu` and `FileExplorerContextMenu` through it for render + both hit-tests. But the `+` new-tab popup (`NewTabMenu`) still has the exact clamp drift this change set out to eliminate, left ungeneralized:

- `render_new_tab_menu` (~render.rs:3999) hand-clamps `menu.position` to the screen when it would overflow the right/bottom edge.
- The hover hit-test (`mouse_input.rs:1030`) and click hit-test (`handle_new_tab_menu_click`, `mouse_input.rs:3510`) use the raw, unclamped `menu.position`.

Open the `+` menu within `NEW_TAB_MENU_WIDTH` (18) of the right edge: it draws shifted onto screen while its clickable/hoverable region stays anchored partly offscreen, so hover highlighting and clicks miss or hit the wrong item. `NewTabMenu` in `types/context_menu.rs` has no `clamped_position` method, unlike the two menus that were fixed. Deeper fix: give `NewTabMenu` a `clamped_position`/`height` too and route its render + both hit-tests through it.

---

crates/fresh-editor/src/app/render.rs
```diff
@@ -3939,24 +3939,12 @@ impl Editor {
...
-        };
-
-        let menu_y = if menu.position.1 + menu_height > screen_height {
-            screen_height.saturating_sub(menu_height)
-        } else {
-            menu.position.1
-        };
+        // Adjust position to stay within screen bounds (same math the mouse
+        // hit-testing uses, so clicks always land on what was drawn)
+        let (menu_x, menu_y) = menu.clamped_position(frame.area().width, frame.area().height);
```
sinelaw, line 3937: **Tab-context-menu clamp is fed two different dimension sources between render and hit-test.**

Render passes `frame.area().width/height` here, but the three hit-test sites pass `self.terminal_width/terminal_height` (`mouse_input.rs:1051`, `:3379`, `:3468`). These are maintained by independent mechanisms: `frame.area()` is ratatui's autoresized true size re-read every draw, while `self.terminal_width/height` is only updated when a crossterm `Resize` event is dequeued (`Editor::resize`). During a resize the draw can already report the new size (menu drawn clamped to it) while the pending `Resize` hasn't been processed (clicks hit-tested against the old size), so a right-click menu near the edge can render in one place and accept clicks in another — the very mismatch `clamped_position` claims to remove.

Low severity (transient, edge-only), but the file-explorer menu already does this correctly by using `last_frame` on both sides; mirror that for the tab menu (pass one consistent dimension source to render and all hit-tests).

---

crates/fresh-editor/src/workspace.rs
```diff
@@ -720,6 +753,131 @@ pub fn get_workspace_path(working_dir: &Path) -> io::Result<PathBuf> {
...
+        };
+        // Authoritative check: the file must itself claim this directory.
+        let claimed = val
+            .get("working_dir")
+            .and_then(|v| v.as_str())
+            .map(PathBuf::from);
+        let Some(claimed) = claimed else {
+            continue;
+        };
+        let claimed = claimed.canonicalize().unwrap_or(claimed);
```
sinelaw: **Asymmetric canonicalization can skip a valid workspace file (PLAUSIBLE, narrow trigger).**

`target` is `working_dir.canonicalize().unwrap_or(raw)` and `claimed` here is `claimed.canonicalize().unwrap_or(claimed)`, computed independently. The common cases stay symmetric and match (both canonicalize, or both fail on the same path). The genuine failure is narrow: the stored `working_dir` string and the lookup path differ but denote the same location via a symlink, and that symlink later fails to resolve on one side only — then one side is canonical and the other raw, they compare unequal at line 845, the candidate is dropped, and `load`/`find_workspace_file_by_root` returns `None`, so the saved session appears lost though the file is present and valid.

Normalize both sides through one helper, or fall back to a raw-string comparison when either `canonicalize` fails, so a one-sided resolution failure can't defeat the match.

---

crates/fresh-editor/src/app/orchestrator_persistence.rs
```diff
@@ -293,22 +303,57 @@ fn discover_sessions(
...
+    // stable-id-bearing files breaking ties. Losers are skipped, not
+    // deleted; `Workspace::save` retires the legacy duplicate on the next
+    // save through the same arbitration.
+    let mut best: std::collections::BTreeMap<PathBuf, Candidate> =
+        std::collections::BTreeMap::new();
+    for c in found {
+        let key = canonical_key(&c.root);
+        match best.get(&key) {
+            Some(cur)
+                if (cur.saved_at, cur.stable_id.is_some())
```
sinelaw: **Duplicate "freshest snapshot wins, stable-id breaks ties" arbitration (cleanup).**

This `(saved_at, stable_id.is_some())` tie-break duplicates the identical logic in `workspace::find_workspace_file_by_root` (~workspace.rs:871). One copy governs boot-time session discovery here, the other the per-root read chokepoint — and the two feed the two separate `stable_id` adoption sites (the shell adopts from `discover_sessions`, `apply_workspace_layout` adopts from `Workspace::load`). If the rule is later tweaked in one place only (e.g. adding a version field), boot discovery and per-root restore can disagree on which file is authoritative for a root, handing a window one id at boot and a different one at materialize. Extract a single shared ranking helper both call.

---

crates/fresh-editor/src/workspace.rs
```diff
@@ -807,9 +965,16 @@ impl From<serde_json::Error> for WorkspaceError {
 }
 
 impl Workspace {
-    /// Load workspace for a working directory (if exists)
+    /// Load workspace for a working directory (if exists).
+    ///
+    /// Resolution goes by the `working_dir` recorded inside each workspace
+    /// file (filenames are stable-id-keyed; legacy files are root-keyed) —
+    /// see [`find_workspace_file_by_root`].
     pub fn load(working_dir: &Path) -> Result<Option<Workspace>, WorkspaceError> {
```
sinelaw, line 1007: **Redundant re-read/re-parse/re-canonicalize on the persistence read path (efficiency).**

`load` resolves the winner via `find_workspace_file_by_root` → `candidate_files_for_root`, which already did `read_dir` + `read_to_string` + `serde_json` parse + canonicalize on every candidate. `load` then re-does `exists()`, a second `read_to_string`, a second parse into `Workspace`, and a second `working_dir` canonicalize on the winner. Because restore calls `load` once per window at boot, an N-window boot runs N full `read_dir` scans of `workspaces/` plus a double parse of each winner.

Have `candidate_files_for_root` hand back the parsed winner (or its content + canonical dir) so `load` deserializes without a second read. Relatedly, the legacy-retirement block (~line 1106, whose own comment notes it "runs on every checkpoint") does a redundant `canonicalize` + `exists` stat on every save even for id-keyed workspaces with nothing to retire — derive the legacy path from the already-canonical `path` and gate retirement behind a one-shot "loaded from legacy filename" flag.

---

sinelaw (commented):
Automated high-effort code review (8 finder angles → adversarial verification). 10 findings posted inline, ranked by severity.

**Would block on:**
1. Extracting a **terminal** tab into a dormant persisted workspace orphans the live PTY — the running shell survives with no UI handle or shutdown path (`clean_orphaned_buffers` drops the terminal buffer, which lacks the file-path protection the file-tab path relies on).
2. `Workspace::delete` aborts on the first `remove_file` error, letting the id-keyed snapshot survive and resurrect a "deleted" workspace.
3. `save` never retires stale id-keyed sibling files, so duplicate/shadowed workspace snapshots accumulate on disk indefinitely.

**Worth fixing before merge:** source-window leaf re-pointing diverges from the real close-tab path (LRU / `hidden_from_tabs` / last-buffer opt-out); `candidate_files_for_root` silently swallows corrupt files (lost diagnostic + `delete` leaves them behind).

**Lower severity:** `NewTabMenu` still has the clamp drift this PR fixed for its sibling menus; the tab-menu clamp mixes `frame.area()` (render) with `terminal_width` (hit-test); canonicalization asymmetry can drop a valid workspace file in a narrow symlink case; duplicated freshest-wins arbitration and redundant persistence I/O.

Four other candidates (poisoned-mutex exit drop, stale-LeafId aliasing in `retarget_leaves_off_buffer`, file-tab extract data loss, cross-window metadata/undo orphaning) were investigated and refuted as safe under the code's invariants. The durable-`stable_id` redesign is sound overall — the sharp edges are all at the seam between the new dynamic workspace lifecycle, the persistence layer, and the pre-existing dormant-window reuse in `create_window_at`.

---

sinelaw:
Worked through the review (pushed in 74e020a). Disposition per finding — fixed most, pushed back on a couple:

**Fixed as described**
- **`delete` aborts on first error** — now best-effort per file: `NotFound` (a concurrent checkpoint retiring a sibling) is ignored, other per-file errors don't abort the loop, first hard error returned after all candidates are attempted. Test: `delete_removes_every_variant`.
- **`save` never retires id-keyed siblings** — `save` now retires *every* other file claiming the root (legacy + rival `<root>.<otherid>.json`), not just the legacy name. Test: `save_retires_stale_duplicate_siblings` (fails without the change). Gated on `stable_id.is_some()`; one `read_dir` next to the existing fsync+rename.
- **Duplicate arbitration logic** — extracted `workspace::workspace_freshness_rank`, now the single source of truth for both `discover_sessions` and `find_workspace_file_by_root`.
- **`candidate_files_for_root` swallows corrupt files** — now logs a `warn!` on an unparseable name-matching file so a damaged sole workspace no longer vanishes into a silent empty start. (Did *not* make `delete` unlink unparseable files — see below.)
- **Canonicalization asymmetry** — root match now also accepts a verbatim raw-path equality, so a one-sided symlink-resolution failure can't drop a valid file.
- **Tab-menu clamp fed two dimension sources** — the three tab-context-menu hit-tests now use `last_frame` (render-time size), matching the render path and the file-explorer menu.

**Fixed, but not the way suggested**
- **Terminal PTY orphan (top finding).** I applied the core suggestion — materialize the reused session *before* the move — on both extract paths. But I want to flag that I could **not** reproduce the orphan you described. In a faithful dormant-reuse repro (seeded persisted workspace at the terminal's cwd, discovered as a pending window), the terminal survives with *and* without the change: the discovered window is seeded (`populated == true`), so materialize takes the non-destructive `apply_workspace_layout` path and `clean_orphaned_buffers` never fires on the moved terminal (instrumented and confirmed). The discard you describe lives in the `populated == false` branch (`Window::from_workspace` replaces the window wholesale), which I couldn't drive a discovered window into. So I've kept materialize-before-move as correct defensive ordering (it does protect that branch), but I'd downgrade the severity from "orphans a live PTY today" to "hardens an ordering that's only reachable via an unseeded pending session." If you have a concrete repro that hits the `from_workspace` branch, I'll turn it into a failing test. Behavioral coverage added regardless: `extract_terminal_into_dormant_session_keeps_live_pty`.
- **Source-window leaf re-pointing diverges from close path.** Harmonized the *selection*: LRU (`focus_history` reversed), skip `hidden_from_tabs`, and honor `auto_create_empty_buffer_on_last_buffer_close` (synthetic placeholder). I deliberately did **not** route through `close_tab_in_split_silent` as suggested — that path calls `close_buffer` on the last viewport and skips modified buffers, which would destroy or refuse to move the very buffer extraction must preserve, including unsaved edits (the headline feature + `extract_tab_preserves_unsaved_edits_and_undo_history`). So the selection logic now matches; the close *mechanism* intentionally stays separate.

**Declined, with reasoning**
- **`delete` should also unlink corrupt files.** Skipped: the filename prefilter is a prefix match (`/a`'s scan name-matches `/a.b`'s files), and the only thing that disambiguates prefix-colliding roots is the parsed `working_dir`. Blindly unlinking an unparseable name-match would risk deleting another root's corrupt file. Logging it (done) is the safe half.
- **`NewTabMenu` clamp drift.** Correct, but as your own note says it's pre-existing and untouched by this PR (`NewTabMenu` has no `clamped_position`). Out of scope here — happy to do it as a follow-up that gives `NewTabMenu` the same helper.
- **Redundant re-read/re-parse on the load path.** Real, but it's an efficiency refactor on a correct path (hand back the parsed winner from `candidate_files_for_root`; gate legacy retirement behind a one-shot flag). The boot cost is bounded (one `read_dir` per window) and the refactor touches the arbitration chokepoint, so I'd rather not fold it into this PR — noting it for a focused follow-up.


---

crates/fresh-editor/src/app/window_actions.rs
sinelaw: Applied the core suggestion — materialize the reused session *before* the move — on both extract paths (14b22af). Caveat worth flagging, though: **I couldn't reproduce the orphan.** In a faithful repro (seeded persisted workspace at the terminal's cwd, discovered as a pending window) the terminal survives with *and* without the change — the discovered window is seeded (`populated == true`), so materialize takes the non-destructive `apply_workspace_layout` path and `clean_orphaned_buffers` never fires on the moved terminal (instrumented and confirmed). The discard you describe lives in the `populated == false` branch (`Window::from_workspace` replaces the window wholesale), which I couldn't drive a discovered window into.

So materialize-before-move is kept as correct defensive ordering (it does protect that branch), but I'd downgrade the severity from "orphans a live PTY today" to "hardens an ordering only reachable via an unseeded pending session." Behavioral coverage added: `extract_terminal_into_dormant_session_keeps_live_pty`. Leaving this thread open — if you have a repro that actually hits `from_workspace`, I'll turn it into a failing test.

---

crates/fresh-editor/src/workspace.rs
sinelaw: Fixed in 14b22af: `delete` is now best-effort per file — `NotFound` (the TOCTOU you describe) is ignored, other per-file errors don't abort the loop, and the first hard error is returned only after every candidate has been attempted. Test: `delete_removes_every_variant`.

---

crates/fresh-editor/src/workspace.rs
sinelaw: Fixed in 14b22af: `save` now retires *every* other file claiming the root (legacy + rival `<root>.<otherid>.json`), not just the legacy name — mirroring what `delete` enumerates. Gated on `stable_id.is_some()`; one `read_dir` next to the existing fsync+rename. Test `save_retires_stale_duplicate_siblings` fails without the change.

---

crates/fresh-editor/src/app/window_actions.rs
sinelaw: Fixed the three divergences in 14b22af: the replacement is now picked in focus-history LRU order, skips `hidden_from_tabs`, and the last-leaf seed honors `auto_create_empty_buffer_on_last_buffer_close` (marking a synthetic placeholder).

I deliberately did **not** route through `close_tab_in_split_silent` as suggested — that path calls `close_buffer` on the last viewport and skips modified buffers, which would destroy or refuse to move the very buffer extraction must preserve, including unsaved edits (`extract_tab_preserves_unsaved_edits_and_undo_history`). So the *selection* logic now matches the close path; the close *mechanism* stays intentionally separate.

---

crates/fresh-editor/src/workspace.rs
sinelaw: Half done in 14b22af: `candidate_files_for_root` now logs a `warn!` on an unparseable name-matching file, so a damaged sole workspace no longer vanishes into a silent empty start.

I did **not** make `delete` unlink unparseable files: the filename prefilter is a prefix match (`/a`'s scan name-matches `/a.b`'s files), and only the parsed `working_dir` disambiguates prefix-colliding roots — so blindly unlinking an unparseable name-match risks deleting *another* root's corrupt file. Logging is the safe half. Leaving open in case you want a safer targeted delete (e.g. only when the name is an exact non-prefix-ambiguous match).

---

crates/fresh-editor/src/app/render.rs
sinelaw: Agreed — and as you note it's pre-existing: `NewTabMenu` is untouched by this PR and has no `clamped_position`. I've kept it out to hold this PR's diff scoped, and would do it as a focused follow-up that gives `NewTabMenu` the same helper and routes its render + both hit-tests through it. Leaving open to track that follow-up.

---

crates/fresh-editor/src/app/render.rs
sinelaw: Fixed in 14b22af: the three tab-context-menu hit-tests now clamp against `active_chrome().last_frame` (the render-time size, written every draw) instead of `terminal_width/height`, matching the render path and the file-explorer menu — so render and hit-test always use one consistent source.

---

crates/fresh-editor/src/workspace.rs
sinelaw: Fixed in 14b22af: the root match now also accepts a verbatim raw-path equality (`claimed == working_dir`) alongside canonical-to-canonical, so a one-sided `canonicalize` failure can't drop a file whose recorded `working_dir` equals the lookup path exactly.

---

crates/fresh-editor/src/app/orchestrator_persistence.rs
sinelaw: Fixed in 14b22af: extracted `workspace::workspace_freshness_rank(saved_at, has_stable_id)` as the single source of truth, now called by both `discover_sessions` here and `find_workspace_file_by_root`, so the two adoption sites can't drift.

---

crates/fresh-editor/src/workspace.rs
sinelaw: Real, but deferring: this is an efficiency refactor on a correct path that reworks the arbitration chokepoint (`candidate_files_for_root` handing back the parsed winner) plus a save-time flag. The boot cost is bounded (one `read_dir` per window), so I'd rather not fold it into this PR. Noting it for a focused follow-up. (The legacy-retirement half it mentions was reworked anyway in 14b22af — that block is now the general duplicate-GC scan.) Leaving open to track.

---

crates/fresh-editor/src/workspace.rs
sinelaw: Heads-up that a later change in this PR **partially reverses** this fix. The branch now supports *multiple workspaces per root* (a tab can be extracted into its own window over the same project — `stable_id`-keyed co-tenants). Under that model a rival `<root>.<otherid>.json` is a **live peer, not a duplicate**, so `save` must not sweep it. The finding-3 GC is therefore narrowed back to retiring only the legacy id-less root-keyed file it re-keys away from; id-keyed siblings are left alone (`6ce217c`, and the test flipped to `save_retires_legacy_but_keeps_co_tenant_siblings`). The "duplicates accumulate" concern you raised is now handled per-identity instead (each window owns exactly one `<root>.<id>.json`).

---

crates/fresh-editor/src/workspace.rs
sinelaw: Related follow-up: with multiple-per-root now supported, whole-root `delete` (remove every file claiming the dir) is still correct for tearing down a project, but closing a *single* co-tenant needs `delete_by_id(root, stable_id)` (added in this PR) so it doesn't take its peers down with it. This best-effort `delete` is unchanged and stays the whole-root teardown.

---

crates/fresh-editor/src/app/window_actions.rs
sinelaw: Resolving — the co-tenant redesign removed this hazard structurally, not just defensively. Both extract paths now go through `create_co_tenant_window`, which mints a brand-new window via `create_window_with_authority` and deliberately does **not** dedup by root (multiple-per-root is the point). So extraction can never land on a pre-existing dormant shell at the terminal's cwd — the `create_window_at` root-dedup this finding depended on is no longer on the extract path. The target is always a freshly-created, scratch-seeded window with no on-disk snapshot, so the destructive `from_workspace` branch is unreachable from extraction. Coverage: `extract_terminal_tab_moves_live_terminal_to_co_tenant`.

---

crates/fresh-editor/src/app/render.rs
sinelaw: Fixed in 47eb439 after all — small and self-contained enough to fold in. `NewTabMenu` now has the same `height()`/`clamped_position()` helpers as `TabContextMenu`, and `render_new_tab_menu` plus both hit-tests (hover and `handle_new_tab_menu_click`) route through `clamped_position`, clamped against `last_frame` so render and input agree during a resize too. Regression test `plus_button_menu_near_right_edge_clicks_land_on_drawn_items` opens the popup pinned to the right edge and clicks the *drawn* "New File" — it fails on the old raw-position hit-test (the click was read as outside-the-menu and dismissed it, no buffer created) and passes now.

---

crates/fresh-editor/src/workspace.rs
sinelaw: Resolving. The resurrection half of the concern doesn't actually occur: `discover_sessions` parses every file to read its `working_dir`, so an unparseable file is skipped at boot and never comes back as a session — a lingering corrupt file is inert disk weight, not a resurrectable session. The `warn!` (14b22af) surfaces it so a damaged sole workspace no longer vanishes into a silent empty start. I'm deliberately keeping `delete` from blind-unlinking unparseable name-matches: the filename prefilter is a *prefix* match and only the parsed `working_dir` disambiguates prefix-colliding roots (`/a` name-matches `/a.b`'s files), so an untargeted unlink risks deleting a **different** root's corrupt file — a worse outcome than leaving an inert orphan. Logging is the complete safe answer here.

---

crates/fresh-editor/src/workspace.rs
sinelaw: Resolving as an explicit follow-up rather than folding it in. This is a pure efficiency refactor on an already-correct path, and it reworks the arbitration chokepoint (`candidate_files_for_root` handing back the parsed winner) plus a save-time "loaded-from-legacy" flag — more moving surface than belongs in this PR, for a boot cost that's bounded (one `read_dir` per window). Tracked separately. The legacy-retirement half it referenced was already reworked in 14b22af (that block is now the general duplicate-GC scan).

---

PR #2724, opened 2026-07-18
fix syntax highlight in markdown code block #2689

### LINKED ISSUE #2689: syntax highlight in markdown code block

asukaminato0721:
fix #2689

Adds language-aware syntax highlighting for recognized Markdown fenced code blocks. Highlight spans are cached per buffer version, while full-buffer fence detection is limited to Markdown files at or below 1 MiB; larger files retain viewport-local raw-code highlighting to preserve huge-file performance.

### DIFF
--- a/crates/fresh-editor/src/primitives/highlight_engine.rs
+++ b/crates/fresh-editor/src/primitives/highlight_engine.rs
@@ -272,6 +272,10 @@ pub struct TextMateEngine {
     last_buffer_len: usize,
     ts_language: Option<Language>,
     stats: HighlightStats,
+    // Markdown's bundled TextMate grammar marks fenced code uniformly as
+    // `markup.raw`. Cache a second, language-aware pass for recognized fences
+    // so ordinary viewport cache hits do not re-parse their contents.
+    markdown_fence_cache: Option<MarkdownFenceCache>,
     // Scope→Category memo. Syntect Scope atoms are append-only-interned
     // globally, so entries never need invalidation.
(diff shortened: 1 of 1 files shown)


sinelaw (changes requested):

- Verified interactively (tmux): fenced ```rust and ```python blocks in a markdown file now get real language-aware highlighting (keywords/numbers/strings colored correctly), vs. the previous uniform `markup.raw` styling. Confirmed edits inside a fence correctly re-highlight (cache invalidation works). No crashes.
- Blocking: `ensure_markdown_fence_cache` calls `buffer.to_string()` (reads the *entire* buffer) and re-scans the whole text for fences on every edit (both `notify_insert`/`notify_delete` unconditionally clear `markdown_fence_cache`, and it's rebuilt in full on the next render). This has no size gate, unlike every other full-file code path in this same file, which is bounded by `MAX_PARSE_BYTES` (1MB). This contradicts CONTRIBUTING.md's "Avoid full-buffer scans" guideline and would cause per-keystroke lag on large Markdown files (an explicitly supported use case per the README's huge-file handling).
- Suggest windowing the fence scan/cache to the parsed range (like `highlight_viewport_base` does) or gating it behind a `buf_len <= MAX_PARSE_BYTES`-style check, consistent with the rest of the engine.
- Minor: PR description is just "fix #2689" — a sentence on the caching approach would help reviewers.

---

sinelaw:
I think this needs a core feature for multi language highlighting, please hold off any more work on this PR.

sinelaw:
Should be fixed now by this PR, merged to master: https://github.com/sinelaw/fresh/pull/2762

PR #2728, opened 2026-07-19
feat(orchestrator): non-blocking new-workspace creation

sinelaw:
Creating a workspace from "Orchestrator: New Workspace" used to park the
whole editor behind a modal "Creating…/Connecting…" dialog with a lone
Cancel button, blocking the rest of the UI until the git worktree was
added (or the remote host connected). That made switching to other
workspaces impossible while one was coming up — and a slow or unreachable
remote host froze everything.

Make submit non-blocking: capture the form's inputs into a self-contained
CreateSpec, close the form immediately, and drop a placeholder row into the
orchestrator dock ("Creating…" for a local worktree, "Connecting…" for a
remote SSH/Kubernetes host that may not be reachable yet). The create runs
in the background; the user keeps working or switches freely. On success the
placeholder is swapped for the real window while keeping focus where the
user was ("stay put, mark ready"); on failure it becomes an error row with
Retry/Dismiss in the row's context menu.

Remote attaches are serialised because the host's cancelRemoteAgent()
cancels every in-flight connect, so two backgrounded connects couldn't be
cancelled independently.

Interrupted *local* worktree creates are persisted and replayed (paused,
resumable) on the `ready` hook, so a quit + restart doesn't silently lose an
in-progress workspace; remote sessions are left to the host's existing
dormant-session persistence to avoid duplicate rows or surprise reconnects.

Adds dock.pending_* / dock.ctx_retry / dock.ctx_dismiss locale strings to
all locales. The local non-blocking e2e is active; the SSH-connecting and
restart-recovery e2es are marked #[ignore] pending a follow-up that routes
the dock's card render path through the pending-row renderer.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_011a69Jrd4u7LHR78z17uhSF


crates/fresh-editor/plugins/orchestrator.ts
```diff
@@ -7402,27 +7753,175 @@ async function submitForm(): Promise<void> {
...
+    !s || !s.pending ||
+    (s.pending.spec.backend !== "ssh" && s.pending.spec.backend !== "kubernetes")
+  ) {
+    return;
+  }
+  const spec = s.pending.spec;
+  s.pending.phase = "creating";
+  s.pending.message = editor.t("dock.pending_connecting");
+  if (s.remote) s.remote.state = "starting";
+  const restoreTo = s.pending.originActive;
```
sinelaw: **Focus steal: remote "stay put" restores to the submit-time window, not the user's current one** (medium)

`restoreTo` reuses `s.pending.originActive`, captured once at submit time in `startPendingWorkspace`. A remote connect can run for many seconds and may additionally wait queued behind another attach — and the whole point of the non-blocking flow is that the user navigates elsewhere meanwhile. On success with `!visit`, the attach makes the born window active and then `restoreActiveWindow(restoreTo)` switches focus back to the *submit-time* window rather than wherever the user has since moved, yanking their focus (and keystrokes) away.

The local worker gets this right by recapturing `editor.activeWindow()` fresh right before its dive (~L7724). Suggest mirroring that here — capture the current active window just before `await editor.attachRemoteAgent(...)` instead of reusing `originActive` — so "stay put" returns to where the user actually is.

---

crates/fresh-editor/plugins/orchestrator.ts
```diff
@@ -7180,108 +7451,227 @@ async function submitRemoteForm(backend: SessionBackend): Promise<void> {
...
     const canonical = await resolveCanonicalRepoRoot(projectPath);
     if (canonical) repoRoot = canonical;
   }
 
-  // Session name resolution: explicit value wins. Otherwise
-  // auto-generate by scanning `refs/heads/session-N` for the
-  // next free index (the same probe that filled the
-  // placeholder).
-  const sessionName = form.name.value.trim() ||
+  const sessionName = spec.name ||
```
sinelaw, line 7669: **Retry/restart-recovery of an auto-named worktree allocates a *new* name, defeating the `rootExists` idempotency and orphaning the first worktree** (medium)

For the common auto-named case (`spec.name === ""`), the name is resolved here via `nextAutoSessionName(...)`, and the resolved name is only pushed into the row's `label` via `relabelPending` — it is never written back into `spec.name` (there's no `spec.name =` anywhere). `savePendingSpecs` (L7885) persists the spec verbatim with `name: ""`.

So if a first attempt already ran `git worktree add` (creating branch+dir `<proj>-N`) but then failed in `createWindowWithTerminal`:
- **Retry** re-enters `runLocalCreate`; because `<proj>-N` now exists, `nextAutoSessionName` returns `<proj>-(N+1)`, `root` points at a new path, `rootExists` (L7663) is `false`, and a **second** worktree is created — orphaning `<proj>-N` on disk.
- **Restart recovery** hits the same path: the persisted `name: ""` spec re-resolves to a new name, so the documented "if the target worktree already exists … the add is skipped" guarantee never fires for auto-named creates.

Only user-typed names resume idempotently. Suggest writing the resolved `sessionName` back into the persisted spec once auto-generated, so retry/resume target the same `root`.

---

crates/fresh-editor/plugins/orchestrator.ts
```diff
@@ -7180,108 +7451,227 @@ async function submitRemoteForm(backend: SessionBackend): Promise<void> {
...
+  launchPendingCreate(id);
+}
+
+// Drop a placeholder for good. If it is the remote connect currently in
+// flight, tear that connect down first (the host's `cancelRemoteAgent`
+// cancels the single in-flight attach); a queued one is just removed.
+function dismissPending(id: number): void {
+  const s = orchestratorSessions.get(id);
+  if (!s || !s.pending) return;
+  if (remoteInFlightId === id) {
```
sinelaw, line 7612: **`dismissPending` doesn't actually cancel an in-flight *local* create — worktree can be orphaned and a window can still spawn** (medium)

This teardown only handles the remote in-flight connect (`cancelRemoteAgent()`). For a *local* placeholder that is still "creating", `dismissPending` just deletes the session from the map; the already-running `runLocalCreate` promise keeps executing, and it only re-checks `pending` at coarse checkpoints (L7643, L7654, L7698).

Two concrete failure modes:
1. **Orphaned worktree** — if the dismiss lands while `git worktree add` (L7664–7696) is running, the checkpoint at L7698 returns *after* the worktree dir + branch are on disk. The session is already deleted (and, being dismissed, not persisted), so the worktree is orphaned with nothing tracking it and no cleanup.
2. **Resurrected window** — there's no `pending` guard between L7698 and `createWindowWithTerminal` (L7727). A dismiss during that final await still spawns the window and re-registers it live via `orchestratorSessions.set(winId, …)` (L7746), so a workspace the user just dismissed reappears as a running row.

Unlike the remote path, local dismissal has no real cancellation. Suggest a per-id cancellation signal checked after each await, plus removing any worktree already added (and a final `pending?` guard before the window is registered).

---

crates/fresh-editor/plugins/orchestrator.ts
```diff
@@ -7402,27 +7753,175 @@ async function submitForm(): Promise<void> {
...
+// keystroke, never an automatic filesystem touch on launch.
+function recoverPendingWorkspaces(): void {
+  const raw = editor.getGlobalState(PENDING_KEY);
+  if (!Array.isArray(raw) || raw.length === 0) return;
+  let restored = 0;
+  for (const e of raw) {
+    if (!e || typeof e !== "object") continue;
+    const spec = (e as Record<string, unknown>).spec as CreateSpec | undefined;
+    if (!spec || spec.backend !== "local") continue;
+    startPendingWorkspace(spec, { restored: true });
```
sinelaw: **Persisted `label` is discarded on recovery, so a restored row shows the stale default name** (low)

`savePendingSpecs` deliberately persists the resolved display name as `{ spec, label: s.label }` (L7885) — capturing the auto-generated / relabelled name. But `recoverPendingWorkspaces` reads only `.spec` here and never reads `.label`; `startPendingWorkspace` then derives the row label from `spec.displayLabel`, which was computed at capture time from the (empty) name field. Net effect: an interrupted, already-relabelled workspace comes back after restart showing the generic default label instead of the name the user last saw, and the persisted `label` field is effectively dead.

Suggest reading the persisted `label` (fallback to `spec.displayLabel`) and applying it to the restored row — or dropping the unused `label` from the persisted shape if that's the intent.

---

crates/fresh-editor/tests/orchestrator_pending_local.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    // Submit with the default project path (the workspace itself). Ctrl+Enter
+    // submits from the initial text field.
+    h.send_key(KeyCode::Enter, KeyModifiers::CONTROL).unwrap();
+
+    // The form is gone and the dock (its "New Task…" toolbar) is showing —
+    // the workspace now lives in the list, not behind a modal.
+    h.wait_until(|h| {
+        let s = h.screen_to_string();
+        s.contains("New Task") && !s.contains("ORCHESTRATOR :: New Workspace")
```
sinelaw: **This e2e never asserts the placeholder row — the core behavior the PR introduces** (low, test coverage)

The assertion checks `contains("New Task") && !contains("ORCHESTRATOR :: New Workspace")` plus absence of the "press Cancel to abort" text. But `"New Task"` is generic dock toolbar chrome (`dock.new_btn`) rendered whenever the dock opens for any reason — so this only proves "form closed + dock opened + no modal". It never asserts that the new workspace's `Creating…` placeholder row was actually added. A regression where submit opens the dock but drops no pending row would still pass.

Suggest also asserting the placeholder row is present — the workspace name/project label or the localized `dock.pending_creating` string — so the test fails if the non-blocking submit opens the dock without adding the row. (The siblings in `orchestrator_dock.rs` do assert row content, so this is a gap specific to this binary.)

---

sinelaw (commented):
## Code review — non-blocking new-workspace creation

Nice change — the CreateSpec-capture + placeholder-row + background-worker structure is clean, and serialising remote attaches around the host's single-cancel model is the right call. I ran a multi-dimension review (concurrency, error/cleanup, state, persistence/recovery, the Rust connection.rs changes, and i18n/tests). connection.rs and the i18n additions look solid — the locale keys are present across all locales and the carrier-reap / `let_underscore_must_use` handling is fine.

Five issues worth a look before merge, all in `orchestrator.ts` except the last. They cluster around two themes the non-blocking design newly exposes — **cancellation/recovery of a partially-created local worktree**, and **focus/label restoration after a long-running background create**:

**Medium**
1. **Remote "stay put" steals focus** — restores to the submit-time window rather than the user's current one (L7827).
2. **Auto-named worktree retry/recovery orphans the first worktree** — the resolved name is never written back to `spec.name`, so retry/restart allocate a new name and the `rootExists` idempotency never fires (L7652).
3. **`dismissPending` doesn't cancel an in-flight local create** — a worktree added after dismissal is orphaned, and the window can still spawn and re-register as live (L7614).

**Low**
4. **Persisted `label` is dropped on recovery** — restored rows show the stale default name (L7902).
5. **Local e2e never asserts the placeholder row** — passes even if submit adds no pending row (`orchestrator_pending_local.rs` L72).

Details inline. (1)–(3) are the ones I'd prioritise; the worktree-orphaning in (2)/(3) leaves filesystem state behind that nothing tracks. These are behavioral/edge-case concerns rather than blockers on the happy path.

---

sinelaw:
Thanks for the thorough review — all five addressed in `6fadaad` (2 commits):

1. **Remote focus-steal** — `runRemoteCreate` now recaptures `editor.activeWindow()` right before `attachRemoteAgent` instead of reusing `originActive` (which is removed). Mirrors the local worker.
2. **Auto-name orphan** — the resolved `nextAutoSessionName` is now pinned back into `spec.name` and persisted, so retry/restart-recovery hit the same `root` and the `rootExists` idempotency fires.
3. **Local dismiss cancellation** — `runLocalCreate` tracks the worktree it added; the post-add and post-spawn checkpoints now `git worktree remove --force` it and tear down the just-born window (SIGKILL + closeWindow) instead of orphaning the worktree / resurrecting the row.
4. **Recovery label** — `recoverPendingWorkspaces` now threads the persisted `label` through `startPendingWorkspace`.
5. **Weak local e2e** — points at a distinct non-git dir and asserts the `task_ws` row, not just `New Task` chrome.

**Reproducer status** (per CONTRIBUTING §1): (4) ships a fail-without-fix e2e reproducer (restored row must show `resumed-alias`, distinct from the spec default/project basename — the wait times out without the fix; verified). (1)–(3) are timing/filesystem edge cases with no deterministic seam in the current harness (no success-delay SSH fixture; local creates can't be paused to time a dismiss; worktree orphans aren't rendered) — validated manually via tmux (§72) and rest on your code analysis.

Separately, while tmux-scripting mouse events for validation I found a **hard crash**: a single SGR mouse sequence with a `0` coordinate (`ESC[<0;0;5M`) panics the editor via crossterm's unchecked `coord - 1` (debug builds abort; release wraps to a bogus coordinate). Filed as #2732 with a deterministic repro — out of scope for this PR.

---

crates/fresh-editor/plugins/orchestrator.ts
sinelaw: Fixed in 6fadaad — `runRemoteCreate` now recaptures `editor.activeWindow()` immediately before `attachRemoteAgent` (and the now-unused `originActive` is removed), so "stay put" returns to the user's current window. No deterministic e2e seam for a mid-connect navigation, so validated via tmux (§72): a remote background create keeps focus on the local window.

---

crates/fresh-editor/plugins/orchestrator.ts
sinelaw: Fixed in 6fadaad — the resolved `nextAutoSessionName` is now written back into `spec.name` and persisted once auto-generated, so both retry and restart-recovery target the same `root` and the `rootExists` skip fires. Validated via tmux: an auto-named worktree create lands exactly one worktree.

---

crates/fresh-editor/plugins/orchestrator.ts
sinelaw: Fixed in 6fadaad — `runLocalCreate` now tracks the worktree it added; the post-add and post-spawn checkpoints `git worktree remove --force` it and tear down the just-born window (SIGKILL the agent + `closeWindow`, after moving focus off it) instead of orphaning the worktree or resurrecting the dismissed row. The existing `orchestratorSessions.get(id)?.pending` checks serve as the per-id cancel signal.

---

crates/fresh-editor/plugins/orchestrator.ts
sinelaw: Fixed in 6fadaad — `recoverPendingWorkspaces` now reads the persisted `label` and threads it through `startPendingWorkspace` (fallback to `spec.displayLabel`). Ships a fail-without-fix reproducer in `orchestrator_pending_recovery`: the restored row must render the persisted `resumed-alias`, distinct from the spec default and project basename — the wait times out without the fix (verified).

---

crates/fresh-editor/tests/orchestrator_pending_local.rs
sinelaw: Fixed in 6fadaad — the test now points the form at a distinct non-git dir and asserts the workspace's own row (`task_ws`) is listed, not just the generic `New Task` chrome, so it fails if the non-blocking submit opens the dock without adding a row.

---

PR #2765, opened 2026-07-23
fix(settings): source theme dropdown from the live registry (#2738)

### LINKED ISSUE #2738: Theme Settings
### Edit > Settings > General > Theme

- dark
- light
- high-contrast
- nostalgia
- terminal

### View > Select Theme

- dark
- dracula
- high-contrast
- light
- nord
- nostalgia
- solarized-dark
- terminal

1. Settings do not list all available themes
2. Settings do not list the themes alphabetically

sinelaw:
## Problem (#2738)

The Settings **Theme** dropdown was schema-driven from a hardcoded static list (`ThemeName::BUILTIN_OPTIONS`), while the **Select Theme** command reads the live `ThemeRegistry`. The two inevitably drifted: Settings showed a stale, hand-maintained subset of built-ins and **never** showed user themes from `~/.config/fresh/themes`. On top of that, the `theme` config field was a frozen JSON-Schema `enum`, so any user-theme config value (a path, a `file://`/`builtin://` URI, a registry key) was rejected as invalid.

## Approach — one source of truth, no static list

Rather than keep patching the static list, the theme list is now sourced dynamically from the registry so Settings == Select Theme by construction.

**Schema — validated string with a dynamic source.** `impl JsonSchema for ThemeName` now emits a `type: "string"` carrying `x-enum-from: "$themes"` (mirroring how `default_language` uses `x-enum-from: "/languages"`) instead of a hard `enum`. Any config form accepted by `ThemeRegistry::resolve_key` (built-in names, `builtin://`, relative `.json` paths, `file://`/http(s) URIs) now validates — including user themes that don't exist at schema-generation time. `plugins/config-schema.json` was regenerated (`ThemeOptions` is now a dynamic string, no enum).

**Registry-backed dropdown.** The generic `x-enum-from` path derives options from object *keys* where display == value — insufficient for user themes, whose **display** (theme name) differs from the stored **value** (portable form). A new `ThemeRegistry::settings_theme_options()` returns `(display = name, value = portable_form)` pairs in `list()` order — the exact order and name→config-value mapping "Select Theme" uses (portable form: `builtin://NAME` for built-ins, a relative path for user themes). This is threaded into the settings render context via a new `BuildContext.theme_options` field and `SettingsState::set_theme_options`, populated from `self.theme_registry` in `open_settings`. The settings item builder resolves the reserved `$themes` source specially to build the dropdown from these pairs (no re-sorting). Legacy bare built-in names (`"dark"`) still pre-select their `builtin://dark` option.

**Retired the drift-prone list.** `ThemeName::BUILTIN_OPTIONS` and the two parity tests that guarded it are deleted — the registry is the only source of truth now.

## Tests
- `schema.rs`: `theme` now asserts it carries the `$themes` dynamic source (was: asserts no `enum_from`).
- `config.rs`: parity test replaced — asserts `ThemeOptions` is a dynamic string (not a frozen enum) and that user-theme values validate.
- New: a user theme in the registry appears in the Settings dropdown in picker order, displayed by name and valued by portable form (`loader.rs` + end-to-end through `items.rs`).

Local results (Linux): `cargo fmt --check` clean; `cargo clippy --all-features --all-targets` passes (errors: none); the config/theme/schema/settings tests pass; regenerating the schema CI-style produces a file identical to the committed one. The prior CI red was the `fmt` job reflowing the now-removed parity test.

Closes #2738

### DIFF
--- a/crates/fresh-editor/plugins/config-schema.json
+++ b/crates/fresh-editor/plugins/config-schema.json
@@ -345,9 +345,12 @@
       "type": "string",
       "enum": [
         "dark",
-        "light",
+        "dracula",
         "high-contrast",
+        "light",
+        "nord",
         "nostalgia",
+        "solarized-dark",
         "terminal"
       ]
     },
--- a/crates/fresh-editor/plugins/config-schema.json
+++ b/crates/fresh-editor/plugins/config-schema.json
@@ -343,16 +343,7 @@
     "ThemeOptions": {
       "description": "Available color themes",
       "type": "string",
-      "enum": [
-        "dark",
-        "dracula",
-        "high-contrast",
-        "light",
-        "nord",
-        "nostalgia",
-        "solarized-dark",
-        "terminal"
-      ]
+      "x-enum-from": "$themes"
     },
     "LocaleOptions": {
       "description": "UI locale (language). Use null for auto-detection from environment.",
--- a/crates/fresh-editor/src/app/settings_actions.rs
+++ b/crates/fresh-editor/src/app/settings_actions.rs
@@ -44,6 +44,21 @@ impl Editor {
                 // Snapshot plugin-registered status-bar tokens for the dual-list picker.
                 let tokens = self.status_bar_token_registry.lock().unwrap().clone();
                 state.set_status_bar_tokens(tokens);
+                // Populate the theme dropdown from the live registry — the same
+                // list "Select Theme" shows (built-ins + user themes, same
+                // order), so the two never drift (#2738).
+                let theme_options = self
+                    .theme_registry
+                    .settings_theme_options()
+                    .into_iter()
+                    .map(
+                        |(display, value)| crate::view::settings::items::ThemeOption {
+                            display,
+                            value,
+                        },
+                    )
+                    .collect();
+                state.set_theme_options(theme_options);
                 state.show();
                 self.settings_state = Some(state);
             }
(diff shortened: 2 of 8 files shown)


sinelaw:
### Follow-up: fixed empty settings dropdowns (interactive verification)

Manual tmux testing surfaced that the Settings **Theme** dropdown (and **Default Language**, and in fact every settings dropdown) rendered **empty** at runtime. Root cause was a **pre-existing bug on `master`**, not this PR: the Settings modal paints dropdowns through the shared plugin-widget path (`render_scalar_via_widget` → `collect_dropdown`), which for an *open* dropdown surfaces the options as floating screen-level pop-overs and **discards** `render_dropdown`'s inline `option_rows` (`let _ = option_rows;`). The Settings modal never draws those pop-overs, so the reserved rows stayed blank. This PR correctly wired the theme list into the settings context, but that data was then dropped by this pre-existing render path.

Fix (`crates/fresh-editor/src/view/settings/render.rs`): in the open-`Dropdown` branch, call `render_dropdown` directly, paint the returned option rows beneath the button, and build one hit-rect per visible row so mouse selection still maps. This restores **all** settings dropdowns.

Verified interactively (screenshot): the Theme dropdown now lists every built-in theme **plus a user theme** (`my-cool-theme` placed under `~/.config/fresh/themes/`), and applying it writes the portable value to `config.json`. `default_language` now lists languages too. Added regression test `open_dropdown_paints_option_rows`; `cargo test -p fresh-editor --lib` = 3016 passed; fmt/clippy clean.

Settings Theme dropdown listing built-in themes + user theme

---

PR #2768, opened 2026-07-23
fix(tabs): truncate long tab names and fix per-split scroll width (#2650)

### LINKED ISSUE #2650: Long file name may mess up list of tabs
Fresh 0.4.3
When multiple tabs are opened and file name is really long, the file name takes all the spaces. Especially when user uses split tabs (left and right)
Image

User can resize tab so the "hidden" name can be displayed.
Image

In both case, the cursor in the image bellow (idk its name :D) barely appeared. Maybe it's not working correctly when split tab horizontally. (this is the left side. The right size has no cursor / tab scroll at all).

Image

Suggestion: If possible, support mouse scroll with combo key (like ctrl + scroll or something) when mouse is on the tab list so scrolling can switch between tabs.

sinelaw:
## Root cause

Issue #2650: a single long filename broke the tab bar. Two independent defects:

1. **Unbounded tab labels.** Tab labels were used at full length with no cap, so one 151-char name consumed the entire strip and hid every other tab.
2. **Per-split scroll width bug.** `ensure_active_tab_visible` was fed `effective_tabs_width()` (the whole editor-content width), but a split's real tab bar is only as wide as its pane (`tabs_rect.width == split_area.width`). In a vertical split the scroll math ran against ~2x the real width, so it under-scrolled and the `>` overflow indicator disagreed with what was visible, sometimes hiding the active tab.

## Fix

**Part 1 — bound each tab label.** Added a shared `elided_tab_name(name, max_cols)` helper in `view/ui/tabs.rs` that truncates the name portion to a fixed 25-column cap by **display width** (not bytes, so CJK/emoji are multibyte-safe), keeping the leading characters and appending a single `…` (U+2026). Applied to the name portion in both label builders — `build_tab_spans` and `calculate_tab_widths` — so their widths stay in lockstep (a mismatch would drift hit-testing/scroll). Also stored into `TabHitArea.label` so mouse hit-testing and the web frontend match the TUI. Disambiguation numbers from `resolve_tab_names` are applied **before** elision so they aren't chopped.

**Part 2 — per-split scroll width.** Added `Window::split_tabs_width(split_id)`, which looks up the focused split's real pane width via `get_visible_buffers` (falling back to `effective_tabs_width` when the split isn't in the visible layout), and fed it to `ensure_active_tab_visible` at the switch/close/focus call sites (`active_focus.rs`, `buffer_close.rs`, `split_actions.rs` x2). The resize path in `lifecycle.rs` already used per-split widths.

Ctrl+scroll-to-switch is intentionally out of scope for this PR.

## Tests

- `elided_tab_name`: short names untouched; 151-char name bounded to the cap and ends with `…`; multibyte/CJK/emoji stays within the width cap without panicking.
- `calculate_tab_widths` and `build_tab_spans` compute matching widths for the same long-named input (the sync invariant).
- Over-long tabs are elided so the active tab always scrolls fully into view in a normal-width bar.
- `split_tabs_width` reports each pane's real width in a vertical split, narrower than the whole-editor `effective_tabs_width`.

All new and existing tab tests pass (`cargo test -p fresh-editor --lib tabs::` and the `ensure_active_tab_visible` / close-others regression tests).

Closes #2650

---

### DIFF
--- a/crates/fresh-editor/src/app/active_focus.rs
+++ b/crates/fresh-editor/src/app/active_focus.rs
@@ -162,8 +162,10 @@ impl Window {
             self.resize_visible_terminals();
         }
 
-        // Ensure the newly active tab is visible
-        let tabs_width = self.effective_tabs_width();
+        // Ensure the newly active tab is visible. Use the focused split's real
+        // pane width, not the whole-editor width, so vertical splits scroll
+        // correctly (issue #2650).
+        let tabs_width = self.split_tabs_width(active_split);
         self.ensure_active_tab_visible(active_split, buffer_id, tabs_width);
 
         if self.file_explorer_visible
--- a/crates/fresh-editor/src/app/buffer_close.rs
+++ b/crates/fresh-editor/src/app/buffer_close.rs
@@ -947,7 +947,7 @@ impl Editor {
         else {
             return;
         };
-        let tabs_width = self.active_window().effective_tabs_width();
+        let tabs_width = self.active_window().split_tabs_width(split_id);
         self.active_window_mut()
             .ensure_active_tab_visible(split_id, active_buffer, tabs_width);
     }
--- a/crates/fresh-editor/src/app/input.rs
+++ b/crates/fresh-editor/src/app/input.rs
@@ -393,6 +393,17 @@ impl Editor {
                 blocks_terminal_input: true,
             });
         }
+        // The close-split confirmation popup gets the same treatment as the
+        // other native context menus: a custom key dispatcher owns the keyboard
+        // while it's open and it blocks PTY routing.
+        if self.active_window().close_split_menu.is_some() {
+            layers.push(Layer {
+                kind: LayerKind::CloseSplitMenu,
+                owns_keyboard: true,
+                key_context: None,
+                blocks_terminal_input: true,
+            });
+        }
         // The centered widget modal (picker / new-session form / plugin
         // overlay) owns the keyboard when focused. It resolves as `Normal`
         // regardless of the underlying buffer's (possibly stale) context so
--- a/crates/fresh-editor/src/app/mod.rs
+++ b/crates/fresh-editor/src/app/mod.rs
@@ -3910,6 +3910,55 @@ mod tests {
         assert!(view_state.tab_scroll_offset <= total_width);
     }
 
+    /// Regression for sinelaw/fresh#2650 (Part 2).
+    ///
+    /// In a vertical split each pane's tab bar is only as wide as the pane,
+    /// but `ensure_active_tab_visible` used to be fed `effective_tabs_width`
+    /// (the whole editor width), so the scroll math ran against ~2x the real
+    /// width. `split_tabs_width` must report the focused split's real pane
+    /// width instead — roughly half the editor after a vertical split.
+    #[test]
+    fn split_tabs_width_reports_per_split_pane_width() {
+        let config = Config::default();
+        let (dir_context, _temp) = test_dir_context();
+        let mut editor = Editor::new(
+            config,
+            80,
+            24,
+            dir_context,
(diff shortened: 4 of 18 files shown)


sinelaw:

Verified in a headless Linux + tmux session on this branch: long tab names are truncated to ~25 columns with a trailing `…` (U+2026), neighbouring tabs stay visible, the active tab stays visible, and the per-split `<` scroll indicator uses each pane's real width (not the whole-editor width).

**Single pane** — a 141-char filename elided (`verylongfilename_abcdefg…`), neighbours (`beta.rs`, `gamma.rs`, `delta.py`, `epsilon.md`) all visible, `<` indicator shows `alpha.rs` scrolled off-left:

tabs — single pane

**Vertical split** — each ~44-col pane truncates and scrolls independently, each with its own `<` indicator:

tabs — vertical split

---

sinelaw:
### Follow-up fixes (interactive before/after)

Three issues found via manual tmux testing in a vertical split, now fixed (commit `d6ab2d9`):

**1. Missing right `>` scroll indicator** — pre-existing: the tab bar drew across the full pane width and the maximize button was painted over the indicator's cell (`>` at `area.width-4` collided with the maximize button). The PR's correct per-split scrolling just made right-overflow happen reliably, exposing it. Fixed by reserving button space and keeping scroll-math in lockstep.

_Before (no `>`):_
before — no right indicator
_After (`<` and `>` both present):_
after — right indicator present

**2. Maximize separated from close-split button** — pre-existing 1-col gap; buttons now grouped.

_Before / after:_
before — buttons separated
after — buttons grouped

**3. Conditional name cap** — the ~25-col truncation now applies **only when the split's tab bar is full**. With room, full names render untruncated; on overflow they elide with `…` + `<`/`>`.

_Full names when they fit / elided on overflow:_
full names when they fit
elided on overflow

New unit tests cover all three (`split_control_reserve_matches_button_count`, `tab_names_full_when_they_fit_and_capped_when_overflowing`, `right_indicator_renders_when_scrolled_off_right`, `no_right_indicator_at_the_right_end`). `cargo test -p fresh-editor --lib` = 3021 passed.

---

sinelaw:
Update: the `visual_regression::visual_comprehensive_b` snapshot has been re-blessed (commit `5111cfc`) — the only diff was the intended split-header button regrouping (`□ ×` → `□×`), and it now passes on all platforms.

On this commit the remaining red `test macos-latest` is **only** the known-flaky `e2e::plugins::orchestrator_new_dialog::ctrl_enter_submits_from_a_text_field` (180s timeout) — `8018 passed, 1 timed out`. That's the same pre-existing macOS-runner flake tracked separately, unrelated to this change. All non-test jobs (fmt/clippy/check/schema/doc) are green; ubuntu + windows test jobs are still running. Not re-running the flaky timeout, per policy.

---

sinelaw:
### Follow-up: split-control reorder + close-split confirmation (interactive)

The right-side control cluster is reordered to **`□ + [sep] > ×`** (maximize · new-buffer · one-column gap · right-scroll · close); the `<` left-overflow indicator stays at the left edge. A single unmaximized split keeps the original inline `+`/`<`/`>` behavior. Hit-testing stays in lockstep (`+`→new tab, `>`→right-scroll, `□`→maximize, `×`→close).

_Before_ (confusing order) / _after_ (`□ + [sep] > ×`):

before
after

Clicking `×` no longer closes immediately — it pops a two-item confirmation (**Close split** / **Cancel**) built on the editor's native context-menu; only "Close split" closes, Cancel/Esc/click-outside dismiss:

close-split confirm

New tests: tab-bar reserve + `external_controls` layout, context-menu, and 3 e2e (`split_close_confirm.rs`). (Maximize `□` is placed just left of `+` — easy to move if you'd prefer it elsewhere.)

---

PR #2772, opened 2026-07-23
fix(lsp): merge hover across all capable servers (#2635)

### LINKED ISSUE #2635: Hover queries only the first LSP server (no merge/fallback) — Tailwind class hover unreachable alongside vtsls
## Summary

Hover (`textDocument/hover`) is dispatched to only the **first** LSP server that advertises the capability, with no merge and no fallback when it returns `null`. Completion, diagnostics, and code actions already merge across all configured servers — hover should too (or at least fall through to the next server on a null result).

## Impact (multi-server setups)

Common React config: `typescript` language with `vtsls` + `@tailwindcss/language-server` (+ `biome`). `vtsls` is registered first and supports hover, so **every** hover request goes to vtsls. Hovering a Tailwind utility inside a `className="..."` string, vtsls correctly returns `null` (no symbol in a string literal), Fresh stops, and the user sees `No hover information available`. The Tailwind server — which would return the generated CSS with pixel equivalents (e.g. `padding: 1rem /* 16px */`, exactly like the VS Code Tailwind extension) — is never queried.

Net effect: you can have vtsls type-on-hover **or** Tailwind class-hover, but not both. VS Code shows whichever server has content for the position.

## Root cause

`crates/fresh-editor/src/app/lsp_requests.rs` — hover uses the single-server helper:

```rust
// with_lsp_for_buffer(): "Dispatch to the first handle that allows this feature"
let sh = lsp.handle_for_feature_mut(&language, feature)?;
Some(f(&sh.handle, &uri, &language))
```

`handle_for_feature_mut` (`crates/fresh-editor/src/services/lsp/manager.rs`) picks the first match:

```rust
self.handles.iter_mut()
    .filter(|sh| sh.handle.scope().accepts(language))
    .find(|sh| sh.feature_filter.allows(feature) && sh.has_capability(feature))
```

Merged features instead use `with_all_lsp_for_buffer_feature` / `handles_for_feature` (returns all matching handles). Hover is not on that path.

Observed (Fresh 0.4.3):
```
Requested hover at .../layout.tsx:157:54
Raw LSP hover response: Null
No hover information available
```
Only one server is consulted per hover.

## Proposed fix

Route hover through the merge path (`with_all_lsp_for_buffer_feature`) and either:
- return the first **non-null** hover (fallback), or
- concatenate non-null hovers from all servers (merge), similar to how completion results are combined.

A simple "first non-null wins" already fixes the vtsls+Tailwind case without changing behavior for single-server languages.

## Environment
- Fresh 0.4.3 (Linux / WSL2)
- Servers: vtsls, @tailwindcss/language-server, biome (lsp-proxy)

sinelaw:
## Root cause

Hover dispatched through `with_lsp_for_buffer`, which queries **only the first** server that advertises `hoverProvider`. When that server returns a null hover (e.g. vtsls on a Tailwind class), the request stops there and a second capable server's hover (e.g. Tailwind LSP) is never shown. Completion and code actions already fan out to every capable server and merge; hover did not.

## Fix

Port hover to the same fan-out-and-merge pattern:

- **State** (`app/hover.rs`): `HoverState` now tracks a `HashSet<u64>` of outstanding request ids plus an accumulator of non-null `HoverPayload`s (each carrying an optional `server_name` so per-server labels can be added later). `claim_pending` removes an id and returns the batch position — kept until the last id is claimed so every response can correlate diagnostics; `pending_is_empty` reports batch completion; `record_requests` replaces the batch and clears the accumulator.
- **Requests** (`app/lsp_requests.rs`): `request_hover` / `request_hover_at_position` fan out via `with_all_lsp_for_buffer_feature`, preallocating one id per server (mirroring `request_completion`) and advancing the id counter by the number of servers. `request_hover_at_position` keeps its `Ok(bool)` contract via `!sent_ids.is_empty()`.
- **Response** (`handle_hover_response`): accumulates each non-null contribution; reports "no hover" **exactly once** (only when all servers answered, nothing accumulated, and no diagnostic overlaps); otherwise rebuilds the popup from **all** accumulated payloads, joining multiple bodies with the existing one-row separator. The symbol range/overlay is taken from the **first** payload carrying a range, so a later rangeless server's word-boundary fallback cannot clobber it.

Decision: non-null hovers are **merged** (concatenated with the one-row separator), not sequential-fallback. Null/`{}` hovers were already normalized to empty contents in `parse_hover_response`, so no changes were needed to the async dispatch/handler layers.

## Tests

- `app/hover.rs` unit tests updated to the set-based API (multi-server batch position retention, accumulator clearing/ordering).
- New e2e suite `tests/e2e/lsp_unified_hover.rs` with three fake-server variants (`spawn_hover_null` returning `result: null`, `spawn_hover_alpha` / `spawn_hover_beta` returning distinct bodies):
  1. First server null, second server content → popup shows the second server's content (the #2635 repro).
  2. Both non-null → popup contains both bodies.
  3. All null → single "no hover" status, no popup.

Closes #2635

---

### DIFF
--- a/crates/fresh-editor/src/app/hover.rs
+++ b/crates/fresh-editor/src/app/hover.rs
@@ -17,16 +17,40 @@
 //! is pure data with no `Editor` or I/O access.
 
 use crate::view::overlay::OverlayHandle;
+use std::collections::HashSet;
 
-/// Owner of in-flight hover request and highlighted-symbol tracking.
+/// One server's contribution to a hover card. Several servers may answer a
+/// single hover request; their non-null payloads are accumulated and merged
+/// into one popup.
+#[derive(Debug, Clone)]
+pub(crate) struct HoverPayload {
+    /// Raw hover body text (markdown or plain), already normalized so that
+    /// null / empty LSP hovers never reach here.
+    pub(crate) contents: String,
+    /// Whether `contents` should be rendered as markdown.
+    pub(crate) is_markdown: bool,
+    /// LSP range of the hovered symbol, if the server provided one.
+    pub(crate) range: Option<((u32, u32), (u32, u32))>,
+    /// Originating server name. Carried so per-server labels can be added
+    /// later; unused for rendering today.
+    #[allow(dead_code)]
+    pub(crate) server_name: Option<String>,
+}
+
+/// Owner of in-flight hover request(s) and highlighted-symbol tracking.
 #[derive(Debug, Default)]
 pub(crate) struct HoverState {
-    /// LSP request id of the in-flight hover request, if any.
-    pending_request: Option<u64>,
-    /// LSP position `(line, character)` of the in-flight request. Retained
+    /// LSP request ids of the in-flight hover requests. Hover fans out to
+    /// every capable server, so multiple ids may be outstanding at once.
+    /// Used to ignore stale responses from earlier mouse moves.
+    pending_requests: HashSet<u64>,
+    /// LSP position `(line, character)` of the in-flight batch. Retained
     /// so the response handler can correlate diagnostics with the hover
     /// point and fuse them into the hover card.
     pending_position: Option<(u32, u32)>,
+    /// Non-null hover payloads collected so far for the current batch, in
+    /// arrival order. Merged into a single popup once responses are in.
+    accumulated: Vec<HoverPayload>,
     /// Byte range `(start, end)` of the currently-highlighted symbol.
     /// Used by mouse-move handlers to detect "still on same symbol" and
     /// skip re-querying.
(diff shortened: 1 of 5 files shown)


sinelaw:
### Interactive verification (colored screenshot)

Verified the fan-out on the PR binary with **two** Python LSP servers (basedpyright + pyright) attached to one file. Runtime logs show two hover requests to the same position (`request_id=13` and `request_id=14` — one per capable server); server 13 returned non-null and server 14 returned null, and the popup **still rendered** the non-null server's hover. That is exactly the #2635 scenario the old first-server-only path failed (the pre-fix binary issued a single request and showed nothing when the first server returned null).

The two-non-null *merged card* path (both servers returning different bodies joined by the separator) is covered by the added unit test (`push_payload_accumulates_in_order`) and the e2e test in this PR.

hover popup — basedpyright over compute_area

---

PR #2773, opened 2026-07-23
fix(plugin-api): openFileInSplit returns false for a dead split id (#2769)

### LINKED ISSUE #2769: `openFileInSplit` against a dead splitId returns `true` and shows nothing
## Summary
[repro.py](https://github.com/user-attachments/files/30323206/repro.py)
[init.ts](https://github.com/user-attachments/files/30323215/init.ts)

Split ids die at runtime: the user can close a split directly, and closing
a split's last tab collapses the split too. Calling
`editor.openFileInSplit(deadSplitId, path, line, col)` afterwards returns
`true` as if it succeeded, but the file is not displayed in any split. A
plugin that cached a split id (the only way to keep targeting "its" pane)
gets a success result and silently shows the user nothing.

Found in practice: an auto-open plugin kept the id of its editor pane;
after the user closed that pane's last tab, every subsequent auto-open
"succeeded" invisibly.

## Environment

- fresh 0.4.3 (Homebrew), macOS (Darwin 25.3.0)

## Reproduction

```
python3 repro.py
```

The script launches `fresh --no-restore` in a pseudo-terminal with an
isolated `HOME` whose `init.ts` (in this directory) does:

```ts
const s0 = editor.listSplits()[0].splitId;
await editor.createTerminal({ direction: "horizontal", ratio: 0.5, focus: false });
editor.closeSplit(s0);                              // s0 is now dead
const ret = editor.openFileInSplit(s0, file, 0, 0); // → true
// …but no split in listSplits() ever shows `file`
```

Exit code 0 and a `BUG REPRODUCED` line mean the call reported success
while the file is displayed nowhere.

## Expected

`openFileInSplit` returns `false` (or throws) when the target split no
longer exists, so the caller can fall back (pick another split, recreate
one, or surface an error).

## Actual

Returns `true`; `listSplits()` shows no split displaying the file. The
buffer may be created in the background, but nothing is visible and the
caller cannot distinguish this from success.

sinelaw:
## Root cause

`editor.openFileInSplit(splitId, path, line, col)` reported success even when the target split no longer existed — closed by the user, or collapsed when its last tab went away — leaving the file displayed in no split.

The culprit is `handle_open_file_in_split` in `crates/fresh-editor/src/app/plugin_commands.rs`. When `set_active_split(target_split_id)` returned `false` (the id doesn't resolve to a live leaf), the handler only `tracing::error!`d and then `return Ok(())`. `Ok(())` is the dispatcher's non-failure result, so the failure was swallowed. The handler also fell through to `open_file(...)` and, on error, returned `Ok(())` again — a path that could load a buffer visible in no split.

## Fix

- When `set_active_split` fails, `anyhow::bail!` so `handle_plugin_command` returns `Err` (the failure signal) instead of masquerading as success.
- The split is validated **before** any buffer is opened, so no orphan buffer is created on the dead-split path.
- The `open_file` error branch now propagates its error (`return Err(e)`) instead of reporting success.

## Test

`open_file_in_split_fails_for_dead_split_id` (in `tests/e2e/plugins/buffer_info_splits.rs`) mirrors the report: open a buffer, split horizontally, close the new split to make its id dangle, then dispatch `OpenFileInSplit` against that dead id. It asserts the dispatch returns `Err`, that the snapshot buffer count is unchanged (no orphan buffer leaked), and that the target file appears in no buffer.

Verified: `cargo build`, the new regression test, `cargo fmt -- --check`, and `cargo clippy` are all clean. Also reproduced end-to-end in the running editor via an `init.ts` plugin (create split → close it → `openFileInSplit` on the dead id): the editor stays intact showing the surviving split, the target file never appears, and the log records `Failed to switch to split 3` from the new failure path.

Closes #2769

---

### DIFF
--- a/crates/fresh-editor/src/app/plugin_commands.rs
+++ b/crates/fresh-editor/src/app/plugin_commands.rs
@@ -1429,7 +1429,12 @@ impl Editor {
         line: Option<usize>,
         column: Option<usize>,
     ) -> AnyhowResult<()> {
-        // Switch to the target split
+        // Validate the target split BEFORE touching any buffer state so a
+        // dead/unknown split id cannot leave an orphan buffer loaded with
+        // nothing on screen. `set_active_split` returns false when the id
+        // doesn't resolve to a live leaf (the split was closed, or its last
+        // tab collapsed it away). Surface that as an error so the failure is
+        // reported instead of masquerading as success (#2769).
         let target_split_id = LeafId(SplitId(split_id));
         if !self
             .windows
@@ -1439,13 +1444,13 @@ impl Editor {
             .set_active_split(target_split_id)
         {
             tracing::error!("Failed to switch to split {}", split_id);
-            return Ok(());
+            anyhow::bail!("openFileInSplit: split {} does not exist", split_id);
         }
 
         // Open the file in the now-active split
         if let Err(e) = self.open_file(&path) {
             tracing::error!("Failed to open file from plugin: {}", e);
-            return Ok(());
+            return Err(e);
         }
 
         // Jump to the specified location (or default to start)
--- a/crates/fresh-editor/tests/e2e/plugins/buffer_info_splits.rs
+++ b/crates/fresh-editor/tests/e2e/plugins/buffer_info_splits.rs
@@ -39,6 +39,36 @@ fn snapshot_active_split(harness: &EditorTestHarness) -> Option<usize> {
     Some(snapshot.active_split_id)
 }
 
+/// Number of buffers the plugin state snapshot knows about. Used to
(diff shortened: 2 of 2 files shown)


sinelaw:
### End-to-end verification (colored screenshot)

Drove the exact repro against the fixed build via an `init.ts` plugin (create split → `closeSplit` → `openFileInSplit` on the now-dead split id). The editor stays intact showing the surviving split, the target file never appears anywhere, and the dispatch now returns an error instead of a false success (log: `Failed to switch to split 3`).

editor intact after openFileInSplit on a dead split

---

PR #2774, opened 2026-07-23
fix(plugin-api): don't panic when setSplitRatio gets a leaf split id (#2770)

### LINKED ISSUE #2770: `setSplitRatio` on a leaf split crashes the editor
## Summary

[init.ts](https://github.com/user-attachments/files/30323266/init.ts)
[repro.py](https://github.com/user-attachments/files/30323267/repro.py)

Calling the plugin API `editor.setSplitRatio(splitId, ratio)` with the
`splitId` of an ordinary (leaf) split — e.g. the id just returned by
`createTerminal` — hits an `unreachable!` in the split manager and aborts
the whole editor process. A plugin bug or a plugin passing the only kind of
id it can obtain (plugins only ever see leaf split ids from `listSplits`
and `createTerminal`; container ids are internal) takes down the entire
session, including every embedded terminal.

## Environment

- fresh 0.4.3 (Homebrew), macOS (Darwin 25.3.0)

## Reproduction

```
python3 repro.py
```

The script launches `fresh --no-restore` in a pseudo-terminal with an
isolated `HOME` whose `init.ts` (in this directory) does only:

```ts
const t = await editor.createTerminal({ direction: "horizontal", ratio: 0.5, focus: false });
editor.setSplitRatio(t.splitId, 0.75);
```

Exit code 0 and a `BUG REPRODUCED` line mean the panic fired.

## Expected

Either the split's ratio changes, or the call returns `false` (the
documented failure mode of the `boolean`-returning split APIs). A plugin
API call should never abort the editor.

## Actual

```
thread 'main' (…) panicked at crates/fresh-editor/src/view/split.rs:1543:17:
internal error: entered unreachable code: ContainerId ContainerId(SplitId(9)) points to a leaf
   2: fresh::view::split::SplitManager::set_ratio
   3: fresh::app::plugin_dispatch::<impl fresh::app::Editor>::handle_plugin_command
   4: fresh::app::async_dispatch::<impl fresh::app::Editor>::process_async_messages
```

The process exits; the tmux pane running fresh dies with it.

## Notes on the root cause

`SplitManager::set_ratio` (`crates/fresh-editor/src/view/split.rs:1543`)
asserts its argument resolves to a *container* node and treats a leaf as
`unreachable!`. The plugin dispatch path (`handle_plugin_command`) passes
the plugin-supplied `SplitId` straight through as a `ContainerId` without
validating the node kind, so any plugin-visible split id triggers the
assertion. A leaf id should either be rejected (return `false`) or be
resolved to its parent container so the call does what plugins plainly
intend ("resize this split").

sinelaw:
## Root cause

`editor.setSplitRatio(leafSplitId, ratio)` aborted the whole editor.

Chain: `handle_plugin_command` → `handle_set_split_ratio` (`app/plugin_commands.rs`) wraps the plugin-supplied `SplitId` in a `ContainerId` and calls `SplitManager::set_ratio` (`view/split.rs`). `set_ratio` matched only `SplitNode::Split` and treated every other node kind — `Leaf`, `Grouped`, or an unknown id — as `unreachable!`, panicking the process. **Every plugin-visible split id is a leaf** (e.g. the id returned by `createTerminal`, `getActiveSplitId`, `BufferInfo.splits`), so *any* `setSplitRatio` call from a plugin crashed the editor.

The sibling `set_fixed_size` already handled a non-`Split` node gracefully via an `if let` no-op; this brings `set_ratio` in line.

## Fix

- **Hardened `SplitManager::set_ratio`**: it no longer has any `unreachable!` arm. It now returns `bool` (`#[must_use]`) — `true` when the id resolved to a resizable `Split` and the ratio was applied (still clamped to `[0.1, 0.9]`), `false` for a leaf / grouped / unknown id — and can never panic. Mirrors `set_fixed_size`'s graceful `if let`.
- **Updated all `set_ratio` callers**:
  - `handle_set_split_ratio` now returns `bool`, propagating `set_ratio`'s result: a non-resizable (leaf/unknown) id is a logged no-op that reports `false` to the plugin. This implements the **return-false** behavior called for in the issue ("either the split's ratio changes, or the call returns false") — the safe, unambiguous option.
  - The separator-drag caller in `app/mouse_input.rs` is already guarded by a preceding `get_ratio(..).is_some()` check (so the id is always a container); it binds and ignores the result.

### Note on the alternative (possible follow-up)

The issue also mentions the enhancement of resolving a leaf id to its *parent* container and actually resizing it. That is deliberately **not** implemented here: a leaf can be either child of its parent split, so which ratio a plugin means (and in which direction) is ambiguous. The return-false behavior is unambiguous and safe; resolve-to-parent can be a separate follow-up if desired.

## Tests

- Unit tests in `view::split`: `set_ratio` on a leaf id and on an unknown id both return `false` and do not panic; on a real container it returns `true`, applies the ratio, and clamps.
- e2e regression test (`tests/e2e/plugins/set_split_ratio_leaf.rs`) driving the plugin `PluginCommand::SetSplitRatio` with a leaf split id: the editor survives (no panic) and the layout is untouched, both for a lone leaf and when a real container also exists in the tree.

Build, `cargo test` (unit + e2e), `cargo fmt -- --check`, and `cargo clippy --tests` all clean for `-p fresh-editor`. Manually verified under tmux with the exact repro (`createTerminal` then `setSplitRatio(t.splitId, 0.75)`): the editor stays alive with the terminal split intact instead of aborting.

Closes #2770

---

### DIFF
--- a/crates/fresh-editor/src/app/mouse_input.rs
+++ b/crates/fresh-editor/src/app/mouse_input.rs
@@ -3301,7 +3301,11 @@ impl Editor {
                 .get_ratio(split_id.into())
                 .is_some()
             {
-                self.windows
+                // Guarded by the `get_ratio(..).is_some()` check above, so
+                // this id resolves to a resizable Split; the bool result is
+                // not actionable here (the drag can only target a container).
+                let _resized = self
+                    .windows
                     .get_mut(&self.active_window)
                     .and_then(|w| w.split_manager_mut())
                     .expect("active window must have a populated split layout")
--- a/crates/fresh-editor/src/app/plugin_commands.rs
+++ b/crates/fresh-editor/src/app/plugin_commands.rs
@@ -1048,16 +1048,32 @@ impl Editor {
         }
     }
 
-    /// Handle SetSplitRatio command
-    pub(super) fn handle_set_split_ratio(&mut self, split_id: SplitId, ratio: f32) {
+    /// Handle SetSplitRatio command.
+    ///
+    /// Returns `true` if `split_id` resolved to a resizable split
+    /// container and the ratio was applied, `false` if it pointed at a
+    /// leaf, grouped, or unknown node. Every plugin-visible split id is
+    /// a leaf, so a plugin calling `setSplitRatio` on one must be a
+    /// graceful no-op that reports failure — never a panic that aborts
+    /// the editor (issue #2770).
+    pub(super) fn handle_set_split_ratio(&mut self, split_id: SplitId, ratio: f32) -> bool {
         // Plugin sends arbitrary SplitId — convert to ContainerId at the boundary
         let container_id = ContainerId(split_id);
-        self.windows
+        let applied = self
+            .windows
             .get_mut(&self.active_window)
             .and_then(|w| w.split_manager_mut())
             .expect("active window must have a populated split layout")
             .set_ratio(container_id, ratio);
-        tracing::debug!("Set split {:?} ratio to {}", split_id, ratio);
+        if applied {
+            tracing::debug!("Set split {:?} ratio to {}", split_id, ratio);
+        } else {
+            tracing::debug!(
+                "setSplitRatio: split {:?} is not a resizable container; ignoring",
+                split_id
+            );
+        }
+        applied
     }
 
     /// Handle DistributeSplitsEvenly command
--- a/crates/fresh-editor/src/view/split.rs
+++ b/crates/fresh-editor/src/view/split.rs
@@ -1533,21 +1533,20 @@ impl SplitManager {
         }
     }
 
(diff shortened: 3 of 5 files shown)


sinelaw:
### End-to-end verification (colored screenshot)

Reproduced the exact repro against the fixed build via an `init.ts` plugin (`createTerminal` → `setSplitRatio(t.splitId, 0.75)` on the returned leaf id). With the fix, the editor **stays alive** with the terminal split intact — no `unreachable!` panic. Status line: `#2770 FIXED: setSplitRatio(leafId=3) did NOT crash -- editor ALIVE, terminal split intact`.

editor survives setSplitRatio on a leaf id

---

Issue #2610, opened 2026-07-07
Editing a large (<10MB) file with line wrap enabled is very laggy — ~1s per keystroke


sinelaw:
## Symptoms

Editing a large file with **line wrap enabled** is extremely laggy — each keystroke/edit takes roughly a second, and a CPU core spikes on every edit. The bigger the file (more lines), the worse it gets.

- Typing, backspacing, or any edit near the top of a large wrapped file stalls for ~1s per keystroke.
- **Scrolling** the same file (arrows / PageUp / mouse wheel) stays smooth — only editing is slow.
- With **line wrap off**, editing the same file is fast.
- Files above the large-file threshold (10 MB) are *not* affected — the slowness is specific to large-but-under-10MB files.

## Steps to reproduce

1. Create a large file that is under 10 MB but has many wrapping lines, e.g.:
   ```sh
   python3 - <<'PY'
   filler = "the quick brown fox jumps over the lazy dog while carrying a very long sentence that keeps going " * 2
   with open('/tmp/pathological_wrap.txt', 'w') as f:
       size = 0
       n = 0
       while size < int(9.7 * 1024 * 1024):
           line = f"line {n:07d}  {filler}"   # ~207 chars, wraps to 2-3 rows
           f.write(line + "\n"); size += len(line) + 1; n += 1
   print(n, "lines")
   PY
   ```
   (~48,000 lines, ~9.7 MB.)
2. Open it: `fresh /tmp/pathological_wrap.txt`
3. Enable line wrap: **View → Line Wrap** (or the palette: `Ctrl+P` → "Toggle Line Wrap").
4. Type some characters near the **top** of the file (or hold a key).
5. Observe: each keystroke lags ~1 second and a CPU core spikes. Scrolling (without editing) is smooth by comparison.

## Notes

- Larger line counts make it worse; smaller files are fine.
- Toggling line wrap **off** makes editing fast again.

## Environment

- Platform: Linux, terminal UI.

Issue #2608, opened 2026-07-07
Split terminal: scrolling one pane while another keeps updating is extremely slow (high CPU), worse as scrollback grows


sinelaw:
## Symptoms

With a split terminal layout where one pane keeps producing output, scrolling in a pane becomes extremely slow and the whole editor feels frozen.

- Scrolling a terminal split lags badly / the UI appears to hang.
- One CPU core sits near 100% the whole time.
- It gets dramatically worse the longer the terminal has been open — i.e. the more scrollback has accumulated. On a terminal that has been running for a while, a single frame can take multiple seconds.

## Steps to reproduce

1. Open `fresh`.
2. Open a terminal and split it so there are two terminal panes side by side.
3. In one pane, run a command that prints continuously, e.g.:
   ```sh
   while true; do date; sleep 1; done
   ```
4. Scroll back (enter scrollback view) in one of the panes while the other keeps updating.
5. Observe: scrolling is extremely slow / the editor hangs, and CPU stays high.
6. Leave the terminal running for a while (so the scrollback grows to tens of thousands of lines) and repeat — the slowness gets much worse.

## Environment

- Platform: Linux, terminal UI.

PR #2775, opened 2026-07-23
fix(terminal): capture full tall lines + grid-wrapped scroll-back with seamless live↔scrollback transition (#2649)

### LINKED ISSUE #2649: Long output doesn't display fully on new terminal
fresh 0.4.3 with ZSH  5.9 on a Debian based distro.
1. Use fresh text editor normally with a Python project
2. Create new terminal tab, drag and drop so it's in the bottom (split screen)
3. Drag and drop so the height of terminal tab is low
4. Run script that print something like "XXXXXXXX" + "A" * 3000

Expect behavior: Output terminal displays correctly
Actual behavior: only a part of output is displayed. It might be something like

AAAAAAAAA
<terminal prompt profile>
AAAAAAAAA

=> Output has no prefix XXXX.

Image

Beside that, the scroll bar doesn't work normally. If user scroll up, it's impossible to scroll down. This is a combo of clicking on upper part of this tab and scroll up. It's really weird so I can't reproduce correctly. However, clicking on other tab might make it work normally.

Image

sinelaw:
## Problem (#2649)

In the integrated terminal, printing a single logical line taller than the pane (the report used `python3 -c "print('X'*8 + 'A'*3000)"` in a short bottom split) exposed two bugs:

- **Symptom 1 — data loss.** The leading `XXXXXXXX` was unreachable at any scroll position and lost on session save ("Output has no prefix XXXX").
- **Symptom 2 — scroll corruption.** "If user scroll up, it's impossible to scroll down" — scroll-back got stuck / collapsed the output.

An earlier revision of this PR fixed symptom 2 by making scroll-back **non-wrapping** — but that traded the bug for a UX regression: the live grid soft-wraps long lines at the pane width, so entering scroll-back visibly reflowed the whole pane (every wrapped line collapsed onto one long row with horizontal overflow). This revision fixes both symptoms **and** makes the live→scroll-back transition seamless.

## Root cause

**Symptom 1.** A tall line's earlier wrapped rows scroll into the alacritty grid *history* while its terminator is still on screen. `flush_new_scrollback` deliberately never commits such an *in-progress* line (the backing file must end on a logical-line boundary), and `append_visible_screen` only emitted the *visible* rows — so those in-history rows reached scroll-back via **neither** capture path.

**Symptom 2.** The scroll math and the renderer disagreed about how scroll-back rows lay out, so scrolling up then back down got stuck. The deeper issue: the editor's only soft wrap was *word-boundary* wrap (gutter budget, hanging indent, EOL-cursor column reservation) — a layout that can never match the PTY grid's *exact-column* wrap, so no assignment of the old flags could make the view seamless **and** consistent.

## Fix

1. **Complete capture** (as before). `append_visible_screen` detects when the first visible row continues a line whose head is in grid history (`row_wraps(Line(-1))`), walks backward to the logical-line start, and *prepends* those rows so the **full** line reaches the backing file. It now returns the prepended head's row *and byte* counts (`PrependedHead`), used for anchoring (below). Session save uses the same method.

2. **Grid-wrapped scroll-back (new).** Scroll-back soft-wraps in a new *grid* mode that reproduces the live grid exactly: rows break at exact column boundaries every `cols` cells (the capture-time PTY width), mid-word, no gutter, no hanging indent, ANSI-aware (the backing file's SGR codes are zero-width). Since the backing file stores full logical lines, re-wrapping them at the grid width yields cell-for-cell the same rows the live grid showed — entering and leaving scroll-back doesn't move a single character. (This is the same model xterm.js/alacritty/kitty use internally: physical rows are exact-column re-chunks of logical lines.)

3. **One row model everywhere (keeps symptom 2 fixed).** A single grapheme walk defines grid row boundaries; the renderer (`apply_grid_wrapping_transform`), all scroll math (`count_visual_rows_for_text_grid` via `WrapConfig::grid` / `Viewport::grid_wrap`), the scrollbar (`VisualRowIndexKey.grid_wrap`), PageUp/Down byte mapping (`grid_segment_source_bytes`), and scroll-back mouse clicks (`terminal_grid_byte_at`) all reduce to it. A unit test pins the transform/count/segment three-way agreement, so renderer-vs-scroll-math drift — the original cause of stuck scrolling — is structurally impossible.

4. **Anchoring.** `sync_terminal_to_buffer` opens the view `PrependedHead::rows` visual rows into the first appended logical line, cursor on the first visible cell — the exit frame lands exactly on the live grid's row 0 and a tall in-progress line's head sits just above, reachable by scrolling up.

### Performance (#2608/#2610) — why wrapping is safe now

The old freeze came from word-wrapping the whole buffer per frame. Grid mode avoids that structurally:
- grid row counting is a single allocation-free pass (no wrap-transform allocations);
- every scroll path (wheel, PageUp/Down, Ctrl+Home/End, ensure-visible) is viewport-local — it only ever counts rows for lines at/near the viewport;
- the only whole-buffer structure, the scrollbar's `VisualRowIndex`, keeps its existing #2610 gates (exact wrapped-row scrollbar only ≤5k lines and ≤2MB; logical-line approximation beyond) and is keyed on buffer version + geometry — the scroll-back snapshot is immutable while parked, so it builds at most once per visit;
- `resolve_line_wrap_for_buffer` now hard-returns **true** (grid mode) for terminal buffers, and the per-frame healer `enforce_terminal_grid_wrap` (né `enforce_terminal_no_wrap`) keeps every terminal viewport in grid mode regardless of which generic path (global wrap toggle, restore, tab drag) touched it.

## Live verification (tmux, real PTY)

Scenario: an exact-column line (`8×Q + 260×W`), a spaced-words line (40 × `wNNxx`), and a tall line (`8×X + A-run + 8×Z`), then Ctrl+Space:

- **Before:** the scroll-back frame differed from the live frame on every content row — each wrapped line collapsed to one row (`XXXXXXXXAAAA…`).
- **After:** `diff` of live vs scroll-back content rows is **empty**; the words line stays split mid-word at the column boundary exactly like the grid; Ctrl+Home reaches `XXXXXXXX` (symptom 1); PageUp×6 → Ctrl+End returns to the identical bottom frame (symptom 2); Ctrl+Space resumes an identical live frame.

## Tests

- `term.rs` unit tests: full character count of an in-progress tall line captured; `PrependedHead` row/byte counts checked; no-op case when the head is on screen.
- `line_wrap_cache` unit tests: grid counting basics (exact fit, mid-word split, zero-width ANSI, wide chars, boundary-straddling wide char) and the transform/count/segment agreement invariant.
- `tests/e2e/terminal.rs`: the existing #2649 test (head reachable at top of scroll-back; scroll round-trip stays reachable) plus a new regression test asserting **row-for-row equality** of the live frame and the first scroll-back frame, and frame stability across a PageUp/Ctrl+End round trip.

`cargo fmt -- --check` and clippy are clean (remaining clippy warnings are pre-existing, verified against the base tree); the full unit suite (3037) passes, including the terminal (74), viewport, and line-wrap suites.

Closes #2649

### DIFF
--- a/crates/fresh-editor/src/app/action_events.rs
+++ b/crates/fresh-editor/src/app/action_events.rs
@@ -651,6 +651,10 @@ impl crate::app::window::Window {
         let active_buffer = self.active_buffer();
         let state = self.buffers.get(&active_buffer)?;
         let vs = self.buffers.splits().map(|(_, vs)| vs)?.get(&split_id)?;
+        // Terminal-grid wrap (fresh#2649): rows are exactly the grid width.
+        if vs.viewport.grid_wrap {
+            return Some(vs.viewport.grid_cols());
+        }
         let gutter = vs.viewport.gutter_width(&state.buffer);
         let wrap = WrapConfig::new(
             vs.viewport.effective_width() as usize,
--- a/crates/fresh-editor/src/app/buffer_management.rs
+++ b/crates/fresh-editor/src/app/buffer_management.rs
@@ -22,18 +22,35 @@ use super::Editor;
 impl crate::app::window::Window {
     /// Resolve the effective line_wrap setting for a buffer, considering language overrides.
     pub fn resolve_line_wrap_for_buffer(&self, buffer_id: BufferId) -> bool {
-        // Terminal buffers never wrap. Their content is column-formatted (ANSI,
-        // aligned output), so wrapping is wrong — and wrapping a large scrollback
-        // makes the scrollbar's visual-row index an O(all-lines) scan on every
-        // frame, freezing the UI (fresh#2608). Forced off here so no global
-        // toggle, language override, or per-buffer pin can turn it on.
+        // Terminal buffers always wrap — in *grid* mode (exact-column rows
+        // at the PTY width, `Viewport::grid_wrap`), so scroll-back lays out
+        // identically to the live grid and entering it never reflows
+        // (fresh#2649). Column alignment is preserved because rows break at
+        // exactly the grid width, and the fresh#2608/#2610 freeze doesn't
+        // return: grid row counting is allocation-free and viewport-local,
+        // and the whole-buffer visual-row index keeps its size gates.
+        // Forced on here so no global toggle, language override, or
+        // per-buffer pin can turn it off (which would reflow scroll-back
+        // into one-row logical lines).
         if self.is_terminal_buffer(buffer_id) {
-            return false;
+            return true;
         }
         match self.buffers.get(&buffer_id) {
             Some(state) => buffer_config_resolve::line_wrap(&state.language, self.config()),
             None => self.config().editor.line_wrap,
         }
+    }
+
+    /// Grid-wrap column count for a terminal buffer: the PTY's current
+    /// width. `None` for non-terminal buffers (or when the terminal state
+    /// is unavailable). Scroll-back entry (`sync_terminal_to_buffer`)
+    /// captures this at sync time; generic paths use it as the fallback.
+    pub(crate) fn terminal_grid_cols(&self, buffer_id: BufferId) -> Option<usize> {
+        let terminal_id = self.get_terminal_id(buffer_id)?;
+        let handle = self.terminal_manager.get(terminal_id)?;
+        let state = handle.state.lock().ok()?;
+        let (cols, _) = state.size();
+        Some(cols as usize)
     }
 
     /// Resolve page view settings for a buffer from its language config.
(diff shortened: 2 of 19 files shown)


sinelaw:

Repro: short bottom-split terminal, `print('X'*8 + 'A'*8000)` (one logical line far taller than the pane).

- **Symptom 1 (data loss) fixed** — the off-screen wrapped rows of the in-progress line are now captured, so the leading `XXXXXXXX` is reachable at the top of scrollback (and the full 8018-byte line is written to the backing file, so session save preserves it):

scrollback top — prefix reachable

- **Symptom 2 (scroll collapse) fixed** — PageUp to top, then PageDown + Ctrl+End returns to a stable bottom frame; round-trips cleanly:

scrollback bottom — stable frame

**Design note:** a full soft-wrap reflow was intentionally *not* done — terminal scrollback is deliberately non-wrapping (column fidelity + avoiding the O(all-lines) per-frame rebuild from #2608/#2610). Symptom 2 was caused by `sync_terminal_to_buffer` enabling line-wrap (disagreeing with the non-wrapping renderer); the fix makes them consistent. Complete-capture + non-wrap consistency is the design-consistent robust fix.

---

sinelaw:
Note on the red `test macos-latest` check: it failed **only** on a known-flaky test — `8011 passed, 1 timed out`, and the timeout is `e2e::plugins::orchestrator_new_dialog::ctrl_enter_submits_from_a_text_field` (180s cap). That test is one of a cluster of `e2e::plugins::*` tests that intermittently time out on the macOS runner (a pre-existing infra flake being investigated separately), and it is unrelated to this change — the terminal e2e suite and all other tests pass here. Not re-running, per the flaky-test policy.

---

PR #2779, opened 2026-07-24
fix(plugin-api): resolve leaf split id to parent container in setSplitRatio (#2770)

sinelaw:
## The gap

PR #2774 (merged) made `editor.setSplitRatio(splitId, ratio)` crash-safe by turning `SplitManager::set_ratio` into a no-op that returns `false` for any non-container id. That stopped the panic — but it left `setSplitRatio` doing nothing useful:

- Every split id a plugin can obtain is a **leaf** id. `getActiveSplitId`, `listSplits`, `BufferInfo.splits`, and `createTerminal` all return leaf ids.
- `set_ratio` only mutates a `SplitNode::Split` **container**.

So a plugin's `setSplitRatio` call always landed on a leaf and always no-op'd — **plugins could never resize a split.** #2774's own body flagged this by deferring the "resolve-to-parent" step. This is that follow-up.

## The fix

`handle_set_split_ratio` (`crates/fresh-editor/src/app/plugin_commands.rs`) now resolves the plugin's leaf id to its parent container via the existing `SplitManager::parent_container_of`, and sets that container's ratio:

- **leaf id with a resizable parent `Split`** → set the parent's ratio (clamped to `[0.1, 0.9]`), report `true`;
- **container id** (rare for plugins) → set directly (legacy behavior preserved by trying the id as a container first);
- **lone top-level leaf** (no parent) or **unknown id** → no-op, `false`. Never panics.

### Ratio orientation
A parent `Split`'s `ratio` is the fraction of space given to its **first** child (larger ratio ⇒ bigger first child, smaller second child). A plugin asking to resize "its" leaf resizes that parent split; whichever side the leaf sits on, adjusting the parent ratio is the single, well-defined knob for that divider. This is documented in the handler and in the plugin-facing `setSplitRatio` doc comment (`quickjs_backend.rs`).

The fire-and-forget JS return contract is **unchanged** — `setSplitRatio` still returns whether the command was *queued*, not the resize result.

## Verification

**Tests** (`cargo test -p fresh-editor --features plugins`, all green):
- Unit tests in `view::split`: a leaf id resolves to its parent + applies + clamps; a lone top-level leaf resolves to `None`; container ids still work.
- E2E tests driving `PluginCommand::SetSplitRatio` with a real **leaf** id: the parent container's ratio actually changes (`0.5` → `0.8`), an out-of-range value clamps to `0.9`, and a lone leaf is a graceful no-op (no panic).

`cargo fmt -- --check` clean; clippy clean for the changed lines.

**Interactive:** an `init.ts` calling `createTerminal({ direction: "vertical", ratio: 0.5 })` then `setSplitRatio(activeLeafId, 0.75)`:
- **before** (origin/master): divider stays at 50/50 — the no-op.
- **after** (this branch): the divider moves to 75/25 — the leaf resolved to its parent and the ratio applied.

Follow-up to #2774 (#2770).

---

### DIFF
--- a/crates/fresh-editor/src/app/plugin_commands.rs
+++ b/crates/fresh-editor/src/app/plugin_commands.rs
@@ -1050,26 +1050,50 @@ impl Editor {
 
     /// Handle SetSplitRatio command.
     ///
-    /// Returns `true` if `split_id` resolved to a resizable split
-    /// container and the ratio was applied, `false` if it pointed at a
-    /// leaf, grouped, or unknown node. Every plugin-visible split id is
-    /// a leaf, so a plugin calling `setSplitRatio` on one must be a
-    /// graceful no-op that reports failure — never a panic that aborts
-    /// the editor (issue #2770).
+    /// Every split id a plugin can obtain is a *leaf* id (`getActiveSplitId`,
+    /// `listSplits`, `BufferInfo.splits`, `createTerminal` all return leaf
+    /// ids), but only a `SplitNode::Split` *container* carries a resizable
+    /// `ratio`. So we resolve the leaf to its parent container and set the
+    /// parent's ratio — that is exactly "resize the pane the plugin is
+    /// pointing at" (it moves the divider between that leaf and its sibling).
+    ///
+    /// Ratio orientation: a parent `Split`'s `ratio` is the fraction of space
+    /// given to its FIRST child (larger ratio => bigger first child, smaller
+    /// second child). A plugin asking to resize "its" leaf resizes that
+    /// parent split; whether the leaf is the first or second child, adjusting
+    /// the parent ratio is the single, well-defined knob for that divider, and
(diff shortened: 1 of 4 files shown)


sinelaw:

Repro via an `init.ts`: `createTerminal({direction:"vertical", ratio:0.5})` then `setSplitRatio(getActiveSplitId(), 0.75)` — i.e. calling `setSplitRatio` with a **leaf** id, which is all a plugin can obtain.

**Before** (origin/master binary) — the call is a no-op, divider stays at 50/50:

before — 50/50, no-op

**After** (this PR) — the leaf id resolves to its parent container and the ratio applies, divider moves to ~75/25:

after — 75/25, applied

A lone top-level leaf or unknown id resolves to no parent → graceful `false`, never a panic (covered by unit + e2e tests).

---

PR #2780, opened 2026-07-24
fix(keybindings): user Alt+letter bindings take precedence over menu mnemonics (#2720)

### LINKED ISSUE #2720: Go to symbol not working when bound to alt+h
- Go to symbol works when called from the cmd pallet
- alt+h is received in the key binding dialog
- alt+h is unbound from the help menu trigger
- alt+h does nothing in the editor (no "go to symbol dialog")

thx

sinelaw:
## Problem

Binding an Alt+letter chord to a command did nothing. With a user config of `alt+h -> command_palette`, pressing Alt+H opened the **Help** menu when menu-bar mnemonics were on, and did **nothing** when they were off — even though the command palette invoked the command fine and the keybinding dialog showed the binding (issue #2720). This reproduces with any bound action (LSP-independent), confirming the fault is in the keybinding precedence layer, not the command or LSP.

## Root cause

The built-in menu-bar mnemonics are not a separate intercept path — they live in the default keymap as **global** bindings:

```json
// crates/fresh-editor/keymaps/default.json
{ "key": "h", "modifiers": ["alt"], "action": "menu_open", "args": {"name": "Help"}, "when": "global" }
```

A user keybinding with no `when` clause loads into the **Normal** context. In `KeybindingResolver::resolve` (`crates/fresh-editor/src/input/keybindings.rs`) the tier order was:

1. custom global
2. **default global**  ← the `alt+h -> menu_open Help` mnemonic matched here
3. custom context (Normal)  ← the user's `alt+h -> command_palette` never reached

So the default-global mnemonic shadowed the user's Normal override. With mnemonics on, the resolved `MenuOpen("Help")` opened the Help menu; with mnemonics off, `MenuOpen` is a no-op in `handle_action`, hence "nothing happens". All key dispatch funnels through `resolve`, so this single ordering bug produced both reported symptoms.

## Fix

Reorder `resolve` so **every custom (user) source is checked before any built-in default source**. Precedence rule enforced:

```
custom global -> custom context -> custom parent-context
  -> default global -> default context -> plugin defaults -> default parent-context
  -> Normal fallthrough
```

A user binding for a chord now always wins over the default keymap (including the menu mnemonics) for the same key+mods. Because a custom-context entry only exists when the user explicitly added it, an un-overridden Alt+letter still resolves to its default mnemonic — the default menu UX is unchanged.

## Verification

- Custom `alt+h -> command_palette` (mnemonics on, the default): Alt+H now opens the command palette.
- No override: Alt+H still opens the Help menu; Alt+F still opens File — no regression.
- `cargo test -p fresh-editor --lib`: 3034 passed, 0 failed, including two new regression tests (`test_user_alt_letter_binding_overrides_menu_mnemonic`, `test_user_global_binding_overrides_default_global`).
- `cargo fmt -- --check` clean; `cargo clippy` reports no new warnings for the changed file.

Closes #2720

---

### DIFF
--- a/crates/fresh-editor/src/input/keybindings.rs
+++ b/crates/fresh-editor/src/input/keybindings.rs
@@ -2003,7 +2003,21 @@ impl KeybindingResolver {
             context
         );
 
-        // Check Global bindings first (highest priority - work in all contexts)
+        // === User (custom) bindings take precedence over ALL built-in defaults ===
+        //
+        // A keybinding present in the user's config must win over the built-in
+        // keymap for the same chord — including default *global* entries such as
+        // the menu-bar Alt-letter mnemonics (`Alt+H → menu_open Help`). We
+        // therefore probe every custom source (global, then the active context,
+        // then its parent context) BEFORE any default source. Previously the
+        // default-global tier was checked between custom-global and
+        // custom-context, so a user's e.g. `alt+h → command_palette` (which
+        // loads into the Normal context when it has no `when` clause) was
+        // shadowed by the default-global mnemonic and never reached dispatch —
+        // it either opened the Help menu (mnemonics on) or silently no-op'd
+        // (mnemonics off). See issue #2720.
+
+        // Custom Global bindings (highest priority — work in all contexts).
         if let Some(global_bindings) = self.bindings.get(&KeyContext::Global) {
             if let Some(action) = global_bindings.get(norm) {
                 tracing::trace!("  -> Found in custom global bindings: {:?}", action);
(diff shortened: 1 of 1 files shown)


sinelaw:

Config: `alt+h → command_palette` (LSP-independent, isolates the keybinding layer).

**Before** — the default-global `alt+h → menu_open Help` mnemonic shadows the user binding, so Alt+H opens the **Help** menu:

before — Alt+H opens Help

**After** — custom sources now resolve before default sources, so Alt+H fires the user binding and opens the **command palette**:

after — Alt+H opens the command palette

Un-overridden Alt+letters are unaffected (Alt+H still opens Help by default, Alt+F still opens File), confirmed interactively and by the two new regression tests.

---

PR #2782, opened 2026-07-24
Support Classic-Mac (CR) line endings (fixes #2736)

### LINKED ISSUE #2736: Classic Mac Line Endings
Fresh can not handle Classic Mac Line Endings (CR).

### New file

1. Set Default Line Ending to CR
2. Create a new file
3. Enter text
4. Save the file
5. The line endings are LF

### Existing file

1. printf '\r' > cr.txt
2. fresh cr.txt
3. Fresh displays <0D>
4. Press ENTER
5. Now you see <0D><0D>
6. Save the file
7. Exit Fresh
8. xxd cr.txt = 0d 0d 0a

sinelaw:
## Summary

Makes Classic-Mac carriage-return (`CR`, `0x0D`) a first-class line-ending mode in the editor — a full peer of `LF` and `CRLF` for detection, display, editing, and save. Fixes #2736.

## Root cause

The editor's in-memory line model splits lines on `\n` only (the piece tree indexes line feeds). `LF` and `CRLF` both contain `\n`, so they split correctly (`CRLF` keeps its raw `\r` as content, stripped at render). A `CR`-only file has **no** `\n`, so two things broke:

1. **Existing CR file did not split.** `TextBuffer::from_bytes` (`crates/fresh-editor/src/model/buffer/mod.rs`) stored the raw `\r` bytes; with no `\n`, the whole file was one row rendered as literal `<0D>` glyphs. Pressing Enter inserted the line ending via `LineEnding::as_str()` — which for `CR` is `\r` — so it added another `<0D>` to the same row instead of creating a new one (`crates/fresh-editor/src/input/actions.rs`).
2. **New file ignored `default_line_ending: cr`.** When opening a non-existent path, `open_file_no_focus_inner` created the empty buffer with `EditorState::new_with_path` but never applied the configured default, so the buffer kept the hard-coded `LF` default and saved `0a` separators (`crates/fresh-editor/src/app/file_open_orchestrators.rs`).

## Fix

The internal model stays `\n`-based; `CR` is normalized on load and reconstructed on save (the same round-trip strategy already used for line-ending conversion).

- **Load:** a `CR`-detected file has its `\r` separators normalized to `\n` in `from_bytes` / `from_bytes_with_encoding`; the large-file lazy path falls back to a full normalizing load for `CR`. The `CR` ending is remembered in `BufferFormat`.
- **Display:** with no literal `\r` left in a `CR` buffer, each segment is its own row and no `<0D>` glyph appears.
- **Edit:** new `LineEnding::insertion_str()` returns the byte(s) to insert into the in-memory buffer — `\n` for both `LF` and `CR`, `\r\n` for `CRLF`. All Enter/open-line/move-line/paste-gap insertion sites now use it, so Enter in a `CR` buffer creates a real row.
- **Save:** a `CR` buffer always converts its internal `\n` back to `\r` (internal storage is `\n`, so the conversion is required even when the ending is unchanged since load). `LF`/`CRLF` save paths are untouched.
- **New file:** the non-existent-file branch now applies `editor.default_line_ending`, so a new file with `cr` saves `0d` separators.

`LF` and `CRLF` behavior is unchanged (regression-guarded by a new test and the existing `migrated_crlf_rendering` suite). A `CR`-detected file that mixes in stray `CRLF`/`LF` is normalized wholesale to the `CR` mode on load, consistent with how the save-time converter already collapses mixed endings.

## Verification

- New unit tests in `model/buffer/tests.rs`: CR detection, CR split on load, blank-row splitting for `\r\r`, `insertion_str` vs `as_str`, CR save round-trip, Enter-inserts-row, new-buffer `default_line_ending=cr` saves `\r`, and an LF/CRLF-unchanged regression guard.
- New e2e/semantic tests in `tests/semantic/migrated_crlf_rendering.rs`: CR file splits into rows with no `<0D>`, Down advances across CR rows, and Enter + Ctrl+S round-trips `\r`.
- Manual (tmux, real `fresh` binary): opening a `Line 1\rLine 2\rLine 3\r` file now renders three separate rows with `CR` in the status bar and no `<0D>` glyphs (before the fix the same file rendered as `1 │ Line 1<0D>Line 2<0D>Line 3<0D>`); Enter creates a new row, and after typing + Ctrl+S the on-disk file is still CR-separated (`od -c` shows `Line 1\rLine 2\rLine 3\rLine 4\r`, no `\n`).
- `cargo fmt -- --check`, `cargo clippy -p fresh-editor` (no new warnings in touched files), buffer unit tests (139 passed) and `migrated_c*` semantic tests (61 passed) all clean.

Closes #2736

---

### DIFF
--- a/crates/fresh-editor/src/app/clipboard.rs
+++ b/crates/fresh-editor/src/app/clipboard.rs
@@ -1273,7 +1273,7 @@ impl Editor {
         let vs_mode = self.active_state().buffer_settings.virtual_space;
         let mut cursor_data: Vec<_> = {
             let state = self.active_state();
-            let line_ending = state.buffer.line_ending().as_str();
+            let line_ending = state.buffer.line_ending().insertion_str();
             self.active_cursors()
                 .iter()
                 .map(|(cursor_id, cursor)| {
--- a/crates/fresh-editor/src/app/file_open_orchestrators.rs
+++ b/crates/fresh-editor/src/app/file_open_orchestrators.rs
@@ -1050,7 +1050,7 @@ impl crate::app::window::Window {
         let mut state = if file_exists {
             // Load from canonical path (for I/O and dedup), detect language from
             // display path (for glob pattern matching against user-visible names).
-            let buffer = crate::model::buffer::Buffer::load_from_file(
+            let buffer = crate::model::buffer::Buffer::load_from_file_for_editing(
                 &canonical_path,
                 self.resources.config.editor.large_file_threshold_bytes as usize,
                 Arc::clone(&self.authority().filesystem),
--- a/crates/fresh-editor/src/app/file_open_orchestrators.rs
+++ b/crates/fresh-editor/src/app/file_open_orchestrators.rs
@@ -1067,11 +1067,23 @@ impl crate::app::window::Window {
             EditorState::from_buffer_with_language(buffer, detected)
         } else {
             // File doesn't exist - create empty buffer with the file path set
-            EditorState::new_with_path(
+            let mut state = EditorState::new_with_path(
                 self.resources.config.editor.large_file_threshold_bytes as usize,
                 Arc::clone(&self.authority().filesystem),
                 path.to_path_buf(),
-            )
+            );
+            // Honor the configured default line ending for the file the
+            // user is about to create — without this the buffer keeps the
+            // hard-coded LF default and saving a brand-new file ignored
+            // `editor.default_line_ending` (issue #2736).
+            state.buffer.set_default_line_ending(
+                self.resources
+                    .config
+                    .editor
+                    .default_line_ending
+                    .to_line_ending(),
+            );
+            state
         };
         // Note: line_wrap_enabled is set on SplitViewState.viewport when the split is created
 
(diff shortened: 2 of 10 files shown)


sinelaw:
### Interactive verification

(Text capture rather than a colored PNG this time — the worktree was missing the gitignored `web-ui/` dir that the `--bin fresh` build script needs; verification used the built binary + `od`.)

**Before** — a CR-delimited file (`Line 1\rLine 2\rLine 3\r`) opens as a single row with literal `<0D>` glyphs:
```
  1 │ Line 1<0D>Line 2<0D>Line 3<0D>
 Trusted  Local  Ln 1, Col 22   Opened classic_mac.txt   CR  ASCII  Text
```

**After** — the same file opens split into real rows; pressing Enter breaks lines, and Ctrl+S round-trips as CR:
```
  1 │ Line 1
  2 │ Line 2
  3 │ Line 3
  4 │ Line 4
  5 │
```
`od -c` after a live edit + save: `Line 1\rLine 2\rLine 3\rLine 4\r` — CR separators, no `\n`. A new file created with `editor.default_line_ending: cr` also saves CR separators (previously ignored → LF).

---

PR #2783, opened 2026-07-24
fix(splits): constrain splits by absolute minimum pane size instead of a fixed ratio percentage

sinelaw:
## Problem

Split container ratios were clamped to a fixed `[0.1, 0.9]` percentage everywhere — split creation, separator drag, keyboard resize, plugin `setSplitRatio`, and distribute-evenly. A percentage floor is the wrong constraint:

- 10% of a huge pane is plenty of space, while 10% of a narrow terminal is only a few columns — too few to render a tab bar or any content.
- It arbitrarily blocks legitimately-small panes even when there is plenty of room to make one.

A plugin asking for `0.99` didn't get "leave the sibling a minimum usable size"; it got a fixed 90%, which on a narrow terminal is still an unusable sliver.

## What changed

Replace the percentage clamp with an **absolute minimum pane size enforced at layout time**.

- Added `MIN_PANE_WIDTH = 10` columns and `MIN_PANE_HEIGHT = 3` rows. A vertical split's separator eats one column and each child needs room for a truncated tab label plus a little content (≈10 cols); a horizontal pane spends one row on its tab bar, so three rows leaves the tab bar plus two content rows — the floor for a pane to be usable rather than just a header.
- `split_rect_ext` now converts the ratio to cells and then pins the split via a new `clamp_first_to_min` helper so that **neither child renders below the minimum** — whatever the first child takes, the sibling always keeps at least the minimum. This is the real guard now.
- **Small-container fallback:** when the container is too small to give both children the minimum (`available < 2 * min`), there is no size that satisfies the constraint, so it degrades to an **even split** rather than collapsing or hiding a pane. Never panics (`total == 0` yields `0`).
- Relaxed the stored-ratio clamps to `[0.0, 1.0]` at all five sites (split creation, `adjust_ratio`, `set_ratio`, `distribute_splits_evenly`, and the mouse-drag path). A plugin asking for `0.99` now **keeps `0.99`**, and the layout leaves the sibling `≥ min` instead of pinning it to a fixed 90%. This applies through the `setSplitRatio` leaf→parent-container resolution added in #2779 as well.

`set_ratio` keeps its `bool` return contract; `get_ratio` round-trips the raw stored ratio.

## Verification

Same action on a 44-column terminal — create a vertical split, then keyboard-resize the divider to the extreme:

- **Before** (`[0.1, 0.9]` clamp): the right pane is pinned to ~4 columns — the `s.txt` tab label and all content are gone, only the close icons remain. Unusable.
- **After** (min-size clamp): the right pane stops at `MIN_PANE_WIDTH` (~10 columns) — the `s.txt` tab, gutter, and (wrapped) content stay visible. Usable.

Tests:
- Unit: `clamp_first_to_min` pins both children; small-container fallback splits evenly; layout never renders a child below min width/height even at ratio `0.99`/`0.0`.
- Updated `test_set_ratio_on_container_applies_and_clamps` and `test_leaf_resolves_to_parent_container_for_ratio` to the new raw-storage semantics (`0.99` stays `0.99`, out-of-range pinned to `[0,1]`).
- Rewrote the `test_split_separator_drag_respects_limits` e2e test: after an extreme drag, the rendered sibling pane keeps at least `MIN_PANE_HEIGHT` rows (rather than asserting the old 0.1/0.9 ratio bounds).
- `cargo test -p fresh-editor` (split/layout focus), `cargo fmt --check`, and `cargo clippy` are clean for the changed files.

---

### DIFF
--- a/crates/fresh-editor/src/app/buffer_groups.rs
+++ b/crates/fresh-editor/src/app/buffer_groups.rs
@@ -856,7 +856,10 @@ impl crate::app::window::Window {
         use crate::view::split::SplitNode;
         for node in self.grouped_subtrees.values_mut() {
             if let Some(SplitNode::Split { ratio, .. }) = node.find_mut(container.into()) {
-                *ratio = new_ratio.clamp(0.1, 0.9);
+                // Raw-storage clamp only: the sibling is kept usable by the
+                // layout-time min-pane-size guard (see `clamp_first_to_min`),
+                // matching `SplitManager::set_ratio` — not a fixed percentage floor.
+                *ratio = new_ratio.clamp(0.0, 1.0);
                 return true;
             }
         }
--- a/crates/fresh-editor/src/app/mouse_input.rs
+++ b/crates/fresh-editor/src/app/mouse_input.rs
@@ -3286,7 +3286,11 @@ impl Editor {
         // The ratio represents the fraction of space the first split gets
         if total_size > 0 {
             let ratio_delta = delta as f32 / total_size as f32;
-            let new_ratio = (start_ratio + ratio_delta).clamp(0.1, 0.9);
+            // Store the raw fraction; the absolute minimum-pane-size guard is
+            // enforced at layout time, so dragging the separator toward the
+            // edge stops when the sibling would drop below the minimum size
+            // rather than at a fixed 10%/90%.
+            let new_ratio = (start_ratio + ratio_delta).clamp(0.0, 1.0);
 
             // Update the split ratio. The container may live in the main
             // split tree or inside a stashed Grouped subtree (buffer group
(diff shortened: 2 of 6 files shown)


sinelaw:

44-col terminal, vertical split, resized toward the extreme:

**Before** (percent clamp) — `0.9` pins the right pane to ~4 cols: no tab label, no content, just the close icons (unusable):

before — pane pinned to ~4 cols

**After** (absolute min, `MIN_PANE_WIDTH=10`) — the pane stops at ~10 cols with its `s.txt` tab, gutter and wrapped content visible (divider moved from col 39 → col 33 = `total − min`):

after — pane stops at ~10 usable cols

The min is enforced at layout time in `split_rect_ext` (the single ratio→rect funnel), so drag, keyboard resize, plugin `setSplitRatio` (via #2779's leaf→parent), split creation, and distribute-evenly all honor it. When the container can't fit two minimums it degrades to an even split. Stored ratios are now raw `[0,1]`; `fixed_first/second` (sidebars/dock) paths are untouched.

---

PR #2785, opened 2026-07-24
fix(tabs): keep the split "+" button after the last tab, not pinned right (follow-up to #2768)

sinelaw:
## Regression

The split tab-bar control-cluster work in #2768 pinned the new-buffer `+` button into the fixed right-side cluster (`+ [sep] > □ ×`), so `+` was **always on the far right** regardless of how many tabs were open. Previously `+` sat **right after the last tab** (and still does in a single, unmaximized split).

## Fix

- `+` is no longer part of the right cluster — the tab renderer (`render_for_split`) draws it **inline, immediately after the last visible tab**, exactly as in a single split.
- The right cluster is reduced to **`> □ ×`** (right-overflow indicator, maximize, close-with-confirm). `<` stays at the left edge.
- `split_control_reserve` shrunk to match (no more `+`/separator columns); the per-split scroll-width math, the `>` overflow indicator, the maximize/close buttons + close-split confirm, and hit-testing (`new_tab_area` now tracks the inline `+`) are preserved.

Files: `view/ui/split_rendering/orchestration/mod.rs`, `view/ui/tabs.rs`, `tests/e2e/split_close_confirm.rs`.

Follow-up to #2768 (#2650). Interactive tmux verification + screenshots to follow in a comment.

---

### DIFF
--- a/crates/fresh-editor/src/view/ui/split_rendering/orchestration/mod.rs
+++ b/crates/fresh-editor/src/view/ui/split_rendering/orchestration/mod.rs
@@ -760,10 +760,11 @@ fn render_split_tab_bar(
     //   Close: shown when multiple splits exist AND not maximized.
     let show_maximize_btn = has_multiple_splits || is_maximized;
     let show_close_btn = has_multiple_splits && !is_maximized;
-    // When a split has any control button the whole right cluster
-    // (`+ [sep] > □ ×`) is owned here (drawn on top after the tabs); the tab
-    // renderer then skips its own `+`/`>`. A single, unmaximized split has no
-    // cluster and lets the tab renderer draw its own inline `+`/`>` (fresh#2768).
+    // When a split has any control button the right cluster (`> □ ×`) is owned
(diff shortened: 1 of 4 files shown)


sinelaw:

Built the branch and drove it in tmux — vertical split with 4 tabs.

**`+` now sits right after the last tab** (`alpha.rs × … delta.rs × +`), with the `□×` control cluster on the far right (the `>` only appears on overflow). Previously the `+` was pinned next to `□×`.

after — + right after the last tab

**Overflow case** (narrowed pane) — `+` follows the last *visible* tab (`delta. +`), with the `<`/`>` overflow indicators and the `□×` cluster still on the right:

overflow — + follows last visible tab

Verified: `+` opens a new buffer, `□` maximizes, `×` pops the close-split confirm. Locally: tabs lib tests (20) pass, visual_regression + split_close_confirm (7) pass after re-blessing `visual_comprehensive_b` for the new `+` position.

---

PR #2791, opened 2026-07-24
git-gutter plugin: force spawned git to use built-in diff format

### LINKED ISSUE #2721: show green bar for changed line when use ext diff
this is how vscode works.

Image

asukaminato0721:
fix #2721

### DIFF
--- a/crates/fresh-editor/plugins/git_gutter.ts
+++ b/crates/fresh-editor/plugins/git_gutter.ts
@@ -207,9 +207,14 @@ async function isGitTracked(filePath: string): Promise<boolean> {
 async function getGitDiff(filePath: string): Promise<string> {
   const cwd = getFileDirectory(filePath);
 
-  // Diff against HEAD to show all changes (staged + unstaged) vs last commit
+  // Diff against HEAD to show all changes (staged + unstaged) vs last commit.
+  // Force Git's native unified format: user-configured external diff/textconv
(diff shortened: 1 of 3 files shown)


sinelaw (changes requested):

- **Root cause fix is correct and manually verified.** Built the branch, configured `git config diff.external <script>` in a scratch repo (simulating delta/difft per the linked issue), and ran the real `fresh` binary in tmux: without `--no-ext-diff`/`--no-textconv`, `git_gutter.ts` gets unparseable output and shows no indicator; with this PR's flags it correctly shows the orange "modified" bar on the changed line.
- **CONTRIBUTING.md #1 ("Reproduce Before Claiming") is not satisfied**: the added assertions in `test_git_gutter_updates_after_save` don't configure any `diff.external`/`core.pager`, so the test exercises the already-working plain-git-diff path — it would pass identically with or without this fix's flags. It doesn't regression-protect the actual bug being fixed.
- Please add a test that sets `diff.external`/`core.pager` in the `GitTestRepo` (mirroring the repro in the issue: delta/difft as difftool) and asserts the gutter indicator appears — this should fail without the `--no-ext-diff --no-textconv` change and pass with it.
- Everything else (scope, diff size, style) looks good — minimal, single-purpose fix, no duplication or architecture concerns.

---

sinelaw:
👍

PR #2823, opened 2026-07-27
fix(input): decode SGR horizontal wheel reports

ttenneb:
## Summary

xterm SGR mouse wheel button codes 66 and 67 denote horizontal wheel left/right. Fresh currently aliases them to up/down by inspecting only the low bit.

Decode wheel values 64–67 by their low two button bits:

- 64: `ScrollUp`
- 65: `ScrollDown`
- 66: `ScrollLeft`
- 67: `ScrollRight`

Legacy X10 and malformed-input behavior are unchanged.

## Regression coverage

The public `InputParser` API parses all four SGR reports across split input boundaries and asserts the exact emitted event. The test fails on the parent for 66/67.

## Validation

- `cargo test -p fresh-input-parser` — 89 passed
- `cargo test -p fresh-input-parser --release` — 89 passed
- `cargo clippy -p fresh-input-parser --all-targets -- -D warnings`
- `cargo check --workspace --all-targets`
- `git diff --check`

### DIFF
--- a/crates/fresh-input-parser/src/lib.rs
+++ b/crates/fresh-input-parser/src/lib.rs
@@ -893,11 +893,11 @@ fn sgr_mouse_event(params: &[u8], pressed: bool) -> Option<Event> {
     );
 
     let kind = if cb & 64 != 0 {
-        // Wheel: low bit selects up/down.
-        if cb & 1 != 0 {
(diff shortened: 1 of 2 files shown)


sinelaw:
:fire: Thanks!

PR #2835, opened 2026-07-28
Migrate ANSI-output test parser from vt100 to rio-vt

raphamorim:
Swaps the terminal emulator behind the `render_real()` → ANSI → grid test pipeline from [vt100](https://crates.io/crates/vt100) to [rio-vt](https://crates.io/crates/rio-vt), Rio's terminal engine.

This PR only moves the **test** parser, but the reason to standardize on rio-vt is that it's a full terminal engine an editor can build real features on — not just a screen-scraper for tests. fresh already ships an embedded terminal (`services/terminal`), and rio-vt is exactly the core that powers, so the same dependency covers both the tests and the product surface:

the features it adds:

- **Correct Unicode width & grapheme clustering** : the same wide-glyph / combining-mark handling that `issue_1577_unicode_width` exists to protect. rio-vt packs width into the grid so cursor, wrapping and selection never desync from width-unaware apps (shells, editors).
- **Grid + scrollback, selection, and regex search** over the visible screen *and* scrollback, the primitives an in-editor terminal needs.
- **Image protocols**: sixel, Kitty, and iTerm2 graphics decoded by the core (inline images in the terminal pane).
- **Shell integration**: OSC 7 (working directory), OSC 8 (hyperlinks), OSC 133 (semantic prompt / command boundaries) jump-to-prompt, cwd-aware splits, clickable links.
- **Per-row damage tracking**, so a consumer repaints only the cells that changed, cheap redraws for a live editor loop.
- **Safe Rust, dependency-light** (no renderer, GPU, or font-shaping deps when built with no features, it's just the terminal), **fast**, and already in production at [Lovable](https://lovable.dev).

It also drops the transitive **`net2` (RUSTSEC-2020-0016, unmaintained)** the old parser pulled in.

- Benchmark against vt100 and alacritty_terminal: https://github.com/raphamorim/rio-vt-benchmark
- Background on the engine: https://rioterm.com/blog/2026/07/27/rio-vt-and-librio

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,7 +30,7 @@ version = "0.8.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
 dependencies = [
- "cfg-if",
+ "cfg-if 1.0.4",
  "const-random",
  "getrandom 0.3.4",
  "once_cell",
@@ -58,7 +58,7 @@ dependencies = [
  "home",
  "libc",
  "log",
- "miow",
+ "miow 0.6.1",
  "parking_lot",
  "piper",
  "polling",
@@ -531,6 +531,12 @@ dependencies = [
 
 [[package]]
 name = "cfg-if"
+version = "0.1.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4785bdd1c96b2a846b2bd7cc02e86b6b3dbf14e7e53446c4f54c92a361040822"
+
+[[package]]
+name = "cfg-if"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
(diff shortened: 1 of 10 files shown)


sinelaw (changes requested):

- `cargo fmt --check` fails on this branch (clean on master) — the new `pub mod vt;` in `tests/common/mod.rs` isn't alphabetically sorted.
- Core rationale doesn't hold up: `services/terminal/term.rs` (fresh's actual embedded terminal) is built on `alacritty_terminal`, not rio-vt — this PR only swaps the *test* ANSI parser, so it doesn't unify anything with the product surface as claimed.
- No product code changes at all; the concrete payoff is dropping one unmaintained transitive dep (`net2`) at the cost of ~30 new transitive crates (incl. `corcovado`, `teletypewriter`, `fuchsia-zircon`, ICU/`url` stack) for test-only ANSI parsing — a questionable trade given this project's stated dependency/size discipline.
- Commit message has no motivation in the body (title only), contrary to the Commit Hygiene guideline.
- Verified: `cargo check --tests --features dev-bins` and `cargo clippy` (no `--fix`) are otherwise clean, and the migration itself correctly covers every `vt100::` call site.

Given no user-facing behavior changes and the rationale for the swap not matching the codebase, this doesn't clear the "fixes a bug / enables something not otherwise possible / meaningfully more usable" bar — requesting changes rather than approving.

---

sinelaw:
Since we're using alacritty for the production terminal I don't see a reason to make this change now

PR #2840, opened 2026-07-28
fix(scroll): prevent blank frame at horizontal scroll bound

ttenneb:
## Summary

After horizontal wheel input reaches the no-wrap right bound, one additional `ScrollRight` can paint a transient blank/shifted frame.

The existing viewport-local bound is refreshed and clamped only after buffer rendering. This change also clamps against the previously discovered `max_line_length_seen` immediately before rendering, preventing an overshot wheel position from being painted first.

This is intentionally an ordering fix:

- uses the existing cached bound
- adds only constant-time arithmetic and `min`
- leaves the post-render scan and clamp unchanged
- does not change width accounting, cache lifetime, scrollbar geometry, wrapping, or mouse routing

The existing monotonic cache behavior after buffer edits is outside this PR’s scope.

## Regression coverage

The screen-only E2E:

1. disables wrapping and enables the horizontal scrollbar
2. sends enough horizontal wheel events to pass the discovered right bound
3. captures the settled rendered frame
4. sends one additional `ScrollRight`
5. asserts that the immediate rendered frame is identical

Verified red/green against parent `44ae19473`:

- parent: fails with a three-column transient frame shift
- this branch: passes

## Validation

- focused blank-frame E2E
- full scrolling E2E suite
- scrollbar-marker E2Es
- `cargo check --all-targets`
- `cargo check --no-default-features --features runtime --all-targets`
- `cargo clippy --all-features --all-targets`
- `git diff --check`

Only pre-existing workspace warnings remain. No performance benchmark is included because the production change adds no scan, allocation, or algorithmic work.

### DIFF
--- a/crates/fresh-editor/src/view/ui/split_rendering/orchestration/mod.rs
+++ b/crates/fresh-editor/src/view/ui/split_rendering/orchestration/mod.rs
@@ -476,6 +476,21 @@ pub(crate) fn render_content(
                 .unwrap_or(&mut empty_folds);
 
             let _render_buf_span = tracing::trace_span!("render_buffer_in_split").entered();
+
+            // Use a previously discovered bound so an extra wheel step cannot
+            // paint a blank frame before the post-render scan refreshes it.
+            if show_horizontal_scrollbar
+                && !viewport.line_wrap_enabled
+                && viewport.max_line_length_seen > 0
+            {
+                let visible_width = viewport.width as usize;
+                let max_scroll = viewport
+                    .max_line_length_seen
+                    .max(visible_width)
+                    .saturating_sub(visible_width);
+                viewport.left_column = viewport.left_column.min(max_scroll);
+            }
+
             let split_view_mappings = render_buffer_in_split(
                 buf,
                 state,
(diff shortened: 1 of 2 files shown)


sinelaw (approved):

- **Bug confirmed & fix verified**: built base commit `44ae194` and this branch, drove both in tmux via raw SGR mouse `ScrollRight` events on a 400-char unwrapped line. Base: settled frame shows a truncated/shifted line, then one more wheel tick jumps to a different frame (blank/shift glitch reproduced). This branch: frame is byte-identical before/after the extra tick — bug is fixed.
- **Code review**: minimal, ordering-only fix — clamps `left_column` against the cached `max_line_length_seen` right before the buffer render call, leaving the existing post-render scan/clamp untouched. No new full-buffer scans (CONTRIBUTING #2), shared `render_content` path covers both TUI and GUI backends (#13).
- **Regression coverage**: new E2E test asserts rendered-output identity only (CONTRIBUTING's "observe, not inspect" rule), matches the interactive repro above.
- **No duplicate/overlapping fix** found in git history for this specific overshoot-at-bound case.
- Approving — scoped, elegant, guideline-compliant fix with a solid reproducer.

---

sinelaw:
Thanks! 🙏

PR #2848, opened 2026-07-28
Wire up three dead config options, remove a fourth

### LINKED ISSUE #2842: Settings UI: "Rainbow Brackets" and "Highlight Matching Brackets" toggles have no effect
Toggling the bracket-matching settings off in the Settings UI updates the config value but bracket colorization keeps rendering exactly as before.

## Steps to reproduce

1. Start `fresh` with an empty buffer.
2. Type `({[]})` on line 1. The six brackets render in the theme's rainbow colors (`bracket_rainbow_1` … `bracket_rainbow_3` and back).
3. Open the Settings UI (`File → Settings`, or the settings action).
4. Press `/`, type `rainbow brackets`, press Enter.
5. Navigate to **Editor → Bracket Matching → Rainbow Brackets** and press Enter to flip the checkbox from `[v]` to `[ ]`.
6. Press `Ctrl+S` to save, then `Esc` to close the Settings UI.

## Expected

The brackets on line 1 are no longer colored by nesting depth.

## Actual

The brackets are still colored with the rainbow palette. The first `(` still renders with fg = `theme.bracket_rainbow_1`.

The same happens for **Highlight Matching Brackets** in the same section, and via the config file instead of the UI:

1. Put this in the user config:
   ```json
   { "editor": { "rainbow_brackets": false, "highlight_matching_brackets": false } }
   ```
2. Start `fresh`, type `({[]})`.
3. The brackets are still rainbow-colored; the first `(` renders with fg = `theme.bracket_rainbow_1`.

## Reproduced on

`2df5598` (main), verified through the e2e harness at 120x40 and 80x24.

sinelaw:
Closes #2842.

Started from "does the Rainbow Brackets toggle do anything?" (it didn't), then audited all 212 `pub` fields across the 22 config structs for the same shape — declared, defaulted, merged, schema-generated, editable in the Settings UI, never read. Four came back. Three are now wired up; the fourth is removed.

## `editor.rainbow_brackets` / `editor.highlight_matching_brackets`

`BracketHighlightOverlay::new()` hard-coded both flags on and nothing assigned them afterwards. Resolved into a `BracketHighlightSettings` carried on `EditorRenderConfig` and threaded to the overlay's per-frame update, rather than latched onto the per-buffer overlay at construction — so a settings change lands on the next frame for every open buffer with no separate invalidation pass. `rainbow_brackets` is gated on `highlight_matching_brackets`, honouring the dependency its own doc text (shown in the Settings UI) states.

Turning a toggle off also has to retract what the enabled pass drew, so `update` no longer early-returns while overlays are live; `colorization_active`/`last_cursor_pos` keep the clear to once rather than every frame.

## `file_explorer.respect_gitignore`

Two overlapping knobs where only one worked: the explorer consulted `show_gitignored` and never this. Given the distinct meaning its name implies — a master switch for whether `.gitignore` rules apply at all — enforced inside `matches_gitignore`, the single fork both the visibility filter and the render status go through, so ignored entries are neither hidden nor grayed out when it is off.

## `textmate_grammar` (per-language)

`BufferConfig::resolve` copied the path out of `LanguageConfig` and nothing read it back, so a `[languages.foo] textmate_grammar = "…"` entry did nothing. Config-declared grammars now go through the same `GrammarSpec` path plugins use, behind a new `GrammarRegistry::apply_languages` that every registry setup site calls: load the files, then re-apply the catalog config the syntax-set rebuild would otherwise drop. The rebuild only runs when a grammar is actually configured. The doc promised TextMate `.tmLanguage`, which syntect cannot parse — corrected to the Sublime format the loader accepts.

## `editor.highlight_timeout_ms` — removed, not implemented

This one was tried and then withdrawn; the two attempts are left in the branch history because CI ran on them and the result is the argument.

A time cap can only be spent past the visible region, or it leaves visible text unhighlighted. But `highlight_viewport` only reports a cache hit once the cached range reaches `parse_end`, so any refresh that stops short leaves the next one re-parsing the remainder. On a large file that costs more total work than a bounded frame saves — the per-frame re-parse cost `a0633bb` had just removed, and it bites hardest on exactly the files the setting was meant to help. CI caught it as a lost cache hit on a 21 KB fixture (`test_perf_cache_hit_no_reparse`, which asserts on counters, not timings).

That left a knob that had to ship off by default, could not be tested in the regime anyone would use it in (a non-zero budget is timing-dependent, so any test of it would be flaky), and hurt when enabled. `highlight_context_bytes` and `large_file_threshold_bytes` already bound highlighting work by size, deterministically.

Removed outright rather than hidden: the config derives no `deny_unknown_fields`, so a stale `highlight_timeout_ms` key in a user config is ignored rather than an error, and nothing can depend on behaviour it never had.

## Tests

Per CONTRIBUTING, every claim has a test that fails without its fix — verified by stashing `src/` and re-running:

- `tests/e2e/bracket_highlight_settings.rs` — 4 tests, one driving the full Settings UI flow (search → toggle → Ctrl+S → Esc), asserting only on rendered cell colors.
- `tests/e2e/file_explorer.rs` — the setting on and off, plus a control test pinning the default so the first isn't measuring a broken `.gitignore`.
- `tests/e2e/language_textmate_grammar.rs` — a configured grammar highlights, and the same file without one does not.

## Not addressed

`editor.nerd_font_icons` is read only by the Settings dialog's own category icons, which is what its doc scopes it to — left alone.

Two `auto_indent` tests (`test_config_increase_indent_pattern_applies`, `test_config_indent_next_line_pattern_applies`) fail when run in a large parallel batch and pass alone. `reload_indent_overrides` writes process-global state, so they are order-dependent. Pre-existing — reproduced on the base commit without any of these changes — and left for a separate fix.

https://claude.ai/code/session_012wmdPY33aHp7fiMFoJwxpr


sinelaw:
`test macos-latest` is red on a **timeout, not an assertion**: 8282/8283 passed, with `e2e::plugins::review_diff_line_staging::test_review_visual_stage_modified_line_pair` hitting the 180s nextest limit.

Not caused by anything on this branch, and I'm not fixing it here:

- That test exercises the review-diff plugin's line staging. This PR touches bracket-highlight settings, the file-explorer gitignore filter, and grammar registry setup — no overlap.
- It passes locally on this branch in **0.85s**, against a 180s limit.
- `test ubuntu-latest` is green on the same commit.
- Master fails the same way. The run for `2df5598` (this branch's original base) also failed only on `test macos-latest`, also with exactly one e2e test timing out at 180s while 8256/8257 passed — a different test that time (`dock_switch_never_paints_two_workspaces_into_one_frame`).

So macOS runners are intermittently blowing the 180s limit on heavy e2e tests, and which test trips varies per run. I'll re-run the macOS job once `test windows-latest` finishes — the whole run has to complete before failed jobs can be re-run.

---

PR #2855, opened 2026-07-29
fix(viewport): use rendered width for cursor visibility

ttenneb:
## Summary

Horizontal cursor visibility reserved a vertical-scrollbar column twice: split layout had already removed that column from the viewport content width, while both the buffer-aware and rendered-layout visibility paths subtracted another cell. This left the final content cell unreachable and could make wide glyphs appear to interfere with the scrollbar edge.

Remove the duplicate reservation from the buffer-aware path. For the rendered-layout path, pass the authoritative `render_area.width` through an internal helper and subtract only the gutter. The existing public `ensure_visible_in_layout` signature remains source-compatible and delegates using the viewport width.

Using the explicit render width is necessary because compose mode can render a narrower area than `Viewport::width`; it is the same coordinate-space correction, not a separate scrolling feature.

## Regression coverage

- unit test proves content-width visibility does not reserve another scrollbar cell
- unit test proves the layout path honors an explicit render width narrower than the viewport
- rendered E2Es cover vertical scrollbar shown and hidden
- every scrollbar row retains track/thumb styling when shown and no scrollbar styling when hidden
- EOL caret and final `Z` reach the correct content edge
- CJK, combining text, and emoji occupy their expected terminal cells beside the edge

The rendered regression fails on the parent, where the shown and hidden cases place the cursor one cell too far left. A subtract-one mutation also makes the focused content-width unit fail (`left_column` becomes 1).

## Validation

- focused viewport unit tests
- vertical scrollbar cursor E2Es
- compose-mode regression
- horizontal scrolling E2Es
- split-view E2Es
- scrollbar-marker E2Es
- `cargo check --all-targets`
- `git diff --check`

## Performance / scope

No scans, allocations, caches, or background work are added. The production change removes duplicate arithmetic and passes an existing width through the rendered-layout call chain.

### DIFF
--- a/crates/fresh-editor/src/view/ui/split_rendering/orchestration/render_buffer.rs
+++ b/crates/fresh-editor/src/view/ui/split_rendering/orchestration/render_buffer.rs
@@ -409,7 +409,12 @@ pub(crate) fn compute_buffer_layout(
     // Ensure cursor is visible using Layout-aware check (handles virtual lines)
     let primary = *cursors.primary();
     let top_byte_before_scroll = viewport.top_byte();
-    let scrolled = viewport.ensure_visible_in_layout(&view_data.lines, &primary, gutter_width);
+    let scrolled = viewport.ensure_visible_in_layout_with_render_width(
+        &view_data.lines,
+        &primary,
+        render_area.width as usize,
+        gutter_width,
+    );
 
     // If we scrolled AND `top_byte` changed, rebuild view_data from the new
     // top_byte (the old view_data no longer matches what's visible).  We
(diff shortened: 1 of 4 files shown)


sinelaw (approved):

- Code quality: minimal, targeted diff (4 files, +178/-15); removes the duplicate scrollbar-column reservation instead of patching around it; both `render_buffer.rs` call sites of `ensure_visible_in_layout` were updated consistently, no stray callers left on the old signature
- Follows CONTRIBUTING.md: reproducer tests included in the same commit (2 unit tests + 2 rendered e2e tests), `cargo check --all-targets` and `cargo fmt --all -- --check` pass cleanly on my machine
- No duplicate/overlapping feature found; this is a distinct fix from the only other related PR (#2690, cursor-reveal timing, different mechanism)
- Interactively validated in tmux: built both `master` and this branch, disabled line wrap, opened a 33-char line with vertical scrollbar shown, pressed End
  - `master`: content stops 2 cells short of the scrollbar column (double reservation bug reproduced)
  - this branch: content/cursor reaches the last cell immediately adjacent to the scrollbar (1 cell gap = just the natural EOL cursor cell)
- Approving

---

sinelaw:
Thanks!

PR #2857, opened 2026-07-30
Plugins can no longer hang the UI: off-loop work, budgeted drains, snapshot-only render

sinelaw:
Fixes UI hangs caused by plugin activity. A test plugin was written to reproduce the problem: dashboard sections that grep the whole repo in a loop, spawn a dozen subprocesses per tick, hit the network, emit 300-column rows, throw, and hang forever. Each change was measured with `FRESH_LOG_SPANS=1` span timings and the existing `paste_timing: render` log line.

## Measurements

Debug build, 200×50 terminal, hostile plugin running, 4–8 minute sessions:

| metric (p99 / max, ms) | before | after |
|---|---|---|
| `terminal.draw` | 444 / 2798 | 120 / 703¹ |
| span `render` | 440 / 2790 | 114 / 699¹ |
| span `render_content` | 62 / 70 | 93 / 109 |
| span `process_plugin_commands` | 2250 / 2300 | 16 / 23 |
| span `process_async_messages` | 2220 / 2300 | 17 / 26 |
| span `editor_tick` (max) | 2300 | 26 |

¹ All samples above 200ms occur in the first ~30s while the startup plugin batch loads; steady-state max is 133ms.

The span hierarchy located the cost: `render` peaked at 2790ms while its child `render_content` peaked at 70ms, so the time was spent in plugin command dispatch inside the draw rather than in painting.

## Changes

**Blocking work moved off the editor thread.** The `plugin_offloop` module holds the handlers for plugin commands that do I/O. It has no `Editor` import, so a handler there has no access to editor state and reports results by posting an `AsyncMessage`. The editor-thread half of a command snapshots what only that thread can read (dirty buffers as `HybridSearchPlan`s; the piece tree is not `Send`) and hands off. `grepProject` was the main offender and now works this way, with coalescing: repeated requests while one is in flight collapse into a single run.

**Drain limits on the dispatch queues.** The plugin-command and async-message drains each run against a 50ms deadline per pass, with the tail deferred in arrival order (FIFO holds; a non-empty backlog keeps the loop on its frame cadence). The limit is a pathology guard rather than a frame pacer: a normal burst — a plugin load, a large file firing hooks, a hundred queued commands — clears in one or two passes instead of being spread over many frames; only a genuine flood gets deferred. Each pass also dispatches at least eight items regardless of the deadline, so one slow handler cannot throttle the queue to one item per frame (simulated: a 232-command startup batch on hardware where each dispatch costs 5ms drains in ~1.5s with the floor versus ~3.7s without; sustained emission that outpaces a one-per-frame drain stays stable). The budgeted drain is used by the interactive loops (`process_async_messages_budgeted`); the existing `process_async_messages()` still drains to completion, since about 80 call sites in tests and the test API depend on that behavior.

**No dispatch inside the draw.** Previously the render path dispatched plugin commands mid-frame, which required a layout recompute and a hold-back list for window switches. That code is deleted (about 170 lines). Command dispatch on the render path now happens once, before layout, through the same budgeted drain as the tick. A frame renders from a single consistent state. Hook replies (overlays, conceals) land on the following frame, which adds one frame of decoration latency independent of what a plugin emits. Plugin locks taken inside the draw use `try_read`/`try_write`; on contention the payload is deferred to the next frame and no hook is marked as seen.

**`CloseBuffer` split into mutation and effects.** Closing a buffer previously did disk work on the editor thread; the recovery-file delete alone blocked it for 1771ms in one measured run. The close path now applies only state mutation synchronously — registry removal, split/tab updates, focus resolution — so the buffer disappears from the UI immediately, and every filesystem effect runs on the tokio runtime via a new `spawn_off_loop_effect` helper (with a synchronous fallback when no runtime exists, as in some test harnesses): the recovery-file delete, the per-file session-state write, the terminal backing-file rename that keeps scrollback searchable, the raw terminal log delete, and the retained-backing GC (a directory scan deleting up to hundreds of files). The remaining synchronous steps were audited: terminal shutdown is a channel send, the `buffer_closed` hook is a post to the plugin thread, and `purge_buffer_state` touches only in-memory maps. The file-save path shares `delete_buffer_recovery` and gets the same behavior.

**Slow-handler detection.** `dispatch_plugin_command_measured` logs a WARN with the command name for any handler that spends more than 50ms on the editor thread, and panics when `FRESH_STRICT_PLUGIN_BUDGET` is set (used in tests). Extracting the command name allocates nothing: the `Debug` writer stops after the variant identifier. This check is what found the `CloseBuffer` stall. The final measurement session produced zero warnings over 7.6 minutes.

## Bugs surfaced by the budget change

Two places depended on a whole command batch being dispatched in one call:

- **Dock focus.** The orchestrator opened the auto-open dock with a mount command followed by a blur command. With budgeted dispatch, a tick boundary can fall between the two, leaving the dock focused for one frame. CI caught this on ubuntu (`auto_open_setting_shows_dock_unfocused_at_startup`). `MountFloatingWidget` now has a `start_blurred` field (serde default false; optional trailing argument in the QuickJS binding and `widgets.ts`; `fresh.d.ts` regenerated), and the orchestrator mounts the dock with it set.
- **Test drain assumptions.** Six `init_script` e2e tests asserted state after one `process_async_messages()` call while the startup batch held 232 registration commands, more than one budget covers. Restoring the drain-to-completion contract on the public method (see above) fixed them.

## Dashboard

- Rows wider than the panel previously pushed the right border out and wrapped the line. All body rows now pass through one `truncate(text, inner)`, and click ranges are clamped to the surviving text.
- Panel size was fixed at `MAX_INNER = 72`. It now targets ~90% of the viewport, clamped to `[24, 110]` and bounded by `viewportW − 2`. The top border is composed against an explicit budget and drops the stamp, then the title, as width shrinks. Vertical overflow clips with a "N more rows — resize to see" marker.
- Sections now receive the panel width (`DashboardContext.width`) and a `columns()` helper that clips each cell to its column. The disk and PR rows use it.
- Refresh was a single loop re-running every section every 5s (134–187 runs/min measured). Each section now declares a TTL (default 30s; git 15s, disk 60s) and a 1s scheduler tick skips fresh or in-flight sections. Measured ambient subprocess rate for the built-in sections: 22.7/min. The clock renders at minute resolution; the seconds display forced a repaint every second.
- A section that throws or exceeds its timeout keeps its last rendered body, with `stale` or `timed out after Nms` in the heading. Other sections are unaffected.
- Glyph badges were replaced with plain-text tags (`pass`/`fail`/`run`, `+3 ahead -0 behind`, ASCII `[###---]` meter). The replaced glyphs are East-Asian-ambiguous width; terminals that render them double-width misaligned the row.

## Tests

- `hostile_plugin_cannot_stall_the_editor_loop` (e2e): runs the grep loop, spawn storm, 300-column rows, throwing and hanging sections against a live editor; asserts the median tick+render duration stays under 250ms and the frame stays well-formed. The median was chosen because inline dispatch raises every iteration, while CI load raises only outliers.
- Width sweep (30–140 columns, `step_by(1)`, with an overflowing section): passes; marked `#[ignore]` due to a ~22-minute runtime, runnable with `--ignored`.
- Plugins e2e suite: 447/448 locally; the one failure (`test_diagnostics_api_with_fake_lsp`) passes in isolation and predates this branch. Terminal suite (140 tests) green after the close-path change.

## Checked and ruled out

The synchronous filesystem API (`readFile`, `writeFile`, `fileExists`, `readDir`, `fileStat`) blocks only the plugin thread: `quickjs_backend`'s `fs_for` resolves a `PluginFilesystem` owned by the plugin runtime. The brief listed these calls as suspects; they are unrelated to the hangs.

## Remaining work

- `render` p99 is 114ms against the original 30ms target. The remaining cost is `render_content` (p50 63ms), the buffer paint itself in a debug build at 200×50. Plugin-attributable frame cost is bounded (`pre_layout_drain` max 19.6ms at the old 4ms budget; the 50ms pathology guard admits more per pass by design). Reducing paint cost is a separate rendering-performance task.
- QuickJS's rejection tracker reports a rejected promise from a synchronously-throwing async function as unhandled even when a handler attaches on a later microtask. In production this is a spurious log line; under the test harness's strict mode it kills the plugin thread. Documented in the hostile test; should get its own issue.

`cargo fmt` clean, `cargo clippy --bin fresh` clean, `tsc` clean on the plugin tree.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BMfQcAE8GNCaDLY6giwnNo)_


sinelaw:
## Iteration 2 — measured results

Same protocol as the baseline (debug build, 200×50 tmux, hostile plugin driving grep loops / spawn storms / wide rows / throwing + hanging sections, ~7.6 min window, `FRESH_LOG_SPANS=1`):

| metric (p99 / max, ms) | baseline | iter-1 (budgets + off-loop grep) | iter-2 (draw eviction + close decomposition) |
|---|---|---|---|
| `terminal.draw` | 444 / 2798 | 461 / 537 | **120 / 703**¹ |
| span `render` | 440 / 2790 | 456 / 530 | **114 / 699**¹ |
| span `render_content` (pure paint) | 62 / 70 | 98 / 119 | 93 / 109 |
| span `process_async_messages` | 2220 / 2300 | 9.6 / 1790 | **17 / 26** |
| span `process_plugin_commands` | 2250 / 2300 | 12 / 1790 | **16 / 23** |
| span `editor_tick` max | 2300 | 1790 | **26** |

¹ Every sample above 200ms sits in the first ~30s (startup plugin batch; the new `render_post_content` span attributes them to first-frame setup). Steady-state max is 133ms.

Also in this run:
- **Zero `plugin_budget` watchdog warnings** over the whole window — the `CloseBuffer` 1771ms stall found in iteration 1 is confirmed fixed by the mutation/effect decomposition.
- Dashboard built-ins' ambient subprocess rate: **22.7/min** while actively driven. (The log's 382/min from `init.ts` is the hostile harness declaring 1s TTLs on every section and receiving exactly the cadence it asked for — bounded by declared per-section rates, which is the invariant.)
- Plugin-attributable frame cost is now bounded: `pre_layout_drain` max 19.6ms, hook firing max 15.1ms; everything else in `render` is content paint.

## What changed since the PR opened

- **Snapshot-only render, fully.** The mid-render drain, its layout-recompute hack, and the `deferred_plugin_commands` hold-back machinery are deleted (−170 lines). The render path has exactly one dispatch point — pre-layout, before anything is painted — sharing the tick's budgeted FIFO drain. Frames can no longer tear by construction.
- **`CloseBuffer` decomposed** (mutation now, disk effect off-loop): recovery-file deletion and session-state writes run on the runtime via `spawn_off_loop_effect`; the recovery mutex is taken inside the effect.
- **Drain contract split**: `process_async_messages()` keeps its historical drain-to-completion semantics (80 direct call sites in tests rely on it); the interactive loops use `process_async_messages_budgeted()`. This surfaced when the frame budget correctly deferred a 232-command startup batch and six `init_script` tests asserted after a single drain call.
- **Hostile-plugin e2e test** added and green: asserts the loop never stalls past 1.5s and the frame never tears under grep loop + spawn storm + 300-col rows + throwing + hanging sections.
- **Width sweep** `#[ignore]`d (passes; ~22 min runtime; run via `--ignored`).
- Plugins e2e suite: 447/448 green (the one failure, `test_diagnostics_api_with_fake_lsp`, passes in isolation — pre-existing parallelism flake with a real spawned fake-LSP process).

## Honest residual vs. the stopping criteria

`render` p99 is 114ms against the 30ms target — but the binding constraint is now `render_content` (pure buffer paint) at p50 63ms / max 109ms in a **debug build** at 200×50, with plugin-attributable cost bounded at ~20ms. The invariant the criteria were proxying for — *worst-case frame time depends on buffer content alone, plus a fixed constant* — holds and is regression-guarded by the watchdog + e2e test. Getting p99 under 30ms from here is a paint-performance project (dirty-region tracking / debug-profile paint cost), orthogonal to plugin-originated hang prevention.

One pre-existing runtime bug surfaced en route, documented in `e2fc8b9`: QuickJS's rejection tracker flags sync-throwing async fns as unhandled even when the handler attaches a microtask later — log noise in production, plugin-thread death under the test harness's strict mode. Worth its own issue.

---

Issue #2713, opened 2026-07-16
Expose a plugin API to paint colored markers on the scrollbar at proportional line positions


RetributionByRevenue:
for example, i'm able to highlight the line green and also the gutter, but i would like functionality that also colors the section in the scroll

<img width="1411" height="885" alt="Image" src="https://github.com/user-attachments/assets/dfa2b8cb-8753-4c24-ba70-818ea884914c" />

sinelaw:
I'm working on a feature to set markers on the scrollbar.

But I just realized you may mean something else here. Did you want the line highlight to extend into the scrollbar area? Or did you mean a marker on the scrollbar, positioned proportionally to show where the marked line sits in the document (what I'm building)?

RetributionByRevenue:
> Or did you mean a marker on the scrollbar, positioned proportionally to show where the marked line sits in the document (what I'm building)?

Yes — exactly this, please.

PR #2868, opened 2026-07-30
Show git hunks and unsaved changes on the scrollbar

sinelaw:
The gutter can only speak for the lines on screen. Scroll away from a change and the editor stops mentioning it. This closes that gap for both kinds of change the gutter shows.

## 1. Git-gutter hunks on the track (`git_gutter.ts`)

Live diff already marks its hunks on the scrollbar (#2713); the git gutter — which most users have on all the time — did not. It now publishes the same hunk set its glyphs come from:

- Own namespace (`git-gutter-scroll`), gutter palette reused, so a hunk's glyph and its mark are the same colour.
- One marker per hunk, spanning `line`..`endLine`. `ScrollbarMarker` gains `endLine` (the inclusive line counterpart to the byte `end`) for exactly this: a `git diff` parser knows a hunk's first and last line but not their byte offsets, and without it the plugin had to emit one marker per changed line.
- One batched whole-namespace replace — the diff covers the whole file, so the set is authoritative and swaps atomically. An empty set (or an untracked file) clears it.

## 2. Unsaved changes on the track (core)

The blue gutter bar for unsaved edits comes from `diff_since_saved`, which already returns **whole-buffer** byte ranges — the gutter just clips them to the viewport afterwards. So the same ranges can feed the scrollbar with no new diffing. The projection grows a second mark source next to the plugin-owned set, painting in the gutter bar's colour at its priority (5), so a git hunk (10) still wins a cell they both want. No plugin and no git repository required.

Three things it needs to stay cheap and correct:

- **The cache is probed before the diff is gathered**, so a steady-state frame pays a key comparison and nothing else. The key gains a new `save_state_version` on the buffer, because a save changes what the marks cover without touching the content version.
- **Ranges within one track cell of each other merge first.** A replace-all across a large file leaves thousands of unsaved ranges, and two coordinate lookups each (`get_line_number` is ~0.5 µs) would put milliseconds on the typing path to draw marks a ≤200-cell track cannot tell apart. Merging is lossless in coverage — a group spans every range inside it — so it is a resolution choice, not a silent cap.
- **Core ranges project through one `endpoints()` helper**, because the logical-line basis pre-resolves byte→line into a map and a lookup it never saw would answer row 0. A benchmark caught exactly that as an index panic in the render loop; the painter is now also guarded against a non-monotonic lookup.

## 3. Two defects found while testing this by hand

- **Inserting lines marked one line too many.** The diff's byte ranges are half-open, and inserting whole lines produces a range ending exactly on a line boundary. `diff_indicators_for_viewport` counted that boundary as changed, so inserting three lines put a bar on four — and the fourth vanished on save, when git reported the real line set. That read as the two surfaces disagreeing; they agree now because the bar is no longer wrong.
- **A hunk cost one marker per line.** Measured on a file with 20 000 changed lines: 28.6 ms in the command handler plus 24.4 ms in the next render, on every save. With `endLine` and one marker per hunk: 17.6 µs and a 2.8 ms render.

## Performance

Measured on this branch (release build). `diff_since_saved` prunes by `Arc` identity, so its cost tracks the *edit* count, not the file size:

| buffer state | diff cost | (10 MB file) |
|---|---|---|
| unmodified | 13 ns | 13 ns |
| 1 unsaved edit | 170 ns | 171 ns |
| 200 unsaved edits | 12 µs | 12 µs |
| ~2 200 unsaved edits | 261 µs | 132 µs |

Whole-frame A/B, worst case (200k lines, 2 000 scattered unsaved edits, 300 renders per sample, 3 samples):

- steady-state frames: 2.34–2.62 ms with, 2.34–2.53 ms without — within noise, as expected from the early-out and the cache probe.
- frames right after an edit: median 2.83 ms with, 2.65 ms without → **≈ +0.19 ms** on the pathological case. A single-edit buffer is ~0.2 µs.

## Tests

- `view::scrollbar_marker` unit tests: core marks paint with no plugin markers; a range paints the rows it spans and not the row after; priority precedence both ways; a save invalidates the projection; `endLine` resolves as an inclusive range and yields to an explicit byte `end`.
- `split_rendering::scrollbar` unit tests: an unsaved edit projects with no plugin markers and clears on save; dense edits coalesce to track resolution without losing the last edit.
- `e2e/scrollbar_markers.rs`: an edit near the top stays marked after scrolling it off screen; saving clears the marks; plugin markers and unsaved marks coexist; and the >5 000-line logical-line regime is covered on its own (that path is where the panic lived).
- `e2e/unsaved_change_gutter.rs`: inserting three lines marks exactly three; editing one line marks one. Fails without the off-by-one fix (4 bars vs 3).
- `e2e/plugins/gutter.rs`: git hunks below the fold show green/red/orange marks in file order while no gutter glyph is on screen; and a clean file stays unmarked until a change is saved.
- Three existing tests updated: two now ignore the scrollbar column, which legitimately changes when the buffer is dirty (`markdown_compose` flicker check, `vi_mode_autostart` line scrape), and `live_diff::test_live_diff_highlights_empty_added_line` now waits for plugin quiescence before reading a cell background instead of sampling a half-applied pass.

Also verified by hand in tmux against the real binary: marks appear in the gutter palette at proportional positions, survive scrolling the edit off screen, hand over from blue to git colours on save, and behave in the >5 000-line logical-line regime.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
   * If you have customized your status bar, open Settings → **Status Bar** and move **Terminal Restart** from *Available* to *Included* (`Shift+→`) under Left or Right.
 * **Tabs for same-named files say which file they are** - open two `mod.rs` and the tabs read `model/mod.rs` and `view/mod.rs` instead of `mod.rs 1` and `mod.rs 2`. Each tab grows only as much of its path as it takes to be unique, and tabs whose name is already unique are untouched (#2851, requested by @anddimario).
 * **Line-ending indicators** - `↵` at every line break and `␍` for the CR half of a CRLF, both off by default (#2798, requested by @akarinotomoshibi).
-* **Scrollbar markers for plugins** - `editor.setScrollbarMarkers` paints marks on the scrollbar track; live-diff hunks and Markdown headings now use it (#2713, requested by @RetributionByRevenue).
+* **Scrollbar markers for plugins** - `editor.setScrollbarMarkers` paints marks on the scrollbar track; git-gutter hunks, live-diff hunks and Markdown headings now use it (#2713, requested by @RetributionByRevenue).
 * **File Explorer sticky parents** - a nested folder's expanded ancestors stay stacked at the top of the sidebar while you scroll (#2705, by @asukaminato0721).
 * **More graphics, docs and build-file grammars** - GLSL `.glslf`/`.glslv`, Wavefront `.obj`, Doxygen, Windows `.rc`, pkg-config, `.cmake.in` and `CMakeCache.txt` (by @asukaminato0721).
 * **Broader LaTeX ecosystem highlighting** - TeX packages/classes, generated `.aux`/`.toc`, BibLaTeX, ConTeXt, BibTeX and `.bst`, plus `latexmkrc` (by @asukaminato0721).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
 * **Tabs for same-named files say which file they are** - open two `mod.rs` and the tabs read `model/mod.rs` and `view/mod.rs` instead of `mod.rs 1` and `mod.rs 2`. Each tab grows only as much of its path as it takes to be unique, and tabs whose name is already unique are untouched (#2851, requested by @anddimario).
 * **Line-ending indicators** - `↵` at every line break and `␍` for the CR half of a CRLF, both off by default (#2798, requested by @akarinotomoshibi).
 * **Scrollbar markers for plugins** - `editor.setScrollbarMarkers` paints marks on the scrollbar track; git-gutter hunks, live-diff hunks and Markdown headings now use it (#2713, requested by @RetributionByRevenue).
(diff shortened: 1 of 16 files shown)


sinelaw:
CI note: `test ubuntu-latest` is red on `e2e::code_tour_dock::test_step_range_is_highlighted_in_the_editor` (180 s timeout). It fails on the base branch too — master's own run on 48f2de9 timed out on exactly the same test:

```
TIMEOUT [180.302s] fresh-editor::e2e_tests e2e::code_tour_dock::test_step_range_is_highlighted_in_the_editor
```
https://github.com/sinelaw/fresh/actions/runs/30820009060/job/91707257249

Nothing in this PR touches it: the test samples a background colour at column 20 of a tour buffer that is never modified, so neither the unsaved-change marks nor the git-gutter markers are in play, and the module passes locally on this branch. The previous ubuntu run hung on a different test in the same module (`test_q_closes_the_tour`), which fits the unbounded waits there rather than anything specific to a test.

Not fixing it here — it's yours to stabilise, and it's already the subject of recent commits on master. I'll re-run this PR's CI when the base recovers.

One failure in the earlier macOS run *was* worth acting on and is fixed in 9a2fec4: `live_diff::test_live_diff_highlights_empty_added_line` read a cell background after waiting only for the `+` gutter glyphs, but `renderHunks` sends gutter indicators before background overlays, so the wait could be satisfied by a half-applied pass. It now waits for plugin quiescence first.

---

PR #2886, opened 2026-08-02
Updates finish through the channel that installed fresh

sinelaw:
## TL;DR

Every `.deb` and `.rpm` install has been unable to update itself. The receipt is written to `PREFIX/share/fresh-editor/`, but `candidate_paths` only looked in `PREFIX/share/fresh/` — the binary is `fresh`, the package is `fresh-editor` — so provenance resolved to `unknown` and users got "installed manually, can't update itself". CI passed because it checked the file *exists*, not that the binary can *find* it.

Fixing that exposed the rest: the update indicator had two states for three outcomes, the `.deb`/`.rpm` path stopped after downloading and handed the user a `dpkg -i` to run, and several channels named commands that could not have worked. Container testing and the package managers' own manuals then showed that most of the command table was wrong, including the two we shipped.

The whole update path now lives in `fresh-update`, next to the provenance that justifies its decisions.

## What updating looks like

1. `fresh` checks for a new release in the background, once a day. When there is one, the status bar shows an indicator: `Update: v0.4.7`.
2. Clicking the indicator opens a confirmation popup. It names the version, says how this copy of fresh was installed, and says what confirming will do — including whether a password will be asked for.
3. The popup offers up to three rows, most automatic first:
   * **Update to v0.4.7 now** — fetch, verify, install.
   * **Download it, then show me the install command** — fetch and verify, then stop. Offered only for `.deb`/`.rpm` installs, where the work splits at a seam: fetching needs the network, installing needs root.
   * **Show the update command** — name the commands and stop. Nothing is requested, nothing is written.
4. Taking any of the first two opens a terminal buffer inside the editor and runs the update in it, as a child process. Progress is visible as it happens: the asset being downloaded, the checksum being verified, then the install output.
5. If the install needs root, `sudo` prompts for a password in that same buffer. That prompt is the only interaction the update asks for; typing the password finishes the update there and then.
6. When the update process exits, the indicator settles into one of three states: `Updated — restart fresh`, `Update needs a command — click for details` when the remaining step is one the user runs themselves, or a failure. Clicking it afterwards reopens the update terminal's output — at "needs a command" it shows the command, at a failure it offers retry, the log, or dismiss.

### What "Update now" does, per install method

The download is always performed by fresh's own code, but in the update child running in that terminal buffer — never in the editor process, and never by a shell command we compose. "Downloads" below means fresh fetches the artifact and checks it against its `.sha256`; everywhere else the package manager does its own fetching.

| Installed via | Platform | Update now | Password |
|---|---|---|---|
| apt (`.deb`) | Linux | Downloads the `.deb`, then `apt-get install` | sudo, in the terminal buffer |
| dnf (`.rpm`) | Linux | Downloads the `.rpm`, then `dnf install` | sudo, in the terminal buffer |
| zypper (`.rpm`) | Linux | Downloads the `.rpm`, then `zypper install` | sudo, in the terminal buffer |
| AUR / pacman | Arch | `git clone` the AUR repo, then `makepkg --syncdeps --install` | yes — makepkg invokes pacman itself, so fresh does not wrap it |
| Homebrew | macOS, Linux | `brew upgrade --formula fresh` | no; brew refuses to run under sudo |
| Nix | any | `nix profile upgrade fresh` | no |
| FreeBSD pkg | FreeBSD | `pkg upgrade fresh` | sudo, in the terminal buffer |
| winget | Windows | `winget upgrade --id sinelaw.fresh-editor --include-unknown` | winget raises UAC itself, in a separate window — Windows has no inline prompt |
| cargo | any | `cargo install --locked fresh-editor` — builds from source, takes minutes | no |
| npm | any | `npm install -g @fresh-editor/fresh-editor@latest` | no, unless the global prefix is system-owned |
| mise | any | `mise upgrade --bump github:sinelaw/fresh` | no |
| tarball / prebuilt | any | Downloads the archive, extracts the binary, replaces the running one | no; needs write access to the install directory |
| AppImage | Linux | Downloads the AppImage, extracts it, swaps the install root | no |
| Flatpak | Linux | Names `flatpak update` for the host — a Flatpak sandbox has no `flatpak` binary | n/a |
| Snap | Linux | Names `snap refresh`; snapd also refreshes on its own schedule | n/a |
| scoop / chocolatey | Windows | Nothing is published for these; points at the releases page | n/a |
| source / unknown | any | Names `git pull && cargo install --path …`, or the releases page | n/a |

Two platform differences worth knowing. On Unix the password prompt appears inline in the update terminal, which is what lets a `.deb`/`.rpm` update finish in one step. On Windows there is no inline equivalent — elevation always opens a separate UAC window — so the Windows channels are the ones where fresh never elevates, and winget handles it itself. Replacing a running executable also differs: Unix renames the new binary over the old one and the running process keeps its old inode; Windows renames the running `.exe` aside first and cleans it up on next launch.

### The original bug

* **`receipt::candidate_paths` also looks in `PREFIX/share/fresh-editor/`.** Both spellings exist on real machines — brew, flatpak and nix use `PREFIX/share/fresh/` — so renaming either side would strand existing installs.
* **`SelfUpdatePhase::ActionRequired`**, keyed off exit code 2 from `fresh --cmd update`. Success and failure alone cannot describe "downloaded and verified, now run this as root".
* **The confirmation popup is derived from the resolved `UpdatePlan`**, instead of one hardcoded row that meant something different per channel.
* **CI asserts resolved provenance** from `config paths` rather than `test -f` plus `grep`.

### Three rungs, and "show the command" means it

`--print-command` used to download and verify the package and stage it on disk before printing — a surprising amount of work, and a file left behind, for a request to *read* something. It now touches nothing: the release feed already carries the asset's URL and filename, so both the download URL and the install command are nameable without fetching a byte. `--download-only` is the new middle rung. `print_command: bool` became `Execution`, which also fixed `--print-command` on a self-contained install swapping the binary anyway, because that branch never consulted the flag.

### Commands that could not have done what they claimed

* **`apt-get install PATH.deb`, not `dpkg -i`.** dpkg is, in its own man page's words, "a medium-level tool": it unpacks and configures but resolves no dependencies. It works today only because the `.deb` is a static binary; the first release that gains a `Depends:` would leave the package half-configured. Verified in a container — `Unpacking fresh-editor (0.5.0) over (0.4.6)`, dpkg state indistinguishable from a repo install. It also reads a bare filename as a *package name*, so the path is forced to look like one.
* **`dnf install PATH.rpm`, not `rpm -U`.** Same missing dependency resolution, plus rpm errors on the version already installed and refuses an older one — which is why `--replacepkgs` had been bolted on.
* **Flatpak could never have worked.** A Flatpak build runs inside `org.freedesktop.Platform`, which contains no `flatpak` binary, so the `flatpak install` we generated could not execute in the process that generated it. The user pressed "Update now" and got ENOENT. `flatpak-spawn --host` needs `--talk-name=org.freedesktop.Flatpak`, a total sandbox escape rather than a scoped permission. The portal's `UpdateMonitor` is the sanctioned route but pulls from the app's remote, and a bundle has no usable one.
* **`npm install -g pkg@latest`, not `npm update -g`.** npm's own v10 docs say a global install is treated as caret-ranged, so `^0.4.6` would not admit `0.5.0`. npm 10 no longer behaves that way, which is worse rather than better — the command's meaning depends on which npm the user has, and a rule that forbids runtime branching cannot rest on that.
* **`mise upgrade --bump`** — plain `mise upgrade` keeps the range in mise.toml, so a pinned user got exit 0 and no upgrade. **`winget --include-unknown`** — winget frequently cannot read an installed version out of a portable/archive package and otherwise answers "No applicable upgrade found" forever. **`brew upgrade --formula`** so it cannot resolve to a cask.
* **Snap names `snap refresh`, not the releases page** — snapd refreshes about four times a day on its own and cannot be told not to.

### Where releases come from

* **https only, host pinned.** Release metadata and artifacts are fetched from a fixed host list or not at all, including across redirects.
* **`$FRESH_RELEASES_URL` / `$FRESH_DOWNLOAD_BASE` no longer reach `sudo`.** Unconstrained they turned "can set an env var" into "can run code as root": point the base at your own server, serve a package *and* the `.sha256` that matches it, and the checksum proves only that you can hash. An override now marks the run untrusted — it still downloads and verifies, so it stays useful for testing, but it prints the install command instead of running it privileged. A release build without the `insecure-endpoints` feature refuses an out-of-policy override outright.
* **One HTTP client for all three fetches.** The artifact used the platform certificate verifier (for TLS-intercepting corporate proxies) while the feed and checksum used the bundled roots — so behind exactly the proxy the first was configured for, the download succeeded, the checksum fetch failed, and verification is fail-closed. That was a live functional bug for enterprise users, not a theoretical one.
* **The feed is parsed as JSON.** It was a string scan for the first `browser_download_url` key in the document; a release body is free text, so a URL written in the prose was indistinguishable from a published artifact. Matching now looks only inside `assets`, requires the architecture to sit where the packaging tool puts it, and requires the match to be unique.

### Getting bytes onto disk

* **A symlink race in the in-place swap.** The staging file was `.NAME.new-PID` written with `fs::write`, which opens `O_CREAT|O_TRUNC` and follows symlinks. Where the install directory is writable by more than its owner, pre-creating that path as a symlink made the updater write 0755 content wherever it pointed. Now: unpredictable name, `create_new` + `O_NOFOLLOW`, 0600 while writing, mode widened on the descriptor once complete, `sync_all` before the rename and an fsync of the parent after.
* **The staged package's directory depends on who installs it.** One we install ourselves goes in a private temp directory that disappears afterwards. One the *user* will install, from a printed command, goes under their cache directory instead — its path is printed to the terminal, so it is not a secret, and temp reapers age `/tmp` out, after which any local user can recreate that path with their own bytes.
* **Running as root no longer needs `sudo` to exist.** Dropping the `doas` probe was right — doas and sudo are different authorization mechanisms, so choosing by what is on `$PATH` makes the update a property of the machine. But `euid == 0` is not another mechanism, it is the absence of a need for one; the command that runs is identical either way.
* **Decompression is bounded.** xz reaches ratios past 1000:1, so a download the HTTP layer considered small was an unbounded `read_to_end`.
* **A pre-release now precedes its own release** (semver §11.3). Stripping the suffix made `0.4.4-rc1` and `0.4.4` compare equal, stranding every pre-release user.

### Structure

`crates/fresh-editor/src/services/updater.rs` and `app/update_prompt.rs` are gone. `fresh-update` gained `endpoint`, `net`, `feed`, `staging`, `elevate`, `archive`, `engine` and `offer`; `engine` is a cargo feature (`net` + `archive`), so a build that only resolves provenance pulls in neither the TLS stack nor the extraction crates. The editor keeps the popup rows, the status-bar indicator and the background-check threading.

## How to Test

Needs `cargo build --bin fresh` and Docker. Each block is self-contained. The endpoint overrides in step 4 work in a **debug** build; a release build refuses them unless built with `--features fresh-update/insecure-endpoints`.

**1. The core bug — provenance resolves**

```bash
docker run --rm -v "$PWD/target/debug/fresh:/usr/bin/fresh:ro" ubuntu:24.04 sh -c '
  mkdir -p /usr/share/fresh-editor
  printf "schema = 1\nchannel = \"apt\"\n" > /usr/share/fresh-editor/install-receipt.toml
  fresh --cmd config paths | grep -E "^(install channel|provenance):"'
```

Expected `APT (.deb) (apt)` / `Authoritative`. On master: `unknown` / `Unknown`.

**2. AUR is the same command on every machine**

```bash
docker run --rm -v "$PWD/target/debug/fresh:/usr/bin/fresh:ro" archlinux:latest sh -c '
  mkdir -p /usr/share/fresh-editor
  printf "schema = 1\nchannel = \"aur-bin\"\n" > /usr/share/fresh-editor/install-receipt.toml
  fresh --cmd config paths | grep ^"update mechanism"
  printf "#!/bin/sh\n" > /usr/local/bin/yay && chmod +x /usr/local/bin/yay
  fresh --cmd config paths | grep ^"update mechanism"'
```

Expected: both lines identical (`git clone` plus `makepkg`). On master the second becomes `yay -S fresh-editor-bin`, and the first names a binary that isn't installed.

**3. Every channel's route, at a glance**

```bash
for ch in apt dnf zypper flatpak aur pacman homebrew winget nix mise npm cargo \
          tarball appimage snap scoop chocolatey source unknown; do
  printf "%-12s %s\n" "$ch" "$(FRESH_INSTALL_CHANNEL=$ch ./target/debug/fresh --cmd config paths \
    | sed -n 's/^update mechanism: *//p' | cut -c1-60)"
done
```

Only `scoop`, `chocolatey`, `source` and `unknown` should mention the releases page. `flatpak` names `flatpak update`, `snap` names `snap refresh`.

**4. The in-editor UX**

The build is already at the latest released version, so nothing is offered by default. Serve a feed claiming a newer one:

```bash
mkdir -p /tmp/fresh-feed
echo '{"tag_name":"v0.9.9","assets":[]}' > /tmp/fresh-feed/release.json

PORT=$(python3 -c 'import socket;s=socket.socket();s.bind(("127.0.0.1",0));print(s.getsockname()[1]);s.close()')
python3 -m http.server "$PORT" --bind 127.0.0.1 --directory /tmp/fresh-feed >/dev/null 2>&1 &
FEED_PID=$!
trap 'kill $FEED_PID 2>/dev/null' EXIT
FEED="http://127.0.0.1:$PORT/release.json"

# Sanity-check the feed before launching the TUI: expect "Latest version:  0.9.9"
FRESH_RELEASES_URL="$FEED" ./target/debug/fresh --cmd update --check

# The background check is debounced to once a day. Note the doubled directory —
# data_dir already ends in /fresh and the stamp path appends "fresh" again.
rm -f "${XDG_DATA_HOME:-$HOME/.local/share}/fresh/fresh/telemetry_stamp"

FRESH_RELEASES_URL="$FEED" ./target/debug/fresh
```

The status bar shows `Update: v0.9.9`; click it. Set `FRESH_INSTALL_CHANNEL` before launching to see each channel's wording and row set — `apt` offers all three rows and says a password will be asked for, `homebrew` offers two and names the command, `flatpak` names the host command, `unknown` offers only details. "Show the update command" exits `ActionRequired` without fetching anything, so the indicator reads "Update needs a command". Delete the stamp again between runs.

If no indicator appears, the `--check` line isolates the feed from the editor. If that prints `0.9.9`, the remaining gates are the stamp file and `editor.check_for_updates` / `--no-upgrade-check`.

**5. Full install against the real release**

Serve the genuine v0.4.6 `.deb` plus a locally generated `.sha256`, install a 0.4.5 package built from this branch, then update. Expected `Unpacking fresh-editor (0.4.6-1) over (0.4.5-1)`, exit 0, `dpkg --audit` clean. Run one container per channel — ubuntu, fedora, tumbleweed, arch, debian — plus a tampered payload (aborts, binary untouched) and the three exit codes (0/2/1). `--print-command` should make no request for the artifact at all; `--download-only` should leave the verified package under the cache directory and print the install command.

## UI Changes

Status-bar indicator, same update, depending on channel:

| Situation | Before | After |
|---|---|---|
| `.deb` downloaded and verified, root install pending | Update failed — click for details | Update needs a command — click for details |
| brew/winget command only printed | Updated — restart fresh | Update needs a command — click for details |
| Flatpak | Update to v0.9.9 → command fails with ENOENT | Names `flatpak update` on the host |
| Update actually installed | Updated — restart fresh | unchanged |

The popup went from one row (`Update to vX`) to plan-derived rows, up to three of them, with body text naming how this copy was installed and whether a password will be asked for.

## Notes for the reviewer

* **Existing `.deb`/`.rpm` users need one manual update.** The fix has to be running to work, so anyone on a package built before this merges stays stuck until they install a post-fix release by hand. Worth a release-note line.
* **Nothing is signed, and the attestation config is inert.** `dist-workspace.toml` sets `github-attestations = true`, but `grep -rn attest .github/workflows/` returns nothing — the release workflow replaced cargo-dist and never carried that step over. So the only integrity control is a `.sha256` served from the same host as the payload, which detects corruption and nothing else. Pinning the host raises the bar to "compromise GitHub" rather than "compromise any host", but it is not a substitute. This is the largest remaining gap and it is deliberately out of scope here.
* **Whether to elevate at all is a real fork.** This branch runs `sudo` in the update terminal so the update finishes in one step. The opposite position — print the command, always, and ship an apt/dnf repo so the package manager's own signed chain is the integrity control — is what uv, gh, zed, VS Code and Docker Desktop all do, and there are several CVEs of the shape "privileged updater trusts a path". The new middle rung narrows the gap between the two positions but does not settle it.
* **snap, scoop and chocolatey are still `Manual`.** Deleting the `Channel` variants is safe — the receipt wire format is a string, and an unknown id already degrades to `Unknown` — but it is a public API change, so it is your call.
* **`nix profile upgrade` is left alone deliberately.** No single invocation works across Nix versions (`--all` from 2.20, a `.*` matcher before), and if the user installed from a locked flake ref no command can help. Picking one would be wrong on some machines.
* **The popup has not been driven by hand.** Step 4 is the recipe; the plan-to-offer mapping and the exit codes it keys off are unit-tested, the rendering is not.
* **No CI job exercises the update flow end to end.** Provenance resolution is asserted; nothing installs an old package, serves a new one and checks the package database moved. Everything in step 5 was done by hand in containers.
* **Unrelated oddity:** the telemetry stamp lands at `~/.local/share/fresh/fresh/telemetry_stamp` because `DirectoryContext::from_system()` appends `fresh` and `telemetry::stamp_file_path()` appends it again. Pre-existing, untouched here.

Follow-ups are tracked in #2918.


sinelaw:
CI is red on all three `test` jobs. I've re-run the failed jobs; the failures look like the pre-existing e2e flakiness rather than anything from this branch.

Four distinct failures, no two the same:

| platform | failure |
|---|---|
| ubuntu | TIMEOUT (180s) `e2e::code_tour_dock::test_next_key_advances_step` |
| macos | FAIL `e2e::plugins::live_diff::test_live_diff_virtual_line_anchored_to_correct_modified_line` |
| windows | TIMEOUT (180s) `e2e::code_tour_dock::test_step_range_is_highlighted_in_the_editor` **and** FAIL `e2e::plugins::review_diff_ux_bugs::test_issue2117_discard_hunk_with_no_trailing_newline` |

Three reasons to read this as flake rather than regression:

1. **Every `fresh-update` test passed on every platform** — all 25, including the new ones, visible in each job log. The failures are plugin/dock e2e tests; this branch touches none of that code.
2. **A regression would fail the same test everywhere.** Four different tests across three platforms is the opposite signature.
3. **`master` is currently red the same way.** Run [31026262433](https://github.com/sinelaw/fresh/actions/runs/31026262433) (commit `a3217c28`) fails macOS with a 180s TIMEOUT in `e2e::plugins::gutter::test_git_gutter_updates_after_save` — different test, same 180s-timeout-in-a-plugin-e2e pattern. #2864 also documented a `live_diff` macOS flake previously.

The 180s timeouts in particular suggest the e2e harness intermittently hangs under CI load rather than asserting anything false.

I'll keep watching and report what the re-run gives. If the same four fail again in the same places, that's not flake and I'll dig into it properly.

---

sinelaw:
CI status on `6f2fc67`: ubuntu green, fmt/clippy/doc/schema/check-no-plugins green, windows still running. macOS failed on the same known-flaky test as before:

```
TIMEOUT [ 180.594s] (5855/8470) fresh-editor::e2e_tests e2e::plugins::gutter::test_git_gutter_updates_after_save
Summary [1371.620s] 8470 tests run: 8469 passed (4 slow), 1 timed out, 160 skipped
```

That is the only failure — one 180s timeout, no assertion failures. Same test is red on master (run `31026262433`, commit `a3217c28`), so it is not caused by this branch. All 25 `fresh-update` tests passed, including the four `6f2fc67` touched (`delegated_commands_template_defaults`, `tool_refs_match_how_each_tool_actually_installed_it`, `install_commands_take_the_downloaded_file`, `reinstall_is_permitted_only_when_asked_for`).

Re-running the macOS job once windows finishes (GitHub rejects `rerun-failed-jobs` while the run is still in progress).

---

PR #2890, opened 2026-08-04
fix toml highlight is confusing #2887

### LINKED ISSUE #2887: toml highlight is confusing
Image

asukaminato0721:
fix #2887

Corrected multiline TOML array state handling and string highlighting.
Prevented bare dots from being classified as floats.
Preserved dotted-key, inline-table, boolean, and date highlighting.

### DIFF
--- a/crates/fresh-editor/src/grammars/toml.sublime-syntax
+++ b/crates/fresh-editor/src/grammars/toml.sublime-syntax
@@ -32,21 +32,30 @@ contexts:
         3: punctuation.definition.table.end.toml
 
   key-value:
-    - match: '([A-Za-z0-9_-]+|"[^"]*"|''[^'']*'')\s*(=)'
+    # Keep top-level keys anchored to the start of the line. Without the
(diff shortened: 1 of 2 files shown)


sinelaw (approved):
- Fixes a real reported bug (#2887): TOML multiline arrays with `=` inside string items (e.g. `"USER=root"`) were breaking highlighting by matching as new key-value pairs.
- Change is small, focused, and matches existing grammar/`highlight_engine.rs` test conventions; includes a regression test that reproduces the bug (per CONTRIBUTING's testing guideline).
- Interactively validated in a local build: multiline-array strings stay green (string), `package.metadata = { enabled = true, label = "v1.2" }` inline-table renders correctly (bool blue, string green), dates render as constants, floats/exponents render as numbers, and a bare `.` is no longer misclassified as a float — all matching the PR's claims.
- CI green (windows/macos/ubuntu tests, fmt, clippy, schema, doc).
- Minor nit (non-blocking): commit message is just "fix" — CONTRIBUTING.md asks for motivation, not just what changed.

---

PR #2894, opened 2026-08-04
Make replace-all finish on files with tens of thousands of matches

### LINKED ISSUE #2893: Replace-all with 60k matches does not complete
## Steps to reproduce

1. Create a 60 000-line file, one match per line:

   ```sh
   python3 -c "open('big.txt','w').write(''.join(f'line {i:06d}: some source code here\n' for i in range(60000)))"
   ```

2. `fresh big.txt`
3. Run **Replace** (`Ctrl+R`).
4. Search for: `source` → Enter. Status bar reads `Found 60000 matches for 'source'`.
5. Replace with: `SRC` → Enter.

## Actual

The replace does not complete. After 120 s the buffer still shows `some source code here`, the prompt line still shows `Replace 'source' with: SRC`, and the UI does not respond to further keys (`Enter`, `Esc`). The process sits at ~68 % CPU.

## Expected

The replace completes, or reports progress and remains interruptible.

## Notes

Reproduced on `claude/git-gutter-scrollbar-markers-qhb34v` (debug build) and again with that branch's scrollbar change compiled out, so it is not specific to those changes.

sinelaw:
Fixes #2893.

Replace-all on a 60 000-line file (one match per line) never completed — the process sat at full CPU and the UI stopped responding. The flow had **four independent quadratic behaviours**, each scaling with the match count, so at 60k matches each one on its own was doing billions of operations.

## The four hot spots

| Where | Before | After |
|---|---|---|
| `app/search_ops.rs` — collecting matches | asked the buffer for "next match after here" per occurrence, and each call re-reads a fresh 64KB chunk out of the piece tree | one streaming pass (`TextBuffer::find_all_in_range`) |
| `state.rs` → `model/marker_tree.rs` — marker adjustment | replayed edit by edit; a deletion must visit every marker at or after it, and the search highlights leave one marker per match | one batched pass that resolves every marker's new position from the edit list, then rebuilds the tree |
| `model/piece_tree.rs` — applying the bulk edit | every edit checked against every leaf; every split point scanned per piece | walking cursor over merged, disjoint delete ranges; binary search for a piece's split points |
| `app/event_apply.rs` — cursor shift | re-summed the whole delta list once per event, and every Delete/Insert carries the primary cursor's id | prefix-summed deltas + binary search |

## Measurements (debug build, 60 000 matches)

- collecting matches: 75s → 97ms
- marker adjustment: ~15 min (extrapolated from the quadratic curve) → 211ms
- piece-tree bulk apply: 70s → 515ms
- end-to-end replace inside the editor: 99s → 3.2s

## Tests

- `tests/e2e/issue_2893_replace_all_many_matches.rs` drives the reported flow (open the 60 000-line file, Ctrl+R, `source` → `SRC`) and asserts only on rendered output: the status line reporting every occurrence, and no match left in the visible text. Without these fixes it never finishes, so it hangs until nextest kills it; with them it passes in ~7s in a debug build.
- `prop_bulk_adjust_matches_sequential` pins the batched marker adjustment to the per-edit semantics. Point markers — all a `MarkerList` ever holds — are also compared against a tree that really did apply the edits one at a time; spanning intervals only against the shadow model, because whether the per-edit path shifts an end that straddles an edit depends on where the node sits in the tree.
- Two unit tests cover `find_all_in_range` against what the repeated single-match search returned, including matches straddling the 64KB chunk boundary and overlapping candidates.

`cargo test -p fresh-editor --lib` passes in full (3199 tests), as does `cargo fmt` and `cargo clippy --all-targets` (no new warnings). The e2e suite was run partially — ~1860 tests passed before I stopped it for time; the four failures seen (`auto_indent::test_config_indent_next_line_pattern_applies` and three `locale::*` tests) are in areas this change does not touch and I have not confirmed them against the base commit.

https://claude.ai/code/session_01KB36GpUMzfbQcWjvNaorfn

---
_Generated by [Claude Code](https://claude.ai/code/session_01KB36GpUMzfbQcWjvNaorfn)_

### DIFF
--- a/crates/fresh-editor/src/app/event_apply.rs
+++ b/crates/fresh-editor/src/app/event_apply.rs
@@ -391,15 +391,21 @@ impl Editor {
         }
         position_deltas.sort_by_key(|(pos, _)| *pos);
 
+        // Prefix-summed deltas so the shift for a position is a binary search
+        // rather than a walk over every edit. Re-summing the whole list per
+        // cursor event made a replace-all quadratic in the match count — one
+        // edit per match, one event per match (issue #2893).
+        let delta_positions: Vec<usize> = position_deltas.iter().map(|(pos, _)| *pos).collect();
+        let delta_prefix: Vec<isize> = std::iter::once(0)
+            .chain(position_deltas.iter().scan(0isize, |acc, (_, delta)| {
+                *acc += delta;
+                Some(*acc)
+            }))
+            .collect();
+
         // Helper: calculate cumulative shift for a position based on edits at lower positions
         let calc_shift = |original_pos: usize| -> isize {
-            let mut shift: isize = 0;
-            for (edit_pos, delta) in &position_deltas {
-                if *edit_pos < original_pos {
-                    shift += delta;
-                }
-            }
-            shift
+            delta_prefix[delta_positions.partition_point(|pos| *pos < original_pos)]
         };
 
         // Apply adjustments to cursor positions
--- a/crates/fresh-editor/src/app/search_ops.rs
+++ b/crates/fresh-editor/src/app/search_ops.rs
@@ -836,25 +836,18 @@ impl Editor {
                 .map(|m| (m.offset, m.len, m.replacement))
                 .collect()
         } else {
-            // Plain text mode - replacement is used literally
+            // Plain text mode - replacement is used literally.
+            // One streaming pass over the buffer: asking for the next match
+            // from scratch per occurrence re-reads a chunk each time, which
+            // dominated the whole replace on files with many matches.
             let state = self.active_state();
             let buffer_len = state.buffer.len();
-            let mut matches = Vec::new();
-            let mut current_pos = 0;
-
-            while current_pos < buffer_len {
-                if let Some(offset) = state.buffer.find_next_in_range(
-                    search,
-                    current_pos,
-                    Some(current_pos..buffer_len),
-                ) {
-                    matches.push((offset, search.len(), replacement.to_string()));
-                    current_pos = offset + search.len();
-                } else {
-                    break;
-                }
-            }
-            matches
+            state
+                .buffer
+                .find_all_in_range(search, 0..buffer_len, usize::MAX)
+                .into_iter()
+                .map(|offset| (offset, search.len(), replacement.to_string()))
+                .collect()
         };
 
         let count = matches.len();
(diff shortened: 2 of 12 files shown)


sinelaw:
CI is green on c7465c0 — all 8 checks.

Getting there took a re-run, so recording what the earlier red runs were, in case it matters for review:

**One was a real bug in this PR, caught by CI.** On the first push, macOS failed `prop_bulk_adjust_matches_sequential` — my own new property test, which is seeded per-run and hit a case my local run didn't. It found the batched marker path mishandling two edits at the *same* position: the per-edit path applies them in turn (a deletion clamps a marker, then an insertion moves it), while the batched path resolves each marker against a single governing edit and computed the net delta instead. Fixed in c7465c0 — `MarkerList` sends such a batch down the per-edit path, and `IntervalTree` documents and debug-asserts the requirement. A bulk edit merges its same-position edits before this point, so nothing reached it that way; the test generator was producing batches outside the contract, which is now fixed with a unit test covering the fallback.

**The rest were flakes.** On c7465c0 ubuntu timed out on `vi_mode_bugs::test_vi_bug_2441_cfr_changes_through_target` (180.7s) and macOS failed two `plugins::live_diff` tests. Both jobs passed on re-run of the same SHA, and windows had already passed the full suite first time. For what it's worth:

- The base commit this branch sits on (48f2de9c) is itself red on master with the same signature — `test ubuntu-latest` timing out at 180.3s, on a different e2e test (`code_tour_dock::test_step_range_is_highlighted_in_the_editor`). Master is red on 5 of its last 15 runs.
- All three tests pass locally on this branch (12/12 `live_diff`, 6/6 `vi_bug_2441`).
- `live_diff` anchors virtual lines to markers, so it was worth ruling out properly: `create_line_anchor` has no callers, so every marker in the codebase is a point marker (`pos..pos`), and for point markers the batched path is provably equivalent to the per-edit one — that equivalence is what `prop_bulk_adjust_matches_sequential` asserts against a tree that really did apply the edits one at a time. It can't move an anchor under any interleaving.

The one thing I could not fully rule out by argument is whether the new 60 000-line reproducer's CPU/memory spike raises the flake rate for timing-sensitive neighbours in the parallel nextest run. Evidence says no — windows passed the full suite with it on the first attempt, and both re-runs passed — but if you'd rather not carry that risk, 40 000 lines still runs minutes past the 180s external timeout without the fixes and would cut the test's footprint by about a third. Happy to shrink it if you prefer.

---

PR #2895, opened 2026-08-04
Tour dock: wheel/scrollbar/highlight/wrap fixes, markdown text-area description, native file picker, reclaimed layout

sinelaw:
Reworks the code-tour dock panel across several review rounds. Every issue was reproduced interactively in tmux before fixing (per CONTRIBUTING.md), and each fix carries a reproducer test asserting rendered output. Rebased onto master, which brought in the parallel key-race fixes (`targetTour()` fallback, `editor.flush()`).

## Interaction fixes

- **Wheel routing is pointer-aware.** Every keyed `List`/`Tree`/markdown document surfaces its geometry as a `ScrollRegion`; the wheel scrolls the widget under the pointer, for floating and split-mounted panels alike.
- **Overflowing panel lists paint scrollbars** (split-mounted panels previously dropped their scroll-region geometry), positioned on the section border, with the selection band clipped to the panel.
- **Long prose word-wraps** instead of ellipsis-truncating; wide glyphs (漢, 😀) no longer shift the section border (shared display-width helpers).
- **Clicking a list row focuses the owning list** — the click-to-focus check compared the row's *item* key against the tabbable set, so arrows after a rail click went to the prose. This was the last CI red (timeouts on all three platforms).
- **Keys can't leak into the source file during a step change**: `revealStep` hands focus back to the panel in the same FIFO turn as the `openFile`, and the panel repaints the new step immediately instead of after the reveal chain.

## The description is a real text area (`markdown` Text mode)

`WidgetSpec::Text` gains `markdown` (forcibly read-only, multi-line): the value renders through the same markdown engine as LSP hover docs (headings, emphasis, inline code, syntax-highlighted fences, new `•`/`N.` list markers) and word-wraps at the host's true width. A shadow `TextEdit` holds the rendered plain text 1:1 with the display: click places a visible block caret (the `reversed` overlay flag was being dropped), arrows/Shift-selection/drag/Home/End/PageUp/Down work, and Ctrl+C copies exactly what's on screen. Mutations are inert.

## Highlight and theming

- The step highlight is back to **one range-wide `extendToLineEnd` overlay**: the render sweep now tail-fills every row an overlay spans (not just its first), and blank-line indentation guides carry the overlay bg instead of punching holes in the band.
- All plugin colours and the step overlay use **theme keys** (`ui.semantic_highlight_bg` etc.), resolved at paint time.

## Loading a tour uses the native file browser

New plugin API `editor.pickFile(label)` — the terminal analogue of a browser's file-input dialog. It opens the real Open File browser (anchored to the active file's directory / window working dir, Backspace walks up the tree, Tab descends, hidden-file toggle) and resolves with the chosen path instead of opening a buffer; cancel resolves null. The browser's input row now renders the prompt's own label instead of a hardcoded "Open file: ". The tour's load command uses it in place of the old blind text prompt.

## Layout

The panel spends its rows on the prose: the full-width divider and the dedicated location row are gone, the step's `file · lines` location leads the description, and Jump-to-code / Re-highlight moved into the header. The Steps rail now shows from 110 columns (was 130). The showcase gif is re-recorded on the final layout.

## Testing

- 30 tour e2e tests (module green on all three platforms as of `173b0e1`), all asserting rendered output; unit tests for the overlay sweep, zipper flag scoping, markdown document rows, caret painting, and wide-glyph borders.
- `cargo check --all-targets`, fmt, clippy clean; `fresh.d.ts` regenerated.
- Verified interactively in tmux: wheel-by-hover, scrollbars, caret, drag-selection, copy, file-picker navigation, rail click + arrow browsing, full-width bands over blank/indented lines.

https://claude.ai/code/session_0151vsVPjZC39JgqD8kV4F8H


sinelaw:
CI note: `test macos-latest` went red on a single-test **timeout** — `e2e::plugins::review_diff_line_staging::test_review_visual_stage_modified_line_pair` (killed at 181s; the other 8390 tests passed). That module doesn't touch widget panels, lists, or any code this PR changes, and the full module passes locally in under 2 seconds, so this reads as a loaded-runner hang rather than a regression from this diff. Clippy, fmt, doc, schema, and check are green; ubuntu and windows are still running. A re-run of the macOS job should clear it — I'll keep watching and will dig in if it reproduces.

---

PR #2905, opened 2026-08-05
Tour commands: browse for the tour file, never guess

sinelaw:
Fixes the tour-opening UX regressions reported after #2902, in two commits:

## 1. `Tour: Load Definition...` opens its picker where tour files actually are

The pick browser anchored at the active file's directory with hidden files off — and tour files are dotfiles at the workspace root (`.fresh-tour.json`, `.tours/*.tour`), so the picker showed nothing pickable and a typed name resolved against the wrong directory (e.g. `src/`) and was rejected.

- `editor.pickFile(label, directory?, showHidden?)` — new optional parameters using the `Opt<Option<..>>` pattern (like `setPromptSuggestions`), so existing single-argument callers are untouched. A relative `directory` resolves against the window's working directory.
- Plumbed through `PluginCommand::StartFilePickAsync` into a new `init_file_open_state_at(dir_override, show_hidden_override)`; `Ctrl+O` keeps its exact behavior via the `None, None` path.
- The tour plugin anchors its picker at the workspace root with hidden files forced visible. `fresh.d.ts` regenerated.

## 2. `Tour: Open Workspace Tour...` always browses — never auto-opens a guess

With exactly one tour discovered, the command opened it immediately; a user who ran it to *choose* got whatever manifest happened to exist, and the multi-match suggestion popup could hand them the wrong file just as easily. The tour to open is now always the user's explicit pick: the command opens the same root-anchored, hidden-files-visible browser as `Tour: Load Definition...`. The discovery scan remains available to scripts via the plugin API's `discoverTours()`.

## Verification

- e2e `test_load_definition_picker_reaches_hidden_tour_files`: fails on the old anchoring twice over (`.fresh-tour.json` never appears; the typed name resolved against `src/` and was rejected); with the fix it selects the manifest by name and waits for the opened tour's settled step.
- e2e `test_workspace_tour_command_browses_instead_of_auto_opening`: fails on the auto-open behavior (the lone `.tours/` manifest opened with no browser at all); with the fix the browser appears with `.tours/` on screen and no tour open.
- Manually verified in a terminal session: the picker opens at the workspace root with **Show Hidden** checked, listing `.tours/` and `.fresh-tour.json`; typing the name opens the tour.
- Plugin typecheck clean; fmt clean. Full e2e module left to CI by request.

https://claude.ai/code/session_01FqsXpbxdKrTFhNxXfh9EuM


sinelaw:
CI status: the one PR-specific failure was `test_load_prompt_is_the_native_file_browser`, which still asserted the picker's old anchoring — fixed in 73e7214, and it passes on the latest run.

The remaining red `test macos-latest` is failing on the base branch too: each macOS run times out one arbitrary e2e test at 180s — on master @ db537d4 (this PR's merge base) it was `e2e::vi_mode_bugs::test_vi_bug_2441_cfr_changes_through_target`, on this PR's runs it was `e2e::plugins::gutter::test_git_gutter_refreshes_after_external_git_checkout`, all in subsystems this diff doesn't touch (recent master pushes show the same single-timeout pattern: 8440/8441 passed). Ubuntu — which runs the identical e2e suite, including the new tour tests — is green on the latest push. I'll re-check when the macOS runner situation recovers.

---

PR #2906, opened 2026-08-05
Re-mint the agent capability token when a terminal is restored

sinelaw:
## The bug

Create Orchestrator workspaces with terminals, quit the editor, reopen it, switch to a previously open workspace, and run `fresh --cmd script …` in the restored terminal:

```
$ echo "return 1+1;" | "$FRESH_BIN" --cmd script run -
no capability token: script evaluation is not authorized
```

A brand-new workspace's terminal works fine. Only restored ones are affected.

Reproduced manually in tmux against a real editor and real PTYs before changing anything: in the restored terminal `FRESH_CMD_TOKEN` is empty while `FRESH_SESSION` and `FRESH_BIN` are both set — so the agent can reach the editor but not drive it.

## Cause

The Orchestrator spawns every workspace's agent with `allowScript`, which mints a `FRESH_CMD_TOKEN` bound to that window and stamps it into the child's environment (`app/terminal.rs::agent_command_env`). The token table (`server/command_access.rs`) is in-memory and process-global by design — nothing about it is persisted.

The saved workspace recorded everything a terminal needs to come back — launch argv, agent-resume argv, transcript paths, title, exited state — *except* that it ever held the grant. So `restore_terminal_from_workspace` respawned the PTY with an empty extra-env map. `respawn_terminal_pty` (the exited-agent restart and the remote-reconnect sweep) had the same gap.

## Fix

Persist the **grant**, never the token:

- `SerializedTerminalWorkspace` gains `script_access: bool` (serde default, skipped when false — workspaces written before this field read back as no grant, the safe direction).
- `Window` gains `terminal_script_tokens: HashMap<TerminalId, String>`. Membership is the grant; the value is kept only so a respawn can revoke its predecessor's token instead of leaving it live in the table. Recorded at both spawn sites from the env `agent_command_env` built.
- Restore and respawn call `Window::remint_terminal_script_env`, which revokes the stale token and mints a new one bound to *this* window — a restored workspace is a new `WindowId`, and `FRESH_SESSION` rides along via the same helper. Token entries re-key onto the real terminal id exactly as the backing/log paths already do.
- `ExitedTerminal` carries `script_access` too, so restarting an agent that had exited re-arms it as well.

## Testing

New integration test `crates/fresh-editor/tests/terminal_restore_script_token.rs`. The terminal's child prints the `FRESH_CMD_TOKEN` it was actually handed, so the test reads back the value the restored agent would itself read, then checks it resolves, carries the script grant, and is bound to the restored window. Verified it fails without the fix (the restored child announces `none`) and passes with it.

`cargo fmt --check`, `cargo check --all-targets` and `cargo clippy --all-targets` are clean. No GUI code touched. Not yet re-run: the manual tmux flow against the patched binary, and the wider test suite.

---
_Generated by [Claude Code](https://claude.ai/code/session_013BFdGBXewxSHGuVDVmUUDb)_

### DIFF
--- a/crates/fresh-editor/src/app/async_dispatch.rs
+++ b/crates/fresh-editor/src/app/async_dispatch.rs
@@ -1236,6 +1236,7 @@ impl Editor {
                 .filter(|argv| !argv.is_empty())
                 .cloned(),
             ephemeral: window.ephemeral_terminals.contains(&terminal_id),
+            script_access: window.terminal_has_script_access(terminal_id),
             title: None,
         }
     }
--- a/crates/fresh-editor/src/app/plugin_dispatch.rs
+++ b/crates/fresh-editor/src/app/plugin_dispatch.rs
@@ -4087,7 +4087,7 @@ impl Editor {
                 persistent,
                 command: command.clone(),
                 title: title.filter(|t| !t.is_empty()),
-                env: terminal_env,
+                env: terminal_env.clone(),
             };
             let spawned = target.create_plugin_terminal(spec);
             // Record the launch/resume argv exactly as `create_window_with_terminal`
@@ -4097,6 +4097,9 @@ impl Editor {
             // respawned a bare shell instead of the agent.
             if let Ok((terminal_id, _, _)) = &spawned {
                 target.mark_terminal_restorable(*terminal_id, command, resume);
+                // File the token this terminal's child was handed, so workspace
+                // capture persists the grant and a restore re-mints it.
+                target.record_terminal_script_token(*terminal_id, &terminal_env);
             }
             spawned
         };
--- a/crates/fresh-editor/src/app/terminal.rs
+++ b/crates/fresh-editor/src/app/terminal.rs
@@ -162,6 +162,77 @@ pub(crate) fn agent_command_env(
     env
 }
 
+impl Window {
+    /// Remember which capability token `terminal_id`'s freshly-spawned child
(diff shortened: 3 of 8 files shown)


sinelaw:
CI on `42c407b`: 7/8 green — `fmt`, `clippy`, `doc`, `schema`, `check (no plugins)`, `test ubuntu-latest`, `test macos-latest`.

`test windows-latest` is red, and it looks pre-existing rather than caused by this PR:

- Here: 8092/8093 passed, one nextest 180s timeout on `fresh-editor::e2e_tests e2e::vi_mode_bugs::test_vi_bug_2441_cfr_changes_through_target`. Suite wall time 1983s.
- Current `master` (`09977e4d`, no changes from this PR) also fails `test windows-latest`, with three 180s timeouts on `fresh-editor::semantic_tests semantic::properties::{property_arbitrary_actions_do_not_panic, property_dispatch_is_deterministic, property_insert_only_undo_is_identity}`. Suite wall time 1464s.

Different tests, same failure mode: the per-test 180s cap being hit on the Windows runner. Nothing in this diff touches vi mode or the semantic property tests — it changes terminal capability-token plumbing, adds one `script_access` field to the workspace schema, and adds a test binary gated on `all(target_os = "linux", feature = "plugins")`, so it does not compile or run on Windows at all.

Will re-run against a green base when Windows recovers on `master`.

---

PR #2917, opened 2026-08-06
fix breadcrumb trail/navigation for buffer #2913

### LINKED ISSUE #2913: breadcrumb trail/navigation for buffer
> I thought about similar mechanism but for the buffers instead. like if you have ts/rs/MD file opened then top bar shows what is the current function/lsp object you are currently in and also shows steak of them if you are in the nested function/block/lsp symbol etc. just like vscode does it too. will provide more positional context for the user. 

 _Originally posted by @PavelLoparev in [#2705](https://github.com/sinelaw/fresh/issues/2705#issuecomment-5197029220)_

asukaminato0721:
fix #2913

---

Adds a breadcrumb trail under the tab bar showing the LSP symbol path at the cursor (e.g. `MyClass > myMethod`), updating live as you move and clickable to jump to that symbol.

Cursor inside `restock`:

```
 File   Edit   View   Selection   Go   LSP   Help
 inventory.py ×   Welcome ×   +
 InventoryStore > restock > current
▾ 1 │ class InventoryStore:
  2 │ ▏   """A tiny store, enough to give the LSP a symbol tree."""
  3 │ ▏
▾ 4 │ ▏   def __init__(self, items):
  5 │ ▏   ▏   self.items = items
  6 │ ▏   ▏
▾ 7 │ ▏   def restock(self, sku, count):
  8 │ ▏   ▏   current = self.items.get(sku, 0)
  9 │ ▏   ▏   self.items[sku] = current + count
 10 │ ▏   ▏   return self.items[sku]
 11 │ ▏   ▏
▾12 │ ▏   def total_value(self, prices):
 13 │ ▏   ▏   return sum(prices[s] * n for s, n in self.items.items())
 14 │ ▏   ▏
```

Cursor moved into `total_value`, trail follows:

```
 File   Edit   View   Selection   Go   LSP   Help
 inventory.py ×   Welcome ×   +
 InventoryStore > total_value
▾ 1 │ class InventoryStore:
  2 │ ▏   """A tiny store, enough to give the LSP a symbol tree."""
  3 │ ▏
▾ 4 │ ▏   def __init__(self, items):
  5 │ ▏   ▏   self.items = items
  6 │ ▏   ▏
▾ 7 │ ▏   def restock(self, sku, count):
  8 │ ▏   ▏   current = self.items.get(sku, 0)
  9 │ ▏   ▏   self.items[sku] = current + count
 10 │ ▏   ▏   return self.items[sku]
 11 │ ▏   ▏
▾12 │ ▏   def total_value(self, prices):
 13 │ ▏   ▏   return sum(prices[s] * n for s, n in self.items.items())
 14 │ ▏   ▏
```


sinelaw (changes requested):

- Interactively verified the feature works: built the branch, opened a TS file with a scripted LSP server, confirmed the breadcrumb row renders below the tab bar, updates live as the cursor moves between `MyClass > constructor` and `MyClass > myMethod`, and clicking the `MyClass` crumb jumps the cursor exactly to `Ln 1, Col 7` (the class name) without modifying the buffer. Also verified `editor.show_breadcrumbs: false` correctly hides the row.
- Build (`cargo build --release`) and `cargo clippy --all-targets` are clean on the changed files; new e2e test observes rendered output correctly per CONTRIBUTING.md.
- Blocking: `attachLineText` in `plugins/lsp_navigation.ts` now calls `editor.getBufferText(bufferId, 0, length)` to read the **entire buffer** into the plugin on every refresh (buffer activate, and 250ms after every insert/delete/revert), replacing the prior per-symbol line-range reads. This directly contradicts CONTRIBUTING.md's "Avoid full-buffer scans" guideline and risks stalling/OOMing on the multi-gigabyte files this editor advertises support for, since it now runs periodically during normal editing, not just on demand. Please rework to read only the touched lines (e.g. resolve the buffer-vs-symbol race some other way, such as a version/generation check, instead of snapshotting the whole text).

---

sinelaw:

Status: the change that was requested has already been made — the author did it on 2026-08-19, a couple of hours after the review, and the review was never updated. So the original blocker is resolved. I did find one separate bug while testing the feature; see the next comment.

sinelaw:
Tested the feature interactively against a Python file with pylsp. The trail renders, nests correctly, and follows the cursor. Clicking a crumb does jump the cursor to the right place (`class InventoryStore` at Ln 1, Col 7).

One bug though: after clicking a crumb, the trail itself does not refresh. It keeps showing the path you came from until you press any arrow key, then it corrects.

```
before click:  cursor 12:35   InventoryStore > total_value > s > n
after click:   cursor  1:7    InventoryStore > total_value > s > n   <- stale
after Right:   cursor  1:8    InventoryStore
```

Reproduced twice. Likely the programmatic jump does not emit the `cursor_moved` event the plugin refreshes on.

sinelaw:
Confirmed your diagnosis: the crumb click moves the caret through `set_buffer_cursor_in_splits`, which writes the view state directly and raises no `cursor_moved`, so the plugin never hears about the jump — the click path has no plugin code in it at all, so the fix has to be host-side. Fixed, along with two more the same path turned up: `cursor_moved.line` is 1-indexed while `getCursorLine()` is 0-indexed and the trail compared both against 0-indexed LSP ranges (so a caret on a symbol's last line dropped that symbol), and a failed symbol fetch was cached permanently, so a server started after the buffer opened never filled the trail in.

I could not push to this PR — the branch is on a fork this session has no write access to — so the work is in #3278, which carries @asukaminato0721's commit unchanged plus the fix on top. Three regression tests, each confirmed failing before and passing after.

Two things I left alone as your call, both noted on that PR: the eager per-symbol line-text attach on every activate and typing pause (lazy would be ~25 lines), and whether to filter the trail by `SymbolKind` — with pylsp it currently runs down to comprehension variables where VS Code stops at container kinds.

---

sinelaw:
Superseded by https://github.com/sinelaw/fresh/pull/3278/commits
Which includes this commits

PR #2927, opened 2026-08-07
fix(lsp): re-sync full-document changes as didClose+didOpen  (avoid at least Roslyn C# LSP crash)

sfjohansson:
**Bug:** Roslyn negotiates INCREMENTAL sync, then throws `NullReferenceException` on a full-document change (`range: None`). That faults its request queue and closes stdout — completion/hover die on the line you're typing, while hover still works on untouched code, so the server looks healthy.

**Fix:** in `handle_did_change_sequential` (the single chokepoint for every didChange) — when a change has `range: None` and the negotiated sync kind isn't FULL, re-sync as **didClose + didOpen**. didOpen carries the whole text with no range, valid under any sync kind.

- Generic LSP layer, **not** C#-gated — applies to any server that negotiated non-FULL sync, i.e. nearly all of them. Roslyn is just the one that crashes instead of silently tolerating the invalid payload.
- Fires only on full-document paths: file reload, workspace restore, plugin edits, `event_apply` resync.
- Per-keystroke ranged edits are untouched.

**Tests:** unit test over every `TextDocumentSync` capability shape, plus a behavioural one — under INCREMENTAL a full-document change resets the tracked version to 0 (proving close+open), a ranged change still bumps it, and under FULL it stays a didChange. Both fail without the fix.

**Refs:** [LSP 3.17 — Text Document Synchronization](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#text-document-synchronization) · [Roslyn `ProtocolConversions.cs`](https://github.com/dotnet/roslyn/blob/main/src/LanguageServer/Protocol/Extensions/ProtocolConversions.cs)

### DIFF
--- a/crates/fresh-editor/src/services/lsp/async_handler.rs
+++ b/crates/fresh-editor/src/services/lsp/async_handler.rs
@@ -149,6 +149,23 @@ fn log_response_error(code: i64, message: &str, server_name: &str, language: &st
     }
 }
 
+/// The change-sync kind the server negotiated (`textDocumentSync.change`).
+/// Unknown / unset defaults to `NONE`, which callers treat as "not FULL" — i.e. the
+/// server requires ranged incremental changes, so a full-document (`range: None`)
+/// change must be re-expressed (see `handle_did_change_sequential`).
+fn negotiated_change_sync_kind(
+    caps: Option<&ServerCapabilities>,
+) -> lsp_types::TextDocumentSyncKind {
+    use lsp_types::{TextDocumentSyncCapability, TextDocumentSyncKind};
+    match caps.and_then(|c| c.text_document_sync.as_ref()) {
+        Some(TextDocumentSyncCapability::Kind(k)) => *k,
+        Some(TextDocumentSyncCapability::Options(o)) => {
+            o.change.unwrap_or(TextDocumentSyncKind::NONE)
+        }
+        None => TextDocumentSyncKind::NONE,
+    }
+}
+
 /// Check if a document is already open and should skip didOpen.
 /// Returns true if the document is already open (should skip), false if it should proceed.
 fn should_skip_did_open(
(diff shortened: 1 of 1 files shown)


sinelaw (changes requested):
This is an autogenerated comment.

Validated interactively in tmux against a stub LSP server that negotiates INCREMENTAL sync and faults on `range: null` (wired via `--config`). No UT/e2e run.

- **The fix works.** Reload-from-disk on this branch sends `didClose` + `didOpen` (full text); reverting only this patch and rebuilding sends `didChange` with `range: null` on the identical step and kills the stub. Per-keystroke ranged edits still go out as `didChange`, and completion returns results after the re-sync.
- **Re-sync skips the `didOpen` grace period.** The new block sits *above* the `pending_opens` wait (`async_handler.rs:1429`), so a full-document change within 200 ms of a `didOpen` fires `didClose`+`didOpen` with no wait — the race that wait exists to prevent (Code Guidelines #3, LSP lifecycle). `notify_lsp_file_changed` (`file_operations.rs:1096` → `1130`) emits `didOpen` and the full-doc change back-to-back, so this is reachable on the fix's own primary path. Please move the block below the grace-period wait. The added test clears `pending_opens` first, so it doesn't cover this.
- **Narrow the change scan.** `.iter().rev().find(|c| c.range.is_none())` accepts a full-replace anywhere in a batch and then silently drops trailing ranged edits — a hazard the comment documents rather than removes (Code Guidelines #15). All five `range: None` sites (`workspace.rs:2238`, `plugin_commands.rs:3916`, `file_operations.rs:1130`, `event_apply.rs:224`/`588`) emit exactly one change; a single-element guard is simpler and makes the drop unrepresentable.
- **Rebase required.** Branch is 1318 commits behind `master` and conflicts in `async_handler.rs` (both sides appended tests at the same anchor — trivial to resolve).
- Title nit: stray double space and "avoid at least Roslyn…" reads awkwardly. The description itself is clear and well-scoped.

Approach and placement are right: `handle_did_change_sequential` is the single chokepoint for every `didChange`, and nothing else in the repo reads `textDocumentSync`, so this is not duplicating an existing path.

---

sinelaw:
This is an autogenerated comment — follow-up to the review above, adding upstream context I should have included the first time.

- **The motivating Roslyn bug is fixed upstream.** [dotnet/roslyn#77002](https://github.com/dotnet/roslyn/issues/77002) → [#78165](https://github.com/dotnet/roslyn/pull/78165), merged Apr 2025 (reverted once for Razor, since re-landed). Current Roslyn `main` handles a range-less change via `SumType<TextDocumentContentChangePartial, TextDocumentContentChangeWholeDocument>`, replacing `TextSpan(0, text.Length)`. Keeping the guard for pinned/older servers is still reasonable, but please reword the code comment — it asserts a live server defect that no longer exists.
- **Missing `openClose` guard.** The re-sync calls `handle_did_close` unconditionally. Neovim's equivalent reload path guards on `client:supports_method('textDocument/didClose')`. Compounding it, `negotiated_change_sync_kind` maps absent capabilities to `NONE`, which the caller treats as not-FULL — so a server advertising no `textDocumentSync` at all now gets a close+open it may not support, where it previously got a `didChange`.
- **Consider a ranged whole-document replace instead of close+open**: `range` = (0,0)..(end of previous content), `text` = new content. Unambiguously legal under INCREMENTAL, needs no `openClose`, works against old Roslyn too, and avoids the version reset to 0 plus the server-side diagnostics/state teardown that close+open causes on every plugin and bulk edit. This is effectively what VS Code emits on reload.
- Prior art, for whichever shape you pick: Neovim does close+open but only for buffer reload; [Helix](https://github.com/helix-editor/helix/blob/master/helix-lsp/src/client.rs) and [vscode-languageclient](https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/common/textSynchronization.ts) never emit a range-less change under INCREMENTAL at all — both derive ranges from their own changeset/edit model. The cleanest fix here is arguably to stop constructing `range: None` at the five call sites; most have old and new text on hand.

Previous review stands: the grace-period ordering and the `.rev().find()` scan are still the blocking items, plus the rebase.

---

sfjohansson:
This change patched the symptom rather than the cause. Roslyn was crashing on a range: None content change, and re-syncing the document with didClose + didOpen let the server recover. Another PR will be setup addressing the root cause if still valid.

yueyinqiu:
I apologize if this reply has bothered you, but this is the only discussion I found related to roslyn-lsp. I'm wondering if fresh currently supports roslyn-lsp? After switching to roslyn-lsp, I feel like many features are missing, perhaps due to its use of numerous proprietary protocols? Or do I need to configure it in a specific way to get the full experience?

sfjohansson:
So far I did get it to work but it seems to be fairly strict/unforgiving when it comes to editor integration.

For instance this change helped it to recover, so code completion continued to work.

Although it did not deal with the root of the issue. I do have another fix for that. I need to setup a PR for it...perhaps this weekend.

PR #2943, opened 2026-08-10
refactor: layer input handling — pure decision core + Editor as imperative shell

sinelaw:
## Summary

`app/input.rs` was 3,584 lines of `impl Editor` mixing four concerns: key routing, action execution, overlay focus queries, and agent-script evaluation. Decisions about what a key *means* were interleaved with the state mutations that carry them out, so none of that logic could be exercised without constructing an editor.

This PR splits it into an **Editor-free decision core** and a high-level **imperative shell**.

### The decision core — `src/input/router.rs` (new, low level)

Pure functions that never see `Editor`. Each takes only the narrow inputs it needs — the key event, the `KeybindingResolver`, a small read-only view struct — and returns a decision value the shell executes:

| Decision | Replaces |
|---|---|
| `layout_reading` | which reading of a chord wins (physical vs layout character) |
| `widget_panel_key` → `WidgetKeyOutcome` | the entire 480-line floating-panel/dock key handler's logic |
| `mode_key_disposition` → `ModeKeyDisposition` | the mode-binding stage (mode chords/bindings, text-input capture, clipboard forwarding, Shift+nav widget selection, read-only blocking) |
| `chord_or_key` → `ChordDisposition` | the final chord-then-single-key stage |
| `unfocused_popup_action` / `completion_popup_action` | popup key precedence |
| `should_dismiss_transient_popup` | hover/signature-help dismissal exceptions |
| `cancels_pending_lsp` | the LSP-cancellation predicate |

Being pure, they carry **14 unit tests that run without constructing an editor** — mode capture/forward/selection precedence, two-key chord walks, dock Esc semantics, dropdown key ownership, modal-swallows vs dock-blurs-through, clipboard chords on text widgets, completion gating, transient-popup exceptions.

### The shell — `src/app/input.rs` (high level)

`Editor::handle_key` is now pure orchestration: each stage either delegates to a dedicated dispatcher or *builds a view → asks the router → applies the named effect*.

### Supporting moves

- **`src/app/action_dispatch.rs`** — `Editor::handle_action`. Executing a resolved `Action` drives buffers, windows, prompts and plugins, so it stays a method; the split keeps *deciding what a key means* separate from *doing what the action says*.
- **`src/app/overlay.rs`** — gains the overlay-stack builder and the shared focus queries, plus a pure `popup_blocked_by_higher_modal` beside the existing layer predicates. This PR extends the pattern that file already established (Editor builds a `Layer` list; pure functions decide precedence).
- **`src/app/agent_scripts.rs`** — `eval_agent_script`, which was never input handling.
- **`Window::claim_next_key`** — the plugin `getNextKey` queue/buffer operation moves onto `Window`, where that state lives.

## Notes for review

Behaviour-preserving; the `handle_action` match and the untouched pipeline stages are byte-identical to the original. Two details are deliberate:

- chord state is cleared only when a mode actually ran its resolution, so Normal-context chord prefixes survive the mode stage;
- `ChordDisposition` separates `Chord` from `Resolved` because a completed chord never cancelled in-flight LSP requests.

The router's tests pin the default keymap so they assert router semantics rather than host-OS defaults — `Config::default()` selects the `macos` keymap on macOS, where Ctrl+A is `MoveLineStart`, not `SelectAll` (same pinning convention as the keybinding-resolver tests).

Rebased onto latest master; two pre-existing clippy `collapsible_match` warnings are fixed as a side effect of the rewrite.

## Testing

- `cargo check` / `cargo clippy --all-targets` / `cargo fmt --check` (`--all-features`) clean — no warnings in touched files
- `cargo test --all-features --lib` — **3,277 passed, 0 failed**

https://claude.ai/code/session_014cyMxXzFLeS94CrnXuXbJf

### DIFF
--- a/crates/fresh-editor/src/app/action_dispatch.rs
+++ b/crates/fresh-editor/src/app/action_dispatch.rs
@@ -1,0 +1,2051 @@
+//! Action execution: carrying out a resolved [`Action`] against the editor.
+//!
+//! Once the input orchestration (`app/input.rs`) — or a menu, a mouse
+//! click, a plugin, a macro — has resolved an [`Action`], this is where
+//! the editor performs it. Executing an action is inherently high-level
+//! `Editor` behaviour (it drives buffers, windows, prompts, plugins), so
+//! unlike the decision layers in `crate::input::router` these are methods;
+//! the split from `app/input.rs` keeps *deciding what a key means*
+//! separate from *doing what the action says*.
+
+use super::*;
+use anyhow::Result as AnyhowResult;
+use rust_i18n::t;
+
+impl Editor {
+    /// Change the current workspace's trust level, persist it, and report it.
+    /// The new policy applies live at the next authority-routed spawn (the
+    /// guarding spawners read the level on every spawn) — there is NO editor
+    /// restart here, deliberately: a rebuild would reset every other
+    /// orchestrator session's buffers/layout (see the body). Trust-gated work
+    /// re-triggers via the `trust_changed` hook instead (e.g. env-manager
+    /// re-activates a now-trusted env). Already-correct selections (e.g.
+    /// confirming the current level) only persist the decision.
+    pub(crate) fn set_workspace_trust_level(
+        &mut self,
+        level: crate::services::workspace_trust::TrustLevel,
+    ) {
+        use crate::services::workspace_trust::TrustLevel;
+        // Trust is a per-window gate: each `Window` owns its own authority +
+        // `WorkspaceTrust` (issue #2280), and the guarding spawners read the
+        // level live at spawn time. Writing the new level here is the whole
+        // change — `set_level` itself documents "no rebuild required". The
+        // next authority-routed spawn (LSP, terminal command, task, formatter,
+        // plugin `spawnProcess`) honours the new level automatically.
+        //
+        // We deliberately do NOT `request_restart` here: that tears down and
+        // rebuilds the *entire* editor — every orchestrator session window,
+        // not just this one — which discarded other sessions' buffers and
+        // reset the layout when toggling a single session's trust (the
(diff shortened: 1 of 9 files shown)


sinelaw:
CI note: `test macos-latest` is red, but the failure is a pre-existing macOS flake on the base branch, not something this PR introduces. I've re-run the failed job.

**On this PR** (`9bcd812`), macOS reported `8692 passed, 2 timed out` — both hitting the 180s nextest wall:
- `e2e::command_palette::test_quick_open_goto_line_live_preview_mouse_click_commits`
- `e2e::plugins::review_diff_line_staging::test_review_visual_stage_modified_line_pair`

**On master** at `d4ff6c8` — the exact commit this branch is rebased onto — the macOS job failed the same way, with a *different* test:
- `8679 passed, 1 timed out` → `e2e::dock_switch_wipes_outgoing_window::dock_switch_never_paints_two_workspaces_into_one_frame`

Why I'm confident it isn't this PR:

1. **Master alone reproduces it** — same job, same 180s timeout mode, ubuntu and windows green in both runs.
2. **A different test times out each run**, which is the signature of a scheduling/contention issue rather than a deterministic routing regression.
3. **ubuntu passed the full 8694-test suite on this exact commit.**
4. **Both tests pass locally in under 3s** (2.26s and 0.92s) — a 60×+ gap from the timeout wall.
5. **Both passed on macOS at the previous commit of this branch**, which already contained the entire input refactor; the only delta since is the rebase onto master.

Per CONTRIBUTING's "be wary of fixes you can't reproduce", I haven't invented a code change for this — there's nothing in this diff to fix. Worth a separate look at macOS e2e stability on master (these tests use semantic waiting, so a timeout means an awaited state never arrived under load).

---

PR #2945, opened 2026-08-10
keybindings: context-specific bindings outrank global ones; gate mnemonics at resolution (#2941)

### LINKED ISSUE #2941: Questions about keybindings
_I have used Claude in working through this issue, but while it was certain there was a bug, I'm not, so I've tried backing down on its language here until I have a better understanding of it._

`KeybindingResolver::resolve` probes the **global** tier before the **context** tier. So when a keymap binds a chord both globally (as a broad fallback) and in a narrower context (for a specific job), the global one seems to win and the narrower one never gets a chance to fire — effectively dead entries, with no warning anywhere.

Is this intentional? 

I'm trying to define a custom keymap that has alt+right as move next word, but due to Reasons:TM: alt+f and alt+right overlap with the incoming esc character. I'm fine with losing the File Menu open in exchange, but it looks like I'd have to kill the Alt+f with a noop first, because of the ordering of the code here.

## Root cause (according to Claude)

`KeybindingResolver::resolve` ([`crates/fresh-editor/src/input/keybindings.rs:2058`](https://github.com/sinelaw/fresh/blob/v0.4.7/crates/fresh-editor/src/input/keybindings.rs#L2058)) probes:

1. custom global
2. custom context
3. custom parent context
4. **default global** ← wins
5. **default context** ← appears unreachable for any chord already in (4)
6. plugin defaults (context)
7. default parent context (e.g. `SearchPrompt` → `Prompt`)
8. Normal fallthrough

Steps 4 and 5 look like the bug. Reading the same ordering, (4) also comes before (6) and (7), which would put a global entry ahead of plugin mode bindings and parent-context bindings as well — though I haven't exercised those two paths.

sinelaw:
## Summary

Fixes #2941: `KeybindingResolver::resolve` probed the **global** tier before the **context** tier, so any chord bound in both resolved to the global entry and the narrower binding was unreachable — the default keymap itself shipped four such dead/hacked-around entries (`Alt+E/F/G/H`), and disabling `menu_bar_mnemonics` turned Alt+F into a dead key instead of freeing it as its docs promise.

## The precedence rule

`resolve` and `resolve_chord` now share one probe order (`probe_order`), defined by two rules:

1. **Source**: every user binding outranks every built-in one (as #2720 established).
2. **Specificity**: within the built-ins, the context chain is walked narrowest-first — exact context, ancestors (`SearchPrompt → Prompt`), then **Global as the root fallback**. A broad `global` entry never shadows a context-specific binding for the same chord. At equal specificity the active keymap outranks plugin `defineMode` defaults; a plugin's mode bindings still outrank the keymap's global entries.

## What that makes possible

- **The prompt Alt+Char bypass is deleted** (`input_dispatch.rs` `resolve_in_context_only` pre-resolve): the browser/prompt Alt toggles (encoding, hidden files, live-grep scopes) now win over menu mnemonics by ordering alone. File-browser actions resolved this way route back through `handle_file_open_action`.
- **`menu_bar_mnemonics: false` actually frees the keys**: the mnemonic bindings are suppressed at *resolution* (skipped as unbound) rather than resolving to a `MenuOpen` that dispatch silently dropped. The dispatch-side gate is removed — which also un-breaks a user's `menu_open` binding on a non-Alt chord.
- **`resolve_chord` inherits the same order**, fixing its leftover pre-#2720 ordering (default-global chords outranked custom context chords) and gaining parent-context chain support.
- **Keymap cleanup**: the dead `normal alt+f → move_word_right` entry is removed from `default.json` (`Ctrl+Right → move_word_end` already covers forward word motion), so Alt+F keeps opening the File menu on the default keymap.
- **Reachability guard**: `test_builtin_keymap_bindings_are_reachable` asserts every entry in every shipped keymap fires in its own context — a keymap can no longer ship silently-dead bindings.

## Behavior changes to be aware of

- **macOS**: `macos.json` deliberately binds `Alt+F`/`Alt+B` to word movement in `normal`, so on the macos keymap **Option+F now moves by word instead of opening the File menu** — the keymap data's stated intent, and the convention Option-as-Meta terminal users expect (this is exactly the issue reporter's scenario). Menus stay reachable via mouse, other mnemonics, and bound menu keys. If mnemonic-priority is preferred on macOS instead, the fix is one line: drop those two entries from `macos.json` like this PR did for `default.json`.
- In prompts, behavior is unchanged (the deleted bypass already delivered context-wins); with mnemonics disabled, Alt+letter chords go from dead to free.

## Testing

- 6 new resolver tests covering all three reproduced scenarios (default-tier shadowing, mnemonics-off dead keys, custom-tier shadowing), plugin-vs-keymap precedence, chord-order parity, and the keymap reachability guard.
- Full lib suite (3236) and e2e suite (3378) pass locally; the two e2e failures observed in a full parallel run are pre-existing global-locale races (pass in isolation, guarded by `lock_locale()`).
- Verified interactively in tmux against a release build: stock Alt+F/Alt+B, file-browser Alt+E / Alt+. toggles, search-prompt Alt+R/Alt+W toggles, mnemonics-off with a user Alt+F binding firing, and custom global-vs-normal precedence.
- Docs updated: `docs/internal/input-keybindings-actions.md` (resolution order), `docs/features/keybinding-editor.md` (user-facing precedence note).

https://claude.ai/code/session_01LcUuPBvHFxWp1tqPJes4Cm


sinelaw:
CI status: everything is green except `test macos-latest`, which is failing on the base branch too — master's tip run ([`5c0d61c`](https://github.com/sinelaw/fresh/actions/runs/31339498217)) fails the same job with a single arbitrary e2e victim (8623/8624 passed there). On this PR's latest run the job reached **8629/8630 passed**; the one miss is a 180s timeout in `test_quick_open_goto_line_live_preview_mouse_click_commits` (a mouse-path test, unrelated to keybindings), and the timing-out test has been a different one on each of the last three macOS runs — a slow-runner flake, not a deterministic failure. Will re-check when the job re-runs or the base recovers.

---

Issue #2244, opened 2026-06-04
LSP hover (and other hover popups) still appear while the LSP info popup is open


sinelaw:
## Description

When the **LSP info popup** is open — the one shown by clicking the **\"LSP\"** indicator in the status bar — symbol hover popups (LSP hover) and other hover-related popups still appear and overlap it. While that popup is open, no hover popup should be shown.

## Steps to reproduce

1. Open a file with an active LSP server.
2. Click the **LSP** indicator in the status bar to open the LSP info popup.
3. Move the mouse over a symbol in the editor.

**Expected:** No hover popup appears while the LSP info popup is open.
**Actual:** The LSP hover popup shows up (and can overlap the LSP info popup).

## Notes / pointers

The LSP info popup is `PopupResolver::LspStatus` (`crates/fresh-editor/src/app/popup_dialogs.rs`, shown via `show_lsp_status_popup`).

LSP hover already has a suppression guard for other popups in `update_lsp_hover_state` (`crates/fresh-editor/src/app/mouse_input.rs:699`), which currently checks `theme_info_popup`, `tab_context_menu`, and `file_explorer_context_menu` — but not the `LspStatus` popup in the popup stack. The fix is likely to extend that guard (and/or `check_mouse_hover_timer`) to also suppress hover when an `LspStatus` popup is on top of the stack, and to make sure all hover-related popups are covered, not just LSP symbol hover.

cc @biscuitvicious

PR #2952, opened 2026-08-10
lsp: unlock auto-import completions; suppress hover under modal dialogs

### LINKED ISSUE #2603: LSP completions never offer unimported symbols, so the documented "auto-imports are applied when you accept a completion" can never trigger
`docs/features/lsp.md` says: *"Code completion: Auto-imports are applied when you accept a completion."* Tested on v0.4.3 (built from `origin/master` @ `11dccfad5`) with rust-analyzer 1.94.1 (Full mode, `cargo check` pre-warmed, workspace Trusted, `LSP (rust) ready`): the auto-importable completion items are never offered in the first place, so the accept path can't be exercised at all.

## Steps to reproduce
1. In a small cargo crate with **no `use` statements** in `src/main.rs`, inside `fn main()`, type `let m = HashMa`.
2. Press `Ctrl+Space` (also tried the palette command "Show Completions").
3. For contrast, try the controls below in the same buffer/session.

## Expected behavior
As in VS Code with the same rust-analyzer version: the popup offers `HashMap` tagged with its import (`use std::collections::HashMap`), and accepting it inserts both the identifier and the `use` line at the top of the file (rust-analyzer's default `completion.autoimport.enable = true` / "flyimport").

## Actual behavior
**No completion popup appears at all** — not for the prefix `HashMa`, not for the exact ident `HashMap` (retried over several seconds, multiple times, valid syntax `let m = HashMap;` too). The same is true for a trait-method that would need an import (`p.type_i` → nothing, where VS Code offers `type_id (use std::any::Any)`).

Controls in the same session proving completion machinery works:
- `p.` + `Ctrl+Space` → member completions (`x i32 (Tab)`, …) ✓
- `Strin` + `Ctrl+Space` → `String` and `stringify!` (both prelude — no import needed) ✓
- `std::collections::HashMa` + `Ctrl+Space` → `HashMap` (fully qualified — no import needed) ✓

So completions work exactly as long as no import would be required; every candidate that would need an auto-import is missing from the results. The documented auto-import behavior is therefore unreachable, at least for Rust.

## Workaround
Type the fully qualified path (or write the `use` line manually).

---
Search performed before filing (0 results each): `auto-import completion`, `flyimport unimported symbol HashMap`, `completion missing import use statement`.

### LINKED ISSUE #2912: Mouse hover popups appear over modal dialogs (Open File, command palette)
## Description

While a modal dialog is open (Open File / command palette), moving the
mouse over the dialog area still triggers LSP hover requests for the
editor buffer behind it. Hover popups render on top of the dialog.

The popups update as the pointer moves, so these are live requests, not
a stale popup left over from before the dialog opened.

The buffer position also doesn't correspond to what's visually under the
pointer — the popups describe symbols that are scrolled out of view.

## Related

Possibly the same class as #2244 (hover popups over the LSP status popup),
fixed in 0.3.12 — but that fix appears specific to the LSP status popup
rather than modals in general. The Open File dialog still leaks.

May relate to the #692 flicker fix, where hover only dismisses when the
mouse leaves the editor area entirely — a dialog drawn over that area
wouldn't trigger dismissal.

## Steps to reproduce

1. Open a C file with an LSP server attached (clangd).
2. Press Ctrl+O to open the Open File dialog.
3. Move the mouse over the dialog area.
4. Hover popups appear for symbols in the buffer behind the dialog.

## Expected

No hover requests while a modal dialog has focus.

## Actual

Hover popups appear over the dialog and update as the mouse moves.

## Workaround

Disable `editor.mouse_hover_enabled`.

## Environment

- fresh 0.4.6
- macOS 26.6 (25G72) (Apple Silicon), iTerm2
- Keymap: macos
- LSP: clangd

Image

Image

sinelaw:
Fixes #2603
Fixes #2912

Three LSP-client fixes, one commit each.

## 1. Auto-import completions were unreachable (#2603)

`create_client_capabilities()` sent `CompletionClientCapabilities` with no
`completionItem` sub-capabilities. rust-analyzer gates its unimported-symbol
candidates ("flyimport") on the client declaring
`completionItem.resolveSupport.properties ⊇ ["additionalTextEdits"]`, so every
candidate that would need an auto-import was silently dropped and the
documented "auto-imports are applied when you accept a completion" flow could
never trigger. The client machinery already applied `additionalTextEdits`
both eagerly and via `completionItem/resolve` — only the advertisement was
missing.

The initialize capabilities now declare:

- **`resolveSupport.properties: ["documentation", "additionalTextEdits"]`** —
  `additionalTextEdits` is what unlocks flyimport and what the accept path
  consumes (`apply_completion_additional_edits` /
  `handle_completion_resolved`). `documentation` is required in practice:
  validated against rust-analyzer 1.94.1, with `additionalTextEdits` alone it
  still offers the candidate and still defers the import edit to resolve, but
  reports `completionProvider.resolveProvider: false` — so a spec-compliant
  client never resolves and the import is silently lost. Deferring
  documentation costs nothing because the completion popup renders
  label/detail only, never `documentation`.
- **`labelDetailsSupport: true`** — with resolve support on, rust-analyzer
  moves the `(use std::collections::HashMap)` annotation into `labelDetails`;
  without this flag it folds that text into the label itself, which the
  accept path would insert literally into the buffer. The popup now renders
  `labelDetails` as the dimmed detail next to the (bare-identifier) label.

Deliberately **not** advertised, because the client does not implement them:

- `snippetSupport` — the accept path detects snippet syntax only
  heuristically (`is_snippet`) and does not track `insertTextFormat`, so
  inviting servers to switch every insertion to snippet format is unsafe.
- other `resolveSupport` properties (`detail`, `command`, …) — the resolve
  response is only consumed for its additional edits.

## 2. Hover popups fired through modal dialogs (#2912)

While a modal dialog (Open File, command palette, menu, …) was open, mouse
motion over the dialog kept arming and firing LSP hover requests for the
buffer *behind* it (positions map straight through to covered content), and
the hover popups rendered on top of the dialog.

Both ends of the hover pipeline are now gated on the overlay stack the input
routers already consult (`Editor::overlay_layers`) via a new
`modal_overlay_active()` predicate — no parallel notion of modality:

- `update_lsp_hover_state` (mouse motion) no longer tracks hover — and
  dismisses any transient hover popup — while a modal overlay is active;
- `check_mouse_hover_timer` no longer fires a request for a hover armed just
  before the modal opened.

The popup band is excluded from the predicate (the hover/completion popups
themselves live there and keep their existing popup-aware handling); the dock
and the editor base are not modal. This generalizes the #2244/#2245 fix,
which covered only the LSP-status popup.

## 3. Accepting a completion applied another candidate's import

Found while testing this PR: with flyimport unlocked, the popup offers several
rows sharing one label —

```
S HashMap (use oxc_allocator::HashMap)
S HashMapStrategy (use proptest::collection::HashMapStrategy)
S HashMap (use std::collections::HashMap)
```

— and accepting the last one inserted `use oxc_allocator::HashMap`.

The accept path passed only the selected row's **label** to
`apply_completion_additional_edits`, which recovered the item with
`completion_items.iter().find(|item| item.label == label)`: the first
candidate sharing the label, not the selected one. Its eager
`additionalTextEdits` were applied — and that same wrong item was the one sent
to `send_completion_resolve`, which is the path that actually decides the
import for rust-analyzer since it defers `additionalTextEdits` to
`completionItem/resolve`. Latent before this PR (without unimported symbols,
duplicate labels were rare); advertising flyimport makes them the norm.

Accept is identity-based now. `Editor::build_completion_popup_rows` builds the
popup rows and records the candidate behind each LSP row
(`Window::completion_popup_lsp_items`) in one pass, and the accept path
resolves the selected row index through that mapping. A bare index into
`completion_items` would not be enough — the popup is a prefix-filtered view
with buffer-word rows appended, so popup row ≠ stored index; recording the
mapping where the rows are built is what makes the index meaningful. All three
popup-building paths (first response, type-to-filter rebuild, buffer-words
only) go through that one function. The `completionItem/resolve` *response*
handler needed no change: it applies the edits from the item the server
returned rather than re-finding one by label.

## Tests

All fail without their fix (verified by stashing the src changes):

- `client_capabilities_advertise_completion_resolve_additional_text_edits`,
  `client_capabilities_advertise_completion_label_details` (unit,
  `async_handler.rs`) — initialize capabilities contain
  resolveSupport/labelDetails.
- `lsp_items_to_popup_items_renders_label_details`,
  `lsp_items_to_popup_items_falls_back_to_detail` (unit, `popup_actions.rs`)
  — labelDetails renders as detail, label/insert text stay bare.
- `test_completion_offers_unimported_symbol_and_applies_auto_import` (e2e) —
  drives a fake LSP that replicates rust-analyzer's observed gating (both the
  flyimport item gating on `additionalTextEdits` and `resolveProvider` gating
  on `documentation`): `HashMa` + Ctrl+Space must offer
  `HashMap (use std::collections::HashMap)` on screen, and accepting must
  insert the `use` line via `completionItem/resolve` (times out without the
  fix — the candidate is never offered).
- `test_no_hover_requests_while_modal_dialog_open` (e2e) — control hover
  works, then with the Open File dialog up neither fresh mouse motion nor a
  pre-armed hover timer produces a hover (failed at the first suppression
  assertion without the fix).
- `test_accepting_second_same_label_candidate_applies_its_own_import`,
  `test_resolve_deferred_import_follows_the_selected_candidate` (e2e) — a new
  fake server (`FakeLspServer::spawn_duplicate_labels`) offers two candidates
  both labelled `HashMap`, the wrong one first, with the import delivered
  eagerly in one test and through `completionItem/resolve` (keyed on the
  item's `data`, as rust-analyzer does) in the other. Down + Tab must leave
  `use std::collections::HashMap;` on screen and never
  `use wrong_crate::HashMap;`. Both insert the wrong import without the fix.

Ran: the four unit tests, all five new e2e tests, the LSP/completion e2e
modules (`-- e2e::lsp_completion e2e::lsp::`, 122 passed) and the `--lib`
`popup_actions` / `async_handler` / completion unit tests (70 passed),
`cargo fmt`, `cargo clippy -p fresh-editor --all-targets` (no new warnings on
the touched files), plus manual validation of the first two fixes against real
rust-analyzer 1.94.1 in tmux with a debug build: `HashMa` + Ctrl+Space offers
`HashMap (use std::collections::HashMap)`, accepting inserts
`use std::collections::HashMap;` at the top of the file; with the Open File
dialog open, mouse motion over the dialog produces no hover popups, while the
same motion without the dialog does.

https://claude.ai/code/session_01D6SkGXXTcJsytTjF1TBf8Y

---
_Generated by [Claude Code](https://claude.ai/code/session_01D6SkGXXTcJsytTjF1TBf8Y)_

### DIFF
--- a/crates/fresh-editor/src/app/async_dispatch.rs
+++ b/crates/fresh-editor/src/app/async_dispatch.rs
@@ -269,12 +269,9 @@ impl Editor {
                 } => {
                     self.handle_lsp_code_action_resolved(action);
                 }
-                AsyncMessage::LspCompletionResolved {
-                    request_id: _,
-                    item,
-                } => {
+                AsyncMessage::LspCompletionResolved { request_id, item } => {
                     if let Ok(resolved) = item {
-                        self.handle_completion_resolved(resolved);
+                        self.handle_completion_resolved(request_id, resolved);
                     }
                 }
                 AsyncMessage::LspFormatting {
--- a/crates/fresh-editor/src/app/editor_accessors.rs
+++ b/crates/fresh-editor/src/app/editor_accessors.rs
@@ -1431,6 +1431,16 @@ impl Editor {
             return false;
         }
 
+        // Suppress hover while a modal overlay (Open File dialog, command
+        // palette, menu, …) covers the editor: the tracked position maps to
+        // the buffer behind the overlay, and the resulting popup would render
+        // on top of the dialog (sinelaw/fresh#2912). This also covers a hover
+        // armed just before the modal opened — the pending timer must not
+        // fire underneath it.
+        if self.modal_overlay_active() {
+            return false;
+        }
+
         let hover_delay = std::time::Duration::from_millis(self.config.editor.mouse_hover_delay_ms);
 
         // Get hover state without borrowing self
--- a/crates/fresh-editor/src/app/lsp_requests.rs
+++ b/crates/fresh-editor/src/app/lsp_requests.rs
@@ -134,35 +134,14 @@ impl Editor {
         }
 
         // Get the partial word at cursor to filter completions
-        use crate::primitives::word_navigation::find_completion_word_start;
-        let cursor_pos = self.active_cursors().primary().position;
-        let (word_start, cursor_pos) = {
-            let state = self.active_state();
-            let word_start = find_completion_word_start(&state.buffer, cursor_pos);
-            (word_start, cursor_pos)
-        };
-        let prefix = if word_start < cursor_pos {
-            self.active_state_mut()
-                .get_text_range(word_start, cursor_pos)
-                .to_lowercase()
-        } else {
-            String::new()
-        };
+        use crate::app::popup_actions::completion_matches_prefix;
+        let prefix = self.completion_word_prefix();
 
-        let matches_prefix = |item: &lsp_types::CompletionItem| -> bool {
-            prefix.is_empty()
-                || item.label.to_lowercase().starts_with(&prefix)
-                || item
-                    .filter_text
-                    .as_ref()
-                    .map(|ft| ft.to_lowercase().starts_with(&prefix))
-                    .unwrap_or(false)
-        };
+        let any_new_match = items
+            .iter()
+            .any(|item| completion_matches_prefix(item, &prefix));
 
-        let filtered_items: Vec<&lsp_types::CompletionItem> =
-            items.iter().filter(|item| matches_prefix(item)).collect();
-
-        if filtered_items.is_empty() && self.active_window().completion_items.is_none() {
+        if !any_new_match && self.active_window().completion_items.is_none() {
             tracing::debug!("No completion items match prefix '{}'", prefix);
             return Ok(());
         }
@@ -178,32 +157,24 @@ impl Editor {
             }
         }
 
-        // Rebuild popup from ALL merged items (not just the new batch)
-        let all_items = self.active_window_mut().completion_items.as_ref().unwrap();
-        let all_filtered: Vec<&lsp_types::CompletionItem> = all_items
+        // Only the *LSP* results open the popup from this path; a response
+        // that leaves nothing matching is not a reason to show buffer-word
+        // candidates. Checked before rebuilding so a late non-matching
+        // response can't drop the mapping behind a popup already on screen.
+        let any_stored_match = self
+            .active_window()
+            .completion_items
             .iter()
-            .filter(|item| matches_prefix(item))
-            .collect();
-
-        if all_filtered.is_empty() {
+            .flatten()
+            .any(|item| completion_matches_prefix(item, &prefix));
+        if !any_stored_match {
             tracing::debug!("No completion items match prefix '{}'", prefix);
             return Ok(());
         }
 
-        // Build LSP popup items, then append buffer-word items below.
-        let mut all_popup_items =
-            crate::app::popup_actions::lsp_items_to_popup_items(&all_filtered);
-        let buffer_word_items = self.get_buffer_completion_popup_items();
-        // Deduplicate: skip buffer-word items whose label already appears in LSP results.
-        let lsp_labels: std::collections::HashSet<String> = all_popup_items
-            .iter()
-            .map(|i| i.text.to_lowercase())
-            .collect();
-        all_popup_items.extend(
-            buffer_word_items
-                .into_iter()
-                .filter(|item| !lsp_labels.contains(&item.text.to_lowercase())),
-        );
+        // Rebuild popup rows from ALL merged items (not just the new batch),
+        // recording which candidate each LSP row came from.
+        let all_popup_items = self.build_completion_popup_rows();
 
         let popup_data =
             crate::app::popup_actions::build_completion_popup_from_items(all_popup_items, 0);
@@ -674,7 +645,7 @@ impl Editor {
                 self.active_window_mut().send_lsp_cancel_request(request_id);
             }
         }
-        self.active_window_mut().completion_items = None;
+        self.active_window_mut().clear_completion_items();
 
         // Get the current buffer and cursor position
         let cursor_pos = self.active_cursors().primary().position;
@@ -737,7 +708,10 @@ impl Editor {
     ///
     /// Called when no LSP servers are available for the current buffer.
     fn show_buffer_word_completion_popup(&mut self) {
-        let items = self.get_buffer_completion_popup_items();
+        // Goes through the shared row builder so the popup-row → LSP-item
+        // mapping is reset (to "no LSP rows") by the same code that fills
+        // it — no separate place to forget.
+        let items = self.build_completion_popup_rows();
         if items.is_empty() {
             return;
         }
--- a/crates/fresh-editor/src/app/lsp_requests.rs
+++ b/crates/fresh-editor/src/app/lsp_requests.rs
@@ -15,6 +15,7 @@ use rust_i18n::t;
 use std::io;
 use std::time::{Duration, Instant};
 
+use crate::app::window::LspCompletionCandidate;
 use crate::model::event::{BufferId, Event};
 use crate::primitives::word_navigation::{find_word_end, find_word_start};
 use crate::view::prompt::{Prompt, PromptType};
@@ -108,18 +109,20 @@ impl Editor {
         request_id: u64,
         items: Vec<lsp_types::CompletionItem>,
     ) -> AnyhowResult<()> {
-        // Check if this is one of the pending completion requests
-        if !self
+        // Check if this is one of the pending completion requests. Removing
+        // it also names the server that answered — the candidates below are
+        // only resolvable against that one.
+        let Some(server) = self
             .active_window_mut()
             .pending_completion_requests
             .remove(&request_id)
-        {
+        else {
             tracing::debug!(
                 "Ignoring completion response for outdated request {}",
                 request_id
             );
             return Ok(());
-        }
+        };
 
         if items.is_empty() {
             tracing::debug!("No completion items received");
@@ -146,14 +149,22 @@ impl Editor {
             return Ok(());
         }
 
-        // Store/extend original items for type-to-filter (merge from multiple servers)
+        // Store/extend original items for type-to-filter (merge from
+        // multiple servers). This is the merge point, so it is also where
+        // each candidate's origin has to be stamped on: once several
+        // servers' results sit in one list, nothing else can tell them
+        // apart.
+        let candidates = items.into_iter().map(|item| LspCompletionCandidate {
+            item,
+            server: Some(server),
+        });
         match &mut self.active_window_mut().completion_items {
             Some(existing) => {
-                existing.extend(items);
+                existing.extend(candidates);
                 tracing::debug!("Extended completion items, now {} total", existing.len());
             }
             None => {
-                self.active_window_mut().completion_items = Some(items);
+                self.active_window_mut().completion_items = Some(candidates.collect());
             }
         }
 
@@ -166,7 +177,7 @@ impl Editor {
             .completion_items
             .iter()
             .flatten()
-            .any(|item| completion_matches_prefix(item, &prefix));
+            .any(|candidate| completion_matches_prefix(&candidate.item, &prefix));
         if !any_stored_match {
             tracing::debug!("No completion items match prefix '{}'", prefix);
             return Ok(());
@@ -636,6 +647,7 @@ impl Editor {
                 .active_window_mut()
                 .pending_completion_requests
                 .drain()
+                .map(|(request_id, _server)| request_id)
                 .collect();
             for request_id in ids {
                 tracing::debug!(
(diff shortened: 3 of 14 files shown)


sinelaw:
## Fixed: accepting a completion applied another candidate's import

Thanks for the repro — that popup is the failure mode exactly:

```
S HashMap (use oxc_allocator::HashMap)
S HashMapStrategy (use proptest::collection::HashMapStrategy)
S HashMapValueTree (use proptest::collection::HashMapValueTree)
  HashMarker (use sha2::digest::HashMarker)
S HashMap (use std::collections::HashMap)
```

Selecting the last row inserted `use oxc_allocator::HashMap`.

### Root cause

The accept path only carried the selected row's **label** down to the edit application:

- `handle_popup_confirm` (`popup_actions.rs`) read the selected row's `text` + `data` and passed the label to `apply_completion_additional_edits(&str)`.
- That function recovered the item with `completion_items.iter().find(|item| item.label == label)` — the **first** candidate sharing the label, never the one that was highlighted.

So the first `HashMap` row's `additionalTextEdits` were applied — and, more importantly for rust-analyzer, that same wrong item was the one handed to `send_completion_resolve`. Since rust-analyzer defers `additionalTextEdits` to `completionItem/resolve`, the resolve path is what decides the import in practice, and the server dutifully resolved the wrong candidate's `use` line.

Latent before this PR: without flyimport, no unimported symbols were offered, so duplicate labels were rare. Advertising them makes duplicates the norm.

### Fix

Accept is identity-based now. `Editor::build_completion_popup_rows` builds the popup rows and records the candidate behind each LSP row (`Window::completion_popup_lsp_items`) in one pass, in row order; the accept path resolves the selected **row index** through that mapping.

An index into `completion_items` alone would not have been enough: the popup is a prefix-filtered view of the stored items with buffer-word rows appended, so popup row ≠ stored index. Recording the mapping where the rows are built is what makes the index meaningful — and all three popup-building paths (first response, type-to-filter rebuild, buffer-words-only) go through that one function, so no second place can order rows differently. Mouse clicks already set the row index before confirming, so they take the same path.

The `completionItem/resolve` **response** handler was checked too: it applies the edits from the item the server returned rather than re-finding one by label, so sending the right item is sufficient there.

### Tests

Two e2e tests drive the real popup against a new fake server that offers two candidates both labelled `HashMap`, the wrong one first (`FakeLspServer::spawn_duplicate_labels`):

- `test_accepting_second_same_label_candidate_applies_its_own_import` — eager `additionalTextEdits`.
- `test_resolve_deferred_import_follows_the_selected_candidate` — no eager edits; the import comes back from `completionItem/resolve`, keyed on the item's `data`, which is rust-analyzer's actual behaviour.

Both press <kbd>Down</kbd> then <kbd>Tab</kbd> and assert on rendered output that the buffer gained `use std::collections::HashMap;` and *not* `use wrong_crate::HashMap;`. Verified both fail on the unpatched branch (the wrong import is inserted). Also ran the PR's existing LSP/completion e2e modules (122 passed) and the `--lib` `popup_actions` / `async_handler` / completion unit tests (70 passed), `cargo fmt`, and clippy on the touched files (no new warnings).

### Noted, not changed here

`send_completion_resolve` picks the first server advertising `completionItem/resolve` for the language rather than the server the item came from — with several completion servers merged into one list that can resolve an item against the wrong server. Separate from this bug; happy to file it.


---

sinelaw:
Two more places in this flow identified a completion by something that isn't its identity. Both pushed to this branch.

**1. The highlight moved while you typed.** `refilter_completion_popup` restored the selection after each keystroke by looking the previously selected row up by its **label**. With duplicate labels — which this PR makes routine by advertising auto-import candidates — the highlight silently snapped back to the *first* row sharing the label. So picking `std::collections::HashMap` and then typing the next character of the word put you back on `oxc_allocator::HashMap`, and since accept follows the highlight, the wrong import landed anyway. The re-filter now carries the selection across by resolving the selected row through the same `completion_popup_lsp_items` mapping the row builder records, and matching candidates rather than labels; a candidate the narrower prefix filters out falls back to the first row as before.

Buffer-word rows deliberately keep label identity: they carry no payload beyond the text they insert, so two rows with the same text are interchangeable. They're matched only *among* the buffer-word rows, though — an LSP row sharing that label is a different candidate with an import attached.

**2. `completionItem/resolve` went to the wrong server.** With several servers configured for one language (fresh merges their completions into one popup), the accept path sent the resolve to the first server advertising resolve support, whichever candidate was accepted. `CompletionItem::data` is an opaque handle only its own server can interpret, and resolve is where the auto-import comes from — so a sibling server answers with its own `use` line. Candidates now carry the id of the server that offered them, stamped on at the merge point in `handle_completion_response` where the origin is still known; `send_completion_resolve` addresses that server and checks *its* resolve capability. Resolve responses are also matched back to the outstanding request now, like the other LSP response handlers.

Tests, both observed failing before their fix: the selection one drives Down + one more character against the duplicate-label fake server and asserts on the rendered selection background that the highlight stayed on the second candidate (and that accepting still applies its import); the routing one configures two fake servers for `rust`, each offering a `HashMap` whose import only it can produce, selects the second server's row and asserts its `use` line is the one that lands.

---

sinelaw:
Pushed two more commits: the selection regression this PR introduced, and the `-32800` log spam.

## 1. Regression: `test_completion_type_to_filter_preserves_selection`

Reproduced on the branch as-is before touching anything (`left: Some("test_alpha")`, `right: Some("test_beta")`), so the fix below is against the reported failure rather than a rewritten test.

**Cause.** Carrying the selection across a re-filter by *candidate identity* is right, but only `build_completion_popup_rows` records the row → candidate mapping in `Window::completion_popup_lsp_items`. A popup shown any other way — a plugin supplying its own list, or a test injecting `PopupData` directly, as this test does — has **no** mapping. Every one of its rows therefore fell into the buffer-word branch of `completion_row_of`, and that branch searched only the rows *past* the LSP mapping:

```rust
rows.iter().enumerate().skip(lsp_rows.len()).find(|(_, row)| row.text == *text)
```

With the mapping empty on entry, `selected_completion_row()` returns `BufferWord("test_beta")`; `build_completion_popup_rows` then fills the mapping with all three LSP candidates, so `skip(3)` skips exactly the rows the selection is in. Lookup yields `None`, `.unwrap_or(0)` resets to row 0 = `test_alpha`. (`Popup::selected_index()` is fine — the popup tracks the selection correctly; it is the re-find that fails.)

**Fix.** Rows with no candidate recorded behind them (`SelectedCompletionRow::Text`, renamed from `BufferWord` since it is no longer only buffer words) now match by text across the whole list. The duplicate-label fix is untouched: identity semantics only buy anything where duplicates actually exist, which is the LSP rows, and those still match by candidate. Widening the text search also cannot quietly promote a plain buffer word into an auto-import candidate — `build_completion_popup_rows` drops any buffer word whose text an LSP row already occupies, so the two never coexist in one popup.

**Test rule 2.** The failing test asserted via `harness.editor().active_state().popups.top().selected_item()` — model state, not the screen. It now locates the rows on the rendered screen and checks which one carries `popup_selection_bg`, matching the `row_is_highlighted` helper the 2952 tests already use, plus a negative assert that `test_alpha` is *not* highlighted. Verified the rewritten test fails on the unpatched tree at the "highlight must stay on the row the user picked" assert, and passes with the fix.

## 2. `-32800 canceled by client` warning spam

```
WARN LSP response error from 'rust-analyzer' (rust) for textDocument/semanticTokens/range: canceled by client (code -32800)
```

Nothing is wrong here: `Window::cancel_pending_lsp_requests` withdraws superseded requests via `$/cancelRequest` on every cursor move, edit and scroll (and the 30s request timeout does the same), and `-32800 RequestCancelled` is the spec-mandated reply to exactly that. `is_suppressed_error_code` already downgraded the two neighbouring routine codes (`-32801 ContentModified`, `-32802 ServerCancelled`) but not this one, so we were warning about our own bookkeeping.

Added `LSP_ERROR_REQUEST_CANCELLED` to the suppressed set with a doc comment in the same style, and extended `test_suppressed_error_codes`. Sanity-checked that nothing relied on it surfacing: `-32800` / `RequestCancelled` appears nowhere else in the tree, and only the log level changes — the response still reaches its caller unchanged. Confirmed the unit test fails without the change.

## Tests run

Selected by area — the tests that exercise `refilter_completion_popup` (reached from `handle_popup_type_char` / `handle_popup_backspace`) and `completion_row_of`, i.e. anything that types or backspaces with a completion popup open, plus the buffer-word popup path. Not by name prefix; that is how this regression slipped through last round (it was named `test_completion_type_to_filter_*`, not `test_lsp_completion_*`).

16 e2e tests, all passing:

- `e2e::lsp::test_completion_type_to_filter_preserves_selection`, `..._basic`, `..._uppercase`, `..._closes_on_no_match`, `e2e::lsp::test_completion_backspace_refilters`
- all four `e2e::lsp_completion_duplicate_label_import_2952::*` (including the two that guard the duplicate-label fix)
- `e2e::lsp_completion_popup_behavior::test_completion_word_chars_still_filter`, `..._underscore_filters`, `..._popup_format_consistent_after_filter`
- `e2e::issue_1901_multicursor_with_completion_popup::test_multicursor_type_char_with_completion_popup`, `..._backspace_...`
- `e2e::dabbrev_completion::test_popup_buffer_words_without_lsp`, `test_popup_buffer_word_accept`

Plus 3 lib unit tests: `services::lsp::async_handler::tests::{test_suppressed_error_codes, test_request_failed_suppressed_only_for_informational_methods, test_request_failed_on_hover_is_not_logged_as_warn}`.

`cargo fmt` clean; clippy clean on the three touched files.

---

PR #2960, opened 2026-08-10
refactor: lift pointer hit-testing out of Editor into a pure layer

sinelaw:
Follow-on to #2943, applying the same decision-core/imperative-shell split to the mouse side.

## The problem

`mouse_input.rs` carried the "what is under the pointer?" ladder as `impl Editor` methods reading `self.active_layout()` / `self.active_chrome()` directly. The precedence rules — which overlay wins where two of them cover the same cell — could only be exercised by constructing a whole editor, so they never were. These are the rules that surface to users as "the button doesn't work".

There were also **two implementations of the same probes**: `compute_menu_dropdown_hover` (menu_actions.rs) and `compute_file_browser_hover` (file_open_input.rs) each duplicated hit-tests the hover ladder also performed.

## The change

`app/hit_test.rs` (new, no `impl Editor`) holds the ladder as free functions:

- `hover_target` = `floating_overlay_target` (context menus → suggestions → popups → file browser) then `chrome_target` (menu bar → open dropdown → explorer → separators → split controls → tabs → scrollbars → status bar → search options).
- `menu_dropdown_target` and `file_browser_target` live here too; their former `impl Editor` wrappers now delegate, so each probe has **one** implementation instead of two.

The view borrows **exactly** the rects and slices the decision reads — deliberately not the layout caches themselves. Naming each field means the shell has to state what it hands over, a newly-added layout field cannot silently drift into the hit-test, and tests construct precise values rather than blanking a cache and trusting that the zeros are meaningful.

One probe cannot be pure: the file-explorer **status indicator** resolves its columns through the theme, the plugin decoration/slot caches and the explorer renderer's slot layout. It stays on `Editor` as `explorer_status_indicator_at` and is passed in pre-resolved — so the *ordering* still lives in the ladder even though that probe's geometry does not.

## Notes for review

- Behaviour-preserving: the ladder is the same sequence of the same comparisons, in the same order.
- `mouse_input.rs` drops ~400 lines net; `PopupAreaLayout` is re-exported `pub(crate)` for the view's popup slice.
- **10 new unit tests** cover precedence rules that had no coverage at all before, because reaching them previously required a whole editor: overlay-beats-chrome-underneath, topmost-popup-wins, scroll-offset suggestion rows (and past-the-end fall-through), explorer close button vs status indicator vs resize border, split controls on the tab row, scrollbar thumb vs track, separator orientation, status-bar segment span.

## Still ahead (not in this PR)

`handle_mouse_click` is a **second ~1,100-line ladder over the same geometry** — menu bar, tabs, scrollbars, status bar, explorer, split controls — with hit-test and effect fused together. Two hand-maintained ladders over identical rects is the drift risk the overlay `Layer` work was created to eliminate. With the pure ladder in place, the click path can become *resolve target → match → apply effect*, the same shape `handle_key` now has.

## Testing

- `cargo check` / `cargo clippy --all-targets` / `cargo fmt --check` (`--all-features`) clean — no warnings in touched files
- `cargo test --all-features --lib` — **3,313 passed, 0 failed**

https://claude.ai/code/session_014cyMxXzFLeS94CrnXuXbJf

---
_Generated by [Claude Code](https://claude.ai/code/session_014cyMxXzFLeS94CrnXuXbJf)_

### DIFF
--- a/crates/fresh-editor/src/app/file_open_input.rs
+++ b/crates/fresh-editor/src/app/file_open_input.rs
@@ -855,62 +855,19 @@ impl Editor {
 
     /// Compute hover target for file browser
     pub fn compute_file_browser_hover(&self, x: u16, y: u16) -> Option<super::types::HoverTarget> {
-        use super::types::HoverTarget;
-
+        // The ladder lives in the pure hit-test layer; this shell supplies
+        // the layout plus the list state (scroll offset, entry count) that
+        // the row probe needs.
         let layout = self.active_window().file_browser_layout.as_ref()?;
-
-        // Check "Show Hidden" checkbox first (priority over navigation shortcuts)
-        if layout.is_on_show_hidden_checkbox(x, y) {
-            return Some(HoverTarget::FileBrowserShowHiddenCheckbox);
-        }
-
-        // Check "Detect Encoding" checkbox
-        if layout.is_on_detect_encoding_checkbox(x, y) {
-            return Some(HoverTarget::FileBrowserDetectEncodingCheckbox);
-        }
-
-        // Check navigation shortcuts
-        if layout.is_in_nav(x, y) {
-            if let Some(idx) = layout.nav_shortcut_at(x, y) {
-                return Some(HoverTarget::FileBrowserNavShortcut(idx));
-            }
-        }
-
-        // Check column headers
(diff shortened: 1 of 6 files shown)


sinelaw:
superseded by migration to retained mode fresh-ui

PR #2962, opened 2026-08-10
Every agent reports in, and you answer from the dock

sinelaw:
Makes Fresh a control centre for a machine full of agents: a mailbox any agent can reach — whoever started it, whatever vendor built it — and a line in the dock to answer them on.

Design: `docs/internal/agent-control-plane.md`.

> **Stacked:** the full-screen **Home** view built on all of this — the agent list over the selected agent's live terminal, beside the chat — is **#2991**, in draft. It was lifted out so this PR is the control plane and the dock rather than a second view built on them. The host's `pane` widget stays here: it is a capability, and its only consumer is what moved.

## The mailbox

One directory per **agent**, in the checkout it works in:

```
<worktree>/.fresh/agents/<name>/
    status        one line — "<state> <summary>" — rewritten in place
    inbox/        one file per instruction for it; inbox/done/ when acted on
    outbox/       one file per thing it wants to say; outbox/read/ once shown
```

Per agent, not per worktree: two agents routinely share a checkout, and one mailbox per root has them overwriting each other's status and racing for each other's instructions. The directory name is also the **address**.

In the checkout rather than the data dir, because that is a path both sides can compute without coordinating: an agent someone started by hand knows its working directory and nothing about the plugin's layout. `.fresh/.gitignore` contains `*`, so the directory and the ignore file both stay out of `git status` (verified, not assumed).

- **`status`** is a state, and it supersedes screen inference wherever it exists — an agent saying `waiting …` outranks any guess made from its terminal, for the glyph *and* the reason text. Screen inference stays as the fallback, so agents that write nothing behave exactly as before.
- **`outbox/`** is a conversation. `status` is one line the agent overwrites; "I'm done pushing the 3 PRs" is said once, at a moment, and must still be there two minutes later. The drain appends to the transcript attributed to the agent and moves the file to `outbox/read/`, which is what makes it idempotent.
- **`inbox/`** is what sending writes.

Agents get three verbs rather than a directory layout to memorise — `fresh --cmd agent status|say|inbox --take` — over the already-authorised script channel, so an agent needs to know nothing about where any of this lives. The env vars (`$FRESH_AGENT_NAME`, `$FRESH_AGENT_INBOX`, …) remain for agents that prefer the files, and `fresh --cmd help agents` documents the convention for the ones we did not launch.

## The chat, in the dock

```
┌ Orchestrator ──────────────────────┐
│ [ New Task… ▾ ]  [Search Tasks   ] │
│ [ ▸ Filters ]  [ view: compact ]   │
├────────────────────────────────────┤
│   · proj                           │
│   * proj-1 · claude                │
│                                    │
│        ↑↓ switch   →← fold         │
│       Enter edit   F2 menu         │
├────────────────────────────────────┤
│ chat · → @proj/claude-main       □ │
│ you → @proj/claude-main            │
│ run the whole suite and report     │
│ which cases fail on the base       │
│ ────────────────────────────────── │
│ @proj/claude-main                  │
└────────────────────────────────────┘
```

**The address is chosen, not typed.** An earlier revision took a leading `@name` out of free text, which meant a typo sent your message nowhere and you found out later. Now typing filters a candidate list and there is no message field at all until you pick one — so an unaddressed message is unrepresentable. `Alt+C` puts the caret in the chat, Tab reaches it through the ordinary focus cycle, Backspace at the name reopens the picker.

Addresses are `workspace/agent`, because a name is only unique inside its own checkout and two workspaces routinely hold a `claude` each. When two workspaces would contribute the same half, every colliding one grows a further path segment until they differ — so `a/fresh/claude` and `b/fresh/claude`, never two identical rows.

**Sending wakes the addressee.** The mailbox is delivery, not receipt: a message to a workspace that is not loaded — every workspace after a restart — used to land in `inbox/` and wait for a turn that was never coming. Sending now restores the workspace, waits for its agent, and focuses its terminal. Nothing is typed at the agent; the inbox file is still the only delivery, so nothing can double-deliver and nothing can answer a permission prompt with a pasted sentence.

The distinction that took the longest to find, and which the UI does not show: a **dormant persisted window** has its agent's argv saved with the workspace, so materialising it brings the agent back and is worth waiting for. A **worktree discovered on disk** has no persisted window and nothing to resume — opening it yields a plain shell. Waiting there would spend the whole timeout to learn nothing, so the wake skips it and says so in the chat.

**Only agents that exist are offered.** A mailbox is evidence an agent *was* somewhere, not that one is there now: the `status` file outlives the process that wrote it. Detection is gated on the same predicate the wake uses — is there a workspace here the host is keeping, live or dormant — so what the picker offers and what sending can deliver are one answer rather than two that drift. Genuinely ambiguous cases stay listed: two mailboxes in one checkout, where neither can be attributed to that window's terminal but both are real.

## The control plane

- **`describeEnvironment()` / `describeWorkspace()`** — every window, its terminals and its panes, so a plugin can answer "what is going on" without being the active window.
- **Terminals addressed by window**, so reads and writes work across workspaces rather than only the focused one.
- **`orchestrator.fleet()`** — who is doing what, most urgent first, with the agent's own words where it wrote them and a guess from its screen otherwise. One call for "who needs me".
- **`editor.emitEvent`** — plugin-to-plugin events on the same bus as host events, so `agent_state_change` is subscribable with `editor.on`.
- **`fresh --cmd script run --as NAME`** replaces a previously-submitted script of that name instead of stacking another plugin, which is what makes an iterating agent's scratch scripts affordable.

No new CLI verbs: everything lands on the existing script channel.

## Host changes

- **A hit area remembers which column it was laid out in.** Hits were `(row, byte-range)` — the layout is 2-D and a hit kept one dimension, so "which column did the pointer mean" was not answerable, only guessable. A panel with a transcript beside a list has no hits in its left half, so every hover there lit a list row and every click selected one. `Row`'s zip already computes each child's span; carrying it on the hit turns resolution into containment. Hits that never went through a zip own their whole row, so single-column panels are unchanged.
- **The mouse wheel scrolls what the pointer is over.** The routing was already position-aware; the transcript emitted no scroll region because it was a `raw()` block rather than a list, so the hit test found nothing and fell through to "first scrollable widget", which was the workspace tree. A keyed `raw()` can now be a bottom-anchored scrollable viewport with its own region and scrollbar.
- **A text field can drop its brackets, and its popup its side borders.** Both are opt-in variants: the shapes are shared with Settings and the dock's search box, so changing the only available one was not on the table.
- **The dock lays itself out against the width the host drew**, not the width it asked for. The two diverge on a dragged dock, and everything downstream of that number — word wrap, truncation, the chrome budget, and so the position of the dock's bottom edge — was correct only while they agreed.
- **Every keyed pane is clickable, interactive or not.** Gating click rectangles on `interactive` is circular: a panel that makes its pane an input target only once it holds focus can then never be clicked into. `interactive` governs where keys go; it is not a statement about whether the rectangle is there.
- **`pane` renders composite buffers**, and clicking a pane focuses it, firing a `focus` widget event so a plugin can move its mode to match.
- **Dormant windows keep their agent's argv** from the persisted snapshot, so a restored workspace is legible before it is materialised.

## Tests

`orchestrator_dock.rs` — the chat's picker, send and transcript on the dock; the hint bar's pinning as the tree grows, folds and overflows. `orchestrator_attach_worktree.rs` — the bulk-action bar, including the Tab cycle that the chat's own Tab handling had been swallowing. `composite_pane.rs` — a composite buffer renders inside a panel. `widgets::registry` — a row hit does not reach across its column; a hit with no column still owns its row. Plus unit coverage for the bare field and popup shapes, the scrollable `raw()` viewport, and its tail-stick.

Beyond the suites, this was driven by hand in a real terminal with real agents in it — two of them sharing one workspace — which is where most of the bugs fixed here came from, including several the code read as correct.

## Deliberately not here

`docs/internal/chat-surface-gaps.md` records what is still wrong with the chat surface and why, so none of it is silently deferred.

`docs/internal/agent-delivery-hooks.md` is the follow-up worth reading before building on the mailbox: what it takes to hand a message to an agent that is *already running*, per vendor. OpenCode's `prompt_async` is verified end to end including the permission case, Claude Code's cross-session messaging socket is purpose-built but its message frame is undocumented, and Codex cannot be reached while idle at all. Each forces a change at launch, so it is a separate piece of work rather than a tweak to this one.

https://claude.ai/code/session_01D22YhixLB4aPupLo5S2K9P


crates/fresh-core/src/api.rs
```diff
@@ -417,6 +417,115 @@ fn default_window_id() -> WindowId {
...
+pub struct TerminalScreenInfo {
+    /// The rows returned, top to bottom. Already tailed and trimmed per the
+    /// request, so `text.len()` is what came back, not the grid height.
+    pub text: Vec<String>,
+    /// Full grid height, before any tailing — so a caller can tell whether it
+    /// is looking at the whole screen or the end of it.
+    pub rows: usize,
+    /// Full grid width.
+    pub cols: usize,
+    /// Cursor position as `(row, col)`, 0-indexed within the *full* grid.
```
sinelaw: The tuple order is documented backwards. `terminal_screen` fills this field straight from `TerminalState::cursor_position()` (`app/terminal.rs`: `cursor: state.cursor_position()`), and that returns `(cursor.column.0, cursor.line.0)` — i.e. `(col, row)`, which is how the rest of the codebase reads it (`render_terminal_content` compares `cursor_pos.1` against the row index and `cursor_pos.0` against the column).

So `TerminalScreenInfo.cursor` is `(col, row)` while the doc — which is exported verbatim into `plugins/lib/fresh.d.ts` as the only description of a bare `[number, number]` — says `(row, col)`. Any plugin that trusts it will index the wrong axis, and on a non-square grid it will silently read out of range rather than fail loudly. Either fix the comment or transpose at the call site.

---

crates/fresh-plugin-runtime/src/backend/quickjs_backend.rs
```diff
@@ -7675,16 +7675,93 @@ impl JsEditorApi {
...
+    /// `lines` returns at most that many rows counted from the **bottom** —
+    /// the tail, which for a full-screen agent is the prompt and input area
+    /// and therefore the part that says whether it is waiting for you.
+    /// `trim` (default `true`) strips trailing blanks, since a grid is a
+    /// fixed rectangle and the padding is pure cost to read.
+    ///
+    /// This is the cheap, reason-over-it view. To *show* a human another
+    /// workspace's terminal, render the window itself rather than pasting
+    /// this into a buffer.
+    /// Describe every open window and the terminals in it.
```
sinelaw, line 7754: The `readTerminal` doc block (lines 7700–7715) has no item under it — it runs straight into this one, so both blocks attach to `describe_environment_start`, and `read_terminal_start` below ends up with no docs at all.

This is not cosmetic: it ships. In the regenerated `plugins/lib/fresh.d.ts`, `describeEnvironment(): Promise<EnvironmentDescription>` is documented with "Read a terminal's live screen as plain text rows… `lines` returns at most that many rows counted from the bottom… `trim` (default `true`)", and `readTerminal(...)` is emitted with no doc comment. A plugin author reading the type definitions gets the wrong description on both calls.

Move the `readTerminal` block down onto `read_terminal_start` and regenerate the `.d.ts`.

---

crates/fresh-editor/src/app/render.rs
```diff
@@ -2624,6 +2624,165 @@ impl Editor {
...
+                (0..rows).map(|row| state.get_line(row)).collect();
+            let cursor_pos = state.cursor_position();
+            drop(state);
+            frame.render_widget(ratatui::widgets::Clear, inner);
+            crate::app::terminal::render::render_terminal_content(
+                &content,
+                cursor_pos,
+                // The caret belongs to whichever view owns input. A pane
+                // is not that until interactive panes land, and two
+                // caretsfor one PTY would be worse than none.
```
sinelaw: This comment is stale as of this same PR: interactive panes *have* landed here (`input.rs::focused_interactive_pane` / `send_key_to_pane`), and the Fleet mounts its pane with `interactive: true` and moves focus onto it while typing.

The consequence is user-visible. `cursor_visible = false` means `render_terminal_content` never applies the `REVERSED` modifier at `cursor_pos`, and nothing else in `render_pane_into_rect` sets a hardware cursor — so in the state the whole "answer a blocked agent" flow exists for, the user is typing into a terminal with no caret anywhere on screen. The "two carets for one PTY" worry does not apply in that state either: the pane is the input target precisely because the focused split is not.

(Also a typo: "caretsfor".)

---

crates/fresh-editor/src/app/render.rs
```diff
@@ -2624,6 +2624,165 @@ impl Editor {
...
+            return;
+        }
+        let Some(window) = self.windows.get(&window_id) else {
+            return;
+        };
+
+        // --- terminal: size the PTY to this rect, then paint its grid
+        if let Some(terminal_id) = window.get_terminal_id(buffer_id) {
+            if let Some(win) = self.windows.get_mut(&window_id) {
+                win.resize_terminal(buffer_id, inner.width, inner.height);
```
sinelaw, line 2674: The docstring above claims "when this pane goes away the owning window reclaims the size on its next frame, with no 'restore' step anywhere". Nothing resizes terminals per frame. `resize_visible_terminals` is only called from layout/lifecycle events — `apply_layout`, split/tab changes, terminal-mode entry — and `Editor::apply_layout`'s own docstring says so explicitly: *"terminals have no such paint-time sync, which is why their PTY size must be pushed here"*. The only two `resize_terminal` calls in the render path are this one and the session-preview one, whose comment correctly says the size is restored *when the user dives into the session*, not next frame.

So after the Fleet closes, the agent's PTY stays at `FLEET_EMBED_ROWS` (12) rows until something independently triggers a relayout. And while the Fleet is open over the workspace that owns the terminal, the split underneath is painting a 12-row grid into its full-height rect. Worth either restoring explicitly on unmount or fixing the claim.

---

crates/fresh-core/src/api.rs
```diff
@@ -2639,6 +2748,49 @@ pub enum WidgetSpec {
     /// Row's horizontal-zip path). Used by Orchestrator's open
     /// dialog so the preview pane shows a live render of the
     /// highlighted session.
+    /// Reserve a rectangle for one **buffer**, from any window.
```
sinelaw, line 2846: `Pane` was inserted between `WindowEmbed`'s doc comment and `WindowEmbed` itself, so the pre-existing block ("Reserve a rectangle in the widget layout for the host to natively paint the editor `Window` identified by `window_id`… Used by Orchestrator's open dialog…") now documents `Pane`, and `WindowEmbed` is left undocumented. Rustdoc will render `Pane` as two contradictory descriptions stacked, one of which describes the wrong widget.

---

crates/fresh-editor/plugins/orchestrator.ts
```diff
@@ -12354,6 +12563,364 @@ editor.on("terminal_exit", (payload) => {
...
+// Claims nothing: every key the fleet needs (Up/Down/Enter/Esc) is already a
+// smart-key default, and claiming them here would only re-implement them.
+editor.defineMode(FLEET_MODE, [
+  // Tab is the only key the navigation mode claims. Everything else it needs
+  // — Up/Down/Enter/Esc — is already a smart-key default.
+  ["Tab", "orchestrator_fleet_type_toggle"],
+]);
+// The typing mode claims exactly the keys that would otherwise steer the list,
+// and forwards them to the PTY instead. Printable characters are not listed:
+// they arrive through the global `mode_text_input` path, which is handled in
```
sinelaw: Two contradictory comment blocks are stacked on this one `defineMode` call, and the first one is wrong.

It says printable characters "arrive through the global `mode_text_input` path, which is handled in `orchestrator_mode_text_input`". That handler is `if (!form || !formPanel || !args?.text) return;` — it is scoped to the New-Session form and is a no-op for the Fleet. The block immediately below then says the opposite ("Everything else falls through to the focused interactive pane"), which matches what the code actually does. Same for the Escape paragraph, which is stated twice.

Looks like the earlier commit's rationale survived the rewrite in `084a26b`. Please drop the stale block — a reader debugging why a keystroke went somewhere unexpected will go looking in `orchestrator_mode_text_input` and find nothing.

---

crates/fresh-editor/tests/e2e/orchestrator_home.rs
```diff
@@ -0,0 +1,266 @@
+//! E2E coverage for the Orchestrator **Fleet** view — the wide panel that
+//! answers "which of my agents needs me", with a live pane showing the
+//! selected workspace's terminal.
+//!
+//! Per CONTRIBUTING.md §2 these drive only the keyboard and assert on
+//! rendered output. The pane's title line (`live · <name>`) is the visible
+//! statement of which workspace the Fleet is pointed at, so it is what the
+//! selection assertions read.
+
+use crate::common::harness::{copy_plugin, copy_plugin_lib, EditorTestHarness};
```
sinelaw: Two of the three headline behaviours in this PR have no reproducer, which CONTRIBUTING §1 requires for every behavioural claim.

1. **Typing at an agent** (`084a26b`, `fc0927c`) — the new host path `focused_interactive_pane` + `send_key_to_pane` in `input.rs`, the `interactive` flag on `EmbedRect`, and the tabbable-only-if-interactive rule in `widgets/render.rs` are not exercised by any test in the repo (`interactive` appears nowhere under `tests/`). Nothing would fail if the `!matches!(outcome, FallThrough)` guard were inverted, or if the pane silently stopped forwarding keys. This is also the flow most likely to break: it depends on the outcome-vs-FallThrough split, on focus landing on `fleet_embed`, and on `FLEET_TYPE_MODE` claiming exactly Tab.
2. **The `waiting` state** (`22bd356`) — neither `matchWaiting`'s bottom-up "prefer the line ending in `?`" rule nor the per-agent registry patterns are covered, and the sort in `fleetRows` that puts waiting rows on top is only reached via the row-removal path used by `fleet_selection_survives_a_reorder`.

The two tests here cover selection stability and pane-vs-embed, both worthwhile — this is about what is missing rather than what is present.

---

crates/fresh-editor/src/app/mod.rs
```diff
@@ -1177,6 +1177,24 @@ pub struct Editor {
...
+    /// by ~20 cross-cutting paths — workspace save, viewport hooks,
+    /// settings broadcasts, buffer-close cascades — and a pane is a
+    /// transient render surface, not a pane the user owns. None of
+    /// those paths should ever see one. `OverlayPreviewState` keeps the
+    /// same isolation for the same reason.
+    ///
+    /// Keyed by target rather than by widget, so two panes showing the
+    /// same buffer scroll together — which is what you want when a
+    /// panel shows one buffer in two places, and harmless otherwise.
+    pub(crate) pane_view_states:
```
sinelaw, line 1213: Nothing ever removes from this map. Entries are inserted lazily in `render_pane_into_rect` and there is no removal on panel unmount, buffer close, or window close — and by design (per the comment) none of the buffer-close cascades can see it, so it will not be cleaned up by anything existing either. CONTRIBUTING §14: a transient render surface still needs a teardown path.

Bounded in practice today only because the Fleet's panes are all terminals, which take the early-return branch and never touch this map. The moment a plugin points a `Pane` at file buffers, the map accumulates a `SplitViewState` per `(window, buffer)` for the lifetime of the process. `BufferId`s are monotonic so there is no aliasing hazard, just unbounded growth — but the isolation argument in the doc comment is exactly why it also needs its own explicit invalidation.

---

crates/fresh-editor/plugins/orchestrator.ts
```diff
@@ -12354,6 +12563,364 @@ editor.on("terminal_exit", (payload) => {
...
+
+/// Sessions in fleet order: waiting first, then working, then everything else.
+///
+/// Sorting by urgency rather than by name is the whole point — a fleet you
+/// have to scan defeats the purpose. Within a bucket the dock's own stable
+/// order is preserved so rows don't shuffle between ticks.
+function fleetRows(): AgentSession[] {
+  const rank = (s: AgentSession): number => {
+    const st = sessionState(s);
+    if (s.pending || s.discovered) return 3;
```
sinelaw: Dead branch: `rank` is only ever called from the `.sort()` below, which runs on a list already filtered by `.filter((s) => !s.discovered && !s.pending)`. No session reaching `rank` can return 3. Either drop the line or drop the filter (keeping rank 3 as the "parked at the bottom" bucket), but as written the two encode different intentions about whether pending workspaces belong in the Fleet at all.

---

sinelaw (commented):
Independent review of the full diff (~2.5k lines): plugin API types, the host handlers, pane render/key routing, the Orchestrator Fleet, tests, and the design doc.

Nine inline findings. The ones I'd act on first:

- **`TerminalScreenInfo.cursor` is documented `(row, col)` but is `(col, row)`** — the doc ships verbatim to plugin authors as the only description of a bare tuple.
- **A dangling doc block in `quickjs_backend.rs`** puts `readTerminal`'s documentation onto `describeEnvironment()` in the generated `fresh.d.ts` and leaves `readTerminal()` undocumented.
- **The interactive pane draws no caret**, guarded by a comment that says interactive panes have not landed yet — they land in this PR.
- **The PTY-resize comment claims a restore that does not exist**; nothing resizes terminals per frame, so the agent's PTY stays at 12 rows after the Fleet closes.
- **No reproducer for the typing flow or the `waiting` state**, the two behaviours the PR is named for.

Things I looked at hard and concluded were fine: the `SendTerminalInput` `window_id: Option<_>` back-compat path and its `None` → active-window resolution; the `trim`-then-tail ordering in `terminal_screen` and its `rows`/`cols` reporting of the *full* grid; `probeTail`'s in-flight set, TTL and keep-last-good error handling (it cannot leak an unhandled rejection); the id-based Fleet selection and the reorder test that covers it; the `FallThrough`-excluded key interception, which does leave Tab as a working escape hatch out of typing mode given `FLEET_TYPE_MODE` claims it; the `fleetLastSpec` repaint skip combined with `holdFleetFocus`; the `void editor.delay(...).then(tick)` ticker, which mirrors the existing `startProbePolling` loop rather than introducing a new pattern; and the two updated visual-regression SVGs, which are regenerated unconditionally on every run and already match the committed insta snapshots on master — drift, not a regression from this branch.

---

crates/fresh-core/src/api.rs
sinelaw: Correct — verified: `cursor_position()` returns `(cursor.column.0, cursor.line.0)`, so the field is `(column, row)` and the doc said the opposite.

Fixed the comment rather than transposing at the call site, so the value keeps matching the terminal's own convention and every other reader of it. Also spelled out *why* the comment is load-bearing, since the type is a bare pair of numbers and this is the only thing that says which axis is which.

---

crates/fresh-plugin-runtime/src/backend/quickjs_backend.rs
sinelaw: Confirmed and fixed. The generated `fresh.d.ts` did indeed emit `readTerminal(...)` with no doc at all — which matters more than usual here, since that file is what `script api` searches, so the call an agent is most likely to reach for shipped undescribed.

Moved the block down onto `read_terminal_start` and regenerated. `describeEnvironment` keeps its own text; `readTerminal` now has its own.

---

crates/fresh-editor/src/app/render.rs
sinelaw: Fixed — this was a real behaviour bug, not just a comment.

`render_pane_into_rect` was called with `show_cursor: false` unconditionally, so a pane you were actively typing into showed no caret at all: keystrokes went to the PTY but there was no visual indication of where they landed, or even which pane had the keyboard. The embed loop now computes it:

```rust
let has_caret = panel_focused
    && emb.interactive
    && emb.key.as_deref().is_some_and(|k| k == panel_focus_key);
```

so exactly the focused interactive pane draws a cursor, and the misleading comment is gone with it. Also fixed the `caretsfor` typo.

Covered by `fleet_typing_reaches_the_selected_terminal` in `crates/fresh-editor/tests/e2e/orchestrator_fleet.rs`.

---

crates/fresh-editor/src/app/render.rs
sinelaw: Fixed. The claim that the owning window "reclaims the size on its next frame" was not true — nothing in the split path re-asserts a size for a terminal that is off-screen, so it simply keeps whatever the pane last set until the owning window renders it again.

The rule that *is* true, and is now what the comment says: whoever renders it, sizes it. The split path sizes it when the window shows it; the pane sizes it when the pane shows it. There is no restore step, and per the "keep it simple" call we are not adding one — if a stale size between views ever bites in practice, that is when to deal with it.

---

crates/fresh-core/src/api.rs
sinelaw: Fixed. Same slip as the `readTerminal` one: adding `Pane` immediately after `WindowEmbed` left `WindowEmbed`'s doc block attached to `Pane`, so `WindowEmbed` — the older, already-shipped variant — was the one left undocumented while `Pane` carried a description of the wrong thing.

Both variants now carry their own doc: `WindowEmbed` embeds a whole window's split tree, `Pane` shows a single buffer from a named window.

---

crates/fresh-editor/plugins/orchestrator.ts
sinelaw: Fixed. The two comments around `FLEET_TYPE_MODE` contradicted each other, and one of them named `orchestrator_mode_text_input` — a form-scoped mode that has nothing to do with the Fleet's typing state. Both are rewritten to say the one thing that is actually true: the Fleet needs a second mode so Tab can toggle between navigating rows and sending keys to the selected agent's PTY, and the mode is what tells the host which of the two the keyboard currently means.

---

crates/fresh-editor/tests/e2e/orchestrator_home.rs
sinelaw: Both gaps are now covered, in `crates/fresh-editor/tests/e2e/orchestrator_fleet.rs`:

- **`fleet_typing_reaches_the_selected_terminal`** — drives Tab into typing mode, types at the Fleet, and asserts the bytes land in the selected workspace's PTY (and only that one).
- **`fleet_shows_why_an_agent_is_blocked`** — asserts the `!` needs-you state and the reason column, i.e. the thing the whole view exists for.

Writing the second one turned up a fourth bug, which is the argument for the comment. `matchWaiting` read `s.terminalTitle`, which the Orchestrator only tracks for terminals *it* created — so an agent adopted into a workspace by any other route never matched its own waiting pattern and silently stayed "working" while blocked on a prompt. It now takes the title from `readTerminal`'s `screen.title`, which is the host's own value and is there regardless of provenance. Same root cause as two earlier bugs on this branch: the plugin's bookkeeping is a subset, `describeEnvironment`/`readTerminal` is the total.

Both verified per CONTRIBUTING §1 — reverting each fix makes its test fail rather than pass-anyway. The waiting test needs its stand-in to emit OSC 0 (`printf '\033]0;claude\007'`) exactly as a real TUI does, otherwise it is asserting on a title no real agent would have set.

---

crates/fresh-editor/src/app/mod.rs
sinelaw: Fixed. `pane_view_states` is keyed by `(window_id, buffer_id)` and nothing ever removed an entry, so closing a window or a buffer left its view state behind forever — a slow leak, and worse, a stale entry that a recycled `BufferId` would inherit scroll position from.

Sweeps added in both teardown paths: `close_window` drops every entry for that window, `close_buffer` drops every entry for that buffer. It stays out of `split_view_states` deliberately — that map is walked by ~20 cross-cutting paths that would then have to understand pane keys — so the two sweeps are the whole cost of keeping it separate.

---

crates/fresh-editor/plugins/orchestrator.ts
sinelaw: Fixed. Correct catch — `fleetRows()` filters out `pending`/`discovered` sessions before sorting, so `rank()`'s `if (s.pending || s.discovered) return 3;` could never fire. It read as if the fleet had a fourth bucket when it has three. Removed, so the ranking now says exactly what it does: waiting, then working, then the rest.

---

sinelaw:
Closing. The feature is still wanted; this implementation is not the way to get it.

**Nobody built this upstream** — `chatComposer`, `agentAddresses`, `wakeAgentWorkspace`, `dock_chat_focus` and the rest are all zero in `master`'s `orchestrator.ts`, and `describeEnvironment` and the `pane` widget don't exist anywhere on `master`. So the gap this PR fills is still open.

**But the framework it was built on is gone.** `master` is 180+ commits past this branch's base, 114 of them touching `plugins/` and 40 the widget subsystem: widget hit identity split from geometry, plugin panels moved onto `fresh-ui` capabilities, the data layer split into `fresh-editor-core`. The machinery this PR's host changes stand on went with it — `scroll_regions` 65→5 references, `ScrollRegion` 20→2, `overlay_hit_test` 4→1.

More to the point, the rewrite **supplies as framework what this branch built by hand**. The clearest case is the upward completions popup, which cost a `completions_above` flag on `WidgetSpec::Text`, an `above_by` field on `OverlayRow`, and a rework of `emit_completion_overlays`. `master` now has:

```rust
/// Anchor `[row, col]` … the host resolves the final screen rect
/// — opening below the anchor, flipping above near the frame
/// edge, clamped on screen
anchor: Option<[u32; 2]>,
```

That is the same feature, and strictly better — it flips automatically, where this branch made the plugin decide. `full_width` likewise replaces the label-padding, and scroll payload now rides the layout box, which was the transcript viewport's whole problem.

**Measured both ways before concluding:** a rebase hits 27 conflicts across 14 files on commit 1 of 12. A merge shows the real surface — 24 files, including 11 conflict hunks in `orchestrator.ts` itself, which `master` rewrote by +4663/−1466. The plugin half doesn't survive either, which was the last hope for porting rather than rewriting.

So this is a rewrite, not a rebase, and a rewrite against v2 should be substantially *smaller* — much of the ~1300 lines of host code here should simply not need to exist.

**What's preserved:** the design and the research are in #3276 — the control-plane design, the chat punch list, and the vendor delivery research (how to hand a message to a running agent across Claude Code, Codex and OpenCode, with OpenCode verified end to end including the permission case). Their status headers say plainly that none of it is on `master` and flag which items v2 has already answered, so a second attempt starts from the reasoning without inheriting the stale plumbing.

---

Issue #2969, opened 2026-08-11
Five UI defects: wheel routing over chrome, tour jump target / border / dock resize, orchestrator hover width


sinelaw:
Five separate UI defects, all reproduced interactively against `fresh 0.4.7` (debug build of `4aa2f1a`) running inside tmux. Filing together since they were found in one pass; each has independent steps to reproduce.

> **Edit:** section 3 has been rewritten. My first pass understated it badly — I only checked whether box borders were drawn, and I only changed the panel width by resizing the window, which happens to trigger the re-layout that is missing on other paths. The dominant symptom is much worse and is described below.

**How these were driven.** Fresh ran in a detached tmux pane and mouse input was injected as raw SGR sequences, e.g. wheel-up at column 30, row 20:

```sh
tmux send-keys -t f:0.0 -H $(printf '\033[<64;30;20M' | od -An -tx1 -v | tr -s ' ' '\n' | grep -v '^$' | tr '\n' ' ')
```

SGR button codes used below: `0` = left press (`M`) / release (`m`), `35` = motion (hover), `64` = wheel up, `65` = wheel down. Screen state was read with `tmux capture-pane -p` (add `-e` to see colors). All of this reproduces the same way with a real mouse; the injection is only so the steps are exact.

---

## 1. Mouse wheel over UI chrome scrolls the focused terminal's scrollback

Wheel events over the menu bar, tab bar and status bar are routed to the **focused pane** instead of being ignored. When that pane is a terminal, the terminal scrolls its scrollback even though the pointer is nowhere near it.

**Repro**

1. `fresh README.md` in a 140x42 terminal.
2. Command palette (`Ctrl+P`) → `Open Terminal to the Right`. The terminal takes focus.
3. In the terminal run `seq 1 300` so there is scrollback to move.
4. Send wheel-up over the **menu bar** (row 1), the **tab bar** (row 2), and the **status bar** (row 42), at a column that sits over the editor, e.g. col 30 — all three are outside the terminal pane.

**Observed** — the terminal's scrollback scrolls on every one of them. Top visible terminal line, 5 wheel notches each:

| wheel target | terminal top line |
|---|---|
| baseline | 264 |
| menu bar (row 1, col 30) | 249 |
| tab bar (row 2, col 30) | 234 |
| status bar (row 42, col 30) | 219 |

**Expected** — wheel over chrome that is not a scrollable pane should do nothing (or scroll whatever is under the pointer), not the focused terminal.

[...]

PR #2970, opened 2026-08-11
fix(mouse): route wheel events to what the pointer is over, not to the focused pane

sinelaw:
Fixes defect **1** of #2969 ("Mouse wheel over UI chrome scrolls the focused terminal's scrollback"). The other four defects in that issue are untouched.

## The bug

Both wheel routers hit-tested the pointer and then fell back to the **active split** when it matched nothing:

```rust
let (target_split, buffer_id) = self.split_at_position(col, row).unwrap_or_else(|| {
    (…active_split(), self.active_buffer())
});
```

So a wheel notch over the menu bar, a tab bar or the status bar scrolled whatever held focus. Reproduced against `4aa2f1a` under tmux with the issue's own steps (`fresh README.md` in 140x42 → `Open Terminal to the Right` → `seq 1 300` → wheel-up at col 30), watching the terminal's top visible line:

| wheel target | before | after |
|---|---|---|
| baseline | 264 | 264 |
| menu bar (row 1) | 249 | 264 |
| tab bar (row 2) | 234 | 264 (tabs pan instead) |
| status bar (row 42) | 219 | 264 |
| **over the terminal pane** | 234 → 219 | **264 → 249** (still scrolls) |

The editor-focused asymmetry the issue describes reproduced too: with the editor focused, a status-bar wheel moved its top line 16 → 9. After the fix it stays put, while a wheel over the editor pane still scrolls it (1 → 9) and leaves the terminal alone.

## The fix

The rule is **the wheel moves what the pointer is over**. One shared hit-test, `Window::wheel_surface_at`, resolves a screen cell to the file explorer, a split, a tab strip, or nothing — and every wheel path routes through it. CONTRIBUTING.md §12: the gate belongs at a single fork rather than as another per-site guard. Four fall-throughs collapse into it:

- `handle_mouse_scroll` (vertical) — its explorer check and its focused-split fallback become arms of the same match.
- `handle_horizontal_scroll` — had drifted into the identical fallback, so Shift+wheel and native `ScrollLeft`/`ScrollRight` over chrome panned the focused split.
- `handle_vertical_scroll`'s terminal branch — consulted `focused_terminal_live()` before any hit-test, so a wheel that never touched a pane (chrome, the tab strip, the explorer) still dropped a focused live terminal into read-only scrollback. It is now scoped to a wheel over a pane.
- The tab bar's `<` / `>` click handlers now nudge the strip through the same helper as the wheel, so a click and a notch move it by the same step and both stop at the last tab (the click path previously incremented the offset without any bound).

Everything upstream of that fork is unchanged: mouse-modal overlays, prompt dropdowns, the popup stack, the orchestrator dock and mounted widget panels all still consume the wheel before the permanent layout is consulted.

### What each region does now

| region under the pointer | wheel |
|---|---|
| editor / terminal split | scrolls that pane (focused or not) — unchanged |
| file explorer | scrolls its viewport — unchanged |
| **tab strip** | **pans the tabs** (new) |
| menu bar, status bar, split separators | nothing |

The **tab strip** is chrome, but chrome that owns scrollable content: it has a per-split `tab_scroll_offset` and `<` / `>` indicators, so under the pointer-wins rule the wheel belongs to it. Wheel up walks toward the first tab, wheel down toward the last, and the **active tab never changes** — the wheel moves the view, matching every other wheel surface in the editor. Scrolling right stops once the renderer reports no remaining right overflow, so the offset can't run off past the last tab and leave you wheeling back through empty space.

The **menu bar**, **status bar** and **split separators** stay inert — not as a policy about chrome, but because nothing scrollable is under the pointer there. Menu dropdowns have no scroll state either, so there is nothing to route to them.

## Tests

`crates/fresh-editor/tests/e2e/issue_2969_wheel_over_chrome.rs` — three e2e tests driving real wheel events, asserting only on rendered output (CONTRIBUTING.md Testing §2): the window of `LINE n` markers on screen *is* the scroll position, and the tab row's text *is* the strip's position.

1. `wheel_over_chrome_does_not_scroll_the_focused_editor` — no PTY needed, so it covers the fix on every runner.
2. `wheel_over_chrome_does_not_scroll_the_focused_terminal` — the issue's exact setup (editor left, focused terminal right, wheel at col 30 over the editor's chrome); skips where no PTY can be opened, like the other terminal e2e tests.
3. `wheel_over_the_tab_bar_pans_the_tab_strip` — enough tabs to overflow the bar; a wheel brings earlier tabs into view, the wheel back stops at the last tab, and the editor beneath does not move.

Chrome rows are located by what they render rather than hard-coded, and each test also wheels over a surface that *should* respond — so a "fix" that swallowed every wheel event fails them. All three fail without the corresponding change (1 and 2 on master: `assertion left == right failed: a wheel over the menu bar (row 0) must not scroll the focused editor`, the pane having jumped 15 lines; 3 without the tab routing: the strip sits still). No timeouts: the terminal test waits on `wait_until_stable` for the shell to go quiet, and every test runs under the harness's per-test temp workdir.

## Checks

`cargo fmt --check`, `cargo clippy --all-features --all-targets`, `cargo check --all-targets`, and `cargo check --all-targets --features gui` all clean (no new warnings in the touched files). Ran the tab/mouse/scroll/terminal/explorer/dock/split/diff e2e slices — 686 + 591 tests, 0 failures — including the pre-existing wheel-routing suites (`issue_2119_wheel_scroll`, `split_view_expectations::test_scroll_wheel_targets_viewport_under_pointer`, `code_tour_dock::test_wheel_scrolls_the_hovered_list`). Also re-verified the whole matrix by hand in tmux: pane scrolls, tab strip pans both ways and stops at the ends, menu bar and status bar inert.

### DIFF
--- a/crates/fresh-editor/src/app/mouse_input.rs
+++ b/crates/fresh-editor/src/app/mouse_input.rs
@@ -595,6 +595,13 @@ impl Editor {
             // its column (never leaks to the window beneath).
         } else if self.handle_split_widget_panel_wheel(col, row, delta) {
             // a mounted widget panel consumed the scroll
+        } else if self.active_window().wheel_surface_at(col, row).is_none() {
+            // Chrome that owns no scrollable content — menu bar, tab bars,
+            // status bar, split separators. The wheel is dropped here rather
+            // than falling through to the focused pane, which used to scroll a
+            // focused terminal's scrollback from a pointer sitting on the
+            // status bar (sinelaw/fresh#2969). Gating at this fork also keeps
+            // the focused terminal out of read-only scrollback below.
         } else {
             if self.active_window().focused_terminal_live() {
                 // Scrolling up drops the focused split into read-only scrollback
--- a/crates/fresh-editor/src/app/mouse_input.rs
+++ b/crates/fresh-editor/src/app/mouse_input.rs
@@ -7,6 +7,7 @@
 //! - Split separator dragging
 //! - Text selection via mouse
 
+use super::scrollbar_input::WheelSurface;
 use super::*;
 use crate::input::keybindings::Action;
 use crate::model::event::{ContainerId, CursorId, LeafId, SplitDirection};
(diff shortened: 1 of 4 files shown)


sinelaw:
CI note: `test macos-latest` is red on this commit, but not on anything in this diff.

The whole run was `8798/8799 passed`, with one **timeout** (nextest's external 180s kill):

```
TIMEOUT [ 180.193s] (5235/8799) fresh-editor::e2e_tests
  e2e::lsp_completion_duplicate_label_import_2952::test_resolve_goes_to_the_server_that_offered_the_candidate
```

Why I'm reading it as runner slowness rather than a regression:

- That test drives LSP completion resolution across two servers; its file contains no mouse, wheel, or scroll code, and this PR only touches wheel routing.
- The whole `lsp_completion_duplicate_label_import_2952` file passes locally against this branch in 1.73s (4 tests, including that one).
- `test ubuntu-latest` is green on the same commit, new tests included.

I'll re-run the macOS job once the run finishes (Windows is still going — GitHub refuses a re-run while any job is in flight) and report back here if it reproduces, in which case it's mine to chase.

---

PR #2989, opened 2026-08-13
file explorer: let plugins narrow which entries are shown

chrismo:
Decorations and slot overrides already flow plugin -> explorer, changing how a row is drawn. Nothing lets a plugin decide whether a row is drawn at all, so "show me only the files in git status" - the view both JetBrains and VS Code offer - cannot be built today. The explorer's own filters are no help: hidden-file and gitignore matching are filename and pattern checks, while git status needs live repository state, which is why every git feature in fresh ships as a plugin shelling out to git.

Add SetFileExplorerFilter / ClearFileExplorerFilter as that same channel one step earlier. A plugin hands over a set of paths; the explorer shows those, the ancestors needed to reach them, and the contents of any directory named. Core stays ignorant of what the filter means, so the same mechanism serves diagnostics, a code tour's files, or anything else a plugin can enumerate.

Three judgement calls worth a look:

- Namespaces union rather than intersect. Intersecting would collapse the tree the moment a second plugin joined - baffling for the user and an easy bug for the plugin author.

- An active filter outranks the explorer's ignore rules, so a matched path appears even when it is gitignored or hidden. A changed file inside a gitignored directory is still a changed file; dropping it would make the filtered view quietly incomplete.

- Filters are cleared when a plugin unloads, tracked separately from the decoration namespaces. A stranded decoration is cosmetic; a stranded filter hides files and outlives the only thing that could clear it.

Filters live on the window rather than on FileTreeView because the view is rebuilt whenever the explorer reopens or the root changes, and a fresh view starts unfiltered - the same reason decorations live there.

No menu entry yet. A "show all files" escape hatch belongs in the Explorer menu next to the existing toggles, but it needs an i18n key in all 15 locales and an e2e test, and where it belongs is worth agreeing on first.

### DIFF
--- a/crates/fresh-core/src/api.rs
+++ b/crates/fresh-core/src/api.rs
@@ -3743,6 +3743,34 @@ pub enum PluginCommand {
         namespace: String,
     },
 
+    /// Narrow the file explorer to a set of paths for a namespace.
+    ///
+    /// Where decorations change how a row is *drawn*, a filter decides
+    /// whether it is drawn at all — letting a plugin offer "only files in git
+    /// status" or "only files with diagnostics" while the explorer stays
+    /// ignorant of what the filter means. Ancestors of the given paths stay
+    /// visible so the matches are reachable, and a given directory brings its
+    /// contents with it.
+    ///
+    /// Namespaces union: a path claimed by any namespace is shown. While any
+    /// filter is active it outranks the explorer's own ignore rules, so a
+    /// matched path appears even when it is gitignored or hidden.
+    SetFileExplorerFilter {
+        /// Namespace for grouping (e.g., "git-status")
+        namespace: String,
+        /// Paths to show. Relative paths resolve against the workspace root;
+        /// paths outside it are dropped. An empty list contributes nothing
+        /// and leaves the explorer unfiltered.
+        paths: Vec<String>,
+    },
+
+    /// Clear the file explorer filter for a namespace. Once no namespace is
+    /// filtering, the explorer returns to its normal view.
+    ClearFileExplorerFilter {
+        /// Namespace to clear (e.g., "git-status")
+        namespace: String,
+    },
+
     /// Set file explorer slot overrides for a namespace.
     ///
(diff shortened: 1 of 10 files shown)


sinelaw (changes requested):

Validated interactively (debug build + tmux + a throwaway plugin calling `setFileExplorerFilter` on a scratch repo). Every behaviour the description claims held up:

- narrows to matched paths, ancestors kept reachable; siblings hidden
- an explicitly matched **gitignored** file reappears — ignore-override works
- two namespaces **union**, not intersect
- `../../../etc` dropped — root confinement holds
- filter survives explorer close/reopen (Ctrl+B twice) — re-application on rebuild works
- clearing restores the original tree exactly, gitignore included

Design is sound: mirrors the existing decorations/slots channel, gates at the single `is_node_visible` fork, reuses `normalize_explorer_plugin_path` / `explorer_path_under_root`, clears on unload. Not duplicating anything — `FileExplorerSearch` only highlights and `IgnorePatterns` isn't plugin-drivable. `cargo fmt` clean, no new clippy warnings.

Requested changes:

- **No consumer.** 544 lines of API with zero callers and no user-reachable entry point. `git_explorer.ts` already parses `git status --porcelain` — wiring it here would deliver the motivating feature and exercise the path.
- **Window/dispatch/JS layer untested.** All 9 new tests sit on `ExplorerFilter`/`FileTreeView`. `handle_set_file_explorer_filter` (normalization, root confinement) and `apply_file_explorer_filters` (re-apply on rebuild) have none, though the description makes specific claims about both. CONTRIBUTING #2 wants an e2e test for a new flow; `tests/e2e/plugins/` already has the harness.
- **Ship the escape hatch with it.** Since a filter outranks ignore rules and hides files, a plugin that sets one and stops responding strands the user. Deferring the menu toggle leaves that gap open.
- Minor: filtered ancestors stay collapsed, so a match deeper than one level isn't visible until manually expanded — VS Code/JetBrains auto-reveal here.

Still draft, so read the above as design feedback rather than a merge block.

---

PR #2991, opened 2026-08-13
Home — every agent at once, with a live pane

sinelaw:
**Draft — stacked on #2962.** Base is `claude/agent-environment-control-lxkrc5`, not `master`; review that one first. The diff shown here is only the Home view.

## What this adds

Home: one fullscreen screen that answers the three questions you actually have about a fleet of agents — who needs you, what did they say, and what is that one doing right now.

- The agent list, sorted by urgency, so whoever is blocked on you is at the top.
- The selected agent's **live terminal** beside it, via the host's pane widget.
- The chat, carrying what every agent has said, with a line to answer them on.

## Why it is a separate PR

It is a *consumer* of the control plane, not part of it. Everything it is built from lands in #2962 and is used there by the dock: the fleet model (also exported as `orchestrator.fleet()`, which is how an agent asks who needs the user), the mailbox, the chat — transcript, address picker, send, and the wake-on-send — and the host's pane widget, which stays in the base as a capability.

Splitting it keeps #2962 to the plane and the dock, and keeps this reviewable as one screen.

## Contents

The Home spec and its pane embed, its focus model and three editor modes, its handlers and its two commands (`Orchestrator: Home`, `Orchestrator: Home Open Selected`), and the `orchestrator_home` e2e module.

Applying this on top of the base reproduces the pre-split tree exactly — verified by diffing against a ref taken before the extraction.

## Status

Draft. The `orchestrator_home::*` e2e tests in this branch are currently timing out and are **not yet diagnosed** — that is the main thing outstanding here. They were failing before the split, so the split neither caused nor fixed them; moving them here is what takes them off the base PR.

---
_Generated by [Claude Code](https://claude.ai/code/session_01D22YhixLB4aPupLo5S2K9P)_


sinelaw:
Closing. This was split out of #2962 as a draft to hold the Fleet/Home UI, and all three reasons to keep it have expired:

- The Home dialog was already flagged as something that might be removed rather than finished.
- Its base, #2962, is now closed — the widget framework it was built on has been rewritten on `master`, so this stack has no bottom.
- It still carries the undiagnosed `orchestrator_home::*` e2e timeouts that were moved here in the first place.

The Home view's design is recorded in `agent-control-plane.md` §7, landing in #3276, so nothing is lost by closing the branch.

---

PR #3002, opened 2026-08-13
fix(review-diff): make a big range reviewable — and fix the per-tick cost that made it unreviewable

sinelaw:
Reviewing `HEAD~100..HEAD` opened on a list of file headers with **no diff at all**. Opening a file from the FILES panel made that one file appear; going back to the unified area still showed nothing else.

The reported bug was a small one in the review plugin. Chasing why the workaround for it was necessary turned up three host defects that made any heavily-decorated buffer cost more per keystroke the more it held.

## The reported bug

Above 20 000 diff lines the stream rendered one file at a time. Which file came from `ensureFocusFile`, which takes the first entry of `visibleFiles()` — the sidebar's dir-grouped order. The stream is emitted in `state.files` order — git's. On any tree deeper than one level those disagree, so the single expanded file sat somewhere in the middle of seventy headers while the cursor stayed at row 1. Nothing on the headers said a body was missing either: they all carried the expanded `▾`.

## Why the cap existed, and why it's gone

The one-file-at-a-time mode was compensating for cost that grew with a buffer's content. Measuring it (`review_diff_layout_bench`) put the shape beyond doubt: an ordinary 100 000-line buffer renders flat at ~1.9ms/frame, while a review of the same size cost 862ms/frame and 3.8s per keystroke. Three causes, each found by measurement rather than reading:

**The plugin state snapshot deep-copied every text property of every buffer, on every tick.** A review row carries one property each. Splitting the harness tick from the paint isolated it: painting a 20k-line review and a 100k-line one cost the same 13ms, while the tick between them went from 38ms to 262ms. Of the tick's twenty-odd steps, exactly one ever exceeded 500µs.

**Rainbow-bracket colorization republished its overlays across the whole buffer every frame.** `replace_range_in_namespace(ns, 0..buffer.len(), …)` asks the marker tree for every marker in the buffer — three per overlay — to retract a few dozen bracket overlays. 98.5% of a frame was inside that one call.

**The overlay set kept a globally sorted vector.** Every add shifted and re-indexed everything after it; every namespace clear scanned the whole set; every range removal re-sorted and rebuilt the index. Bracket matching and the review's own cursor-line highlight do clear+add on every cursor move, so holding an arrow key paid for 40 000 unrelated overlays each time.

With those fixed, a 100k-line review costs what a 20k-line one does on every interactive path — 35ms per cursor move either way — so the layout budget bought nothing but a review that hid files from the reader. It, its config field, its "not loaded" headers and its i18n keys are all removed. `focusOnly` now means only what its name says: the side-by-side composite, which is per-file by construction.

## Changes

**Review plugin** — the budget removed; `visibleHunkIndexAtCursor` and `jumpToFile` read hunk rows from the build's own map instead of counting hunks while skipping collapsed files (collapse is a conceal, so those rows are still in the stream and the count drifted by exactly the hunks it skipped — clicking the sticky header with a collapsed file above landed the cursor inside that file).

**Bracket colorization** — caches what it computed from and returns early when none of it moved; replaces over the bytes it scanned plus what the previous pass wrote, rather than the whole buffer. `OverlayManager` counts removals globally and per namespace so that cache can tell when its overlays were removed from under it.

**Overlay set** — unordered, with three indexes: marker → slot, handle → slot, namespace → slots. `add` is O(1), `remove_by_handle` a lookup, a namespace clear O(k) in its own overlays. Nothing needed the global priority order; the renderer sorts the handful it draws. Also fixes a pre-existing leak: `clear()` deleted two of each overlay's three markers, orphaning one interval per overlay in the tree forever.

**Text properties** — `TextPropertyManager` holds `Arc<[TextProperty]>` and hands out a clone of that handle, so the snapshot takes one per buffer per tick as a refcount bump. An intermediate version of this skipped the copy when a version looked unchanged; the version turned out to be the property *count*, so a panel repainting the same number of rows with new values served the old ones. Sharing removes the question rather than answering it.

## Tests

`PerfCounters` counts what the snapshot copies, and `plugin_snapshot_scaling` asserts that no tick copies a property and that a review five times the size costs the same per tick — counters rather than timings, because the same defect measured 53ms in debug and 4ms in release and a duration can't hold a line in CI.

`plugin_text_property_freshness` asserts the contract the snapshot sits in the middle of, from outside the host: a plugin writes rows with properties, reads them back, rewrites the same number of rows with different values, and must read the new ones.

Plus regression tests for the sticky-header jump, the marker leak, and range reviews laying out every file. Each new assertion was checked by reverting the fix under it — three of my first attempts passed against the broken code and were rewritten.

`review_diff_layout_bench` (ignored by default) keeps the durations, with the numbers and the reasoning in its header.

https://claude.ai/code/session_01S7D3NqdLCSCqqbTC3FPwcn


crates/fresh-editor/plugins/audit_mode.ts
```diff
@@ -5504,26 +5498,23 @@ function updateReviewStatus(): void {
...
-    // Map back to the global state.hunks index.
-    let visited = 0;
+    let best = -1;
+    let bestRow = 0;
     for (let i = 0; i < state.hunks.length; i++) {
-        const h = state.hunks[i];
-        if (state.collapsedFiles.has(fileKeyOf(h.file, h.gitStatus || 'unstaged'))) continue;
-        if (visited === visibleIdx) return i;
-        visited++;
+        const row = state.hunkRowByHunkId[state.hunks[i].id];
```
sinelaw, line 5487: The bug this rewrite fixes is still live in the sibling function `jumpToFile` (lines 2547-2565), which maps into `hunkHeaderRows` by counting hunks while skipping collapsed files:

```ts
if (state.collapsedFiles.has(hKey)) continue;   // ← not skipped in the buffer
...
const row = state.hunkHeaderRows[visibleIdx];
```

`buildDiffLines` never consults `collapsedFiles` (collapse is a conceal applied by `applyFolds`), so `hunkHeaderRows` still contains every collapsed file's hunk headers — exactly the drift the new doc comment describes.

Sequence: unified stream with file A (2 hunks) above file B. Enter on A's header to collapse it. Scroll so B is the pinned file and click the sticky header → `jumpToFile(B)` skips A's two hunks while counting, gets `visibleIdx = 0`, and jumps to `hunkHeaderRows[0]` — A's first hunk, a folded row in the wrong file, instead of B's first hunk.

Same one-line fix as here: once `foundGlobal >= 0`, use `state.hunkRowByHunkId[state.hunks[foundGlobal].id]` and drop the `visibleIdx` counting entirely.

---

crates/fresh-editor/plugins/audit_mode.ts
```diff
@@ -2994,6 +2981,13 @@ function review_toggle_file_collapse() {
     const headerFile = fileHeaderUnderCursor();
     if (headerFile) {
         const key = fileKey(headerFile);
+        // Nothing under the header to fold: the file is past the stream's
+        // layout budget, so the key that expands a header loads it.
+        if (!fileBodyRendered(key)) {
```
sinelaw: This branch can never fire, and its comment describes a mechanism deleted in this same PR.

`review_toggle_file_collapse` is only reachable from `review_enter_dispatch`, which returns early at the `state.centerComposite` check (line ~3217) whenever the composite is up. So it only ever runs on the unified path, where `focusOnly` is false and `fileBodyRendered()` is unconditionally true. There is no "past the stream's layout budget" case left.

Two sibling `fileBodyRendered` branches added by the budget commit are dead for the same reason:

- `on_review_mouse_click` line 2631 — that loop only runs under `if (data.buffer_id === diffId)`, and the unified text buffer is not the mounted center buffer in side-by-side, so `focusOnly` is always false there.
- `jumpToComment` line 2734 — the composite case already returned at line 2702, so this is reachable only in the transient window while an async composite build is in flight.

The two genuinely live call sites (`onFilesTreeSelect` 2338, `jumpToGlobalHunk` 5529) still say "the stream's budget stopped short of it". Suggest deleting the three dead branches and rewording those two to name side-by-side, so the only surviving mention of the budget is the deliberate historical note in `syncFocusMode`.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
```diff
@@ -6609,11 +6628,23 @@ impl Window {
...
-                    .buffer_text_properties
-                    .insert(*buffer_id, state.text_properties.all().to_vec());
+            // Store text properties if this buffer has any, and only if the
+            // copy in hand is stale.
+            if state.text_properties.is_empty() {
+                snapshot.buffer_text_properties.remove(buffer_id);
+                snapshot.buffer_text_property_versions.remove(buffer_id);
+            } else {
+                let version = state.text_properties.version();
+                if snapshot.buffer_text_property_versions.get(buffer_id) != Some(&version) {
```
sinelaw: This guard is equivalent to comparing property *counts*, so a re-render that keeps the count serves the plugin the previous render's properties forever.

`TextPropertyManager::version` does not survive a whole-manager replacement, and replacement is the only way a buffer's properties are ever set. `set_virtual_buffer_content` (`virtual_buffers.rs:455`) does `state.text_properties = properties`, where `properties` came from `TextPropertyManager::from_entries` — a manager built by `Self::new()` (version 0) plus one `add()` per property. So after every `setBufferContent` the version is exactly `len()`, and this comparison can only ever notice a change in the number of properties.

Concrete sequence:
1. A plugin renders a panel whose rows carry properties — e.g. the review plugin's sticky header, one row with `{file: "a.rs", line: 10}`. `from_entries` → version 1, snapshot copies it.
2. The cursor moves to another file; the plugin re-renders the same panel with `{file: "b.rs", line: 400}`. `from_entries` → version 1 again.
3. `snapshot.buffer_text_property_versions.get(id) == Some(&1)`, so the copy is skipped. `snapshot.buffer_text_properties` still holds `{file: "a.rs", line: 10}`.
4. `getTextPropertiesAtCursor` (`quickjs_backend.rs:304`) answers from that stale vec, so "open this row" navigates to the wrong file/line.

It is not limited to one-row panels: any re-render with an unchanged number of property-bearing entries hits it — a list whose selection moved, a re-sorted tree, a diff re-laid-out at the same row count. The stale `start`/`end` offsets also no longer line up with the new text, so the cursor can match a different property or none at all.

The pre-existing behaviour was a full copy every tick, so nothing masked this before. `TextPropertyManager::version`'s doc ("moves on every mutation, and on nothing else") is what is untrue here — the counter needs to survive replacement. Sourcing it from a process-global `AtomicU64` bump inside `add`/`set_all`/`clear`, or adding a `replace_with(&mut self, other)` that keeps `self.version + 1` and using it at `virtual_buffers.rs:455`, would both close it.

---

crates/fresh-editor/tests/e2e/plugins/review_diff_ux_bugs.rs
```diff
@@ -2016,3 +2016,154 @@ fn test_issue2117_discard_hunk_with_no_trailing_newline() {
...
+    open_review_range_head(&mut harness);
+
+    let screen = harness.screen_to_string();
+    assert!(
+        screen.contains("DEEP_MARKER"),
+        "the stream must open on real diff content, not a wall of headers. \
+         Screen:\n{screen}"
+    );
+    assert!(
+        !screen.contains("not loaded"),
```
sinelaw: This assertion cannot fail. The `not loaded` marker, its i18n keys and the code that emitted it are all deleted in this PR — the string does not exist anywhere in `audit_mode.ts` or `audit_mode.i18n.json` any more — so it is vacuous here and again at line 2162.

The other half of the gap: this test is documented as the reported `HEAD~100..HEAD` case, but the repo it builds is two files / six diff lines. The defect it names was size-triggered (`totalVisibleDiffLines() > 20000`); reintroduce that exact `syncFocusMode` and this test still passes, because the changeset never reaches the threshold.

What does carry weight is `screen.contains("DEEP_MARKER")`: `fileGroups()` sorts `src/` before `src/a/`, so the sidebar's first file is `src/zeta.rs` while the diff's first is `src/a/deep.rs` — the assert fails the moment unified renders only `filesCurrentKey`. Worth saying that in the comment as the actual contract, dropping the `not loaded` canary, and — if the size axis is meant to be covered — sizing the repo past a plausible budget rather than implying it already is.

---

crates/fresh-editor/src/app/mod.rs
```diff
@@ -428,7 +428,32 @@ pub(crate) const PLUGIN_COMMAND_HANDLER_HARD_LIMIT: std::time::Duration =
     std::time::Duration::from_millis(500);
 
 /// The main editor struct - manages multiple buffers, clipboard, and rendering
+/// Counters for work whose cost must not grow with a buffer's content.
```
sinelaw, line 430: `PerfCounters` was inserted inside `Editor`'s doc comment. Line 430 (`/// The main editor struct - manages multiple buffers, clipboard, and rendering`) now documents `PerfCounters`, and `pub struct Editor` at line 451 is left with no doc comment at all. Moving that line back down to sit directly above `pub struct Editor` restores both.

---

sinelaw (commented):
Reviewed the review-diff plugin, its i18n, and the three e2e test files. The unified stream's final state is correct — the reported bug (cursor at row 1 seeing the diff-order first file while only the sidebar-order first file had a body) is fixed, since unified now lays out every file unconditionally; `filesCurrentKey`/`ensureFocusFile` stay coherent across both layouts; the `visibleHunkIndexAtCursor` rewrite is right and also fixes the filtered-file case the old counting loop got wrong. `check-types.sh audit_mode.ts` is clean apart from the pre-existing `lib/widgets.ts` errors, and no `maxExpandedDiffLines` / `not loaded` leftovers remain in the plugin or i18n.

Three things worth changing: the same visible-hunk counting bug still live in `jumpToFile`, a set of `fileBodyRendered` branches that can no longer fire (plus comments still describing the deleted budget), and a vacuous assertion in the new e2e tests.

---

crates/fresh-editor/src/view/overlay.rs
```diff
@@ -597,67 +638,99 @@ impl OverlayManager {
...
-        self.rebuild_marker_index();
     }
 
     /// Clear all overlays and their markers
     pub fn clear(&mut self, marker_list: &mut MarkerList) {
+        self.removal_epoch = self.removal_epoch.wrapping_add(1);
         for overlay in &self.overlays {
             marker_list.delete(overlay.start_marker);
             marker_list.delete(overlay.end_marker);
         }
```
sinelaw, line 657: `clear()` deletes `start_marker` and `end_marker` but never `span_marker`, so every span marker this PR introduces is leaked into the `MarkerList` permanently. Every other removal path (`swap_remove_at`) deletes all three; only this one doesn't.

Verified with a throwaway test on this branch:

```rust
// 10 overlays added -> query_range(0,1000) == 30 markers
mgr.clear(&mut ml);
assert_eq!(ml.query_range(0, 1000).len(), 0);
// left: 10, right: 0
```

The consequence is not just memory. `set_virtual_buffer_content` (app/virtual_buffers.rs:443) does `overlays.clear()` then `overlays.extend(...)` on every content set — that is the review-diff path, with a comment there noting "a big git-show diff can be ~500k overlays". Each reload leaves one orphan interval per overlay in the tree forever. Since `candidates_in`, `query_viewport`, `remove_in_range`, `remove_in_range_for_namespace` and `replace_range_in_namespace` are now *all* driven by `marker_list.query_range`, those orphans are returned by every one of those queries and then silently dropped by the `marker_to_idx.get(mid)` filter — so the `k` in the O(log N + k) this PR is buying grows without bound across reloads, and `IntervalTree` ids are monotonic (`next_id`) so they never get reclaimed either.

Fix is one line: `marker_list.delete(overlay.span_marker);` in the loop.

---

crates/fresh-editor/src/view/overlay.rs
```diff
@@ -503,19 +544,19 @@ impl OverlayManager {
...
-        // O(log N + k) for the lookup; restoring the priority-sorted
-        // invariant after `swap_remove` is O(N) (adaptive sort on a
-        // near-sorted vec) plus O(N) marker_to_idx rebuild. For typical
-        // markdown_compose workloads where overlays in a buffer share
-        // the same priority, the adaptive sort is a no-op pass.
+        self.removal_epoch = self.removal_epoch.wrapping_add(1);
+        // O(log N + k) throughout: the marker tree narrows the candidates
+        // and each removal patches only its own index entries.
         // Spanning overlays (start < range.start && end > range.end) are
         // not detected — same precondition as ConcealManager.
```
sinelaw, line 563: This comment is no longer true as of this PR, and the behaviour change it hides is silent.

At the merge base, `marker_to_idx` held only `start_marker` and `end_marker` (`git show <base>:.../overlay.rs` — the two `marker_to_idx.insert` pairs), so an overlay with `start < range.start && end > range.end` produced no candidate and survived. This PR registers `span_marker` too (`index_entry`), and the span is an interval covering the whole extent, so `query_range` now *does* return spanning overlays as candidates — and the filter below (`start < range.end && range.start < end`) accepts them. Same for `remove_in_range_for_namespace` and `replace_range_in_namespace`.

Concrete effect: a plugin's `clear_overlays_in_range(start, end)` for one line inside a multi-line overlay (a hunk-wide `extend_to_line_end` background, a code-tour step range, a multi-line diagnostic) used to leave that overlay alone and now deletes it. Nothing in the test suite pins either behaviour.

Either drop the comment and note the widened semantics in the docstring, or keep the old precondition by skipping candidates reached only via `span_marker`. Please pick deliberately — right now it reads as if nothing changed.

---

crates/fresh-editor/src/view/overlay.rs
```diff
@@ -340,109 +356,133 @@ impl OverlayManager {
...
+        self.marker_to_idx.insert(end, idx);
+        self.marker_to_idx.insert(span, idx);
+        self.handle_to_idx.insert(handle, idx);
+        if let Some(ns) = namespace {
+            if let Some(slots) = self.namespace_to_idx.get_mut(&ns) {
+                slots.remove(&from);
+                slots.insert(idx);
+            }
+        }
     }
```
sinelaw, line 449: `reindex_moved` clones the handle `String` and the namespace `String` on every swap-remove, but both keys are already present in their maps — the clones only exist to dodge a borrow that `rustc` allows anyway (`self.overlays` and `self.handle_to_idx` are disjoint fields). This compiles as-is on this branch (`cargo check -p fresh-editor --lib`):

```rust
fn reindex_moved(&mut self, idx: usize, from: usize) {
    let overlay = &self.overlays[idx];
    self.marker_to_idx.insert(overlay.start_marker, idx);
    self.marker_to_idx.insert(overlay.end_marker, idx);
    self.marker_to_idx.insert(overlay.span_marker, idx);
    if let Some(slot) = self.handle_to_idx.get_mut(&overlay.handle) {
        *slot = idx;
    }
    if let Some(ns) = &overlay.namespace {
        if let Some(slots) = self.namespace_to_idx.get_mut(ns) {
            slots.remove(&from);
            slots.insert(idx);
        }
    }
}
```

Same in `index_entry` for the namespace: `entry(ns).or_default()` forces an owned key on every add, though the key is new only for the first overlay of a namespace. That one runs once per `add`/`extend` element, so the bulk path here — `set_virtual_buffer_content`'s `extend` over the whole diff — pays a `String` allocation per overlay that the pre-PR `extend` (sort + rebuild) did not.

---

crates/fresh-editor/src/view/overlay.rs
```diff
@@ -597,67 +638,99 @@ impl OverlayManager {
...
-        for (i, o) in self.overlays.iter().enumerate() {
-            self.marker_to_idx.insert(o.start_marker, i);
-            self.marker_to_idx.insert(o.end_marker, i);
-            self.marker_to_idx.insert(o.span_marker, i);
+        self.handle_to_idx.clear();
+        self.namespace_to_idx.clear();
+        for idx in 0..self.overlays.len() {
+            self.index_entry(idx);
         }
     }
```
sinelaw: `rebuild_indexes` has no callers left — the sorts that used the old `rebuild_marker_index` are gone, and no test invokes it. It builds with a warning:

```
warning: method `rebuild_indexes` is never used
   --> crates/fresh-editor/src/view/overlay.rs:675:8
```

Delete it; `index_entry` covers every live path.

---

sinelaw (commented):
Reviewed `view/overlay.rs` and its call sites. The index maintenance itself holds up — I could not construct a sequence that leaves a stale or aliased slot. Four things worth acting on: a leaked `span_marker` in `clear()`, a silent widening of `remove_in_range` semantics that its comment still denies, avoidable per-add/per-swap `String` clones on the bulk path, and a dead `rebuild_indexes`.

---

crates/fresh-editor/src/app/mod.rs
sinelaw: Fixed — the doc line is back directly above `pub struct Editor`, and `PerfCounters` has its own.

Not acting on the sibling finding about `remove_in_range` semantics, though, because it was read against the wrong baseline. `origin/master` in this clone is two commits behind where the branch actually started (`eaeab81`), and the span marker landed in between:

```
$ git show eaeab81:crates/fresh-editor/src/view/overlay.rs | grep -n span_marker
165:    pub span_marker: MarkerId,
361:            self.marker_to_idx.insert(o.span_marker, i);
388:            self.marker_to_idx.remove(&overlay.span_marker);
```

So `marker_to_idx` already held span markers before this branch, and `remove_in_range` already returned spanning overlays as candidates. The stale comment about them not being detected predates this work; nothing widened here.

The `clear()` leak in the same file was real and is fixed, with a test that counts markers before and after — that one holds against `eaeab81` too, where `clear()` deletes only the two endpoints.

---

sinelaw:
`test ubuntu-latest` is red here, and it isn't this branch. Recording the evidence rather than just asserting it.

Two runs of the same commit (`18d9bc0`) failed on **different** tests, and a master run from before this branch's base fails the same way:

| run | commit | platform | failure |
|---|---|---|---|
| master `560c435` (predates this branch) | base | macOS | `devcontainer_spec_conformance::no_user_means_no_dash_u_flag` — 8890/8891 passed |
| this PR, run 1 | `18d9bc0` | Linux | `devcontainer_spec_conformance::lifecycle_hooks_fire_in_spec_order_during_up` — 9089/9090 passed |
| this PR, re-run | `18d9bc0` | Linux | `remote_channel_timeout_tests::test_reconnection_via_replace_transport` — **timed out at 180s**, 9089/9090 passed |

All three are wall-clock-bound: the devcontainer ones poll for a file a spawned lifecycle hook writes, against a 10s deadline; the remote-channel one has a 180s cap. Total suite time swung 1401s → 1630s between two runs of identical code, which is the kind of variance that pushes them over. `master` is red in 6 of its last 15 CI runs.

Everything this branch adds passed on all three platforms, including in the runs that failed elsewhere: text-property freshness, snapshot scaling counters, sticky-header jump, the marker-leak test, and the range-review layout tests. `test macos-latest` and `test windows-latest` are green on `18d9bc0`, as are fmt, clippy, doc, schema, actionlint, check-no-plugins and build-musl.

I've re-run the job once; happy to re-run again, but at a ~40% base failure rate that's luck rather than signal. If these tests are worth stabilising, `retries` in the nextest profile would stop them gating unrelated PRs — out of scope for this one.

---

crates/fresh-editor/plugins/audit_mode.ts
sinelaw: Addressed in `74635d0`. `jumpToFile` now finds the file's first hunk and reads its row from `hunkRowByHunkId`, with the `visibleIdx` counting gone.

Worth noting the reachability, since it shaped the test: `jumpToFile` has exactly one caller — a click on the sticky header. My first regression test drove `.` instead and passed against the broken code; so did the second, which asserted on the sticky header's text (with a small diff the viewport top sits above every file header, so it shows the fallback summary either way). The test that landed clicks the sticky row and asserts on the plugin's own hunk index; reverted, it reports "landed on hunk 1 of 2, inside the collapsed file above".

---

crates/fresh-editor/plugins/audit_mode.ts
sinelaw: Mostly addressed in `74635d0`.

Deleted the two branches that are provably unreachable — `review_toggle_file_collapse` and the file-header case in `on_review_mouse_click`, both of which only run against the diff buffer, which isn't mounted in side-by-side.

Kept the one in `jumpToComment`, on your own reasoning: you note it's reachable in the transient window while an async composite build is in flight. A guard that costs one call and is correct in both layouts is worth keeping over a race I'd have to prove impossible.

Reworded the comments at the two live sites (`onFilesTreeSelect`, `jumpToGlobalHunk`) plus `jumpToComment` to name side-by-side, so the only surviving mention of the budget is the deliberate note in `syncFocusMode`.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
sinelaw: Correct, and the best catch on this branch — thank you. Verified before acting: `from_entries` builds `Self::new()` and `add()`s one property each, and `virtual_buffers.rs:455` replaces the whole manager, so the version really was the property count.

Fixed twice, and the second time is the one that matters:

`74635d0` took your first suggestion — a process-global `AtomicU64`, so no two managers can share a version and a replacement always reads as a change.

`18d9bc0` then removed the version entirely. `TextPropertyManager` holds `Arc<[TextProperty]>` and hands out a clone of that handle, so the snapshot takes one per buffer per tick as a refcount bump. That deletes the version, the counter, the snapshot's parallel version map, the dead-buffer pruning and the carried-across-ticks state — the operation the check was skipping is now free, so there is no freshness question to answer. Two designs and two invalidation bugs in one branch was the argument for it.

`plugin_text_property_freshness` is the test this was missing: a plugin writes rows with properties, reads them back, rewrites the same number of rows with different values, and must read the new ones. It asserts from outside the host, so it holds whatever the host does underneath — against the count-based check it reports reading `a.rs` after `b.rs` was written, which is your sequence exactly.

---

crates/fresh-editor/tests/e2e/plugins/review_diff_ux_bugs.rs
sinelaw: Half addressed, half deliberately not.

**Addressed:** both `not loaded` assertions are gone, and the comment now states what you identified as the load-bearing contract — the sidebar sorts `src/` above `src/a/` while the diff starts at `src/a/deep.rs`, so `DEEP_MARKER` fails the moment unified renders only `filesCurrentKey`.

**Not addressed, on purpose:** I'm not sizing the repo past a budget. The threshold it would test no longer exists, so the test would be pinning a mechanism this PR deletes rather than a contract — and paying twenty thousand diff lines of setup per run to do it. What's worth pinning is that per-tick cost doesn't scale with review size, and `plugin_snapshot_scaling` does that with counters at two sizes, which is exact and fast where a size-triggered e2e test would be neither.

Your wider point stands though, and it cost me: three of my new assertions passed against the code they were meant to catch. Each one now gets reverted-and-run before it lands.

---

crates/fresh-editor/src/view/overlay.rs
sinelaw: Real bug, fixed in `74635d0` — one line, plus `clear_deletes_every_marker_it_created`, which counts markers before and after and fails with `left: 10, right: 0` on the old code.

One correction to the attribution: this is **pre-existing**, not introduced here. Your baseline was `origin/master`, which in this clone is two commits behind where the branch actually started, and the span marker landed in between:

```
$ git show eaeab81:crates/fresh-editor/src/view/overlay.rs | sed -n '/pub fn clear(&mut self, marker_list/,/^    }/p'
    pub fn clear(&mut self, marker_list: &mut MarkerList) {
        for overlay in &self.overlays {
            marker_list.delete(overlay.start_marker);
            marker_list.delete(overlay.end_marker);
        }
```

Your reasoning about *why it matters more now* holds exactly, though, and is why it was worth fixing here rather than filing: every query is driven by `marker_list.query_range` after this branch, so the orphans are returned and discarded by every one of them.

---

crates/fresh-editor/src/view/overlay.rs
sinelaw: Rejecting this one — it rests on a baseline that isn't this branch's.

`origin/master` in this clone is two commits behind where the branch started (`eaeab81`), and the span marker landed in between. At the real branch point `marker_to_idx` already held all three markers:

```
$ git show eaeab81:crates/fresh-editor/src/view/overlay.rs | grep -n span_marker
165:    pub span_marker: MarkerId,
202:        let span_marker = marker_list.create_span(range.start, range.end);
361:            self.marker_to_idx.insert(o.span_marker, i);
388:            self.marker_to_idx.remove(&overlay.span_marker);
```

So `query_range` already returned spanning overlays as candidates before this branch, and `remove_in_range` already accepted them. The stale comment about spanning overlays "not being detected" predates this work — it was already wrong when I got here, and `query_viewport`'s own docstring at the same commit describes relying on the span marker for exactly the case the comment denies.

You were right that it reads as if nothing changed, so I'd rather not leave the contradiction sitting there — but correcting a comment that was already stale is a separate change from this PR, and it deserves a test pinning whichever behaviour is intended rather than a doc edit that guesses. Leaving it.

---

crates/fresh-editor/src/view/overlay.rs
sinelaw: Addressed in `74635d0`, both halves.

`reindex_moved` re-points the keys in place via `get_mut`, as you wrote it — the disjoint-field borrow does compile, and the clones are gone.

`index_entry` now takes the `get_mut` path when the namespace is already known and only owns a key for the first overlay of a namespace, so the bulk `extend` over a whole diff no longer pays a `String` per overlay.

---

crates/fresh-editor/src/view/overlay.rs
sinelaw: Addressed in `74635d0` — deleted. `index_entry` covers every live path, as you say.

For the record on how it got there: I'd marked it `#[cfg(test)]` when the last caller went away, which silences the warning in a normal build but not in a test build — so it was dead code wearing a disguise. Deleting is right.

---

PR #3008, opened 2026-08-14
Widget framework v2: registered components + one event model — box-tree dispatch, kind-owned handlers, chrome registration with one z-order for all gestures

sinelaw:
## Summary

Implementation arc of the widget-framework v2 plan (`docs/internal/widget-framework-v2-review.md`), developed commit-by-commit with suites green at every step. The goal: replace hardcoded UI event handling with a standard model — components REGISTER (geometry + capabilities), events DISPATCH by hit-testing derived structure, PROPAGATE consume-or-fall-through, and per-component HANDLERS own behavior. That model is now real and consistently applied for the pointer press/wheel walks, hover, and the keyboard; drag/release are grab-keyed by ruling with component-owned bodies. A staff-level audit (21 items), an adversarial re-audit, the keyboard arc (K1–K4), and three review waves (R1–R9, including the maintainer&#39;s inline findings and must-land list) each closed or explicitly ruled on everything they found. **Scope note:** the chrome box collection is still FLAT (no parent links — the guard boxes are the deliberate flat-world containment encoding, documented at `ChromeTreeBuilder`); relational precedence, phases, tree unification and chrome focus are split out to the stacked forward-design PR #3024.

## The plan (`docs/internal/widget-framework-v2-review.md` + `docs/internal/chrome-event-model-plan.md`)

Argues the recurring focus / wheel-routing / text-input bugs share structural causes: the widget tree flattens to rows with no vertical axis, z-order, or clipping; per-kind behavior smeared across 280+ match sites; popups that are not part of the tree; and event precedence expressed as hand-ordered control flow. Hardened by two adversarial review passes; carries a living implementation-status section.

## Widget framework (panel-local)

1. **`WidgetImpl` extraction (phase 1)** — 19 kinds, one dispatch point (`kinds::behavior`). Adding a kind touches exactly one dispatch line for rendering, pointer, wheel, key interiors, AND (post-R9) panel key navigation and plugin value-setting: the arrow/Enter policies are capability-declared (`arrows_advance_focus`, `picker_nav`, `activates_on_picker_enter`/`picker_activate_event`) and the typed mutation arms delegate to kind fns — the router and dispatch arms hold no kind matches.
2. **Typed text-input fast lane** — the `mode_text_input@:` string encoding survives only as deprecation-warned keymap compat.
3. **Host-supplied panel height** — `visibleRows` optional; `search_replace.ts` migrated off its chrome arithmetic.
4. **The `LayoutBox` tree (panel-local)** — per-widget rects/z/parent-links/flags built during collection; the `ScrollRegion` channel deleted; vertical flex; display-width-correct columns.
5. **Wheel = hit-path bubble** — `WidgetImpl::on_wheel`, scroll chaining by construction.
6. **Kind-owned keys, pointers, FOCUS, and CAPABILITIES** — `on_key` (`KeyFx`), `on_pointer` (`PointerFx`), `on_focus_change`, and declared capabilities (`HitArea.row_target`/`.context_click`, `BoxMeta.picker_scroll_target`, the R9 key/value set). Clipboard has ONE owner.
7. **Derived focus ring + traps** — the rules exist once (`collect_tabbable` asks `box_meta`); `focus_trap` produced and consumed; the residual dual traversal is oracle-guarded with its retirement condition recorded.
8. **Popups** — completion popup and dropdown pop-over parented as boxes; plugin `WidgetSpec::Popup` renders through `PanelPopup`, and the panel-clipped form promotes (one predicate; regression test). Documented limitation: buffer-mounted panels drop the overlay/popup channels (recorded arc).

## Chrome registration (app-level)

9. **Chrome surfaces REGISTER; behavior lives WITH registration** — 17 `ChromeComponent`s, one registry. Every per-surface behavior body lives in its component&#39;s file: press, wheel, hover, key, and drag. `mouse_input.rs` holds engines, grab routing, terminal forwarding, and shared queries. **The chrome collection is FLAT by ruling** (see Summary): boxes have no parents; z bands + push order are the whole ordering; `parent`/`focusable`/`focus_trap`/`scroll` are reserved-but-unset at chrome level.
10. **One engine per channel for press and wheel; drag/release grab-keyed by ruling.** `dispatch_pointer` serves all four press kinds with per-surface `(owner, kind)` dedup and the `pointer_walk_step` opacity contract (pure fn, unit-tested); `dispatch_wheel` serves both axes (deliberate no-dedup/no-opacity for chaining, documented); hover is the same walk in query form. Drag/release dispatch on the derived `PointerGrab` (no re-hit-testing mid-drag) with component-owned bodies. ONE `chrome_tree` per mouse event.
11. **Opaque surfaces absorb** — `pointer_opaque` gates the pointer walk, the panel probe, and terminal mouse-forwarding; modal surfaces absorb the horizontal wheel axis they have no content for.
12. **Modal band in RANK order** — capture offered over the owner-stamped `overlay_stack()`; every capturing component pairs capture gate and layer on ONE activity predicate.
13. **The overlay stack is DERIVED and tested** — per-component `layers()` with explicit ranks; unit tests pin the deliberate rank relations, rank distinctness, and the always-owning base layer; **golden e2e tests** (`e2e/dispatch_precedence.rs`) pin routing OUTCOMES: wheel/hwheel over a popup never move the buffer, popup-chrome clicks never move the cursor, workspace-trust keys beat an open prompt.
14. **Full pointer-grab slot** — 13 press-to-release drag classes, drag AND release grab-keyed, terminal-forward suppression covers every grab.
15. **Geometry from layout — complete for the migrated caches; the paint-recorded class is a CLOSED, ruled list.** The `screen_space`-class surfaces (popups, suggestion rects, overlay toolbar, trust dialog, file-browser dialog) are enumerated in ONE roster on `ChromeLayout`&#39;s doc with "additions require a ruling"; retiring the class is #3024&#39;s compositing arc.

## Keyboard registration arc (K1–K4)

The staged, hand-ordered key pipeline is gone: ONE walk over the owner-stamped `overlay_stack()` via `on_layer_key`, with `handle_key` reduced to the documented pre-band plus the walk, degrading (not panicking) on the tested base-layer contract. Pre-band membership is restricted BY RULING to whole-pipeline observers and custom-dispatcher modals. A documented ruling explains why the keyboard walk does NOT get one-stack-per-event (handlers may mutate then decline; the handler-level context rebuilds are load-bearing freshness — invalidation-aware derivation is #3024).

## One text engine (§4.5 step 1, with undo)

Prompt&#39;s private editing engine deleted (~500 lines) for the shared `TextEdit`; the prompt&#39;s text state IS an embedded single-line `TextEdit`; review-found regressions fixed with tests.

## Review + remediation ledger

Four audit/review passes, three remediation waves. "Closed" below means fixed in code **or converted to a written ruling at its site** — the rulings are enumerated (plan doc "R wave" section), never silent:

- The 21 staff-audit items: fixed, except the four documented-by-ruling at their sites (mounted-panel channels, router dock branch, panel-hover pipeline, tie-break divergence).
- The re-audit residue: closed by the B/C waves; the keyboard item by K1–K4.
- The duplicate-mechanism/partial-migration audit: closed by R1–R7 (rank inversions, half-wired flags, hand-synced encodings, ~264 lines of dead old-model code deleted).
- The maintainer&#39;s inline wave: closed by R8; the must-land list by R9 (flat-tree ruling, key/value capabilities, claim trims, golden tests, screen_space roster, keyboard-rebuild ruling, PassAfter contract test).
- Remaining recorded residue, each with a ruling: `is_mouse_over_any_popup`&#39;s parallel rect query, panel hover beside the chrome walk, the `hit_path`/`hit_stack` tie-break divergence, and the structural items split to #3024.

## Known-red CI

The **playwright suite** failure is pre-existing on the base branch (bisected; documented in the PR triage comment). The long-flaky quick_open goto-line click test was root-caused (late background file-scan delivery) and fixed with an async-quiescence wait. One Windows-only failure of master&#39;s own new compose-marker test (`1ec1d34`, via the merge ref) is master-side code racing its pre-scan on slow runners.

## Validation

- Per-commit: `cargo check --all-targets` clean; lib suite green at every slice; full suites arbitrate in CI on three platforms (ubuntu + macos green on the full suite; Windows&#39; sole failure master-side, above).
- Unit tests pin: `hit_stack` ordering, overlay precedence resolution, layer-rank relations + distinctness, the base-layer contract, the `pointer_walk_step` opacity contract, panel-clipped popup promotion; golden e2e tests pin the routing outcomes named above; standing debug oracles cover slice-7 geometry parity and focus-ring parity.

## Remaining arcs

The forward design is PR #3024 (stacked): relational precedence replacing `layer_rank`/z-band integers, propagation phases (capture/target/bubble with observer registration, dissolving the pre-band), parent links + chrome focus ring, tree unification + retiring the paint-rect path, Settings/prompt interiors onto the framework, mounted-panel compositing.

https://claude.ai/code/session_01Dji9pGVSzHj2FkJdgFgva2


crates/fresh-editor/src/widgets/layout_box.rs
```diff
@@ -0,0 +1,330 @@
...
+    /// consumed here when the widget is not already at its bound.
+    pub scrollable: bool,
+    /// An opaque surface: a pointer event inside this box that no
+    /// descendant consumes stops here rather than falling through to
+    /// content beneath (today: overlay-promoted popup rows).
+    pub pointer_opaque: bool,
+    /// A focus scope boundary: Tab cycles among focusables inside the
+    /// nearest enclosing trap instead of the whole panel.
+    pub focus_trap: bool,
+}
```
sinelaw, line 107: Two of these four dispatch flags are only half-wired, which is exactly the kind of drift this tree was built to eliminate:

- `focus_trap` has a live **consumer** (`focus_ring_scoped`, called from `handle_widget_focus_advance` at widget_runtime.rs:1517) but **no producer**: no `box_meta` impl ever sets it — the only `focus_trap = true` in the codebase is this file's unit test. The PR body's "`focus_trap` scoping live" overstates this; today it is unreachable machinery. 
- `pointer_opaque` has **producers** (the Overlay container, the Text completion popup box) but **no consumer**: the actual opacity decision in `probe_floating_widget` (mouse_input.rs:4590) tests `boxes[i].z > 0`, not this flag, and only tests assert it.

Neither is wrong per se — staged foundations are the PR's stated method — but a flag whose producer and consumer live on opposite sides of a convention can silently diverge (e.g. a future z>0 box that is *not* meant to be opaque). Suggestion: either make `probe_floating_widget` consume `pointer_opaque` instead of `z > 0` (making the flag real), and give `focus_trap` its first producer (the Overlay subtree is the obvious candidate), or trim both until their phase lands.

---

crates/fresh-editor/src/widgets/render.rs
```diff
@@ -459,22 +517,37 @@ pub fn render_spec_with_options(
...
+    // computed above exists because focus must resolve *before*
+    // collection (widgets style by focus); the two are equal by
+    // construction — box_meta mirrors collect_tabbable's rules — and
+    // the debug assert keeps them that way until the pre-pass ring
+    // can be retired with the constraint-layout phase.
+    let derived_tabbable = crate::widgets::layout_box::focus_ring(&collected.boxes);
+    debug_assert_eq!(
+        derived_tabbable, tabbable,
+        "box-tree focus ring diverged from collect_tabbable"
+    );
```
sinelaw, line 644: This is the one place the PR adds a parallel mechanism *without* a structural guarantee. The focusability rules now live in two independent copies — the central `collect_tabbable` match (line 645) and each kind's `box_meta` — kept equal by convention plus this `debug_assert_eq`. In a release build, divergence is silent and asymmetric: `focus_key` is resolved against the spec-walk ring (lines 505–512) while the *published* `tabbable` is the derived ring, so focus could land on a key that Tab cycling can no longer reach.

The chicken-and-egg (focus must resolve before collection) justifies keeping the pre-pass, but not keeping two copies of the *rules*. `collect_tabbable` could be reduced to a spec walk that asks the single source of truth:

```rust
let meta = super::kinds::behavior(spec).box_meta(spec);
if meta.focusable { if let Some(k) = meta.key { out.push(k); } }
```

Then the two rings are equal by construction, the per-kind `focusable`/`disabled` rules exist exactly once, and the assert becomes a tree-shape check rather than a rules-sync check. Worth doing now rather than at the constraint-layout phase — it deletes the central match this PR's own doc (kinds/mod.rs) says should not exist.

---

crates/fresh-editor/src/app/widget_runtime.rs
```diff
@@ -2688,101 +2427,57 @@ impl Editor {
...
+                        .map(|(row, col)| {
+                            crate::widgets::layout_box::hit_path(&p.boxes, row, col)
+                                .into_iter()
+                                .rev()
+                                .filter(|&i| p.boxes[i].scrollable)
+                                .filter_map(|i| p.boxes[i].key.clone())
+                                .collect()
+                        })
+                        .unwrap_or_default();
+                    (p.spec.clone(), along_path)
```
sinelaw, line 1380: The hit-path bubble is a genuine improvement (scroll chaining by construction, ~300 lines of kind-probing deleted). But note what still sits *above* it: the `focused_text_completions_open` pre-check at the top of this loop absorbs the wheel for the whole panel, position-blind, before the tree is ever consulted. The `text_completions` box this PR adds (z=1, pointer-opaque) can't participate — it is neither `scrollable` nor keyed, so the `filter(|&i| p.boxes[i].scrollable)` here can never route a wheel to it, and `scroll_focused_text_completions` below re-implements the popup scroll including a duplicated "5 if zero" mirror of the renderer's visible-rows default.

That's the same hand-simulated-depth pattern the key path just deleted, one dispatch surface over. Follow-up suggestion: mark the completion popup box scrollable, let `Text::on_wheel` handle it while the popup is open, and delete both `focused_text_completions_open` and `scroll_focused_text_completions` — the tree already carries everything needed.

Minor: `p.spec.clone()` clones the full spec tree per panel per wheel notch. Pre-existing pattern and fine at this scale, but worth remembering when the box tree grows to app level.

---

crates/fresh-editor/src/app/widget_runtime.rs
```diff
@@ -2178,49 +1920,46 @@ impl Editor {
...
+        self.with_dropdown_helper(panel_key, widget_key, |spec, key, panel, fx| {
+            crate::widgets::kinds::dropdown::set_open(spec, key, panel, open, fx);
+        });
+    }
+
+    /// Shared shell for the Dropdown click paths: resolve the widget's
+    /// spec node, run the kind-owned mutation (the same helpers
+    /// `Dropdown::on_key` uses — one copy of the cycle/open logic),
+    /// repaint, and fire the queued events.
+    fn with_dropdown_helper(
```
sinelaw: "One copy of the cycle/open logic" is true for cycle and open — but `handle_widget_dropdown_set` just above is still a hand-rolled third copy of the same pattern: resolve current selection from instance state with spec fallback, clamp into the option set, write `WidgetInstanceState::Dropdown`, fire `change` with `{index, value}`. That's `kinds::dropdown::cycle_selection` with an absolute index instead of a delta.

A `dropdown::set_selection(spec, key, panel, index, fx)` next to `cycle_selection`, routed through this same `with_dropdown_helper` shell, would erase the last duplicated dropdown mutation and make the claim fully true.

---

crates/fresh-editor/src/app/mouse_input.rs
```diff
@@ -4460,43 +4573,53 @@ impl Editor {
...
+        let on_overlay = self
+            .widget_registry
+            .panels_for_buffer(slot.buffer_id())
+            .first()
+            .and_then(|k| self.widget_registry.get(k))
+            .is_some_and(|p| {
+                crate::widgets::layout_box::hit_path(&p.boxes, brow, local_screen_col as u32)
+                    .last()
+                    .is_some_and(|&i| p.boxes[i].z > 0)
+            });
```
sinelaw: Deleting `overlay_hit_test` and unifying right-click through this probe is one of the PR's clearest wins (including the real click-through-popup fix). One question about this derivation, though: `panels_for_buffer` filters a `HashMap<PanelKey, _>` (registry.rs:303), so its order is arbitrary — `.first()` picks a nondeterministic panel whenever a buffer ever hosts more than one panel, and the panel whose `boxes` decide `on_overlay` here isn't guaranteed to be the same panel whose `overlays`/`entries` the code below reads via `self.panel(slot)`.

The caller already holds the slot's own `panel_key` (`fwp.panel_key`) — resolving the boxes through that key directly would remove both the nondeterminism and the possibility of the surface decision and the text lookup disagreeing.

---

crates/fresh-editor/src/widgets/kinds/containers.rs
```diff
@@ -0,0 +1,1509 @@
...
+    for mut b in std::mem::take(&mut child_out.boxes) {
+        b.row += 1;
+        b.col += LEFT_BORDER_PREFIX.chars().count() as u32;
+        // Scrollable boxes widen two more columns — through the right
+        // padding onto the `│` border — matching the scroll-region
+        // widening below so a wheel over the section border still
+        // scrolls the widget inside it.
+        if b.scrollable {
+            b.width += 2;
+        }
```
sinelaw: This +2 widening is the sharpest example of the box tree's current cost: it exists to keep the *box* rectangle in sync with the *scroll-region* rectangle, by hand, in prose ("matching the scroll-region widening below"). Every container path in this file now shifts three column-addressed channels in lockstep — embeds, scroll_regions, boxes (plus byte-shifted hits/overlays) — and nothing structural stops them drifting; a future container that shifts two of the three compiles fine and mis-routes the wheel.

Two cheaper shapes worth considering before the next arc piles more on: (a) a single `shift_geometry(&mut CollectedOutput, row_off, col_off)` helper that all container paths call, so a channel can't be forgotten; or (b) since `ScrollRegion`'s geometry is now redundant with the scrollable box (only its `scroll`/`total`/`item_height` payload is still needed for scrollbar paint and text-drag), fold that payload onto the box or key it off instance state and delete the region *rectangle* — the plan defers deleting the channel, but the duplicated geometry is deletable now.

---

crates/fresh-plugin-runtime/src/backend/quickjs_backend.rs
```diff
@@ -9197,7 +9197,54 @@ impl QuickJsBackend {
...
+    }
+
+    /// Typed fast lane for a mode's printable text input. Carries the
+    /// mode and the typed text as structured fields instead of the
+    /// legacy `mode_text_input@<mode>:<char>` action-name encoding
+    /// (which `start_action` still parses for keymap-bound plugin
+    /// actions). Rides the same request channel as every other
+    /// dispatched action, so a mode's own bindings (Backspace, Space,
+    /// ...) and plain characters stay strictly ordered.
+    pub fn start_mode_text_input(&mut self, mode: Option<&str>, text: &str) -> Result<()> {
```
sinelaw, line 9229: The typed lane is a real improvement for the hot path, and sharing `start_action_resolved` keeps it honest. But note the net shape: +172/−15 — the legacy `mode_text_input@<mode>:<char>` string-parse in `start_action` survives in full, and after this PR nothing in-tree produces that encoding anymore (the only remaining references are doc comments in `input/router.rs` / `input/buffer_mode.rs`, which are now stale and worth updating). If the parse is kept solely for user keymaps that bound the string form, consider logging a deprecation warning when it actually fires, so there's a signal for when the parallel path can be deleted rather than it living forever as compat of unknown necessity.

---

sinelaw (commented):
Adversarial review focused on one question: **does each piece move the codebase to a more correct, robust, clean, generic implementation — or does it add bloat?** Every claim below was verified against the actual diff, not the commit messages.

## Overall verdict

**Mostly genuine improvement, honestly staged — with a recurring, fixable weakness: new mechanisms that coexist with the old ones they're meant to replace, synced by convention instead of construction.** The PR deletes real things (the `overlay_hit_test` entry point, ~300 lines of Editor kind-probing on the wheel path, both popup key short-circuits, `search_replace.ts`'s viewport arithmetic, the `mode_text_input@:` string splicing at its producer) and fixes two real bugs it found structurally (right-click through open popups; dead wheel over fits-everything lists). The e2e evidence of behavior preservation is unusually strong. The net +3.2k lines is mostly extraction boilerplate, doc comments, and the 880-line plan doc — not duplicated logic.

The bloat that *does* exist is concentrated and nameable: two half-wired `LayoutBox` flags (`focus_trap` has no producer, `pointer_opaque` has no consumer — the PR body's "focus_trap scoping live" overstates), a second focus-ring computation per render whose rules exist in two copies guarded only by a `debug_assert`, and a third parallel column-addressed geometry channel that every container must now shift in lockstep with the other two. None of it is speculative dead-ends — each has a recorded follow-up in the plan doc — but all of it is drift surface until the deletions land, and two of the three are cheaply fixable *now* (see inline comments).

## Per-piece scorecard

| Piece | Verdict | One line |
|---|---|---|
| WidgetImpl trait extraction | **GENUINE** | Real decomposition: render.rs non-test code 6.8k→3.5k, one total dispatch, helpers moved with their only consumers |
| Typed text-input fast lane | **GENUINE (modest)** | Kills string-splicing at the producer; legacy parse retained (+172/−15) with stale doc references — needs a deprecation signal |
| Host-supplied panel height | **GENUINE** | Real new capability; deletes plugin-side chrome arithmetic; honest fallbacks (two-auto = legacy 20 rows) |
| LayoutBox tree | **GENUINE, conditionally** | Sound substrate, already load-bearing for 3 features — but it's a *fourth* geometry channel until ScrollRegion's rectangles die |
| Wheel = hit-path bubble | **GENUINE** | −324 lines of Editor probing, scroll chaining by construction; the completion-popup absorb still bypasses the tree, position-blind |
| Kind-owned key handling | **GENUINE** | Popup short-circuits actually deleted; panel-level kind ladder remains (acknowledged); one dropdown mutation still hand-rolled |
| Derived focus ring + traps | **NET NEUTRAL** | Second ring computed per render, rules duplicated, release divergence is silent; traps unreachable (no producer) |
| Popups as boxes | **NET NEUTRAL (staged)** | +92/−0; correct structure, deferred deletions are defensibly wire-format-bound; `pointer_opaque` is write-only |
| overlay_hit_test deletion | **GENUINE** | Deletes a duplicate entry point *and* fixes the right-click blind spot; one nondeterminism question inline |
| WHEEL_SURFACES table | **GENUINE (modest)** | Ordering is data and documented; containment still per-handler, so relocation with better shape, not unification |
| CLICK_SURFACES table | **GENUINE (modest)** | Same; guards-as-variants is slightly odd but explicit and positionally faithful |

## Top recommendations

1. **Make the two focus-ring copies equal by construction, not by debug_assert** — have `collect_tabbable` ask `behavior(spec).box_meta(spec)` instead of maintaining its own central match (render.rs:532 comment). This deletes the duplicated rules and the silent release-divergence mode in one move, and it's small.
2. **Wire or trim the two half-flags** — give `focus_trap` a producer or remove it until §4.4 lands; make the probe consume `pointer_opaque` instead of `z > 0` (layout_box.rs comment). Dead flags on the dispatch substrate are exactly where the next mysterious mis-route will hide.
3. **Don't let the geometry channels multiply further** — before the app-level arc, add a single shift helper (or fold ScrollRegion's rectangle into the box) so containers can't shift two of three channels and compile (containers.rs comment). The `+2`-widening-matched-in-prose pattern is the drift class this whole PR exists to kill.

---

sinelaw:
CI triage for the two red checks on `4dc55ac`:

**`playwright suite` — pre-existing breakage on the PR's base, not this branch.** Bisected locally by running the full web suite (`web-ui/test/run.sh`) at four points: branch head, `618a999` (before the dispatch-table arc), `b7c2b40` (before the box tree / auto-sizing), and the merge-base **`093d1e9` (master)**. All four fail identically — `session list fills the dock` at ratio 0.54, `dropdown option highlight fills the row`, and the fatal 30s timeout clicking the dock's "New Task…" option at `drive.mjs:862`. The Web UI workflow is path-filtered and last ran on master Aug 10 (`5dbd4837`, where the same section already had two FAILs — the Escape-close pair); none of the ~10 master commits since (the text-property/perf cluster: `873613c`, `e707270`, `f9b8e52`, `1489176`, `093d1e9`, …) touched web files, so the suite never ran on them and the breakage went unnoticed until this branch touched `web-ui/js` and triggered it. The dock regression on master needs its own bisect of that range — happy to take that as a separate fix against master if wanted.

**`test macos-latest` — flake.** One failure, `command_palette::test_quick_open_goto_line_live_preview_mouse_click_commits` (0.86s, not a timeout); passes 8/8 consecutive runs locally on this head, and the mouse path it exercises is platform-independent harness code. Re-queued.

---

sinelaw:
All seven review findings are now addressed on the branch; resolving the threads. Where each landed:

- **`focus_trap` producerless / `pointer_opaque` unconsumed** — `probe_floating_widget` now consumes `pointer_opaque` (not `z > 0`), and `WidgetSpec::Component` is the `focus_trap` producer (transparent scope kind, plugin-buildable via `component()`).
- **Dual copies of the focusability rules** — `collect_tabbable` is now the suggested spec-walk over `box_meta(spec)`; the per-kind rules exist once and the debug assert is a tree-shape check.
- **Position-blind completion-wheel pre-check** — `focused_text_completions_open` absorb and `scroll_focused_text_completions` are deleted; the popup box is scrollable and `Text::on_wheel` handles it through the ordinary hit-path bubble (the "5 if zero" default now lives only with the kind).
- **`handle_widget_dropdown_set` third copy** — `dropdown::set_selection` next to `cycle_selection`, routed through the same mutation shell.
- **`panels_for_buffer().first()` nondeterminism** — the probe resolves boxes through the slot's own `fwp.panel_key`.
- **Three column-addressed channels shifted by hand** — `CollectedOutput::absorb_child` is the single fold for all channels, and the `ScrollRegion` rectangle is gone entirely: the scroll payload rides the scrollable widget's own box (`BoxScroll`), which the scrollbar painters and drag read.
- **Legacy `mode_text_input` string parse** — fires a deprecation `tracing::warn` when the string form is actually used, and the stale doc comments in `input/router.rs` / `input/buffer_mode.rs` now describe the typed lane (`0093855`).

Since the review snapshot, the same arc also finished: all five pointer gestures (hover/wheel/click/right-click/double-click) dispatch over z-ordered chrome boxes with no hand-ordered chains left, and every kind's keyboard vocabulary is kind-owned in `on_key` with a `KeyFx` effects channel (`focus_advance`, `flash_scrollbar`, `clipboard_copy`) — `handle_widget_key` retains only cross-widget panel policy.

---

crates/fresh-editor/src/app/mouse_input.rs
```diff
@@ -1082,385 +731,477 @@ impl Editor {
...
-    /// options. Called only after floating overlays have been ruled out.
-    fn hover_target_in_chrome(&self, col: u16, row: u16) -> Option<HoverTarget> {
-        // Check menu bar (row 0, only when visible)
-        // Check menu bar using cached layout from previous render
-        if self.active_window().menu_bar_visible {
-            if let Some(ref menu_layout) = self.active_chrome().menu_layout {
-                if let Some(menu_idx) = menu_layout.menu_at(col, row) {
-                    return Some(HoverTarget::MenuBarItem(menu_idx));
+    /// Handle mouse drag event
+    pub(super) fn handle_mouse_drag(&mut self, col: u16, row: u16) -> AnyhowResult<()> {
```
sinelaw, line 939: **The drag/release channel never joined the registration model, and the PR description reads as though it did.**

`ChromeComponent` has no `on_drag` / `on_release`. This function is ~325 lines (856–1180) of `match PointerGrab` with most bodies written out *inline in the walk file*: the vertical-scrollbar loop over `split_areas`, ~85 lines of horizontal-scrollbar thumb math, popup selection-extend, the prompt scrollbar, and the popup scrollbar. Only `SplitSeparator`, `ExplorerWidth`, `TerminalSelectPending` and `TabDrag` delegate out.

The consequence is that a single surface's gesture handling is split across two files by *channel* rather than kept with its registration:

- prompt scrollbar **press** → `chrome/prompt.rs::handle_click_prompt_scrollbar`
- prompt scrollbar **drag** → here, `PointerGrab::PromptScrollbar` arm
- popup scrollbar **press** → `chrome/popups.rs::handle_click_popup_scrollbar`
- popup scrollbar **drag** → here, `PointerGrab::PopupScrollbar` arm

So adding a drag surface still means editing four central places (the `PointerGrab` enum, `pointer_grab()`, this match, and the mouse-up finalizer match) — not "one component file + one registry line".

To be clear about what I'm *not* claiming: the grab-slot design itself is right, and the ruling behind it (no re-hit-testing mid-drag — the btop-resize bug) is well argued. The issue is scope of the claims. "ONE engine per channel, ALL gestures" and "#3 dispatch-shim depth — fully closed: behavior bodies live with their components; the walk file holds engines and routing only" are accurate for the four press kinds and both wheel axes, but drag and release are neither registration-based nor body-free here. Suggest either an `on_drag(&self, ed, grab, col, row)` / `on_release` pair on the trait keyed by the grab's owning component, or — cheaper and still honest — moving each arm's body to its component's file (as the press halves already are) and downgrading the claim to "press and wheel are one engine each; drag/release are grab-keyed by ruling, with bodies owned by their components".

---

crates/fresh-editor/src/app/mouse_input.rs
```diff
@@ -196,16 +54,45 @@ impl Editor {
...
+                ) {
+                    return result;
+                }
+            }
+        }
+
+        // Cancel the LSP-rename prompt on ANY mouse interaction.
+        // RULING — pre-band whole-channel observer, the mouse analogue
+        // of the keyboard's transient-popup dismissal: it must fire on
+        // every event kind (click, wheel, even bare motion) wherever it
```
sinelaw: **This observer's ruling and its position disagree.**

The comment says it is a "pre-band whole-channel observer" that "must fire on every event kind (click, wheel, even bare motion) wherever it lands" and that "routing proceeds to capture/walk as if it weren't here". But the modal capture band directly above (lines 66–85) `return`s the capturing component's result, so this never runs while any capturing modal is up — Settings, KeybindingEditor, CalibrationWizard, WorkspaceTrust, or FloatingModal.

Compare its stated keyboard analogue, the transient-popup dismissal in `input.rs` (~247): that one really is placed ahead of `dispatch_layer_keyboard`, and its own comment leans on exactly that ("typing under Settings still dismisses a hover popup"). Here the equivalent — mousing while Settings is open — does *not* cancel an LSP-rename prompt.

Practical impact is small (an open rename prompt underneath a full-screen modal is a narrow state), so this may well be fine as-is. But one of the two should change: either hoist the block above the capture band to match its ruling, or reword the ruling to say it observes everything the capture band lets through. As written, the next reader will trust the comment.

---

crates/fresh-editor/src/app/mouse_input.rs
```diff
@@ -1082,385 +731,477 @@ impl Editor {
...
+        press: super::chrome::PointerPress,
+        col: u16,
+        row: u16,
+        modifiers: crossterm::event::KeyModifiers,
+    ) -> AnyhowResult<()> {
+        let tree = super::chrome::chrome_tree(self);
+        let mut seen = std::collections::HashSet::new();
+        for i in crate::widgets::layout_box::hit_stack(&tree, row as u32, col as u32) {
+            let b = &tree[i];
+            if !seen.insert(b.lb.kind) {
```
sinelaw: **Dedup keyed on a `&'static str` imposes a global uniqueness constraint that nothing enforces — and the three walks don't agree on the policy.**

`seen` is keyed on `b.lb.kind`, so the first hit box of a given kind string is the only one whose owner is consulted. That is sound *today* for the reason the doc comment gives: multi-box surfaces resolve by position over their whole collection inside the handler (`Popups::on_pointer` re-scans `popup_areas`, `handle_click_global_popups` walks the whole list, etc.), and kind strings happen to be component-prefixed so two components never collide.

But it's an implicit invariant across a flat string namespace shared by 17 independently-authored components, with no compile-time check and a silent failure mode: a new component that reuses an existing kind string, or an existing component that starts pushing the same kind at two different z bands with position-dependent handling, gets one dispatch and the second box is quietly never offered. That is precisely the class of bug this arc exists to remove — a precedence fact encoded somewhere other than the tree.

Two cheap options: key on `(b.owner, b.lb.kind)`, which preserves the intended "dispatch a surface once" semantics without the cross-component coupling; or drop the dedup and require handlers to be idempotent per box (they mostly already are, since they re-resolve by position).

Related: the three walks now carry two different dedup policies — `dispatch_pointer` dedups, while `dispatch_wheel` (400) and `compute_hover_target` (734) offer every hit box. That asymmetry is defensible (wheel needs chaining) but is currently undocumented at any of the three sites; worth a line in each so the difference reads as deliberate.

---

crates/fresh-editor/src/app/mouse_input.rs
```diff
@@ -254,11 +138,27 @@ impl Editor {
...
+        // pointer-opaque chrome box over the cell (an info popup, the
+        // suggestions dropdown, the theme-info popup) must take the
+        // event in the walk — forwarding it would inject mouse codes
+        // into the PTY *through* the popup. This replaces growing the
+        // hand list one surface at a time; the context-menu check
+        // above stays NAMED by ruling because its boxes are
+        // deliberately not opaque (its close-guard backdrop owns
+        // outside clicks), so opacity cannot express it.
+        let opaque_chrome_over_point = {
+            let tree = super::chrome::chrome_tree(self);
```
sinelaw: **`chrome_tree()` is built twice per mouse event — including on bare motion.**

This gate builds the full tree (17 components' `collect`) and runs a `hit_stack` sort purely to answer "is any `pointer_opaque` box over this cell". It runs unconditionally on *every* `MouseEvent` that gets past the capture band, whatever the kind. Then whichever arm handles the event builds the tree again: `dispatch_pointer` (815), `dispatch_wheel` (407), or `compute_hover_target` (741) via `update_hover_target`. So a mouse-move stream costs two full collects and two `hit_stack` sorts per event.

The derivation itself is a clear improvement over the hand-grown surface list it replaced — no argument there. The observation is only that the "rebuilt per event from live state, deliberately never persisted" ruling (`chrome/mod.rs:12`) is stated once and paid twice. Caching the tree for the duration of a single `handle_mouse` call is free with respect to that ruling — same event, same state, no staleness window — and halves the cost:

```rust
let tree = super::chrome::chrome_tree(self);   // once, after the capture band
// ...threaded into the forward gate and the dispatch arms
```

The blocker is that `dispatch_pointer` / `dispatch_wheel` take `&mut self` and would need the tree passed in, which is mechanical but touches all three engines — reasonable to defer, worth recording.

For what it's worth, I don't think this is currently hot: most `collect` bodies are rect reads off paint caches, and `Menu::collect` already gates the expensive `menu_layout_now()` walk on "menu open" with exactly this reasoning ("recomputing dropdown geometry on every mouse move with no menu open would be pure waste"). Same argument applies one level up.

---

crates/fresh-editor/src/app/widget_runtime.rs
```diff
@@ -1223,255 +1000,63 @@ impl Editor {
...
+                    // every other menu in the dock. Previously
+                    // ↑/↓ were silently dropped here.
+                    self.handle_widget_focus_advance(panel_key, delta);
+                }
+                if let Some(target_key) = scrollable {
+                    let target_kind = self.widget_registry.get(panel_key).and_then(|p| {
+                        crate::widgets::find_widget_by_key(&p.spec, &target_key).cloned()
+                    });
+                    match target_kind {
+                        Some(fresh_core::api::WidgetSpec::List { .. }) => {
```
sinelaw: **Per-kind *key* policy still lives outside `kinds/` — narrower than "adding a kind touches exactly one dispatch line".**

I grepped every `WidgetSpec::` site to check the single-dispatch claim. It holds where the PR says it does: `render.rs`'s 125 occurrences are essentially all inside its `mod tests`, so `render_collected` really is pure delegation to `behavior()`, and the pointer / wheel / focus-change paths all route through it too. `actions.rs` and `registry.rs` are structural traversals (keys, children), not behavior.

The exceptions are here and in one other place:

- `handle_widget_key` (1001–1088): the Up/Down/Enter picker-forwarding policy hand-matches `Button`/`Toggle` (1010–1011), `List`/`Tree` (1030, 1036), and `Text` (1055, 1073, 1076). So *which* keys a kind participates in for panel-level nav is decided by a central `match`, not by the kind.
- `plugin_dispatch.rs:5366–5417`: `Number` / `Dropdown` / `DualList` arms for `WidgetAction` value-setting (min/max clamp, option resolution).

`docs/internal/chrome-event-model-plan.md` does record this honestly as residue ("the widget_runtime central kind policies ... candidate `arrow_peek` / `activate_on_picker_enter` capabilities if drift bites"), and adding a kind today is genuinely one line *for rendering*. It's the PR summary that generalises past the code — worth qualifying there, since a new kind that wants arrow participation or a settable value does still touch two central matches.

The `BoxMeta.picker_scroll_target` capability already added in this PR looks like the right precedent for closing it: the same treatment for "participates in arrow peek" and "activates on picker Enter" would let these arms become capability checks.

---

crates/fresh-editor/src/app/input.rs
```diff
@@ -252,126 +259,29 @@ impl Editor {
...
+        // capture-all modals, workspace trust, the prompt rungs, the
+        // popup rungs (including the unfocused popup-cancel/-focus
+        // interception), down to the editor base, whose handler is the
+        // pipeline tail (mode bindings, composite routing,
+        // chord/keybinding resolution — see `dispatch_base_key` in
+        // `chrome/base.rs`) — offered the key top-down in declared
+        // rank order. See `dispatch_layer_keyboard`.
+        let result = self
+            .dispatch_layer_keyboard(&key_event)
+            .expect("editor base layer answers every key");
```
sinelaw: **This `.expect` is load-bearing on two invariants in a different file, with nothing pinning either.**

It holds only while both of these stay true in `chrome/base.rs`:

1. `Base::layers` pushes `EDITOR_BASE` **unconditionally** (79–97) — no state gate, ever; and
2. `Base::on_layer_key` returns `Some(..)` for every key (99–111).

Both are correct today and both are commented as deliberate ("the keyboard owner of last resort ALWAYS answers — the walk terminates here (`handle_key` relies on it)"). But the coupling is cross-file and invisible from here: someone adding a plausible-looking guard to `Base::layers` — say, skipping the base layer while a capture-all modal is up — turns every keystroke into a panic on the main input path, in release builds too.

Given the walk is the whole keyboard pipeline now, degrading rather than panicking seems worth it:

```rust
let result = self
    .dispatch_layer_keyboard(&key_event)
    .unwrap_or(Ok(crate::input::handler::InputResult::Ignored));
```

An unhandled key is a far better failure mode than a crash, and the invariant is still worth asserting — just as a `debug_assert!` on the stack, or a unit test over `overlay_stack()` asserting the last entry is always `LayerKind::Editor` with `owns_keyboard: true`.

---

crates/fresh-editor/src/app/chrome/mod.rs
```diff
@@ -0,0 +1,505 @@
...
+    /// from the same activity predicate its capture gates on), so the
+    /// `layer_rank` consts are the ONE precedence source for capture
+    /// and keyboard alike — registry order no longer encodes it.
+    ///
+    /// Whole-event capture (rather than per-gesture boxes) is the
+    /// honest intermediate: the modal handlers own presses, drags,
+    /// releases, and hover as one unit; decomposing them onto the
+    /// gesture walks needs the pointer-grab slot and the scan's
+    /// opacity gate, recorded as the residue of this slice.
+    fn capture_mouse(
```
sinelaw, line 400: **The contract that makes the capture band correct is enforced only by this comment, and violating it fails silently.**

Since R3, `handle_mouse` offers `capture_mouse` by walking `overlay_stack()` and calling `components()[owner]` for each owner it finds (`mouse_input.rs:66–85`). That means a component whose `capture_mouse` returns `Some` but whose `layers()` declares **nothing** in the current state is *never offered the capture at all* — its `capture_mouse` is unreachable code. The doc here states the requirement ("every capturing component declares a layer from the same activity predicate its capture gates on"), which is what makes rank the single precedence source, and I verified all five implementors currently honour it:

| component | `capture_mouse` gate | `layers` gate |
|---|---|---|
| `Settings` | `settings_up(ed)` | `settings_up(ed)` |
| `KeybindingEditor` | `keybinding_editor.is_some()` | same |
| `CalibrationWizard` | `calibration_wizard.is_some()` | same |
| `WorkspaceTrust` | `workspace_trust_on_top()` | same |
| `FloatingModal` | `floating_widget_panel.is_some()` | same (presence; `owns_keyboard` keys off `focused`) |

The failure mode if that pairing ever drifts is the bad kind: no compile error, no panic — the modal simply stops being modal for the mouse, and events start leaking to the buffer/terminal beneath it, which is the exact bug class `dispatch_modal_mouse` existed to prevent. The two gates are also textually far apart (different methods, sometimes different files), so drift is easy.

Suggestion: hoist the predicate into one method — e.g. an `fn active(&self, ed: &Editor) -> bool` that both `layers()` and `capture_mouse()` consult — so the pairing is structural rather than a convention. Failing that, a test that drives each modal's state up and asserts `overlay_stack()` contains a layer owned by that component would catch drift.

---

crates/fresh-editor/src/app/chrome/mod.rs
```diff
@@ -0,0 +1,505 @@
...
+/// Overlay-layer ranks, DESCENDING = higher keyboard/modal precedence.
+/// Deliberately independent of both registry order and box z: the menu
+/// LAYER outranks the prompt's and the popups' (an open menu owns the
+/// keyboard over them) while its boxes sit in a lower pointer band,
+/// and the native context menus rank BELOW the popup layer (the
+/// unfocused-popup guard's `take_while` must not see them) while their
+/// boxes sit at the very top. Event-debug (1000) is hardcoded in
+/// `overlay_layers` — a debugging instrument, not a component.
+pub(crate) mod layer_rank {
+    pub(crate) const SETTINGS: u16 = 900;
```
sinelaw, line 49: **This 12-line const block is now the single precedence source for three channels, and nothing tests it. There are zero tests under `app/chrome/`.**

These ranks drive, all at once: the keyboard walk (`dispatch_layer_keyboard`), the mouse capture band (post-R3), the PTY gate via `any_layer_blocks_terminal_input`, `get_key_context()` via `resolve_focus_context`, and `popup_blocked_by_higher_modal`'s `take_while`. Consolidating five encodings into one is the right move — but it also means a one-character edit here silently changes behavior in five places.

What *is* covered, and well: `widgets/layout_box.rs` has 11 tests pinning `hit_stack`'s band/depth/document-order rules, effective-z lifting, `screen_space` exclusion, and `focus_ring_scoped` trap scoping. `overlay.rs` has 5 tests over `resolve_focus_context` / `any_layer_blocks_terminal_input` on hand-built layer slices.

What isn't covered anywhere:

- the **rank ordering itself** — nothing asserts `WORKSPACE_TRUST > PROMPT > POPUP > CONTEXT_MENU`, even though two of those relations are called out in the PR body as deliberate behavior changes ("workspace-trust keys now rank above an open prompt"). A regression there is invisible.
- the ordering constraint this block's own comment describes as load-bearing: context menus must rank *below* `POPUP` so `popup_blocked_by_higher_modal`'s `take_while` doesn't see them. That's a real coupling between a const and a `take_while` in another file, asserted nowhere.
- the `Disposition` contract — specifically `PassAfter` vs `Pass`-on-an-opaque-box, which mod.rs:82–94 documents as "NOT interchangeable" and notes only diverges "if an opaque surface adopts observer semantics". That's the kind of latent rule that wants a test more than a comment.
- `stack.sort_by(|a, b| b.0.cmp(&a.0))` relying on sort stability for intra-rank registry order (`overlay.rs:265`).

The PR's diff touches 4 test files and adds no new ones, so the whole chrome layer is verified only end-to-end through the existing e2e suites. For a behavior-preserving refactor of this size that's a defensible strategy — but a handful of pure unit tests over `overlay_stack()` and the rank relations would be cheap and would pin exactly the facts that are now single-sourced here.

---

sinelaw (commented):
Read the branch at `06c4199` against the code rather than the plan docs; `cargo check --all-targets` is clean. The model is real and consistently applied: `chrome_tree()` + `hit_stack` for the four press kinds and both wheel axes, the same walk in query form for hover, and `overlay_stack()` + `on_layer_key` for the keyboard, with the old `dispatch_modal_keyboard` / `dispatch_modal_input` / `chrome_boxes` genuinely gone (they survive only in comments). The z-band ×10 scale and the derived rank stack are doing what the plan claims — one ordering per channel instead of hand-ordered ladders.

Eight inline findings, roughly in descending order of how much they matter:

1. **`mouse_input.rs:856`** — drag/release never joined the model: no `on_drag`/`on_release` on the trait, ~325 lines of grab bodies inline in the walk file, and press/drag halves of the same scrollbar split across two files. This is the main place the "ALL gestures" / "walk file holds engines and routing only" claims outrun the code.
2. **`widget_runtime.rs:1030`** — per-kind *key* policy still hand-matched centrally (plus `plugin_dispatch.rs:5366–5417`). The single-`match` claim does hold for render/pointer/wheel/focus — I checked every `WidgetSpec::` site — but not for panel key nav or value-setting.
3. **`chrome/mod.rs:43`** — the rank table is now single-sourced for five consumers and untested; there are zero tests under `app/chrome/`. `layout_box.rs` and `overlay.rs` are well covered, which makes the gap more conspicuous.
4. **`chrome/mod.rs:349`** — the `capture_mouse` ↔ `layers` predicate pairing that makes the capture band correct is comment-enforced; drift makes a modal silently non-modal for the mouse.
5. **`mouse_input.rs:819`** — dedup keyed on a global `&'static str` namespace, unenforced, silent failure mode; and the three walks carry two undocumented dedup policies.
6. **`input.rs:282`** — `.expect("editor base layer answers every key")` is load-bearing on two invariants in another file; a panic on the main input path if either changes.
7. **`mouse_input.rs:156`** — `chrome_tree()` built twice per mouse event, including bare motion.
8. **`mouse_input.rs:90`** — the LSP-rename observer's "pre-band, fires on everything" ruling contradicts its placement after the capture band.

None of these is a blocker on its own; 1–3 are the ones I'd want resolved (or the claims trimmed) before this is treated as the finished state of the arc, since each is a place where a future reader would trust a stated invariant that the code doesn't hold.

---

sinelaw:
## Gap list against the end goal

Measured against the stated goal — **generic event registration, propagation and dispatch with no hand-specified exceptions** — rather than against the previous state. Under that goal a documented ruling isn't closure, it's a deferred exception, so several items this PR closes "by ruling" are counted open below.

### 1. Precedence is still hand-authored data

The biggest one, because everything else inherits from it. Four separate manual orderings:

- **`layer_rank`** (`chrome/mod.rs:42–54`) — 11 magic numbers picked by hand, with prose explaining why the menu outranks the prompt.
- **Box `z` values** — ~30 hard-coded integers spread across the 17 `collect()` bodies on a ×10 scale (195 dock blur, 175 transient guard, 155 suggestion capture, 15 overlay scrim…), where the *gaps* are load-bearing and only comments explain them.
- **Registry array order** — "within a band, registry order IS precedence", so ordering is encoded twice: z number *and* array position.
- **`pointer_grab()`** (`chrome/mod.rs:223`) — a 13-arm if-ladder whose arm order encodes drag precedence, explicitly preserved from the ladder it replaced.

Generic version: precedence declared *relationally* by each component (`above(Popup)`, `modal`, `guard_over(self)`) and topologically sorted, or derived from a real containment/modality relation — not integers a human assigns. Until then every new surface is a hand-placed exception, and nothing detects a collision or an inversion.

### 2. Two channels have no registration at all

- **Drag + release** — no `on_drag`/`on_release` on the trait. A 13-variant enum, a hand-ordered derivation, a central `match` with bodies inline (`mouse_input.rs:856–1180`), plus a second `match` for release finalizers. Adding a drag surface = 4 central edits.
- **Chrome focus** — three unrelated focus states (dock `focused`, `Prompt.toolbar_focus`, popup `focused`) and no chrome focus ring, even though chrome boxes already carry `focusable`/`focus_trap` and `focus_ring_scoped` would work on them today. Nothing sets those flags at chrome level.

Also **hover reactions** are a broadcast to all 17 components (each self-filtering by target variant) rather than propagation along the hit path — a different mechanism from every other channel.

### 3. The pre-band: nine stages that bypass the walks by design

Keyboard: event-debug intercept (a hardcoded rank-1000 layer with `owner: None`), terminal-input gate, plugin `getNextKey` capture, the `on_key` grab loop, transient-popup dismissal.
Mouse: capture band, LSP-rename observer, terminal-forward gate, terminal-link Ctrl+click — plus three content trackers (`update_lsp_hover_state`, `update_terminal_link_hover`, `update_widget_hover`) beside the hover walk.

Two structural problems inside this, beyond the count:

- **`on_key` is a second keyboard entry point** whose members outrank every declared rank by pipeline position. That is why `ContextMenu`'s rank (830) deliberately isn't its keyboard precedence — the rank table lies for one component.
- Nothing expresses **observers** (act-and-continue, see-everything) or **captures** (whole-channel) as propagation *phases*. `Disposition::PassAfter` is the pointer-side version of exactly this and it works; the keyboard has no equivalent, so observers get hoisted out of the walk instead. A generic model needs capture → target → bubble with observer registration, and then most of these nine stages become ordinary registrations.

### 4. `capture_mouse` + bespoke interiors

Whole-channel capture is the exception mechanism itself: 5 modals bypass all four gesture walks, and the trait doc concedes decomposing them needs the grab slot and the opacity gate. Behind it, ~8 bespoke `InputHandler` interiors (Settings, keybinding editor, calibration wizard, trust dialog, menu, file browser, query-replace confirm, prompt) are invoked from components but route nothing generically. Settings-onto-the-framework and prompt-as-widgets are the unstarted work that would let capture be deleted.

### 5. Three parallel pointer paths

- **`screen_space` boxes never hit-test** → a shadow path against paint-recorded rects (`popup_hits`, `popup_rect`, `prompt_toolbar_boxes`, `file_browser_layout`, `workspace_trust_dialog`, suggestion rects). Still open for exactly the surfaces most likely to gain new controls.
- **Buffer-mounted panels** drop the overlay/popup channels entirely and click with `on_overlay=false` (`plugin_dispatch.rs:5142`) — a third path with different capabilities from dock/floating.
- **Federated chrome vs panel trees**, with *opposite* tie-breaks (`hit_stack` earlier-wins, `hit_path` later-wins), so moving a surface between them silently changes semantics.

One tree, one primitive, one tie-break is the target.

### 6. Handlers re-hit-test what the walk just resolved

`is_mouse_over_any_popup`, `is_mouse_over_file_browser`, `is_mouse_over_transient_popup`, plus most `on_pointer` arms re-scanning `popup_areas`/`suggestions_area` to find the row. The walk resolves the box, then the handler re-derives the same fact from rect caches. Generic version: the walk hands the handler its resolved target (box + row/index) so a handler never queries geometry.

### 7. Remaining per-kind central matches

`handle_widget_key`'s arrow-peek / picker-Enter policy (Button/Toggle/List/Tree/Text, `widget_runtime.rs:1001–1088`) and `plugin_dispatch.rs:5366–5417`'s Number/Dropdown/DualList value-setting. `BoxMeta.picker_scroll_target` is the right precedent — these want the same treatment (`arrow_peek`, `activates_on_picker_enter`, `settable_value`) and then the matches disappear.

Minor: the focus ring is now genuinely single-ruled (`collect_tabbable` asks `box_meta`), but two *traversals* remain, guarded by a `debug_assert`, with retirement gated on constraint layout.

### 8. Invariants held by comment instead of structure

`capture_mouse` ⇄ `layers` predicate pairing; base-always-answers (an `.expect` on the main input path); box-kind string uniqueness (the dedup key); the rank relations (WT > Prompt > Popup, and ContextMenu < Popup for `popup_blocked_by_higher_modal`'s `take_while`); sort stability for intra-rank order. Each fails silently or panics rather than failing to compile.

### 9. Efficiency

- **`chrome_tree()` built 2× per mouse event**, including bare motion (forward gate, then the dispatch arm).
- **`overlay_stack()` built 3–4× per keystroke** — once by the walk, then again by every `get_key_context()` inside a handler (`base.rs:131`, `dock.rs:182`) and by the transient-dismiss stage. Each rebuild runs all 17 `layers()` bodies plus a sort.
- **`hit_stack` re-walks parent chains inside the comparator** → O(n log n × depth); effective z wants computing once per box per event.
- **Hover broadcasts to 17 components per motion event.**

Not hot today (most `collect` bodies are cached-rect reads), but "efficient" in the goal means the derived structures get built **once per event and threaded** — which is also what makes single-tree dispatch affordable later.

---

**Suggested order**, since these unlock each other: relational precedence (1) first, as it's the data every channel reads → then phases + observer/capture registration (3), which lets the pre-band and `capture_mouse` (4) dissolve into ordinary registrations → then drag/release and chrome focus onto the same mechanism (2) → then unify the trees and retire the paint-rect path (5), which is what finally lets handlers stop re-hit-testing (6). Items 7–9 are independent and can land anytime; 8 is cheap and worth doing first, since it's what catches regressions during 1–5.

---

sinelaw:
## What must land here regardless

`1776f4c` and `a052f49` closed seven of the eight inline findings — I re-verified each against the branch: drag bodies now live in their components (`handle_mouse_drag` is ~104 lines of routing, down from ~325), dedup is `(owner, kind)`, one `chrome_tree` per mouse event, `handle_key` degrades instead of panicking with the contract under test, capture/layer gates paired on one predicate per modal, first unit tests under `app/chrome/`, and the LSP-rename observer resolved by rewording with a real rationale (a click aimed at Settings must not reach through it) rather than moving the block. That last one is the better answer; I'd have argued for moving it.

The forward design is now split out as **#3024** (stacked on this branch). This comment is the other half of that split: what should **not** wait for it. The filter I applied — either it's free and corrects the record, or it prevents the model from rotting the first time someone adds a surface or a kind.

---

### MUST

**1. Say that the chrome tree is still flat.** `chrome-event-model-plan.md` opens by listing chrome's five deficiencies — "a central enumeration (no registration), **flat (no parents, no bubble)**, ordered per-gesture, guarded by pre-walk ladders, handlers in central match arms" — and says the doc is the design for closing that gap. Four are closed. That one isn't: no component sets `parent`, so all ~50 chrome boxes are rootless, and `hit_stack`'s two structural rules (effective z = max along the ancestor chain, children-above-parents) are dead code at chrome level — every box has depth 0 and effective z = own z. The 15 full-frame guard boxes are the workaround for the missing containment test.

I'm not asking for parent links here — that's real work and it changes routing. I'm asking that it not merge *implying* it's done, because the type has a `parent` field, the doc calls the result a tree, and the next person to write a chrome surface will either wonder why nesting doesn't work or add a 16th guard box thinking that's the pattern. A ruling at `ChromeTreeBuilder` plus a line in the status section, noting that `parent` / `focusable` / `focus_trap` / `scroll` are reserved-but-unset at chrome level, costs nothing and is the single highest-value paragraph left in this PR.

**2. Close the per-kind policy matches.** 13 `WidgetSpec::` sites still sit outside `kinds/`: `handle_widget_key`'s arrow-peek and picker-Enter policy (Button/Toggle/List/Tree/Text) and `plugin_dispatch`'s Number/Dropdown/DualList value-setting. The headline claim is "adding a kind touches exactly one dispatch line", which is true for rendering and false for keys — so the first new kind that wants arrow participation drifts them. `BoxMeta.picker_scroll_target` is already the precedent in this PR; `arrow_peek`, `activates_on_picker_enter` and `settable_value` finish the job cheaply, and then the claim is simply true.

**3. Trim the claims that outrun the code.** R8 downgraded the drag claim, which is exactly the right move — the same treatment is owed to "#3 dispatch-shim depth — **fully closed**", "geometry from layout (slice 7) — **complete**", "**ALL 21 ITEMS CLOSED**", and the one-dispatch-line claim above. An overstated claim is worse than an open TODO: a TODO invites a look, a "closed" stops one. This is a text edit and it's what makes the doc trustworthy as the reference for the next arc.

**4. Precedence golden tests.** The three new chrome tests pin rank *relations*; nothing pins routing *outcomes*. The regressions this arc exists to prevent are all of the form "event reached the wrong surface": a click on a popup's border doesn't move the buffer cursor beneath; Esc with a prompt open and the dock focused cancels the prompt; wheel over a popup doesn't pan the buffer behind; a right-click under an overlay prompt is swallowed; workspace-trust keys beat an open prompt (a deliberate change in this PR, currently unpinned). Half a dozen tests over the real dispatch entry points, and the next refactor can't silently invert them.

**5. Enumerate the paint-recorded surfaces in one closed list.** The `screen_space` class is documented at its sites and named in the PR body (`workspace_trust_dialog`, `prompt_toolbar_*`, `file_browser_layout`, plus popup and suggestion rects). It's the parallel geometry path, so it needs one list with one owner and "additions require a ruling" — otherwise it grows surface by surface, which is how it got here.

### SHOULD — same class as fixes already made

**6. `overlay_stack()` is still rebuilt 3–4× per keystroke.** Once by the walk, then again by `get_key_context()` in `chrome/base.rs` and `chrome/dock.rs`, plus the transient-dismiss stage and `popup_blocked_by_higher_modal`. Each rebuild runs all 17 `layers()` bodies and a sort. R8 fixed precisely this on the mouse side ("ONE chrome_tree per mouse event"); the keyboard side is the same fix, threaded through the walk.

**7. Pin the `Disposition` contract.** `chrome/mod.rs` documents `PassAfter` and `Pass`-on-an-opaque-box as "NOT interchangeable" and notes they only diverge if an opaque surface adopts observer semantics. That's a latent rule with no test; one test is cheaper than the bug.

### DEFER — and say so, so they don't read as oversights

Paint-time compositing and the mounted-panel overlay/popup channels; constraint layout; the `LayoutSpec` intermediate; chrome-as-nodes; chrome focus unification and deleting `layer_rank`; Settings and prompt interiors onto the framework. All of these are #3024's territory and none of them should hold this PR.

---

Net: items 1, 3 and 5 are documentation and cost an hour. Items 2, 4, 6 and 7 are small code changes that either make an existing claim true or keep it true. Everything structural can wait.

---

crates/fresh-editor/src/app/mouse_input.rs
sinelaw: Fixed (R8 wave), taking the "grab-keyed by ruling, bodies owned by their components" shape. Every inline drag body moved to its owning component's file: prompt scrollbar → `chrome/prompt.rs::handle_prompt_scrollbar_drag`, popup scrollbar + selection-extend → `chrome/popups.rs`, vertical/horizontal scrollbar + separator → `chrome/splits.rs`, dock resize → `chrome/dock.rs`. The `PointerGrab` match in `handle_mouse_drag` is now pure routing — every arm is a one-line delegation — and the release finalizer is grab-keyed the same way (R4). The PR description's channel claims were also trimmed to match (press and wheel are one engine each; drag/release are grab-keyed by the no-re-hit-testing ruling, with bodies component-owned).

---

crates/fresh-editor/src/app/mouse_input.rs
sinelaw: Fixed by rewording the ruling to match the position (the second option). The observer's comment (`mouse_input.rs` ~107) now states it fires on "every event kind the capture band above lets through" and explains why sitting BELOW the band is deliberate — a capturing modal owns its events wholly, and a click aimed at Settings must not reach through it to cancel a rename prompt the user cannot see — with the explicit contrast against the keyboard's transient-popup dismissal, which does observe under modals. The comment and the code now agree.

---

crates/fresh-editor/src/app/mouse_input.rs
sinelaw: Fixed (R8 wave), taking the first option: the press-walk dedup is now keyed on `(b.owner, b.lb.kind)` (`dispatch_pointer`), so the "dispatch a surface once" semantics hold per component with no cross-component string-namespace coupling. The per-walk asymmetry is also now documented as a deliberate DEDUP POLICY at all three sites: presses dedup (multi-box surfaces resolve by position in their handler), wheel deliberately offers every box (scroll chaining needs it), and hover is first-`Some`-wins so dedup would be a no-op.

---

crates/fresh-editor/src/app/mouse_input.rs
sinelaw: Fixed (R8 wave), and then some. First, the mechanical fix suggested here: ONE `chrome_tree` per mouse event, built after the pre-band observers (the LSP-rename cancel can close a prompt and change geometry) and threaded as `&[ChromeBox]` into the forward gate and all dispatch arms — a motion stream no longer pays two collects + two `hit_stack` sorts per event. Then the R10 slice went further: `chrome_tree` is now memoized across events under a validated claim (coarse `ui_gen` epoch match + a fresh `overlay_stack` build equal to the snapshot the cached tree was built from), with every hit oracle-checked against a full rebuild in debug. A quiet motion stream now reuses one tree across many events, not just within one.

---

crates/fresh-editor/src/app/widget_runtime.rs
sinelaw: Fixed (R9 wave), taking exactly the `picker_scroll_target` precedent suggested here. Both central matches became capabilities/kind fns: `handle_widget_key`'s Up/Down/Enter arms now consult `arrows_advance_focus()` (Button/Toggle), `picker_nav()` (`PickerNav::{Skip, Peek, TakeFocus}` — List peeks, Tree takes focus), and `activates_on_picker_enter()` + `picker_activate_event()` (List/Tree), so a new kind opts into panel nav from its own `WidgetImpl`. The `plugin_dispatch.rs` value-set arms now delegate to kind-owned mutation fns (`number::set_value_state`, `dropdown::set_index_state`, `dual_list::set_included_state`) — the dispatch arms hold routing only, no clamp/resolve policy. The PR summary was also qualified in the same wave (claim trims).

---

crates/fresh-editor/src/app/input.rs
sinelaw: Fixed (R8 wave), exactly as suggested: `handle_key` now degrades — `dispatch_layer_keyboard(...)` falls back to `Ok(InputResult::Ignored)` in release, with a `debug_assert!` that screams if the walk ever falls past Base. The invariant itself is pinned both ways: the load-bearing sites in `chrome/base.rs` are commented as such, and the `overlay_stack_always_ends_with_an_owning_base_layer` unit test in `chrome/mod.rs` asserts the stack's last entry is always the Editor layer with `owns_keyboard: true`, across a matrix of overlay states.

---

crates/fresh-editor/src/app/chrome/mod.rs
sinelaw: Fixed (R8 wave), taking the hoist option: each capturing modal now has ONE named activity predicate consulted by BOTH `layers()` and `capture_mouse()` — `settings_up` / `kb_editor_up` / `calibration_up` (+ workspace-trust) in `chrome/modals.rs`, `panel_up` in `chrome/floating_modal.rs` — each with a comment stating exactly this contract ("ONE activity predicate, consulted by BOTH `capture_mouse` and `layers` — the pairing the capture band's correctness rests on"). Drift now requires editing the shared predicate, which changes both gates together; a component can no longer gate its capture on state its layer doesn't declare.

---

crates/fresh-editor/src/app/chrome/mod.rs
sinelaw: Fixed (R8/R9 waves), on both levels. Unit tests now exist under `app/chrome/` (`mod.rs::tests`): `deliberate_rank_relations_are_pinned` asserts every relation called out here — `WORKSPACE_TRUST > PROMPT > POPUP > CONTEXT_MENU`, menu-over-prompt, context-menu-below-popup (the `take_while` coupling, cited in the test) — `ranks_are_distinct` guards accidental equality, `overlay_stack_always_ends_with_an_owning_base_layer` pins the base-layer contract over real editor states (which also exercises the sort's intra-rank ordering), and `pass_after_is_not_pass_on_an_opaque_box` pins the `PassAfter`-vs-opacity divergence via the extracted `pointer_walk_step` — that contract is now code, not prose. On top of those, `e2e/dispatch_precedence.rs` (R9) pins the routing *outcomes* end-to-end: wheel and h-wheel over a popup never move the buffer, a popup-chrome click never moves the cursor, and workspace-trust keys beat an open prompt — the deliberate behavior-change relations this comment flagged as invisible to regression.

---

sinelaw:
Status on the must-land list (#5318034056) — all seven items are on the branch; every inline review thread is now replied to and resolved.

**MUST — all landed (R9 wave):**
1. **Flat-tree ruling** — recorded at `ChromeTreeBuilder`: no component sets `parent`, the full-frame guard boxes are the deliberate flat-world containment encoding, and `parent`/`focusable`/`focus_trap`/`scroll` are called out as reserved-but-unset at chrome level (parent links + chrome focus are #3024). The plan doc's status section says the same.
2. **Per-kind policy matches closed** — `arrows_advance_focus`, `picker_nav` (List peeks, Tree takes focus), `activates_on_picker_enter` + `picker_activate_event` on `WidgetImpl`; the `plugin_dispatch` value-set arms delegate to `number::set_value_state` / `dropdown::set_index_state` / `dual_list::set_included_state`. The central matches are gone.
3. **Claims trimmed** — the overstated "fully closed" / "complete" / "ALL 21 ITEMS" / one-dispatch-line claims were qualified in the PR body and plan doc.
4. **Precedence golden tests** — `e2e/dispatch_precedence.rs`: wheel and horizontal wheel over a popup never move the buffer, a popup-chrome click never moves the cursor, workspace-trust keys beat an open prompt (the deliberate rank change, now pinned as an outcome).
5. **Closed `screen_space` roster** — one list on `ChromeLayout`'s doc, additions require a ruling.

**SHOULD:**
6. **Repeated `overlay_stack()` builds** — resolved half by ruling, half by making the expensive part cheap. The ruling (recorded at `dispatch_layer_keyboard`): handler-level rebuilds are load-bearing freshness — handlers may mutate then decline, so a stack snapshot threaded through the walk would route the next rung against stale claims; invalidation-aware derivation is #3024's territory. The cost side landed as the R10 slice: `chrome_tree` (the expensive derivation) is memoized across events under a checked claim — coarse `ui_gen` epoch match plus a fresh `overlay_stack` build equal to the snapshot the cached tree was built from, every hit oracle-checked against a full rebuild in debug. `overlay_stack` itself deliberately stays unmemoized: it is ~17 cheap predicate reads and is the ground truth the tree cache validates against. (A first cut that trusted a hand-bumped counter for both was caught by its own oracle in CI — 82 debug-build failures across four unbumped mutation families — which is now recorded as a ruling in the plan doc: staleness is checked, not trusted.)
7. **`Disposition` contract pinned** — the `PassAfter`-vs-opacity step is extracted as `pointer_walk_step` and unit-tested (`pass_after_is_not_pass_on_an_opaque_box`); code, not prose.

The DEFER list is recorded as #3024's territory in the plan doc's status section.

---

sinelaw:
Merged manually

Issue #3006, opened 2026-08-14
Fix: Selection Range Does Not Expand When Selecting Lines at the Top and Bottom of a File


akarinotomoshibi:
## Fix: Selection Range Does Not Expand When Selecting Lines at the Top and Bottom of a File
In VSCode and other editors, the following behavior is implemented:
- When pressing Shift+Up on the topmost line, the selection range is extended from the current cursor position to the leftmost position (the 0th character)
- When pressing Shift+Down on the bottommost line, the selection range is extended from the current cursor position to the rightmost position (the final character)

Currently, this functionality does not operate as expected, and no selection is made.

I respectfully believe this to be a significant issue, as it necessitates the verification of the selection range each time one works with selections, which substantially impairs the user experience.

## Additional
This issue also manifests when a user drags the mouse cursor beyond the boundaries of the text area.
Additionally, there is a further problem wherein automatic scrolling fails to function when the mouse cursor is dragged outside the specified region

## Note:
* This issue is translated by Claude Haiku 4.5
* If this issues including problem sentences, I will fix and apologize. I really sorry.

sinelaw:
Reproduced on `claude/fresh-bug-repro-sweep-gdleyn` (e8001dc), debug build (`cargo build`), Linux, tmux 3.4 100x20, config `{ "editor": { "restore_previous_session": false } }` with `HOME` pointed at an empty temp dir and `--no-restore --no-upgrade-check`.

Treated as two separate defects, as reported. **(a) reproduces exactly as described. (b) reproduces in the upward direction and for the autoscroll half; downward drag does extend and scroll, so that half is reported as measured, not as a repro.**

## (a) Shift+Up on the first line / Shift+Down on the last line — reproduced

Steps:

1. `printf 'first line of the file\nsecond line here\nthird line here\nlast line of the file' > sel2.txt` (deliberately no trailing newline, so line 4 really is the last line)
2. `fresh --config <cfg> --no-restore --no-upgrade-check sel2.txt` in a 100x20 pane
3. `Right` ×10 — status reads `Ln 1, Col 11`
4. `Shift+Up`
5. `Down` ×3 — status reads `Ln 4, Col 11`
6. `Shift+Down`

Observed, captured panes verbatim.

Before step 4:

```
  1 │ first line of the file
  2 │ second line here
  3 │ third line here
  4 │ last line of the file
[...]

PR #3016, opened 2026-08-15
fix(mouse): let a drag past the top/bottom edge keep selecting and scrolling

sinelaw:
Addresses the "Additional" half of #3006 — dragging the mouse beyond the text area neither extends the selection nor scrolls. The keyboard half of that issue (Shift+Up/Down at the file edges) is a separate change.

## Two causes, not one

The issue comment traced this to the clamp in `click_geometry.rs`, where `content_row = row.saturating_sub(content_rect.y)` folds every row above the content area onto the topmost visible line, with no counterpart to the explicit "below last visible line" branch. That is real, but it is not what makes the drag freeze in practice. Measuring against the unfixed code turned up a second cause that dominates:

* Drag up after scrolling the viewport **with the keyboard** — already worked. Top line walked `14 → 13 → 10 → 7 → 4 → 1`, selection growing.
* Drag up after scrolling **with the wheel** — the reported symptom exactly. Top pinned at 7, cursor pinned at `Ln 7`, selection frozen at 40 cells across 6+ motion events.
* Drag **down** after scrolling with the wheel — frozen the same way. The original measurement in the issue started from an unscrolled viewport, which is why the downward direction looked healthy there.

The wheel and the scrollbar set `skip_ensure_visible`, and until now only a key press ever cleared it. So the drag moved the cursor and the viewport simply refused to follow, in both directions.

The clamp is a genuine second defect, and isolating it shows how much it was hidden: with `editor.scroll_offset = 0`, dragging past either edge moved the viewport **zero** lines across 8 motion events. Every bit of scrolling in the default config was a side effect of the scroll-off margin — which also explains the fixed 3-line steps regardless of how far past the edge the pointer sat.

## The change

* `ClickTarget` gains `row_undershoot` (rows *above* the content area), the counterpart to the existing `row_overshoot` that the `saturating_sub` used to swallow silently.
* New `position_offset_by_lines`, which walks `n` buffer lines from a position while preserving the pointer's visual column, via `visual_column_of` + a grapheme-cluster walk so multibyte lines stay on a char boundary. Clamps at buffer start/end.
* `handle_text_selection_drag` reads the overshoot/undershoot instead of the clamped position, so the selection head can leave the viewport in both directions, and clears `skip_ensure_visible` on the dragged split before applying the move — mirroring what `handle_key` already does.

Scrolling is now proportional to how far past the edge the pointer is, rather than a fixed scroll-off-sized step.

## Still missing: autoscroll while the pointer is held still

Not implemented here. With the button held and the pointer parked outside the text area, nothing moves until a new motion event arrives, and a stationary pointer emits none — so this needs a tick-driven step.

The groundwork exists: `editor_tick` already hosts deadline-driven checks, and the test harness can drive it deterministically through `tick_and_render()` plus `advance_time()` on the injected `TestTimeSource`, so it can be covered without a time-based test. The shape would be recording the last outside-the-rect pointer position on each drag motion, clearing it on mouse-up and on re-entry, and replaying `handle_text_selection_drag` from `editor_tick` once a rate interval has elapsed on the injected clock — folding the next deadline into `next_periodic_redraw_deadline()`. One trap: existing tick timers such as `check_mouse_hover_timer` use real `Instant::elapsed()` rather than the injected time source, so an autoscroll timer must use `time_source()` or it will not be testable.

## Tests

`crates/fresh-editor/tests/e2e/issue_3006_drag_beyond_text_area.rs`, driving raw mouse `Down`/`Drag` events and asserting only on rendered output — topmost `LINE nnn` marker for scroll position, `theme.selection_bg` cells for the selection. Per-test temp fixture, no timeouts or sleeps.

* `drag_above_the_text_area_scrolls_up_after_a_wheel_scroll`
* `drag_below_the_text_area_scrolls_down_after_a_wheel_scroll`
* `drag_past_the_edges_scrolls_with_scroll_off_disabled` — isolates the clamp with `scroll_offset = 0` and a keyboard-parked viewport, so the skip flag is not involved

Each also drags to a row *inside* the text area and asserts the viewport does not move, so a fix that scrolled on every drag event could not pass.

## Verification status — please read

**The reproducer tests were compiled but never executed, in either direction.** No pass or fail is being claimed for them; CI is the first real check. This was a deliberate instruction to skip local e2e runs, not an oversight.

What *was* measured is the behaviour of the unfixed code described above, using throwaway print-only tests that are not part of this commit. The committed assertions were designed from those numbers but derived by reasoning over `Viewport::ensure_visible` / `check_nowrap_visibility`. If CI fails, the likely spots are the exact selection-cell threshold and the "drag inside the text area must not scroll" guard rows.

The neighbouring `e2e::mouse` / `e2e::selection` / `e2e::block_selection` suites were also not run, so any regression from the changed drag-below behaviour — the head now lands one line *below* the last visible line rather than on it — is unverified.

Ran and clean: `cargo fmt --all`, `cargo check --all-targets` (only two pre-existing warnings, present in the baseline), `cargo clippy --all-targets` (no new warnings in the touched files).

https://claude.ai/code/session_014RASTj1HKSpMtmL2ZpjJAN

---
_Generated by [Claude Code](https://claude.ai/code/session_014RASTj1HKSpMtmL2ZpjJAN)_

### DIFF
--- a/crates/fresh-editor/src/app/chrome/splits.rs
+++ b/crates/fresh-editor/src/app/chrome/splits.rs
@@ -1334,16 +1334,37 @@ impl Editor {
                 let gutter_width = state.margins.left_total_width() as u16;
                 let target_position = match terminal_grid_target {
                     Some(pos) => pos,
-                    None => crate::app::click_geometry::screen_to_buffer_position(
-                        col,
-                        row,
-                        content_rect,
-                        gutter_width,
-                        &cached_mappings,
-                        fallback,
-                        true, // Allow gutter clicks for drag selection
-                        compose_width,
-                    )?,
+                    None => {
+                        let target =
+                            crate::app::click_geometry::screen_to_buffer_position_with_overshoot(
+                                col,
+                                row,
+                                content_rect,
+                                gutter_width,
+                                &cached_mappings,
+                                fallback,
+                                true, // Allow gutter clicks for drag selection
+                                compose_width,
+                            )?;
+                        // Pointer outside the text area: the row→line lookup
+                        // can only name lines that are on screen, so it clamps
+                        // to the first/last visible one and the selection head
+                        // stops dead at the viewport edge. Carry the rows past
+                        // the edge into lines past the edge, so the head keeps
+                        // moving and the viewport follows it (issue #3006).
+                        //
+                        // Without this the drag only ever scrolls as a side
+                        // effect of the scroll-off margin, so a configured
+                        // `scroll_offset = 0` means dragging past the edge
+                        // does nothing at all.
+                        let rows_past_edge =
+                            target.row_overshoot as isize - target.row_undershoot as isize;
+                        crate::app::click_geometry::position_offset_by_lines(
+                            &state.buffer,
+                            target.position,
+                            rows_past_edge,
+                        )
+                    }
                 };
                 let (new_position, anchor_pos) = if drag_by_words {
                     if target_position >= anchor_position {
@@ -1396,6 +1417,21 @@ impl Editor {
 
         if let Some(event_log) = self.active_window_mut().event_logs.get_mut(&buffer_id) {
             event_log.append(event.clone());
+        }
+        // A drag is cursor motion, so it owns vertical placement from here on
+        // — exactly like a key press, which clears this same flag in
+        // `handle_key`. A wheel or scrollbar scroll sets `skip_ensure_visible`
+        // so the render pass won't yank the viewport back to the cursor, and
+        // nothing on the mouse path used to clear it again: after any scroll
+        // by wheel or scrollbar, a drag-select moved the selection head but
+        // the viewport stayed frozen, in *both* directions (issue #3006).
+        if let Some(view_state) = self
+            .windows
+            .get_mut(&self.active_window)
+            .and_then(|w| w.split_view_states_mut())
+            .and_then(|states| states.get_mut(&leaf_id))
+        {
+            view_state.viewport.clear_skip_ensure_visible();
         }
         self.active_window_mut()
             .apply_event_to_buffer(buffer_id, leaf_id, &event);
--- a/crates/fresh-editor/src/app/click_geometry.rs
+++ b/crates/fresh-editor/src/app/click_geometry.rs
@@ -86,6 +86,12 @@ pub(crate) struct ClickTarget {
     /// rendered row). Vertical virtual space uses this to place the cursor
     /// on lines below the end of the buffer.
     pub row_overshoot: usize,
+    /// Visual rows *above* the content area (0 when the click hit a
+    /// rendered row). The row→line lookup clamps to the topmost visible
+    /// line, so without this the caller cannot tell "pointer on the first
+    /// text row" from "pointer six rows up on the menu bar" — which is
+    /// exactly what a drag past the top edge needs to know (issue #3006).
+    pub row_undershoot: usize,
     /// The clicked column within the content area (viewport-relative,
     /// after the gutter).
     pub text_col: usize,
@@ -126,6 +132,10 @@ pub(crate) fn screen_to_buffer_position_with_overshoot(
     // Calculate relative position in content area
     let content_col = col.saturating_sub(content_rect.x);
     let content_row = row.saturating_sub(content_rect.y);
+    // How far the pointer is *above* the first text row. `content_row`
+    // saturates to 0 there, so this is the only record that the pointer
+    // was outside the content area at the top.
+    let row_undershoot = content_rect.y.saturating_sub(row) as usize;
 
     tracing::trace!(
         col,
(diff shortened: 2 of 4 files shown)


sinelaw:
CI failure fixed, and the branch is rebased onto current master (`7a9065b`).

**All three failures were test bugs — no production code changed.** The strict-monotonic viewport assertions, the ones that actually carry the regression, passed untouched throughout. Every panic was the same wrong expectation: that the screen row at the very edge of the text area is the row the selection reaches.

1. `drag_above_..._after_a_wheel_scroll` — the head did leave the viewport (31 → 29) and the viewport did follow (31 → 26), but with `scroll_offset = 3` the head then sits three lines *inside* the viewport, so the first content row draws `LINE 026`, above the selection and correctly unselected.
2. `drag_below_..._after_a_wheel_scroll` — the same margin mirrored, compounded by the placement change this PR makes: the head now lands one line *below* the last visible line, and the assertion was written against the old placement.
3. `drag_past_the_edges_scrolls_with_scroll_off_disabled` — the too-strict threshold. The head genuinely is on the top visible line here, but the drag anchors it at the pointer's column, so that line contributes ~5 selected cells rather than the 8 demanded.

In each case the head moved the correct number of lines past the edge (2 up — the menu and tab-bar rows above the content area; 1 down — the status-bar row below it) and the viewport followed in the right direction, so the behaviour was right and the assertions were wrong.

The replacements are stronger than what they replace: instead of counting cells on the edge row once, each motion event must push the outermost `LINE nnn` still carrying `theme.selection_bg` further out **and** move the top `LINE nnn` marker — every step, in both directions, independent of `scroll_offset`. The scroll-off-disabled test keeps the "head is on the top visible line" claim, which is only true there, as an equality against the rendered top line. The third test also gained the "a drag inside the text area must not scroll" guard the other two already had.

Verified on the rebased tree, all six `issue_3006` tests — the three here plus the three shift-select tests that arrived with the merged #3017:

```
test e2e::issue_3006_drag_beyond_text_area::drag_below_the_text_area_scrolls_down_after_a_wheel_scroll ... ok
test e2e::issue_3006_drag_beyond_text_area::drag_above_the_text_area_scrolls_up_after_a_wheel_scroll ... ok
test e2e::issue_3006_shift_select_at_buffer_edges::shift_down_on_last_line_selects_to_buffer_end ... ok
test e2e::issue_3006_shift_select_at_buffer_edges::shift_up_on_first_line_selects_to_buffer_start ... ok
test e2e::issue_3006_shift_select_at_buffer_edges::shift_up_extends_existing_selection_to_buffer_start_then_stops ... ok
test e2e::issue_3006_drag_beyond_text_area::drag_past_the_edges_scrolls_with_scroll_off_disabled ... ok

test result: ok. 6 passed; 0 failed
```

Not vacuous: with `click_geometry.rs` and `mouse_input.rs` reverted to master and the test file kept, all three fail on the viewport assertions —

```
drag step 0 above the text area must scroll up (top line was 31, now 31)
drag step 0 below the text area must scroll down (top line was 31, now 31)
with scroll_offset = 0, drag step 0 above the text area must still scroll up (top line was 26, now 26)
```

Two caveats. Only the `issue_3006` tests were run locally, so the cross-platform claim rests on the tests containing nothing platform-variant (`LINE nnn` marker parsing and `selection_bg` cell reads only) rather than on a macOS/Windows run — CI settles that. And autoscroll while the pointer is held still is still not implemented; the PR description covers the `editor_tick` + `TestTimeSource` route for it.

---

sinelaw:
Rebased onto current master (`11fce39`). This was not a mechanical rebase — worth a note for review.

Master's new commits include the chrome/mouse dispatch refactor: `app/mouse_input.rs` went from 4749 to 1091 lines and `handle_text_selection_drag` moved out to `app/chrome/splits.rs`. Git could not place this PR's hunk, so rather than resolving the conflict textually I took master's `mouse_input.rs` as-is and re-applied both halves of the fix in the function's new home:

- the overshoot/undershoot → `position_offset_by_lines` call, now against `crate::app::click_geometry::screen_to_buffer_position_with_overshoot`
- the `clear_skip_ensure_visible` call on the post-drag path

The `click_geometry.rs` half rebased cleanly — master already carries `row_overshoot`, and this PR's `row_undershoot` plus `position_offset_by_lines` applied without conflict.

Verified on the rebased tree — all six `issue_3006` tests, this PR's three plus the three shift-select tests from the merged #3017:

```
test e2e::issue_3006_shift_select_at_buffer_edges::shift_down_on_last_line_selects_to_buffer_end ... ok
test e2e::issue_3006_drag_beyond_text_area::drag_below_the_text_area_scrolls_down_after_a_wheel_scroll ... ok
test e2e::issue_3006_drag_beyond_text_area::drag_above_the_text_area_scrolls_up_after_a_wheel_scroll ... ok
test e2e::issue_3006_shift_select_at_buffer_edges::shift_up_on_first_line_selects_to_buffer_start ... ok
test e2e::issue_3006_shift_select_at_buffer_edges::shift_up_extends_existing_selection_to_buffer_start_then_stops ... ok
test e2e::issue_3006_drag_beyond_text_area::drag_past_the_edges_scrolls_with_scroll_off_disabled ... ok

test result: ok. 6 passed; 0 failed
```

Since the fix now sits inside refactored dispatch code, the port is the part worth a reviewer's eye — the behaviour is unchanged but the surrounding call path is not the one originally reviewed.

---

PR #3020, opened 2026-08-15
Composite buffers in a pane, and three web UI faults

sinelaw:
Four fixes to behaviour that predates the agent-control-plane work, split out so they can be reviewed without it. **#2962 is now stacked on this.**

## `pane` showed nothing for a composite buffer

Side-by-side diffs go through a separate render branch, which painted blank inside a panel.

## Three faults in the web UI's playwright suite

Each was hidden behind the one before it — the first killed the run, so nothing after it had executed in a long time.

**The New Workspace step could never find its menu option.** It clicks `New Task… ▾` to open the create dropdown, then the `New Task…` option. Menu options are full-width buttons, so the option's real text is padded — `"  New Task…                    "` — and Playwright does **not** trim before testing a RegExp `hasText`. The anchored pattern matched nothing and the click waited out its full 30s timeout. Checked against a real browser rather than reasoned about:

```
anchored /New Task…$/    -> 0 matches
tolerant /New Task…\s*$/ -> 1 match
```

The anchor still has to stay: without it the pattern also matches the `New Task… ▾` button that opened the menu, and `.first()` takes that one and closes the dropdown again.

**The dock's toolbar was overflowing.** The plugin sizes the search field in terminal cells against the dock's column count; the web lays the dock out in CSS pixels, and the two do not agree — measured at 457px of content in a 216px row. It wrapped onto four lines and the session tree paid for them, which is what the "tree ≥ 55% of dock height" check was failing on. The field shrinks now and the button keeps its natural width. Squeezing the button alone was not enough — it just moved the wrap inside the button and cost the same row.

**The dropdown highlight was the width of its text, not the width of the menu.** Options are built as `col(...rows)`, so they land in a `.w-col`, but both the CSS rule and the test looked for `.w-row > .w-button`. The rule matched nothing, and the test's `closest('.w-row')` returned null — reporting a failure that had nothing to do with width. Both accept either container now.

## Verification

The full web UI suite runs green against a real browser — **181 passed, 0 failed** — three consecutive times. (Running it locally needs loopback excluded from a proxy, `NO_PROXY=127.0.0.1,localhost`, or Chromium proxies the bridge and the scene never populates.)

## Scope note

The hit-area column fix and the pane click/focus fixes are also pre-existing-behaviour fixes, but they could not come with this PR: `HitArea` gains a required `column` field, which forces every construction site in `widget_runtime.rs` to change, and that file's other changes need the widget vocabulary from `api.rs`. Separating them would have meant splitting `api.rs` hunk by hunk and landing API with no consumer. They stay in #2962.

https://claude.ai/code/session_01D22YhixLB4aPupLo5S2K9P

---
_Generated by [Claude Code](https://claude.ai/code/session_01D22YhixLB4aPupLo5S2K9P)_

### DIFF
--- a/crates/fresh-editor/src/view/ui/split_rendering/mod.rs
+++ b/crates/fresh-editor/src/view/ui/split_rendering/mod.rs
@@ -253,6 +253,45 @@ impl SplitRenderer {
         )
     }
 
+    /// Render a **composite** buffer into an arbitrary screen rect.
+    ///
+    /// The composite sibling of `render_phantom_leaf`, and it has to
+    /// exist separately for the same reason the split path branches:
+    /// a composite carries no text of its own — its content is the
+    /// source buffers named by `composite.sources`, laid out into
+    /// columns. Sent through the per-leaf renderer it paints one
+    /// empty leaf, which is what a `Pane` pointed at a side-by-side
+    /// diff used to show.
+    ///
+    /// The caller owns the `CompositeViewState` (scroll and cursor
+    /// row), exactly as it owns the `SplitViewState` for a leaf.
+    #[allow(clippy::too_many_arguments)]
+    pub fn render_phantom_composite(
+        buf: &mut ratatui::buffer::Buffer,
+        area: Rect,
+        composite: &crate::model::composite_buffer::CompositeBuffer,
+        buffers: &mut std::collections::HashMap<BufferId, EditorState>,
+        theme: &crate::view::theme::Theme,
+        view_state: &mut crate::view::composite_view::CompositeViewState,
+        use_terminal_bg: bool,
+        show_tilde: bool,
+    ) {
+        orchestration::render_composite::render_composite_buffer(
+            buf,
+            area,
+            composite,
+            buffers,
+            theme,
+            // A phantom composite is never the focused split; the
+            // caret belongs to whatever owns input.
+            /* is_active */
+            false,
+            view_state,
+            use_terminal_bg,
+            show_tilde,
+        );
+    }
+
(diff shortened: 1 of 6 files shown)


sinelaw:
Closing — `master` has since fixed three of these four independently, and better.

| this PR | `master` today |
|---|---|
| `/New Task…\s*$/` for the padded option label | `/^\s*New Task…\s*$/` (`drive.mjs:405`, `:962`), anchored both sides and driven off the semantic tree rather than a CSS locator |
| `.w-row{flex-wrap:nowrap}` + `.w-text{flex:1 1 auto;min-width:0}` for the overflowing dock toolbar | present verbatim at `40-widgets.css:24` and `:58` |
| `.w-col>.w-button` added to the overlay fill rule | the selector approach was deleted outright and replaced by a `full_width` field on `Button` |

That last row is the interesting one. `full_width`'s own API docs say it replaced "padding the label with spaces out to a width the caller had to supply" — the very padding that broke the matcher in the first place. Upstream fixed the cause; this PR was patching the symptom, which is also why the fix had to be spelled as a regex tolerating trailing spaces.

The fourth change — the 39-line composite-buffer fix in `split_rendering/mod.rs` — is genuinely still absent from `master`. But its test failed CI here on all three platforms (`e2e::composite_pane::a_pane_over_a_composite_buffer_renders_both_columns`, a real assertion failure at 0.4s, not a timeout), because it was split away from the `HitArea`/`widget_runtime` work it depends on. It belongs with that work, not on its own, and that work is not being revived — so it is not worth carrying this PR to hold it.

The research this branch was supporting is preserved in #3276.

---

Issue #1528, opened 2026-04-09
Add new language syntax support for verilog and system verilog


bqinTT:
title

rkaletka:
bumping up the issue
@bqinTT any chance that you're from tiny tapeout?

rkaletka:
@sinelaw any chance I can help without being fluent in rs?

sinelaw:
also added VHDL because

rkaletka:
nice, thanks, SV support means a lot to me!

PR #3026, opened 2026-08-18
Add OCaml ecosystem, Dafny, Nginx, and SMT-LIB 2 highlighting

asukaminato0721:
## Summary

- Add built-in Coq/Rocq syntax highlighting for declarations, Gallina terms, tactics, nested comments, strings, attributes, numbers, and operators.
- Add built-in Dune syntax highlighting for stanzas, fields, actions, variables, comments, strings, and atoms.
- Detect `.ocamlinit` as OCaml; detect `.coq`, `_CoqProject`, and `_RocqProject` as Coq/Rocq; and detect standard Dune filenames and `*.dune`.
- Preserve `.v` as V by default, as documented by #1528, while automatically detecting `.v` as Coq/Rocq when an ancestor directory contains `_CoqProject` or `_RocqProject`.
- Add built-in Dafny highlighting and `.dfy` detection, covering declarations, verification clauses, nested comments, modern and legacy attributes, normal/verbatim strings, types, numbers, calls, and operators.
- Add built-in Nginx highlighting for directives, contexts, variables, strings, regular expressions, addresses, values, and comments.
- Detect `nginx.conf`, `*.nginx`, `*.nginx.conf`, and conventional files beneath an `nginx/` directory without claiming the generic `.conf` extension.
- Add built-in SMT-LIB 2 highlighting for solver commands, declarations, sorts, attributes, quoted symbols, strings, bit-vector and numeric literals, operators, and comments, including common Z3 optimization and fixedpoint commands.
- Detect `.smt2` automatically and accept `smt`, `smt2`, `smt-lib`, and `z3` as manual language aliases, without claiming the SMT-LIB 1 `.smt` extension.
- Keep detected language IDs, status-bar languages, language-specific settings, and selected highlighting grammars synchronized.

## Testing

- `git diff --check`
- `cargo test -p fresh-editor-core test_smtlib2 --lib`
- `cargo test -p fresh-editor-core test_default_languages_map_smtlib2_files --lib`
- `cargo test -p fresh-editor-core test_nginx --lib`
- `cargo test -p fresh-editor-core test_ocaml_ecosystem --lib`
- `cargo test -p fresh-editor-core test_dafny --lib`
- Interactive tmux validation of Coq/Rocq, Dune, and OCaml highlighting.


sinelaw (changes requested):

- Tested locally in tmux: Coq (`identity.v`), Dune (`dune`), and OCaml `.ocamlinit` all highlight correctly — grammars are well-formed and status bar detects them as expected.
- Regression: reassigning `.v` from `vlang` to `coq` (config.rs) silently breaks existing V-lang files. Verified interactively — a real V source file (`module main / fn main() {...}`) now gets detected+highlighted as Coq/Rocq, not V. `.vv` is not V's real extension, so this isn't a safe fallback.
- This reverses a previous, deliberate decision (CHANGELOG: "HDL language support" #1528 — "`.v` left mapped to vlang for compatibility") without acknowledging it, updating the changelog, or discussing the tradeoff in the PR description.
- Please keep `.v` → vlang and give Coq/Rocq distinct extensions/filenames only (`.coq`, `_CoqProject`, `_RocqProject` — already present), or otherwise resolve the conflict without regressing existing V-lang users.
- PR title/description ("add highlight" / "coq, ocaml related highlight") could be more descriptive of the actual scope (new Coq + Dune grammars, OCaml filename detection, vlang extension change).

---

asukaminato0721:
fix #3105

sinelaw:

The `.v` blocker is already resolved, by the author, in 3220f5542 ("Preserve V language .v detection", authored 2026-08-20 — the day after the review). Verified at the current head:

- `config.rs`: `vlang` is `["v", "vv"]`, `coq` is `["coq"]`. Matches master.
- `coq.sublime-syntax`: `file_extensions: [coq]`.
- No `.v`→Coq mapping left in `language_detect.rs`, `detected_language.rs`, the LSP manager, or any of the five new grammars.
- The `_CoqProject`/`_RocqProject` ancestor rule survives (`language_detect.rs:37-60`) and fires only when detection already returned `vlang`. That part is a good feature and was kept.

One correction to the review: `.vv` was not introduced by this PR. master already had `extensions: vec!["v", "vv"]` for vlang (config.rs:7230). The PR only ever removed `"v"`; it never added `"vv"`.

Pushed 9b572ca38 to close a real gap in coverage. `.v` staying V-lang was only asserted at the config-key level (`detect_language` → `"vlang"`); nothing asserted that a plain `.v` file resolves to the V **grammar and highlighter**, which is the layer where the regression was observed interactively. The new assertions go through `DetectedLanguage::from_path` and check `name`, `display_name`, and token categories. Confirmed non-vacuous: reintroducing either half of the regression makes the tests fail with `left: "coq", right: "vlang"`.

`cargo check -p fresh-editor-core --tests` clean, 13 relevant tests pass. The two `fresh-editor` crate tests at `manager.rs:2265` were not run locally; they exercise the same `detect_language` path and CI covers them.

sinelaw:
I built this branch at `9b572ca` and drove the actual editor in tmux against real sample files for each language, checking the status bar for detection and ANSI-preserving screen captures for the highlighting itself.

Everything detects correctly: `.ocamlinit` as OCaml, `.coq` and `_CoqProject` as Coq/Rocq, `.v` staying V outside a Coq project and switching to Coq/Rocq inside one, `dune`/`dune-project` as Dune, `.dfy` as Dafny, `nginx.conf` / `*.nginx` / files under `nginx/` as Nginx (and a plain `application.conf` correctly still INI), and `.smt2` as SMT-LIB 2. Highlighting is genuinely present and readable in all of them — nested comments in OCaml, Coq and Dafny stay nested, doubled-quote escapes work in Coq, Dafny and SMT-LIB, and nothing crashed, hung or garbled the screen, including on deliberately broken files with unclosed comments and strings, and while typing live.

No regressions: Rust, Python, JSON, Markdown, V, and shebang-based shell detection are all untouched — I checked the shebang path specifically since `detected_language.rs` now resolves the config language's grammar before falling through to `find_by_path`. `cargo fmt --check` is clean, clippy on `fresh-editor-core` reports no new warnings, and `cargo test -p fresh-editor-core --lib` is 1253 passed / 0 failed.

Two small cosmetic nits if you feel like tightening them, neither blocking:

- Nginx log-level words match inside file paths, so `error_log /var/log/nginx/error.log warn;` colours the `error` in the path as a constant. `\b` fires after `/` and before `.`, so something like `(?<[\w./-])(debug|info|…)(?[\w./-])` would fix it.
- In Dune, digits inside a flag token get read as numbers: `(flags (:standard -w +a-4-9))` lights up the `4` and the `9`.

One expectation-setting note rather than a defect: several scopes the description mentions — variables, quoted symbols, operators, attribute contents — are mapped to the default foreground by the default theme, so they render the same as plain text on screen even though the scopes are applied correctly (the unit tests assert them and pass). Same is true of existing Rust and Python, so it's nothing this PR introduced.

Nice work, this is a lot of careful grammar for one PR.

---

sinelaw:
Following up on my earlier comment — the grammars themselves test out fine. The remaining gaps are all in the one non-grammar change, the new block in `detected_language.rs`.

It fixes something real: `config_lang_id` sets the buffer's name while `find_by_path` sets the grammar, and `override_name` only stamps the name, so a `.v` in a `_CoqProject` tree ends up labelled `coq` with V-lang highlighting. Worth fixing. Three things about the shape of the fix:

1. **It changes detection precedence for every file type, not just the five new languages.** That's a wide blast radius for a PR that's otherwise additive, and it's the part that most needs review attention.

2. **`first_line` becomes unreachable whenever a config rule matches the path.** Shebang detection lives inside `find_by_path`, which the early `return` skips. The doc comment on `from_path_with_fallback` still lists "5. Shebang / first-line regex" in its priority order, so it's now inaccurate. Worth either a test for a config-matched extension whose first line points elsewhere, or updating that comment to describe the new order.

3. **It overlaps with #3018**, which reconciles the same two halves of detection in the same function. That one keeps `find_by_path` first and re-resolves afterwards, and deliberately keeps the path-resolved grammar when the config id's grammar is unknown ("an unknown config key must never downgrade working highlighting"). Whichever lands second will conflict. Might be worth deciding which approach to keep before either merges.

Minor: `.or_else(|| registry.find_by_name(id))` means a typo in a `grammar = "..."` setting silently resolves via the language id instead of surfacing.

None of this affects the grammars, which are good work. If it'd be easier, splitting the `detected_language.rs` hunk out so the five grammars can land on their own seems reasonable — happy to discuss either way.

---

PR #3028, opened 2026-08-18
Migrate the editor's chrome onto fresh-ui

sinelaw:
## Summary

Moves the editor's chrome off its hand-written painters and hit-tests and onto the
`fresh-ui` retained tree. What began as five surfaces is now nearly all of the
chrome: the frame, the menu bar and its dropdowns, context menus, the status bar,
the search-options row, the file explorer, the file browser, the prompt and its
suggestions, every popup (completion, LSP hover, signature help, code actions, LSP
status, theme info), the modals (workspace trust, calibration wizard, the settings
dialog in full, the keybinding editor), the orchestrator dock, and plugin panels —
both the floating one and the buffer-mounted ones — plus the split grid's per-leaf
seam.

The PR opened as a planning document (`docs/internal/fresh-editor-ui-migration.md`)
and grew into the migration it planned; this description replaces the original,
which described the docs alone.

## What moved

**The seam.** `view/shell/fold.rs` turns a `LayoutSpec` into ratatui cells; the
`Host` leaf hands the text pipeline back its own rectangle, so the keep/migrate
boundary the plan named is untouched. `view/shell/input.rs` is the same seam the
other way: crossterm events become `fresh_ui::Input`. The fold runs in two bands
(background before the remaining legacy painters, overlay after), which is what
let surfaces migrate in any order rather than strictly top-down.

**One tree per frame.** The editor keeps a single `Ui` for the life of the window,
so every frame after the first is a reconcile. Geometry, provenance for the theme
inspector, and the web projection all read out of that one tree instead of the
two-to-three parallel walks each surface used to keep in step by hand.

**A theme name is real theme keys.** `fg_key/bg_key` (+attrs), resolved through the
editor's existing `resolve_theme_key`. This replaced a twenty-arm match over
invented names, and it converges with the theme inspector, which already recorded
provenance as exactly that pair. A plugin's `ThemeKey` is *data*, not one of our
names, so it resolves as `Paint::Asked` over whatever was beneath it rather than
tripping the `debug_assert` that guards our own typos.

**Plugin panels keep their API exactly.** `WidgetSpec` is unchanged; the adapter in
`view/shell/widgets.rs` turns a spec into a description, and `widgets::covered`
states which kinds it covers — a panel is described or painted, never half of each.
Existing plugins go on working untouched.

**Keyboard by containment, not by rank.** `Modality::Exclusive` layers replaced the
z-banded modal capture band and the hand-listed key walk; `on_layer_key` and
`layer_rank`'s routing role are gone.

## What this is worth

Fifteen files are deleted outright — three `app/chrome` painters, the whole
`view/popup/input` tree and `popup_input`/`popup_mouse`, `view/settings/layout.rs`,
`calibration_wizard.rs`, `event_debug.rs`, `menu_input.rs`, `suggestions.rs` — and
`view/ui` is net −2577 lines while `app/chrome` is net −2471.

Deletions land where the duplication was. Several surfaces computed the same
geometry twice — once to paint, once to hit-test — with a `debug_assert` holding
the copies together; the tree makes that structurally impossible. `MenuRenderer::render`,
the second frame layout, `clamped_position`, `SettingsLayout`, `SuggestionsRenderer`
and the coordinate-driven menu click paths are all gone.

The remaining `+` is mostly two things: `view/shell` (+25675), which is the new
descriptions and the tests that pin them, and `docs/internal` (+7324), which is the
plan plus the review write-ups. `fresh-ui` itself grew by +1965/−76 — the library
needed additions (a stateful row builder, `Ink`, dismissal rules), not surgery.

## Bugs this found

Every one of these was **a rule the old code enforced that the description did not
carry**, and none of them was a type error — which is the argument for the parity
oracles rather than for reading harder. Each is pinned by a test.

- **`fresh-ui`: a root re-measure is not a whole-tree measure.** `flush_layout`
  cleared `layout_dirty` whenever it measured from the root. Path-marking stops at
  the nearest relayout boundary, so the root walk short-circuits above every
  boundary marked from below and leaves it holding the previous frame's
  measurement — and `TextRender` shapes its rows at measure time and paints from
  them. Eight e2e failures and a timeout.
- **`fresh-ui`: a change is not always a message.** A `List` writes its own
  `hovered` through an updater and produces no message, so a host that decides
  staleness from messages alone never draws the hover. Found by driving the
  completion popup; `Ui::needs_frame` is the library's own answer and is now what
  the host asks.
- **A migrated surface acts on the press.** The web frontend forwards a chrome
  click as a synthetic mouse-down and never sends the release; the handlers this
  replaced fired on `PointerPress::Left`. Two surfaces asked for `Click` and one for
  `SecondaryClick`, which the library derives from press *and* release — so every
  terminal test passed while the browser silently stopped responding.
- **A row fills its container; a row in a box does not.** A row-wide hit filler that
  is right for a panel's flow makes a pop-over's box grow to whatever it is offered.
  `Fill::{ToRowEnd, ToText}` names the difference.
- **A stamped row theme is a ground, and a row that paints its own cells covers it.**
  The dock's compact rows showed the active-session band only in the gaps between
  glyphs; so did the completion popup's hover. `List::windowed_stateful` hands the
  builder the `RowState` so the row is built *from* its state.
- **A caret marker is zero-width, and `rect_of` drops zero-size elements** — so
  every described surface's caret lookup had been answering `None`.
- **The LSP hover tracker owned a repaint it never reported.** `update_hover_target`
  used to return "the target moved, redraw" and the content trackers rode on it;
  nothing replaced that half.
- **A wheel notch is `mouse_wheel_scroll_lines`, not 1**; **the explorer's caret must
  not repaint the row's background**; **a modal drawn over the dock owns its own cells**.

## Verification

`ui_shell_frame_parity` checks the shell's rectangles against the real ratatui
layout across the frame's cases; `scene_parity` checks the web projection against
the TUI cells; `every_popup_theme_name_is_a_real_key` and its siblings guard that
no description invents a theme name.

Beyond the suites, every surface in the list above was driven by hand in a terminal
against master's behaviour — menus and submenus, the explorer and its context menu,
the command palette, settings and its search, dropdown and entry dialogs, the
keybinding editor, the trust dialog, the calibration wizard, split-divider drag, the
orchestrator dock in both densities, five mounted-panel plugins, and (with `gopls`)
the completion, hover, signature-help, code-action and LSP-status popups.

## Still to come

Tracked in the migration doc, deliberately left for follow-up branches:

- **F.6** — Ctrl+Right-click theme inspection is blank over a *described* surface:
  every writer of the per-cell theme map is a painter. The fix belongs in `fold_band`.
- **F.2** — `Paint::Lit`, the one colour in the display list with no theme name,
  retires into a dynamic theme tier once plugins can register named keys.
- **C.1's remaining half** — what the widget runtime still does for a described panel.
- **A.4** — `popup_blocked_by_higher_modal` and `get_key_context`, the last two
  readers of `layer_rank`.
- The dead painters in `view/ui` are not all deleted yet.


sinelaw:
**`playwright suite` is red, and it is red on `master` too — same test, same line.**

It times out in *New Workspace dialog is modal* waiting for the dock's create-dropdown option after clicking the "New Task… ▾" trigger:

```
locator.click: Timeout 30000ms exceeded.
  - waiting for locator('.widget-surface.w-dock .w-button').filter({ hasText: /New Task…$/ })
  at crates/fresh-editor/web-ui/test/drive.mjs:867
```

`master`'s own last Web UI run ([32067093429](https://github.com/sinelaw/fresh/actions/runs/32067093429), commit `a192420`) fails at the identical locator, at what was then line 862 of the same file. It predates this branch, and it predates the `sendClick` change here — it was already failing at `56476c4`, before that change landed.

It is also not the TUI path: the e2e test that drives the same button, `e2e::orchestrator_dock::mouse_click_on_dock_new_button_opens_form`, passes. Whatever is wrong is in the web dock's dropdown, which this branch does not touch — no fix ported, because there is no fix on `master` to port.

Everything the suite asserts about this branch's own surfaces passes, including the one that would catch a palette regression from the suggestion-list migration in these commits:

```
PASS command palette (Ctrl+P) still renders its suggestion list (no regression)
```

`scene parity (web/TUI)` is green as well, which is the check that would catch the web scene and the TUI disagreeing about the palette's geometry.

I have no means to re-run the failed check from here. Keeping the PR watched until it is green.

---

PR #3033, opened 2026-08-19
Make a selection over empty lines visible (#2797)

### LINKED ISSUE #2797: Proposal: Display a visual indicator on each line within the selected range to show that a line break has been selected
## Proposal: Display a visual indicator on each line within the selected range to show that a line break has been selected
In the current Fresh Editor, even when selecting an empty line, it is difficult to visually confirm whether the line is truly selected.
When I first used Fresh Editor, I was very confused and spent several hours searching through the settings, because I believed that the selection range had not been created. In the end, I realized that the selection had in fact been made.

Mainstream and niche editors alike, including VS Code, Obsidian, and Micro Editor, support this feature in some form. It is likely not even something that can be turned off.

For the reasons above, I believe this is a feature that should clearly be enabled by default.
That said, since some users may not prefer it, it might be better to allow the feature to be toggled on and off.

## Note
* This text was translated by ChatGPT-5.4 mini. The original text is in Japanese.
* If there are any inappropriate or impolite expressions in the text, I sincerely apologize.

sinelaw:
Fixes #2797.

Selecting empty lines left the screen untouched: on a blank line the line break is the only thing selected, and the newline cell renders as nothing — so a selection over blank lines looked exactly like no selection at all. The reporter spent hours in the settings believing the selection had never been made.

## What changed

**1. Selected line breaks paint one highlighted column.** Every line break inside a *linear* selection now highlights the single column it occupies — column 0 on an empty line, just past the text otherwise (what VS Code shows). Block-selection rects are deliberately excluded: a rect can cover the newline's column without the break being part of what would be copied, so `SelectionActiveSet` grew a `contains_linear()` alongside `contains()`.

**2. Whitespace inside a selection shows its indicators.** Spaces and tabs inside a selection draw `·` / `→` even when the per-position settings (or the `whitespace_show` master toggle) keep them hidden elsewhere, so the extent of a selection over runs of blanks is legible without turning indicators on for the whole buffer. New `editor.whitespace_in_selection` setting, **on by default**, independent of the master toggle and of the per-buffer whitespace overrides. Line-ending indicators stay out of it — the break gets the highlighted column instead of a `↵`.

The flag rides along the existing buffer-config resolution (`BufferConfig::resolve` → `BufferSettings::apply_config`), so the renderer reads it per buffer with no new render plumbing.

## Tests

`tests/e2e/issue_2797_selection_visibility.rs` — four e2e tests that drive keys and assert only on rendered cells:

* selected empty lines carry `selection_bg` at column 0, and an empty line just outside the selection does not;
* a selected line break after text highlights exactly one column past the text;
* selected spaces render `·` while unselected spaces on the same line stay blank;
* `whitespace_in_selection = false` keeps selected spaces blank.

Three of the four fail on `master` (the fourth is the negative control for the config-off path). Verified by reverting `cells.rs` alone:

```
failures:
    selected_empty_lines_show_a_highlighted_column
    selected_line_break_shows_a_highlighted_column_past_line_end
    whitespace_inside_selection_shows_indicators_by_default
test result: FAILED. 1 passed; 3 failed
```

All four pass with the change. `cargo fmt --all` and `cargo clippy --all-targets` are clean; `config-schema.json` regenerated with `./scripts/gen_schema.sh`; the settings label added to all 15 locales; docs and CHANGELOG updated.

---
_Generated by [Claude Code](https://claude.ai/code/session_014hm5DUeVUowDumVGH8oC1m)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
 
 ### Features
 
+* **A selection that spans empty lines is visible** - every selected line break highlights the column it occupies (column 0 on an empty line), and whitespace inside a selection draws its `·` / `→` indicators even when the indicators are otherwise off (`whitespace_in_selection`, on by default) (#2797, reported by @akarinotomoshibi).
 * **Installing and updating Fresh** - the preferred install method for Linux is now a single self-contained, self-updating, statically linked (musl) binary - one install that works everywhere, instead of a dozen fragile packaging channels.
 * **LSP completions can offer auto-imports** - servers like rust-analyzer now suggest unimported symbols, and accepting one inserts the `use`/import line (#2603).
 * **The scripting API can manage the whole Orchestrator dock** - create, rename, move, archive, delete and list workspaces, including over SSH, not just create them.
--- a/crates/fresh-editor/docs/visual-regression/screenshots/Comprehensive_UI_A_01_state_a.svg
+++ b/crates/fresh-editor/docs/visual-regression/screenshots/Comprehensive_UI_A_01_state_a.svg
@@ -876,7 +876,7 @@
   <text x="559" y="122" fill="#ffffff" class="terminal" style="">&quot;</text>
   <rect x="567" y="108" width="9" height="18" fill="#000000"/>
   <text x="568" y="122" fill="#ffffff" class="terminal" style="">;</text>
-  <rect x="576" y="108" width="9" height="18" fill="#000000"/>
+  <rect x="576" y="108" width="9" height="18" fill="#323c5a"/>
   <rect x="585" y="108" width="9" height="18" fill="#000000"/>
   <rect x="594" y="108" width="9" height="18" fill="#000000"/>
   <rect x="603" y="108" width="9" height="18" fill="#000000"/>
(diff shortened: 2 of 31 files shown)


sinelaw:
Drove the built binary by hand in tmux (`fresh demo.rs`, cursor to line 3, `Shift+Down` ×4 so the selection covers lines 3–6, three of them empty), and captured the pane before and after the change.

The plain-text capture shows the whitespace indicators; the highlighted line-break column is a background colour, so it only shows in the colour capture — described below each block.

**Before** (`master`) — lines 3, 5 and 6 are inside the selection and look completely untouched:

```
▾ 1 │ fn main() {
  2 │     let a = 1;
  3 │
  4 │     let b = 2;
  5 │
  6 │
  7 │     let c = 3;
```

Only line 4 carries `selection_bg`, ending flush with the `;`.

**After** — same selection:

```
▾ 1 │ fn main() {
  2 │     let a = 1;
  3 │
  4 │ ····let·b·=·2;
  5 │
  6 │
  7 │     let c = 3;
```

In colour, lines 3, 5 and 6 each carry one `selection_bg` cell at column 0 (the selected line break), and line 4's highlight now runs one column past the `;` for its own break — the ANSI capture shows `48;5;17` (the high-contrast theme's `selection_bg`) for exactly one cell on each blank row:

```
 3 │ ESC[48;5;17m ESC[49m                    ← one highlighted column, then normal bg
 4 │ ESC[48;5;17m····let·b·=·2; ESC[49m      ← text + one column for the break
```

The `·` in line 4 are the in-selection whitespace indicators: `whitespace_spaces_*` are all off in this run (defaults), and the unselected indentation on lines 2 and 7 stays blank — only the selected run shows dots.

The screenshots themselves are not committed to the repo, per the working agreement on this branch.

---

PR #3036, opened 2026-08-19
fix: count UTF-8 chars not bytes for keybord shortcuts

georglauterbach:
This enables multi-byte UTF-8 keys (e.g., for German) to be recognized.


sinelaw (approved):

- Root cause confirmed: `parse_key` used `s.len() == 1` (byte count), so any 2+ byte UTF-8 char (ü/ö/ä, etc.) was silently rejected — matches CONTRIBUTING's "avoid raw byte-offset string ops" guideline
- Fix is minimal, correct, and reproduced-before-fixed per the new `chars().count() == 1` check and added tests (`test_parse_key`, `test_umlaut_key_from_config_resolves`); ran full `input::keybindings` suite: 49/49 pass
- No duplicate/existing helper for this; no architectural impact — single-file, single-purpose fix
- Manually validated end-to-end in tmux: built the branch, configured `Ctrl+ü` → `duplicate_line` in `config.json`, sent the real terminal key-press (kitty CSI-u escape sequence) into a running `fresh` instance — the line duplicated as expected, confirming the fix works for actual German keybindings, not just in unit tests
- Minor nit (non-blocking): PR title has a typo, "keybort" → "keyboard"

---

georglauterbach:
@sinelaw Can this be merged?

sinelaw:
Thanks!
Part of this was included in the massive refactor of version 0.5.0, and the rest is now merged here

PR #3060, opened 2026-08-21
fix(input): don't let a plugin mode's single-key fallthrough steal in-flight chords

dbactual:
## Problem

While a buffer-local mode is active (e.g. `markdown-source`, which auto-activates on `.md` buffers), every keymap chord whose **second key** resolves to a ui-fallthrough action is swallowed by the mode stage.

Under the emacs keymap in a markdown buffer:

- `C-x C-c` → status bar shows "Copied line" (the pending `C-x` was abandoned and `C-c` ran `Copy` with no selection)
- `C-x C-f`, `C-x C-s`, `C-x C-w`, `C-x 2`, `C-x 3` — all dead

Chords with a plain second key (`C-x b`, `C-x k`) work, which made this look like a random subset of broken bindings.

## Cause

`mode_key_disposition` runs before Normal-context chord resolution. For the second key of an in-flight chord:

1. mode-context chord lookup → `NoMatch` (modes define no chords)
2. mode **single-key** resolution in `Mode(...)` context → `allows_ui_fallthrough()` is true for `Mode(_)`, so `C-c` resolves to `Copy` through the whitelist → `Run(Copy)`

The pending chord is cleared and the whitelisted action runs instead.

## Fix

When `chord_state` is non-empty, resolve the in-flight chord against the `Normal` context before the mode's single-key stage; only a `NoMatch` lets that stage run. Mode bindings still take precedence for fresh keystrokes, and mode chords still take precedence for their own sequences — the added lookup only handles the chord already in flight.

## Testing

- Verified manually: `C-x C-c` / `C-x C-f` / `C-x 2` now work in markdown buffers; `Enter`/`Tab` still run the markdown-source mode handlers.
- A clean-room resolution test (chord `C-x C-f` → `Action::Open`, `C-x C-c` → `Action::Quit` against the real emacs keymap) passes both before and after — the bug is in the mode-stage ordering, not the resolver, which is why the existing resolver tests didn't catch it.


sinelaw (changes requested):

- Interactively verified the fix (tmux, emacs keymap, `.md` buffer w/ markdown-source mode active): `C-x C-c` now quits correctly (previously fell through to `Copy`, matching the reported bug), `C-x C-f` opens the file picker, and `C-x b` (already-working plain-second-key chord) still works — no regression observed.
- The fix itself is minimal, well-commented, and mirrors the existing mode-chord-resolution pattern directly above it — no complexity/architecture concerns.
- Requesting changes on test coverage only: CONTRIBUTING.md's Testing guideline #1 ("Reproduce Before Claiming") requires a test that fails without the fix and passes with it. This PR touches only `router.rs`, but `mode_key_disposition` is pure and already has unit tests in this same file (e.g. `read_only_global_mode_blocks_unbound_keys`). Please add one covering the in-flight-chord-vs-mode-fallthrough case this PR fixes.

---

sinelaw (commented):

Reproduced against the installed 0.5.1 binary, confirming the report independently.

emacs keymap, plugins on, 80x24.

- `doc.md` (markdown-source mode active), C-x C-c: editor did not quit, pid unchanged, status bar showed "Cop..." (Copied line).
- `notes.txt`, same keymap, same C-x C-c: editor quit, pane left the alternate screen.

Same chord, same keymap, differing only by buffer type. Matches the described cause: the mode stage consumes the second key of an in-flight chord.

sinelaw:

Status: this PR is ready. The unit test you asked for is now on the branch as 873bb3d1d — it fails without the fix and passes with it.

sinelaw:
Thanks!

PR #3068, opened 2026-08-22
fix(review diff): pin git's patch format so local diff config can't blank the review (#3066)

### LINKED ISSUE #3066: maybe support diffview
https://github.com/sindrets/diffview.nvim

asukaminato0721:
## Summary

**Motivation.** Review Diff shells out to `git diff` and parses the unified patch it gets back, then rebuilds `a/`…`b/` patches to feed `git apply`. Each of those invocations inherited whatever the user's git config does to diff output, so a perfectly dirty working tree could render as **"No changes"**. Four settings cause this independently:

| setting | effect on the parser |
|---|---|
| `diff.external` | output replaced by the tool's, nothing parseable |
| textconv filter | both sides converted, so hunk line numbers no longer address the real file — `git apply` then gets a patch that cannot land |
| `color.diff = always` | every line wrapped in ANSI escapes |
| `diff.noprefix` / `diff.mnemonicPrefix` | strips or renames the `a/`/`b/` prefixes the parser matches on |

`--no-ext-diff` alone only covered the first row.

**Goal.** Pin the patch format on every diff the reviewer parses, in one place rather than seven.

**Contents.**
- A single `diffArgs()` helper, with all 7 call sites routed through it: working tree (staged / unstaged / untracked), range, stash, and both side-by-side paths.
- The flags are spliced in *after* the `range.command ||` fallback so command overrides inherit them — the stash path previously needed a hand-patched argv for exactly this reason.
- Prefixes are pinned with `-c diff.srcPrefix=` / `diff.dstPrefix=` plus `noprefix=false` / `mnemonicPrefix=false`, **not** `--src-prefix=` / `--dst-prefix=`. `git stash show` garbles the latter two — it emits a truncated prefix, and which side is mangled depends on argument order — so passing them there would have broken stash review instead of protecting it. The `-c` form also predates `--default-prefix`, so it keeps working on older git.
- Regression tests configure a hostile git (external diff + textconv + forced colour + prefix-less headers) and cover both the range and the working-tree path.

---

fix #3066 

same as https://github.com/sinelaw/fresh/pull/2791

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -236,7 +236,7 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
   * **A remote workspace's buffers load off the editor loop.**
 * **Opening a C# file can no longer run project code** - `dotnet restore` is now gated behind Workspace Trust (#2063, reported by @z3moo).
 * **git-gutter works with an external difftool** - it now forces git's own diff format (#2721, by @asukaminato0721).
-* **Review Diff works with an external difftool** - range, working-tree, stash, and single-file reviews now force Git's native patch format instead of mistaking tools such as `difft` for an empty diff (#3066, by @asukaminato0721).
+* **Review Diff works with an external difftool** - range, working-tree, and stash reviews now force Git's native patch format instead of mistaking tools such as `difft` for an empty diff (#3066, by @asukaminato0721).
 * **No blank frame at the horizontal scroll bound** (by @ttenneb).
 * **Modal dialogs are no longer clipped**, and worktree trust isn't re-asked.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -236,6 +236,7 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
   * **A remote workspace's buffers load off the editor loop.**
 * **Opening a C# file can no longer run project code** - `dotnet restore` is now gated behind Workspace Trust (#2063, reported by @z3moo).
 * **git-gutter works with an external difftool** - it now forces git's own diff format (#2721, by @asukaminato0721).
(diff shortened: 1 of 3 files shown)


sinelaw:
Manually verified in a real terminal against a genuinely hostile git config, with the released 0.4.10 binary as an A/B baseline.

Test repo configured with **all four** hazards at once:

```
diff.external       = <script printing "EXTERNAL DIFF TOOL OUTPUT">
diff.collapse.textconv = <script collapsing every .rs to a constant>
color.diff          = always
diff.noprefix       = true
diff.mnemonicPrefix = true
```

with one modified tracked file. Then **Review Diff** in both builds:

| | Released 0.4.10 | This branch |
|---|---|---|
| Header | `Review Diff — 3 files, +0 / -0` | `Review Diff — 5 files, +2 / -0` |
| `src/main.rs` | `+0 / -0` → `(no diff available)` | `+1 / -0`, real hunk, `+// MODIFIED-LINE-3068` |

So the "No changes on a dirty tree" symptom reproduces exactly on the baseline and is gone here.

I also independently reproduced the `--src-prefix` hazard this PR deliberately avoids, on git 2.55.0:

```
git stash show -p --src-prefix=a/ --dst-prefix=b/   ->  diff --git Vf.txt f.txt
git stash show --src-prefix=a/ --dst-prefix=b/ -p   ->  diff --git f.txt Uf.txt
git -c diff.srcPrefix=a/ -c diff.dstPrefix=b/ ...   ->  diff --git a/f.txt b/f.txt   ✓
```

The garbled prefix character even differs between runs, which points at uninitialised memory in `builtin/stash.c` rather than defined behaviour. Using the `-c` config form instead of the command-line flags is the right call — the flag form would have broken stash review while fixing the others.

sinelaw:
:fire:

PR #3070, opened 2026-08-22
fix(install): drop the GitHub API; anonymous access is now the only requirement

sinelaw:
## The report

A user running the documented install line got this:

```
[INFO] Installing the universal build (x86_64-unknown-linux-musl)...
curl: (22) The requested URL returned error: 403
[INFO] Downloading [ERROR] could not fetch the latest release metadata....
curl: (3) bad range in URL position 3:
[ERROR] could not fetch the latest release metadata.
  ^
[ERROR] download failed: [ERROR] could not fetch the latest release metadata.
```

## What went wrong

`install.sh` called the GitHub API to find out what to download. Unauthenticated, that API is metered at **60 requests per hour per source IP**, a budget shared by everyone behind a NAT, a CI runner, or an office network — so it answers `403`.

It then failed in a way that hid the cause. `log_error` wrote to **stdout**, and the lookup was reached through `$(...)`:

```sh
URL=$(asset_url_exact "$ASSET")   # -> fetch_release_json -> log_error
if [ -z "$URL" ]; then ... fi     # never fires: URL holds the error message
```

`exit` inside a command substitution ends only the subshell, so the error text was captured **as the URL**, the emptiness guard passed, and curl was asked to fetch `\033[0;31m[ERROR]…` — reading `[0;31m` as a glob range and reporting `bad range in URL position 3`, which names curl's parser and nothing actionable.

## Fix 1 — assets come from the public download endpoints

`/releases/latest/download/NAME` serves the current release's asset of that name, and its `.sha256` sidecar the same way, for anonymous requests. The version embedded in the `.deb`/`.rpm`/`.AppImage` filenames comes from the 302 `/releases/latest` already returns (`Location: …/releases/tag/v0.4.10`).

That covers everything the API was doing, so `fetch_release_json`, `asset_url_exact`, `asset_url_match`, the JSON scraping, and the `GITHUB_TOKEN`/`GH_TOKEN` handling are removed. Anonymous access to `github.com` is the whole requirement.

This also fixes a latent mis-selection: naming the AppImage exactly avoids the `-gui-` build, which shares the suffix and arch token and was excluded only by sort order.

## Fix 2 — the published one-liner could execute a 404 page

Testing this the way a user actually installs it turned up a second failure, upstream of the script. The command we publish used bare `curl` or `curl -sL`; both let GitHub's error page be piped into `sh` **and executed**.

Measured against a real 404 (`… | sh`, stderr to the terminal):

| command | bytes piped to `sh` | what the user sees |
|---|---|---|
| `curl URL \| sh` | 14 | progress-meter noise, then `sh: 1: 404:: not found` |
| `curl -sL URL \| sh` | 14 | `sh: 1: 404:: not found` |
| `curl -fsSL URL \| sh` | **0** | `curl: (22) The requested URL returned error: 404` |

Why each flag is there:

- **`-f`** is the one that matters. On HTTP ≥400 curl emits **no body**, so nothing bogus reaches the shell — that is the 14 → 0 in the table. Without it, a 404 page, a captive-portal login page, or a proxy error page is whatever `sh` runs next.
- **`-S`** restores the error message that `-s` suppresses. This is why `-sL` was the worst of the three: silent *and* uninformative. Bare `curl` at least printed `curl: (22) … 403` — the only reason the original report showed a diagnosable error at all.
- **`-s`** drops the progress meter, which renders as several partial lines when stdout is not a terminal (visible in the first row).
- **`-L`** follows redirects, so a redirected fetch delivers the script rather than an empty body.

Applied everywhere the command is published: `README.md` (×4), `homepage/index.html` (×3 — both visible `<code>` blocks **and** the `INSTALL_COMMANDS.linux` string behind the copy-to-clipboard button, which is what most people actually paste), and the script's own usage header.

It is the same fault as the original bug — a failure carried forward as if it were data — sitting in the line we ask people to paste.

## Fix 3 — honest failure reporting

- Diagnostics go to **stderr**. Helpers return values by echoing them, so a message on stdout was indistinguishable from the value.
- `download_verified` reads the **HTTP status** via `-w`, which separates a rate-limited download from a missing asset — GitHub meters anonymous downloads as well as the API.
- It accepts only `http(s)` URLs.
- A failed fetch returns to the caller; bytes that fail verification are fatal — retrying those elsewhere would only launder the problem.

## Smaller, and plainer to read

| | master | this PR |
|---|---|---|
| code lines (comments/blanks excluded) | 528 | **512** |
| hosts contacted on its own paths | `api.github.com`, `github.com` | **`github.com`** |
| credentials read | `GITHUB_TOKEN`, `GH_TOKEN` | **none** |

Two rules were applied to the comments.

**Only what the code cannot say.** Cut: the URL-shape guard, the `-w` flag, the version-free asset name — each restated the line beneath it. Kept: that GitHub meters anonymous downloads, that the `.sha256` sidecar follows the same URL rule, what `/releases/latest` redirects to and what "latest" resolves to, and which methods use another registry or root.

**State what is there, not what is absent.** An absence cannot be checked against the code and goes stale the moment the code grows the thing it promised to avoid. So "no GitHub API, nothing needs a credential" became "assets come from the public download endpoints, which serve anonymous requests; the API meters those at 60 per hour per source IP"; "not `-f`, which collapses every HTTP error" became "`-w` reports the HTTP status, which separates a rate-limited download from a missing asset"; and so on.

Structure that was not earning its place also went: `release_asset_url()` was a function whose whole body was one `printf` (now a `RELEASE_DOWNLOAD` constant, dropping four command substitutions), and the download-status `case` had `return 1` repeated four times (now one flat `case`). An intermediate version used `${_dl#2}` to test for 2xx — it works, but it is the kind of shell golf this file should not contain, and it was reverted.

Since this file is audited by strangers about to pipe it into a shell, the header now answers their question directly:

```
#   downloads   github.com/sinelaw/fresh/releases, over anonymous HTTPS
#   verifies    the .sha256 published beside each asset; a mismatch aborts
#   writes      ~/.local/share/fresh-editor and a `fresh` symlink in
#               ~/.local/bin, plus a .desktop entry and icons under
#               ~/.local/share (skip those with --no-desktop-integration)
#   records     every path written outside its own directory, in
#               installed-files.txt, which a later run reads to clean up
```

Every claim there was checked against an instrumented run rather than asserted. A first draft said "nothing else is contacted" and "root only for deb/rpm/aur"; both were wrong — `npm` can escalate, and `brew`/`nix`/`cargo`/`aur` fetch from their own registries — and the text now records what each method actually does.

## Tests

`scripts/test-install-failure-modes.sh` shims `curl` and reduces `PATH`, so it runs with no network. It pins two properties:

1. **Anonymity** — every method installs without contacting the API or sending a credential. A token is *planted in the environment* for every run, so a regression that starts using one is caught rather than silently taken.
2. **Honest failure** — every failure path yields a real URL or a clear diagnosis, and 403 / 404 / offline are reported distinctly.

All 19 checks pass here; the suite fails on `master`. Assertions are gated on an `install.sh ran` check first, because most are negative and would otherwise pass vacuously — an earlier draft did exactly that when `timeout` was missing from the reduced `PATH`.

A new `installer` CI job runs it and lints `install.sh`; the existing shellcheck step covers workflow `run:` blocks only.

## End-user verification

Run as an unprivileged `testuser` on a clean home, piped into `sh`, against the exact current bytes:

| | result |
|---|---|
| `curl -fsSL … \| sh` | musl universal build; receipt confirms `x86_64-unknown-linux-musl`, `static-pie` linked, `fresh 0.4.10` |
| `… \| sh -s -- --method=appimage` | version via the redirect, picks `fresh-editor-0.4.10-x86_64.AppImage` (not the gui build), editor runs |
| `… \| sh -s -- --method=deb` | constructs `fresh-editor_0.4.10-1_amd64.deb`, downloads, checksum verified (dpkg stubbed) |
| open a file in the TUI | alt-screen, terminal title set to the file, renders |
| re-run over an existing install | upgrades in place, manifest still 10 paths, no duplicated icons |

One check had to be redone: the first pass fetched from `raw.githubusercontent.com` and the md5 did not match the working tree — the CDN was serving an older revision, so it had validated the previous commit. Re-run over a local HTTP server against the exact bytes.

Also confirmed: web and API agree on which release is "latest" (`v0.4.10`), and both resolve to the newest full release — so which release users are offered does not change, which `rehearse-self-update.sh` depends on.

`shellcheck -S warning`, `sh -n`, `dash -n`, `actionlint` all clean.

**Not covered here:** macOS/Homebrew, `aarch64`, and `fresh --cmd update` — no way to exercise those in this environment.

https://claude.ai/code/session_01Uved5vGDdEH9YfmEkssau9


sinelaw:
Re-tested the way a user actually installs — `curl … | sh`, unprivileged account, clean home, fetching this branch from `raw.githubusercontent.com` rather than running the working tree. That turned up a failure in the **documented command itself**, upstream of anything the script does.

`curl -sL URL | sh` hides curl's own errors, so when the fetch fails GitHub's `404: Not Found` body is piped into `sh` and *executed*:

```
$ curl -sL https://raw.githubusercontent.com/sinelaw/fresh/refs/heads/no-such-branch/scripts/install.sh | sh
sh: 1: 404:: not found

$ curl -fsSL https://raw.githubusercontent.com/sinelaw/fresh/refs/heads/no-such-branch/scripts/install.sh | sh
curl: (22) The requested URL returned error: 404
```

Same fault this PR already fixes inside the script — a failure carried forward as if it were data — sitting in the line we ask people to paste. Now `-fsSL` everywhere it is published: README, homepage (including the copy-to-clipboard command), and the script's usage header. As a bonus it drops the progress meter from the fetch, which rendered as several partial lines when stdout is not a terminal.

**What the end-user pass covered** (all as `testuser`, not root, against the branch URL):

| | result |
|---|---|
| `curl -fsSL … \| sh` on a clean home | installs, `fresh 0.4.10` |
| `curl -fsSL … \| sh -s -- --method=appimage` | installs — exercises `latest_version()` and the constructed name for real |
| open a file in the TUI | alt-screen, terminal title set to the file, renders |
| re-run over an existing install | upgrades in place, manifest still 10 paths, no duplicated icons |
| binary | `static-pie` linked, as advertised |

The appimage run matters most: it is the first real-network exercise of the redirect-based version discovery that replaced the API, through the documented pipe-with-arguments form.

---

sinelaw:
Simplification pass, applying the rule that a comment should only carry what the code cannot.

**Comments cut** (each restated the line beneath it): the URL-shape guard, the `-w` flag, and the version-free asset name. **Kept**, because reading the code cannot tell you any of it: that GitHub rate limits anonymous downloads, that the `.sha256` sidecar follows the same URL rule, what `/releases/latest` redirects to and what "latest" excludes, and which methods can ask for root or pull from another registry. Also dropped a `root never used` line from the header that the paragraph four lines above already said.

**Structure that was not earning its place:**

- `release_asset_url()` was a function whose entire body was one `printf`. A `RELEASE_DOWNLOAD` constant reads more plainly and removes four command substitutions.
- The download status was one `case` arm per status with `return 1` repeated four times → one flat `case` with the returns inline, six lines shorter.

An intermediate version used `${_dl#2}` to test for 2xx. It works, but it is exactly the kind of shell golf this file should not contain, so it went back to a plain `2??)` arm.

| | master | now |
|---|---|---|
| code lines | 528 | **512** |
| comment lines | 126 | 158 |

**Re-verified on the simplified code, not the previous commit.** Worth noting because the first attempt at this check was invalid: `raw.githubusercontent.com` served a cached older revision, and the md5 did not match the working tree. Re-run against the exact current bytes over a local HTTP server, as `testuser`, piped into `sh`:

- `curl -fsSL … | sh` → musl universal build, receipt confirms `x86_64-unknown-linux-musl`, `static-pie` linked, `fresh 0.4.10`
- `… | sh -s -- --method=appimage` → resolves the version through the redirect, picks `fresh-editor-0.4.10-x86_64.AppImage` (not the gui build), checksum verified, editor runs
- `… | sh -s -- --method=deb` → constructs `fresh-editor_0.4.10-1_amd64.deb`, downloads, checksum verified (dpkg stubbed, so the install step itself was not completed in that run)

Full suite green: 19 failure-mode checks, `shellcheck -S warning`, `dash -n`, `actionlint`.

---

sinelaw:
Added a real end-to-end install test on clean distro images — `scripts/test-install-e2e.sh` plus a paths-filtered `install e2e` workflow.

Everything so far tested the installer on *this* machine: same distro, same shell, an account I made. The failure-modes suite shims curl and asserts on messages, which cannot tell you whether the install works, because nothing is ever installed. This runs the other half — a clean image, an unprivileged user, the published one-liner piped into `sh`, a real download from GitHub, and the installed editor executed at the end.

| image | result |
|---|---|
| `ubuntu:24.04` | ok — `fresh 0.4.10` |
| `debian:12` | ok — `fresh 0.4.10` |
| `fedora:41` | ok — `fresh 0.4.10` |
| `alpine:3.20` | ok — `fresh 0.4.10` |
| `archlinux:base` | ok — `fresh 0.4.10` |

Each also asserts the install directory ends up owned by the unprivileged user rather than root, since "needs no root" is a claim worth testing rather than assuming.

**Alpine is the case worth having.** `/bin/sh` there is BusyBox ash, not bash — so it is the real check that this script is POSIX. `dash -n` only parses; it never runs anything. Alpine also carries no glibc, so it doubles as the check that the musl build runs where it is meant to.

Two details that mattered:

- **The script under test is served from the working tree over HTTP, not pulled from a branch.** A CDN will serve an older revision of a just-pushed file — that happened earlier in this PR and quietly validated the previous commit. Serving locally makes "what ran" equal "what you edited".
- Behind a proxy, `E2E_PROXY`/`E2E_CA_BUNDLE` pass one through with the host rewritten to the docker bridge, because the container's loopback is not the host's. CI needs neither, and the install reaches GitHub anonymously there — which is itself a check on this PR's central claim.

One harness gap found and fixed along the way: `fedora:41` ships no `su` or `useradd`, so the prep installs `shadow-utils util-linux`. That was my test being wrong, not the installer.

`shellcheck` and `actionlint` clean.

---

PR #3071, opened 2026-08-23
fix(explorer): stop the scrollbar and caret stealing row hit targets

### LINKED ISSUE #2859: scroll bar in explorer is not visible
Image

asukaminato0721:
## Summary

The scrollbar itself now comes from master (#3161 and the `viewport` widget), so this PR is no longer about drawing or dragging it. What is left are two hit-testing bugs the scrollbar exposed: the bar sits in the panel's last column, and two things in the explorer answer the mouse in places they should not.

## Bug 1 — the scrollbar reports the file hiding behind it

**Before.** Point at the scrollbar and a tooltip appears for a file, saying it has unstaged changes. Nothing about the bar should say that. It happens when a file's name is just long enough to fill the row: its git-status marker is pushed out of the visible area and ends up underneath the bar, still answering the mouse.

**After.** Pointing at the scrollbar shows nothing. The status marker of a normal row still works.

**Code.** The explorer drew its rows into the content lane but did not clip them to it, so a row's hit targets could extend past the lane and under the bar, even where nothing was drawn. The rows are now clipped for hit-testing as well as for ink.

## Bug 2 — the selected row cannot show its own tooltip

**Before.** Point at the git-status marker of any row and a tooltip appears — except on the row you have selected, where nothing happens. Tooltips work everywhere but where the cursor is.

**After.** The selected row behaves like every other row.

**Code.** The selection caret `▌` is painted as an overlay on top of the selected row. The overlay is only meant to be ink, but it was taking the mouse as well, so it covered the whole row and swallowed the marker's tooltip underneath. The overlay is now pointer-transparent.

## Verified

Both were reproduced by running two builds of this branch side by side, identical except for `view/shell/file_explorer.rs`, and checked by hovering the real UI.

Bug 1 is narrow: it needs a modified file whose name lands on exactly the boundary width. In a sweep of name lengths from 8 to 26 characters, only 19 triggered it. Shorter names keep the marker inside the lane; longer ones push it out entirely.

fix #2859


asukaminato0721:
image

sinelaw:
Manually verified in a real terminal, with the released 0.4.10 binary as an A/B baseline.

Scratch project with 80 entries in an explorer too short to show them all. I sampled the actual cell background colours in the scrollbar column rather than eyeballing it, since the bar is drawn as a styled blank rather than a glyph.

**This branch**

| State | Scrollbar column |
|---|---|
| unscrolled | thumb rows 2–17, track rows 18–37 — thumb pinned at the top |
| scrolled down to `file_65` | thumb rows 16–31, track above and below — thumb tracks the viewport |

**Released 0.4.10 baseline**: zero scrollbar cells painted, at any scroll position. The bar simply isn't there, which is the reported bug.

One incidental observation from the A/B: the row's trailing status marker sits one column further left on this branch (`●  │`) than on the baseline (`● │`). That's the `content_width` change, and it's the only user-visible effect of that particular hunk — worth knowing that it's a deliberate reservation rather than a drift.

Note also that `test_file_explorer_...`-style waits on the unsaved marker need the project root canonicalised, or they hang on macOS where the temp dir is reached through a symlink; that bit me in CI here and is fixed in `8cfcaec80`.

sinelaw (approved):

- Reviewed against CONTRIBUTING.md: reuses the existing `ScrollbarState`/`ScrollbarMouse` abstraction (already used by other panels), no `std::fs`/`std::process::Command` in editor code, tests read rendered output only (no model/view accessors)
- No duplicate feature: confirmed the File Explorer previously drew no scrollbar at all (issue #2859); this is a genuine gap fill, not a near-duplicate of existing UI
- Manually tested the built binary in tmux: with 60 files in an explorer too short to fit them, the scrollbar renders and its thumb correctly tracks scroll position from top to bottom; collapsing the tree to fit makes the scrollbar disappear as expected
- Diff is clean and appropriately scoped; commit messages are mostly well-described, with a couple of early terse ones ("fix", "support drag") — minor, non-blocking
- Approving

---

sinelaw:
Pushed b9585c941 with review follow-ups.

The overflow test only proved something if the row was due a status marker, and nothing checked that — it now asserts the buffer is unsaved and the tree shows it, and I confirmed the test still fails with `clip(true)` removed. The short-name hover test was never the clip guard (its slot never reached the bar either way), so it is renamed to say what it does check. Added back a case master had and the rewrite dropped: a tree that fits from the first frame, which takes a different path than one that just stopped overflowing. The unit test now also asserts the bar column answers nothing. `glyph_col` replaces `chars().position()`, which counted symbols and drifted left of the real column once a wide glyph sat to its left.

PR #3076, opened 2026-08-25
fix(settings): honor forward-Delete in entry-dialog Text fields (#2875)

### LINKED ISSUE #2875: input box can use backspace, but cannot use "delete" key
Image

56steve:
## What

Fixes #2875 — in the settings **Edit Item** dialog, a Text field accepted **Backspace** but the **Delete** key did nothing.

## Root cause

The entry dialog's text-editing handler (`crates/fresh-editor/src/view/settings/input.rs`) routed `Delete` to `TextList` row-removal for every control that wasn't a JSON editor:

```rust
KeyCode::Delete => {
    if is_editing_json { dialog.delete(); }   // forward-delete, JSON only
    else { dialog.delete_list_item(); }       // remove a TextList row
}
```

A scalar `Text` field has no list rows, so that branch was a silent no-op — the Delete key looked dead. Backspace worked because it goes through the per-control dispatch (`dialog.backspace()`), which does handle `Text`.

Note the same class of bug in the **main settings body** was already fixed (see `test_body_text_field_home_end_and_forward_delete`); the entry dialog was missed.

## Fix

Route `Delete` the way the main-body handler and the dialog's navigation-mode handler already do: forward-delete the character at the caret for caret fields (`Text`/`Json`), and remove the focused row only for a composite `TextList`.

## Tests (both fail without the fix, pass with it)

- **Unit** `test_entry_dialog_text_field_forward_delete` — drives `KeyCode::Delete` through the dialog's key dispatch and asserts the field forward-deletes.
- **E2E** `test_entry_dialog_text_field_delete_key_issue_2875` — drives the real editor into the python-LSP **Edit Item** dialog (the dialog in the issue screenshot), types into `Command`, presses Left+Delete, and asserts on the **rendered screen**:
  - without fix: `[pylspZZZq` (the `q` survives)
  - with fix: `[pylspZZZ` (the `q` at the caret is deleted)

## Checks

`cargo fmt` ✓ · `cargo clippy -p fresh-editor --all-targets` ✓ (0 errors) · `cargo check --all-targets` ✓ · settings suite 115 passed.


sinelaw:
Manually verified in a real terminal, with the released 0.4.10 binary as an A/B baseline.

Both builds driven through the identical path: Settings → `/lsp` → Enter → Edit Value → `[+] Add new` → **Add Item**, whose focused `Command:` field is the scalar Text control from #2875.

| Keystrokes | Released 0.4.10 | This branch |
|---|---|---|
| type `abc`, `Home`, `Delete` | `abc` — **key is dead** | `bc` ✓ |
| `Home`, `Shift+End`, type `X` | `abcX` — no selection, `X` appended | `X` ✓ — selection replaced |

The first row is exactly the reported bug and it's fixed. The second is one of the chords this PR grants for free by moving onto the shared key table.

I also hand-checked the other newly-granted chords on this branch, since they're claimed in the commit message and the CHANGELOG:

- `Ctrl+Delete` — `foo bar` → `Home` → `Ctrl+Delete` → `[ bar` ✓ (word deleted forward)
- `Ctrl+Right` — `alpha beta` → `Home` → `Ctrl+Right` → type `|` → `alpha| beta` ✓ (word motion)

All four behave correctly. Worth noting for the record that `Ctrl+Delete`, `Ctrl+Right` and `Shift+End` currently have no automated test that drives *this dialog* — they're exercised only against a raw `TextEdit` in the shared `text_key.rs` suite, which doesn't prove the dialog reaches the table. That's the exact seam #2875 was about, so it's worth a few extra lines in `test_entry_dialog_text_field_word_and_selection_keys`.

sinelaw:
Thanks!

Issue #779, opened 2026-01-15
UX enhencement: displaying lines after EOF


hadmir:
It would look much better if lines after EOF are in slightly different shade of background color, instead of being marked with ~ character.

PR #3082, opened 2026-08-26
fix(theme): keep the space below the last line on the theme background

sinelaw:
## The reported problem

The area between the last line of a short file and the status bar renders as a grey band that matches no colour the active theme defines.

Those rows are painted with the theme colour `editor.after_eof_bg`. No built-in theme actually *names* that colour, so it was derived: `editor.bg` shifted ten steps toward contrast (added in #1673 for #779, "lines after EOF should be a slightly different shade instead of `~`"). On the `dark` theme that puts a `#282828` strip under a `#1e1e1e` buffer, covering everything from the last line down to the status bar.

## Reproduced

Installed from `scripts/install.sh`, opened a 7-line file in tmux, and read the real cell colours out of `tmux capture-pane -e`:

```
content row : ESC[48;2;30;30;30m   <- editor.bg    #1e1e1e
post-EOF row: ESC[48;2;40;40;40m   <- after_eof_bg #282828
```

## The change

When a theme does not name `after_eof_bg`, those rows now follow the background the *content* rows are actually painted on. `Theme::post_eof_bg(effective_editor_bg)` makes that call at the two paint sites — `fill_eof_rows` and `render_past_end_row` — so it holds for `editor.use_terminal_bg` too, where the effective background is `Color::Reset` and an opaque band would otherwise sit under the last line on a transparent terminal. Measured on a patched build:

```
use_terminal_bg=false   content ESC[48;2;30;30;30m   post-EOF ESC[48;2;30;30;30m
use_terminal_bg=true    content ESC[49m              post-EOF ESC[49m
```

and across themes with `use_terminal_bg` off — `dark` `30,30,30`, `light` `255,255,255`, `dracula` `40,42,54` — each post-EOF row identical to its own content rows.

#779 is not lost: a theme that names a colour distinct from `editor.bg` keeps it, an explicit choice being different from a fallback. The `~` markers (`editor.show_tilde`, on by default) still mark where the buffer ends either way.

## Tests

Two tests that fail on the mutation they exist to catch, rather than only passing on the fix:

| mutation | failing test |
| --- | --- |
| `post_eof_bg` returns `self.after_eof_bg` unconditionally | `post_eof_rows_follow_use_terminal_bg` |
| `fill_eof_rows` paints `effective_editor_bg`, ignoring the theme | `distinct_after_eof_bg_paints_only_the_post_eof_rows` |

The second installs a themes-dir theme with `bg` `[30,30,30]` and `after_eof_bg` `[90,20,20]` and asserts the content row keeps `bg` while the post-EOF row takes the shade. Both sample a row away from the cursor's own line, which carries `current_line_bg` rather than `editor.bg`.

## Also in this PR

- `theme.schema.json` regenerated via `scripts/gen_schema.sh`.
- The theme-editor field description updated in all 15 locales that carry it.
- `docs/features/editing.md` — the Post-EOF Background section, which also named the key wrongly as `post_eof_bg`.
- The dashboard plugin's comment about why it pads below its frame.
- Comprehensive visual-regression baselines regenerated: 971 changes are this fix (`#0a0a0a` -> `#000000` on post-EOF rows under `high-contrast`); 12 are pre-existing drift — whitespace indicators inside a selection picking up the derived `selected_indicator_fg` the baselines had not been refreshed for.
- CHANGELOG notes one migration gap: the theme editor saves fully-resolved themes, so a theme saved through the GUI before this release has the derived shade written into its JSON and keeps it until that field is cleared. A load-time migration that dropped a stored `after_eof_bg` equal to `shade_toward_contrast(bg, 10)` was considered and rejected — it cannot tell a stale default from a deliberate subtle shade that lands on the same value.

## Testing

- `cargo fmt --all --check`, `cargo clippy -p fresh-editor --all-targets` — clean.
- 67 theme unit tests, the 6 `issue_779` e2e tests, and the theme / visual-regression / blog-showcase e2e groups pass.
- Full e2e suite (before the review round): 3548 passed, 7 failed. All 7 are pre-existing parallel-execution flakes, not this change: 6 pass in isolation, and the 7th (`auto_indent::test_config_indent_next_line_pattern_applies`) fails identically on the parent commit with this change reverted — which test in that module fails moves between runs, and `live_grep::test_live_grep_uses_working_dir` behaves the same way. The symptom (a per-language indent config not in effect) points at shared global state leaking between concurrent tests; worth a separate look.


crates/fresh-editor/src/view/theme/types.rs
```diff
@@ -1678,14 +1680,16 @@ impl From<ThemeFile> for Theme {
...
+            // theme's own editor background: the area below the last line is
+            // still the editor, and a derived shade there reads as a grayed
+            // out strip that belongs to no theme color.
             after_eof_bg: file
                 .editor
                 .after_eof_bg
                 .clone()
                 .map(|c| c.into())
-                .unwrap_or_else(|| shade_toward_contrast(file.editor.bg.clone().into(), 10)),
+                .unwrap_or_else(|| file.editor.bg.clone().into()),
```
sinelaw, line 1692: Resolving the default here — at `ThemeFile` → `Theme` construction — means it can't see `editor.use_terminal_bg`. With that option on, the editor background becomes `Color::Reset` (`render_buffer.rs:128`, `render_composite.rs:38`) but `theme.after_eof_bg` is still the theme's opaque `editor.bg`, and both paint sites use it verbatim: `fill_eof_rows` (`render_line/trailing.rs:254`) and `render_past_end_row` (`render_composite.rs:294`).

Built this branch and read the cells out of `tmux capture-pane -e`, `dark` theme, 2-line file:

```
use_terminal_bg=false   content ESC[48;2;30;30;30m   post-EOF ESC[48;2;30;30;30m   <- fixed
use_terminal_bg=true    content ESC[49m              post-EOF ESC[48;2;30;30;30m   <- band remains
```

`use_terminal_bg` is documented as "allowing transparency or custom terminal backgrounds to show through" (`config.rs:1316`), so on a transparent terminal the space below the last line is still the one opaque rectangle on screen — the same complaint this PR is fixing, just in a different colour than before (`#1e1e1e` instead of `#282828`).

Two ways out, both small:

- Keep `Theme::after_eof_bg` as `Option<Color>` and let the paint sites fall back to `effective_editor_bg`. That value is already computed in both places, and at the `render_past_end_row` call site (`render_composite.rs:81`) it's literally in scope and just isn't passed.
- Or keep the eager resolution and mirror it where `effective_editor_bg` is derived: `if use_terminal_bg && theme.after_eof_bg == theme.editor_bg { Color::Reset }`. `themes/terminal.json` already names `after_eof_bg: "Default"`, so that equality test does the right thing there too.

---

crates/fresh-editor/tests/e2e/issue_779_after_eof_shade.rs
```diff
@@ -1,35 +1,55 @@
...
+            "name": "post-eof-shade",
+            "editor": { "bg": [30, 30, 30], "after_eof_bg": [60, 60, 60] }
+        }"#,
+    )
+    .expect("theme fixture should parse");
+    assert_eq!(theme.editor_bg, Color::Rgb(30, 30, 30));
+    assert_eq!(
+        theme.after_eof_bg,
+        Color::Rgb(60, 60, 60),
+        "a theme that names after_eof_bg still gets the post-EOF shade (#779)"
```
sinelaw, line 61: This new test covers `ThemeFile` → `Theme` conversion only, and after the default change it's the *only* thing left in this file that can distinguish `after_eof_bg` from `editor_bg`. The two rendering tests below (lines 56–99) go vacuous:

`expected` is `theme.after_eof_bg`, which is now `theme.editor_bg` for every built-in theme, and the editor block paints the whole render area with `editor_bg` anyway (`render_buffer.rs:661`). So `assert_eq!(style.bg, Some(expected))` at column 5 of a post-EOF row is indistinguishable from the same assertion on any blank cell of any buffer row. Those tests would keep passing if `fill_eof_rows` dropped `theme.after_eof_bg` and hard-coded the editor background, and they'd keep passing if `first_post_eof_row` returned a row *inside* the buffer.

Net result: nothing in the suite pins the PR's claim that "a theme that wants the post-EOF area shaded sets `editor.after_eof_bg` … and those rows use it" — the render path is unpinned in exactly the direction the change makes fragile.

One test would fix it: install a theme JSON with `bg` and a clearly different `after_eof_bg` into a harness themes dir (same pattern as `tests/e2e/theme.rs:470`), render a short buffer, and assert the post-EOF cell bg equals the override **and** differs from `theme.editor_bg`. That covers both `fill_eof_rows` and the regression this PR is about.

---

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
   * Compose mode no longer frames the paragraph after a fenced code block as code when the buffer opens or scrolls into view mid-document (#3001).
   * Bold, italic, and inline-code markup no longer stay visible as raw `**`/`` ` `` until you click elsewhere (#2968, reported by @mommysgoodpuppy).
   * Comments inside fenced code blocks are no longer misdetected as Markdown syntax, and scrolling a large file in Compose mode is smooth on the first pass instead of only after it's been scrolled through once.
+* **The space below the last line follows the theme background** - rows past the end of the buffer were painted with a shade derived from `editor.bg` rather than with a colour the theme actually names, so a short file sat above a grey strip that belonged to no theme. Those rows now use the theme's editor background; the `~` tildes still mark where the buffer ends. A theme that wants the post-EOF area shaded says so explicitly with `editor.after_eof_bg` (#779).
```
sinelaw: "A theme that wants the post-EOF area shaded says so explicitly" is true of the eight in-repo themes, but a lot of user themes already say so without meaning to, and they won't get this fix.

`ThemeRegistry::to_json_map` (`view/theme/loader.rs:310`) hands the theme-editor plugin a fully-resolved `ThemeFile`, `after_eof_bg` included, and `saveTheme` writes *every* field back out — the comment at `plugins/theme_editor.ts:2121` says so explicitly ("This ensures we always write every field"). So anyone who has customised a theme through the graphical editor before this change has `"after_eof_bg": [40, 40, 40]` (or whatever `shade_toward_contrast(bg, 10)` produced) baked into their theme JSON in the user themes directory. After upgrading they keep the grey band, and under the new semantics their file now reads as a deliberate opt-in to it.

Worth either a sentence here naming the key to delete, or a load-time migration that drops a stored `after_eof_bg` when it exactly equals `shade_toward_contrast(bg, 10)`.

---

sinelaw (commented):
Reviewed the head commit `95ef53b` (GitHub shows this PR as one commit / 10 files, so the "Scope note" about earlier unmerged work no longer applies — nothing else to skim).

Three inline findings; details on the lines. Summary comment to follow.

---

sinelaw:
## Overall read

I think the diagnosis is right and I'd land the direction, but the fix is resolved one layer too early and the tests no longer hold it in place. Three inline comments; the first is the one I'd want addressed before merge.

**The case for the change is stronger than the description makes it.** Two things I checked that back it up:

- On `dark`, `shade_toward_contrast([30,30,30], 10)` is exactly `[40,40,40]`, which is `dark.json`'s own `current_line_bg`. So the post-EOF band wasn't just "a colour no theme names" — it was rendering in the *current-line highlight* colour, i.e. the whole area below the buffer read as one giant highlighted line.
- `color_to_rgb` returns `None` for `Color::Reset` and `Color::Indexed` (`types.rs:108`), so `shade_toward_contrast` was a no-op for any theme built on terminal-default or 256-colour palette entries. #779's shade only ever existed for truecolor themes. It was never a coherent default.

**Where it falls short.** The default is resolved in `From<ThemeFile> for Theme`, before `editor.use_terminal_bg` is applied, so under that option the fix does nothing: content rows go to `ESC[49m` while post-EOF rows still paint the theme's opaque `editor.bg`. Verified against a build of this branch in tmux. That option exists specifically so terminal transparency shows through, so the band is arguably more conspicuous there now than before. Resolving against `effective_editor_bg` at paint time fixes it and costs about five lines — details inline on `types.rs:1692`.

**Tests.** `explicit_after_eof_bg_is_honored` is theme-parsing only; the two rendering tests now assert `cell.bg == theme.after_eof_bg` where that equals `theme.editor_bg`, which is what every blank cell in the content area carries. They can't fail any more. Inline on the test file with a concrete replacement.

**Migration.** The theme editor writes every field explicitly, so themes users have already saved through it carry the derived shade in their JSON and won't get this fix. Inline on the changelog line.

**Should it be a default change at all?** I'd say yes for the reasons above, but flag one consequence: this silently changes the rendering of every third-party and package theme that never opted out, and the only route back is hand-editing a theme JSON — there's no user-level toggle. If the intent is narrower ("the bundled themes shouldn't ship this band"), setting `"after_eof_bg"` equal to `bg` in the eight in-repo theme files would achieve the same visible result while leaving other people's themes alone. Related: with `editor.show_tilde: false` there is now no end-of-buffer cue at all, and several views force tildes off (session preview at `app/render.rs:3162` and `:4248`, terminal scrollback at `split_rendering/orchestration/mod.rs:216`). For terminal scrollback that's an improvement — the comment there says those rows "should look blank" — but a user running with `show_tilde: false` loses the boundary entirely. Your call; just worth being deliberate about it rather than a side effect.

**Things I checked that are fine:** all 15 locales carry the updated `field.after_eof_bg_desc`; the schema description matches the Rust doc comment verbatim and the `schema` CI job is green; the `shade_toward_contrast` doc comment update is accurate (its only remaining caller is `selected_indicator_fg` at `types.rs:167`); `docs/features/editing.md` also fixes a pre-existing wrong key name (`post_eof_bg` → `editor.after_eof_bg`); the SVG baselines are regenerated docs artifacts, not assertions (the assertion is the insta text snapshot), and the 12 whitespace-indicator glyphs really are drift from `9e47e23`, which is already on master and did not refresh them.

---

crates/fresh-editor/src/view/theme/types.rs
sinelaw: Confirmed and fixed in d8e5282. I reproduced it exactly as you did before touching anything — `use_terminal_bg=true` gave content `ESC[49m`, post-EOF `ESC[48;2;30;30;30m`.

Took the first route, with the fallback expressed as a method so both paint sites share one rule:

```rust
pub fn post_eof_bg(&self, effective_editor_bg: Color) -> Color {
    if self.after_eof_bg == self.editor_bg {
        effective_editor_bg
    } else {
        self.after_eof_bg
    }
}
```

Equality rather than `Option` keeps `Theme.after_eof_bg` a plain `Color` for the plugin API, the theme-color key table and the theme editor, and it has a useful side effect you pointed at: a theme saved through the GUI writes `after_eof_bg` equal to `bg`, and that now behaves identically to leaving it unset, `use_terminal_bg` included. `themes/terminal.json` is `Reset == Reset`, so it takes the same path.

`effective_editor_bg` now rides in `LineRenderInput`, as it already did for the gutter, and reaches `fill_eof_rows`; `render_past_end_row` takes it as an argument at the call site where it was already in scope. Re-measured after the change:

```
use_terminal_bg=true    content ESC[49m              post-EOF ESC[49m
use_terminal_bg=false   content ESC[48;2;30;30;30m   post-EOF ESC[48;2;30;30;30m
```

---

crates/fresh-editor/tests/e2e/issue_779_after_eof_shade.rs
sinelaw: Right on both counts — the two rendering tests were passing on a tautology. Added the test you describe, plus one for the `use_terminal_bg` path, in d8e5282.

`distinct_after_eof_bg_paints_only_the_post_eof_rows` installs a themes-dir theme with `bg` `[30,30,30]` and `after_eof_bg` `[90,20,20]` (the `tests/e2e/theme.rs:470` pattern), then asserts the content row keeps `bg` and the post-EOF row takes the shade — so it fails in both directions, not just one.

`post_eof_rows_follow_use_terminal_bg` asserts both rows are `Color::Reset`.

I checked each one actually fails on the mutation it exists to catch rather than trusting that it would:

| mutation | failing test |
| --- | --- |
| `post_eof_bg` returns `self.after_eof_bg` unconditionally | `post_eof_rows_follow_use_terminal_bg` |
| `fill_eof_rows` paints `effective_editor_bg`, ignoring the theme | `distinct_after_eof_bg_paints_only_the_post_eof_rows` |

One wrinkle worth recording: both sample a row *away* from the cursor. Row `first` is the cursor's line and carries `current_line_bg`, so the first draft of the `use_terminal_bg` test failed on `Rgb(20,20,20)` — a two-line buffer and `first + 1` fixed it.

I left the two original assertions in place. They are weak now, but they still pin that the rows go through `after_eof_bg` at all, and the new pair covers the direction that matters.

---

CHANGELOG.md
sinelaw: Verified — `to_json_map` serialises a resolved `ThemeFile` (`after_eof_bg: Some(theme.after_eof_bg.into())` in `impl From<Theme> for ThemeFile`), so a GUI-saved theme does bake the shade in.

Took the sentence, naming the field to clear:

> One exception: a theme you saved through the graphical theme editor before this release has the old derived shade written into its JSON, so it keeps it until you clear that field.

Not the load-time migration, deliberately. Dropping a stored `after_eof_bg` when it equals `shade_toward_contrast(bg, 10)` cannot tell a stale default from a deliberate choice that happens to land on the same value — ten steps off the background is exactly what someone reaching for a *subtle* post-EOF shade would pick by hand, and silently deleting it would be the same class of bug as the one this PR fixes, pointed the other way.

Going forward the problem does not recur: the equality check in `Theme::post_eof_bg` means a theme saved from the GUI *after* this change writes `after_eof_bg == bg` and behaves exactly as if it were unset.

---

PR #3093, opened 2026-08-26
feat(welcome): an onboarding ladder as the empty-workspace screen

sinelaw:
Replaces the `[No Name]` scratch buffer / blank-pane hint with an interactive welcome buffer, structured as a **ladder**: a zero-anxiety first viewport, then three bannered levels descending from everyday editing through IDE features to agent orchestration.

Design note: `docs/internal/welcome-screen-design.md`.

> **Status.** All eleven CI jobs green, `mergeable_state` clean, rebased onto current master as **two commits**: the host and plugin-API changes, then the feature. Manual QA across four passes found seventeen interaction and rendering defects, several of which contradicted claims earlier revisions of this description made. **Fourteen are fixed and verified against the real binary.** Two remain open (below). Three turned out **not to be this page's** and are reproduced on stock editor behaviour.
>
> The last round went the other way round from the rest of this work: **the host was fixed first**, three new primitives, and the page then deleted the workarounds they replaced. Those primitives were put through an adversarial review that found eight problems with them — seven fixed, one recorded.
>
> Commit SHAs cited in earlier revisions of this description are gone: the branch was squashed from 54 commits to 2 and rebased onto master's `fresh-editor-core` split. Where history matters it is described rather than linked.

## The shape

The first viewport is a wordmark, one line of positioning, the "show this screen on startup" switch, three numbered doors (`1`/`2`/`3` jump straight to a level), the plain verbs with their **live** keybindings, and a reassurance — and it names no LSP, git, worktree or agent. You reach those by scrolling, which is the gesture that means "show me more". The measure is capped at 88 columns and centred in the pane, the way markdown compose mode sets a page; framed cards sit narrower still, at 72.

## The page rides the painter, deliberately

The `fresh-ui` migration moved the editor's chrome onto a retained tree and took the plugin panels with it. A pane whose mounted panel is **described** now answers its own hover, presses and drags with nodes.

The flip is gated on `pane_panel_owns_its_scroll`, and **this page's scroll is the buffer's**: it carries a real cursor and the host scrolls it. So it stays on the painter, in the same class as Git Log.

That is not a deferral of taste. A described pane's rows are laid out by the tree, so the mirror's line *n* stops being screen row *n* the moment a list inside it scrolls — and the screen→line projection a caret needs (`view_line_mappings`) is filled by the very text pass a described pane does not run. A page with a caret cannot cross until that projection is replaced. The panels that crossed are the ones that never had a caret to place.

## Three primitives, and what a skeptical review did to them

The last three defects were all the same shape: the page had grown a workaround for something the host should answer. So the host answers it now.

**1. A focused text widget takes a printable key ahead of the mode's bindings.** A mode declaring `allow_text_input` contains a text widget, and while that widget has focus a character can only mean "type this". Mode bindings resolved first, so a mode binding a character silently stole it from its own focused field — which is what ate `0`–`3`, `/` and Space out of the finder. The page's `typeIfFinding` and its six call sites are gone.

**2. A panel declares whether "nothing focused" is a real state** — `WidgetPanelOptions { autoFocusFirst }` on the mount. `RenderOptions::auto_focus_first` already existed with exactly the right documented semantics and was hardcoded `true` for every panel, so a plugin clearing focus could not clear it: the next repaint re-seeded it onto the first tabbable, the plugin's record and the host's drifted, and keys went to a widget nobody had focused.

**3. `ScrollToWidget` learns to reveal, not only to jump** — `ScrollAlign::Minimal`. The command only ever had "put it on the top row", whose own doc comment says it is a jump, "not reveal this somewhere on screen". Following focus is that other request. The page's `cardForWidget` table is gone with it.

**The second took three attempts.** The mount and the update paths honoured the option and the bug survived both, because `rerender_widget_panel` — where focus changes actually go — renders *mounted* panels through `render_floating_spec`, which seeded unconditionally. Fixing the two obvious paths left the bug exactly as it was.

### The review

The three were handed to an agent told to find reasons they are wrong, judged as editor API rather than as this page's needs. It found eight things; seven are fixed:

- **`autoFocusFirst: false` shipped with a regression attached.** `handle_widget_focus_advance` resolved the focus index with `position(...).unwrap_or(0)`, so an empty key read as "focused on `ring[0]`" and the first Tab landed on `ring[1]` — skipping the first widget entirely. Unreachable while focus was always seeded; the option is what made it reachable. Nothing focused now sits *outside* the ring: forward goes to the first, backward to the last.
- **A fourth focus-seeding path.** `rerender_widget_panel` returns early for a panel the shell tree describes, and that branch called `resolve_panel(..., true)` with the policy sitting unread beside it. So the option was discarded on every repaint of exactly the panels the migration has already taken over. This page escaped only because its buffer scrolls.
- **`Minimal` could silently do nothing.** Defining it as "the cursor move alone" made it inherit `skip_ensure_visible` — a latch set by any wheel or scrollbar scroll, and by `ScrollAlign::Top` itself, cleared only at the top of `handle_key` and only for the active split. A reveal that declines to scroll looks exactly like one with no need to, so this would have failed invisibly. It clears the latch on the target splits first now.
- **A read-only or markdown `Text` counted as somewhere you type** — and that predicate now decides whether a printable key reaches the mode's bindings at all, so focusing a rendered document would divert every character into a field that discards it.
- **The modifier test was a third hand-rolled copy.** Both capture clauses go through one `typed_char` now, deferring to `is_text_input_modifier` — which knows that crossterm reports Windows AltGr as Ctrl+Alt. `!intersects(CONTROL | ALT)` excluded `@ [ ] { }` on German, French and Italian layouts.
- **The options bag could not grow the way its own doc promised.** `bool` plus a struct-level default generated a *required* TypeScript field, so every future option would break every plugin constructing the bag; and `deny_unknown_fields` meant a plugin built against a newer host failed to deserialize wholesale on an older one — falling back to all-defaults, which is the behaviour the option exists to turn off, while still reporting success. Now `Option<bool>` + `#[ts(optional)]`, no `deny_unknown_fields`, and a malformed bag fails the mount like a malformed spec does.
- **The chord carve-out claimed more than it delivered.** Returning `TextInput` before `resolve_chord` means a character-prefixed chord can never be *started* while a field has focus. That is the right reading — in a focused field, `g` is the letter g — but the comment implied otherwise and the test asserted a state the router cannot reach.

**Recorded, not fixed:** the rule is gated on `has_focused_text_widget`, computed against the *active buffer*, so it never fires for a floating or dock panel on a sentinel buffer. `orchestrator.ts`'s New Session form is in that class. Narrowing the gate to the panel holding the keyboard is a larger change than the bug in front of it.

The review also caught that a test-validation experiment had left the router clause disabled in the working tree, and confirmed the second router clause (below) was correctly reverted.

### Two primitives designed and then rejected

**Unbound modified keys falling through to Normal when no widget is focused.** It would have deleted the page's forwarding table too. Two existing tests and a comment in the same function rejected it, correctly: a mode with `allow_text_input` captures characters whether or not a widget has focus, so "no focused widget" does not mean "not typing", and Ctrl+O reaching the buffer there could hijack a mode mid-input. The forwarding table is a mode naming what it wants to let through, which is a legitimate thing for a mode to do.

**Centring the compose column on what is actually painted.** Implemented, measured, reverted — see **Known defects**.

## The rest of the host changes

Twelve more, all but two in the widget runtime, and none of them about this page. Each was found by this page and each is a bug or a gap on its own.

1. **`update_widget_hover` never reached a mounted panel.** `hoverStyle` on any of its widgets was dead spec.
2. **`labeledSection` gains a `hover_style`**, so a section can carry the key of the control inside it.
3. **`Button` gates focus rendering on `focusable`.** A widget dropped from the Tab cycle can never be what focus is on.
4. **`widget_panel_width` sized panels to the split, not the compose column.**
5. **`zip_row_blocks` skipped `RowPiece::Flex`**, so `flexSpacer()` was a no-op in any row holding a block child.
6. **The gutter was reserved twice when composing.**
7. **An input's background was gated on `focused`**, so only the field you were already typing in looked like a field. Its key also moves off `ui.prompt_bg`, which several shipped themes set to a *foreground* colour.
8. **A focused input marks itself on its frame** — there is no caret to do it on a panel mounted into a buffer.
9. **`scrollToWidget`**, so a page can navigate to its own content.
10. **`labeledSection` takes a width in columns, not an integer percent.**

Plus two outside the widget runtime: **`Window::workspace_restored`** (closing every tab and quitting left the on-disk workspace untouched, so the next launch resurrected the files you had just closed) and **navigation is not a mode's to swallow** (a mode with `allowTextInput` was dropping plain arrows, page keys and Home/End).

## The demos are live, not illustrated

| Card | What actually runs |
|---|---|
| Pick up where you left off | a real `TextInput`; fuzzy-finds over `git ls-files` as a subsequence |
| Language smarts | a Rust sample highlighted by the real grammar engine |
| Review your diff | reads `git status --porcelain -b` — real branch, real changed files |
| Make it yours | the theme buttons call `applyTheme`; the whole editor restyles |
| The Orchestrator dock | the real workspace list with agent state |

Outside a repo the finder and the git card each degrade to one sentence rather than an error.

## Lifecycle

Opens at startup with nothing to restore, and after the last buffer closes. An auto-opened screen **nobody touched** steps aside when a real file opens; one the reader engaged with stays until they close it. `Welcome` in the palette brings it back, and the page's scroll hint says so — along with the fact that `Esc` closes it.

## What the build taught

1. **A panel repaint keeps the pane's scroll position.**
2. **Hiding the caret cost more than it saved.** Built with `showCursors: false`, the page reimplemented scrolling wholesale; all deleted once it took a real cursor. The caret is load-bearing twice over — it is what puts the page on the painter side of the migration's gate.
3. **Mounting a panel clears `showCursors`.** Re-assert after the panel exists.
4. **`setBufferCursor` reveals; it does not jump.**
5. `viewport_changed` fires on height changes too, including the row the command palette takes. It also used to dedupe on raw *width*, which sounds like the layout input and is not.
6. **Shift+Tab is its own key code.**
7. **Revealing a focused widget is not the host's job here** — and the page's own answer had three bugs in it.
8. `getAllThemes()` answers with the registry object, not a list.
9. **Closing the page reopened it.**
10. A `List` inside a `labeledSection` cannot reach the section's right border.
11. `Enter` on a single-line `Text` widget is advance-focus.
12. **`executeAction` with an unknown name fails only in the log** while reporting success. Every merge from master re-checks each target by hand; the trap bit once for real, when #3098 renamed `start_review_branch` and git reported no conflict because the breakage is semantic.
13. **An underline runs the width of the button's cells.**
14. **`cargo build` does not compile test code** — and `clippy --all-targets` compiles it but never **runs** it. That is how a broken test reached CI after a clean local lint. `cargo test -p <crate> --lib` is the check that catches it; twenty call sites in test code needed new arguments for one host change.
15. **Green CI said nothing about any of the seventeen defects** — because this plugin had no tests. `tests/e2e/plugins/` has dozens for other plugins. It has three for this one now.
16. **Wrapping measured in UTF-16 units.**
17. **A card's text budget is spent in three places.**
18. **A key a mode binds never reaches the mode's own text-input handler** — bindings are consulted first. Now the host's problem, not the page's.
19. **A hit area is the control's own cells, not the row it sits on.**
20. **A layout hint has to match what the host actually reserves.**
21. **`getViewport()` is the *active split*, not your buffer.** Split the window, work in the other pane, and every measurement this page took came from that pane. The `viewport_changed` payload carries the right number, addressed to this buffer — it was being destructured and thrown away.
22. **A workaround can hide a regression in the thing that replaces it.** `autoFocusFirst: false` was correct and exposed a latent first-Tab bug that could not fire while focus was always seeded. Both halves of a swap need testing, not just the half you changed.

## Testing

**Automated:** all eleven jobs green on the current head across ubuntu, macOS and Windows. Locally on the rebased tree: `cargo test -p fresh-editor --lib` (2499) and `cargo test -p fresh-editor-core --lib` (1294 — the widget runtime moved crates in master's data-layer split), the three welcome e2e tests, `cargo fmt` and `clippy --all-targets` all clean.

**The plugin now has e2e tests** — `tests/e2e/plugins/welcome_screen.rs`, the first for this plugin. Three, each pinning something that was *observed broken* rather than a happy path: the page renders its three doors; the finder receives the characters the page also binds (the slash, the digit and the space); leaving the finder does not park focus on the startup switch. The character test was verified to **fail** with the router clause disabled, so it catches the bug rather than merely passing beside it.

**Manual:** an independent QA pass drove the real binary in tmux at 60/80/100/120/160 columns, in a git repo and outside one, across restarts; re-driven after each migration merge and after each fix round. Confirmed across those passes: startup with no stray `[No Name]`; the numbered jumps; line and page scrolling; `Ctrl+Home`/`Ctrl+End`; Tab/Shift+Tab in both directions with exactly one thing highlighted; the finder filtering, editing, walking hits and opening a pick; Enter *and* Space activating; theme chips restyling live and persisting; cards structurally intact at every width with correct CJK measurement; the startup checkbox round-tripping to disk; mouse click, hover enter/leave and wheel via SGR sequences.

**On the primitives round**, each fix driven against the real binary and read back in escape codes: `/` then Escape then Space leaves `showOnStartup` true where it used to write false; the first Tab changes exactly one row — the startup switch — where it used to skip it; the first Shift+Tab wraps to the end of the page; the wordmark survives the first Tab and the focused door keeps its own title; six Tabs and six Shift+Tabs return the view; the six top-of-page stops and the five theme chips scroll not at all, while every stop below the fold still reveals; `src/main1 x` types into the finder with every digit, the slash and the space; `2` still jumps to Level 2 when nothing is focused.

**After the rebase onto master's `fresh-editor-core` split**, all of the above re-run plus a click 40 columns right of a finder result's filename, which still opens it.

**Earlier revisions of this section claimed the page was centred "to within half a column", that the only ellipsis on it was the finder's own, and that PageDown "paints the caret on the status-bar row". All three were wrong**, and are corrected here and below.

## Fixed

**First QA pass:** seven focus stops never revealed; the startup switch now says what it did; the finder list never scrolled; wrapped card prose dropped words; the theme swatch lost its `●` after restart.

**Second pass:** digits `0`–`3` swallowed by the level jumps; `/` silently dropped, so `src/main` typed as `srcmain`; Space opened the marked result instead of typing; finder rows clickable only on their filename glyphs; the finder field had no focus indicator; the page leaned, its level rules and door cards a column apart; `Esc` closed the page and the page never said so.

**Primitives round:** `/` then Escape then Space silently turned off "Show this screen on startup" — a persisted setting changed with the control off screen; the focus reveal scrolled even when the target was already visible, clipped the focused control's own card title, and made Tab and Shift+Tab stop being inverses; `viewportWidth()` read the wrong pane when the window was split; `layoutKey` ignored height while `air()` keys on it; and the dead code the review found — `toMeasure()`, `ART_W`, a `void`-discarded accumulator, `Workspace.dirty`.

## Known defects (remaining)

1. **One column of horizontal parity.** Traced properly: the pane area is 119 columns (not the terminal's 120), the compose column 89, the painted panel 88, `left_pad` 15 and the right margin 16. So the residual is **one unavoidable column** from an odd slack — and at 120 the "centre what is painted" formula gives the *same* answer as the current one; they differ only at other parities, where the new one is right. That fix was implemented (centre on the painted width, with a shared `composed_panel_width`) and then **reverted**: keeping render and click mapping in step requires the painted width at ~8 click-path call sites, and a mismatch there is an off-by-one click bug across every composed buffer — strictly worse than a column of cosmetics. The clean version stores the painted width on the split view state rather than threading it.

2. **Light theme paints the page's side margins on only some rows.** Narrowed considerably: the banding is **content-tied and per-card** (the git card's rows lack the desk colour, the theme card's have it), it follows the content when the view moves, and it **survives a full redraw** — so it is not stale cells. `dark` is uniform because `compose_margin_bg` is close to the editor background there. Not root-caused. No text contrast failures were found (WCAG scan over seven screens, nothing below 2.0).

## Not this PR's

- **After a page motion the viewport and the cursor disagree by exactly `PAGE_OVERLAP`.** On this page the first `PageDown` leaves the caret *hidden* — `cursor_flag` is 0, so no rendered line owns that byte — and the next keypress slides the view about three rows. One bug: the landing position is the last column of a visual row, and the column before it renders three rows *below* the viewport top, which is the overlap. **It reproduces on a plain wrapped text file with no widget panel involved**, which is what rules this page out. This supersedes the earlier "PageDown leaves the cursor above the viewport" and "the view slides ~3 rows" entries; the "column jumps unasked" half of that report was wrong — landing mid-line is the documented behaviour of a page motion over wrapped text.
- **Dismissing the page opens the File Explorer.** Host behaviour on an emptied workspace: closing the last file tab with the mouse, in a session that never showed this page, opens the explorer too.
- **A raw i18n key `status.cancelled` reaches the status bar.** `i18n::t()` returns the key itself on a miss, so the `editor.t("x") || "fallback"` idiom is a **dead fallback in 63 places across 2 plugins**.
- **A click inside a panel publishes the hardware cursor at the wrong widget.** Carried from earlier testing; no pass since has reproduced it on single clicks.
- **`heading_marks_survive_exploring_the_document` intermittently times out on slow runners**, at the 180 s per-test cap rather than on an assertion. It has done so on three heads of this PR and passed on a re-run and, on the current head, without one. Master shows the same signature on its own commits with a different test. Nothing here to port a fix into; if the cap needs raising or the test splitting, that belongs on master beside the commit that made it heavy.

## Still aspirational

The LSP card has no live hover/diagnostic, the git card links to review rather than staging hunks inline, and the Orchestrator card lists workspaces without embedding a terminal. All three are called out in the design note.

---

https://claude.ai/code/session_016LUgxK1669ZAv2jsbkE59z


sinelaw:
**`test macos-latest` is red on `361d3b8`, and I don't believe it belongs to this PR.**

What failed — one test, and it timed out rather than asserting:

```
TIMEOUT [180.271s] (5550/9313) fresh-editor::e2e_tests
Summary [1995.417s] 9313 tests run: 9312 passed (9 slow), 1 timed out, 162 skipped
```

Why I read it as not this PR's:

- **`test ubuntu-latest` passed the same suite on the same commit** (`361d3b8`), as did clippy, doc, fmt, schema, `check (no plugins)`, actionlint, installer and build-musl. Only macOS failed.
- **The diff touches nothing markdown-compose.** This head adds a plugin change (`welcome_screen.ts`: an action rename and a button label) on top of a merge of `master`. There is no path from a plugin string to a markdown compose scrollbar test.
- **That test file was one of the nine deflaked in #3097** (`95ccc19`), which is already merged into this branch — so the existing stabilisation is present here, and there is no further fix to port.
- **The runner looks loaded**: nine tests flagged slow in this run, one taking 152s, and this one hitting the 180s ceiling exactly. It passed on the three previous heads of this same PR (`356dad6`, `7fe07db`, `1edf6ce`).

Re-run: attempted immediately and refused with `403 This workflow is already running`, because `test windows-latest` is still in progress on the same run. I'll re-run the failed job once the run completes, and will keep watching this PR until it is green rather than leaving it here.

To be clear about what green would and would not mean: this PR is **not ready to merge** for reasons unrelated to CI — six defects from a manual QA pass remain open, and they are listed under *Known defects (remaining)* in the description. None of them is visible to any check in this suite.

---

sinelaw:
**`test macos-latest` is red on `1856d7f`, and it is not this PR's.** One e2e test hit the 180 s per-test cap:

```
Summary [1553.197s] 9882 tests run: 9881 passed (6 slow), 1 timed out, 162 skipped
TIMEOUT [ 180.622s] (5955/9882) fresh-editor::e2e_tests
```

A timeout, not an assertion — and the evidence points at runner speed rather than the diff:

- **This branch touches nothing in markdown compose.** `git diff --name-only master HEAD` matches no file with `markdown`, `compose` or `scrollbar` in its path.
- **That test was made markedly heavier by master itself.** `93741f1` ("test: give the scrollbar-marker tests a real grammar registry") added `.with_full_grammar_registry()` to this suite, which is what makes the editor resolve embedded regions. A test whose name is *survive exploring the document* now does that exploration against a full grammar load.
- **The same test passed on ubuntu in this very run**, and on macOS on the previous head (`1194a27`, all eleven green).
- **Master shows the same signature on its own commits.** Two before the tip merged here, `22df6e31`'s Windows job failed with exactly one e2e test at the cap — `review_diff_line_staging::test_review_visual_stage_line_in_second_file`, `9439/9440 passed, 1 timed out` — a different test, same shape. The slow runners intermittently push one long e2e test over 180 s.

Master at `47483e8`, the commit merged here, is green.

**Action:** re-running the failed job once, which is currently blocked because `test windows-latest` is still in progress on the same workflow run. I will spend that single re-run as soon as the run completes and report the outcome here rather than leaving it. No fix is being ported because there is nothing in this PR to port one into — if the cap needs raising or that test needs splitting, it belongs on master beside the commit that made it heavy.

---

sinelaw:
**The re-run passed.** `test macos-latest` is green on `1856d7f`, and with it all eleven jobs — `heading_marks_survive_exploring_the_document` completed well inside the cap on the second attempt. That confirms the reading above: a runner-speed timeout on a test this branch does not touch, not a defect here. The one permitted re-run is now spent; if that test times out again on a future head it gets treated as real rather than re-run.

New head `e451131` merges master `18c98b9` — four render/geometry commits (grid prepared once per frame, panes reconciled before the frame, the dead `build_view_data` pass dropped, the gutter resolved per pane). That is the machinery this page's layout and its hover tracker sit on, so the clean merge was not taken as proof. Verified on it: `clippy --all-targets` and `cargo fmt` clean, every `executeAction` target and `getPluginApi` name still resolving, `pane_content_at` still present — and by hand in tmux at 120 columns, the page still renders, the block is still **exactly one column** left of centre (the gutter rework did not shift it), hover lights on enter and clears on leave, a click still jumps to Level 2, the caret still tracks (`Ln 6` → terminal row 7), and the finder still takes a click, filters live and opens `src/parser.rs` on Enter, with the git card showing the live branch and changed files.

---

sinelaw:
**`test windows-latest` is red on `6be7831` with a timeout, and the evidence says it is not this PR's.** Re-run triggered; this is a different test from the one discussed above, so it gets its own first re-run.

```
Summary [1945.686s] 9073 tests run: 9072 passed (4 slow, 1 leaky), 1 timed out, 433 skipped
TIMEOUT [ 180.065s] (4580/9073) fresh-editor::all_tests
    e2e::orchestrator_dock::dock_new_folder_and_move_session_into_it
```

A timeout at the 180 s per-test cap, not an assertion. Four separate lines of evidence:

**It runs in 1.44 seconds locally.** Not "fast enough" — 125× under the cap. A change that pushed this test to 180 s would have to have made it two orders of magnitude slower, and it would then be slow on ubuntu and macOS too:

```
test e2e::orchestrator_dock::dock_new_folder_and_move_session_into_it ... ok
test result: ok. 1 passed; ... finished in 1.44s
```

**The other two platforms passed the same suite on the same commit.** `test ubuntu-latest` and `test macos-latest` are both green on `6be7831`, as are clippy, doc, fmt, schema, `check (no plugins)`, actionlint, installer and build-musl. Only Windows, and only this one test.

**This exact branch content already passed Windows twice.** `6be7831` is a pure rebase of `b6f1392` onto master `fa4efb7` — I verified the plugin, docs and API trees are byte-identical across it. Windows was green on `b6f1392`, and on `d52bae6` before it. The only delta since is master's own commit.

**And the diff cannot reach this test.** The dock is a *floating/dock* panel, and all three new primitives are inert for it: the key-precedence rule is gated on `has_focused_text_widget`, computed against the **active buffer**, and dock panels live on a sentinel buffer — noted as a known limitation under *Three primitives* in the description. `autoFocusFirst` is hardcoded `true` at both floating/dock mount sites, so the dock resolves focus exactly as before. And the empty-focus branch of `handle_widget_focus_advance` cannot fire for a panel whose focus is seeded.

This is the same signature this PR has hit before on slow runners — a single e2e test at exactly 180 s while everything else passes, with one test flagged `leaky` in the same run. Master shows it too, on its own commits, with different tests each time. Nothing to port a fix into here: if the cap needs raising or these tests need splitting, that belongs on master.

I will report the re-run outcome here rather than leaving it, and keep watching until the PR is green.

---

sinelaw:
**The re-run passed.** `test windows-latest` is green on `6be7831`, and with it all eleven jobs — `dock_new_folder_and_move_session_into_it` completed inside the cap on the second attempt, on the same commit and the same runner image.

That confirms the reading above: a runner-speed timeout on a test this branch cannot reach, not a defect here. This test's one re-run is now spent; if it times out again on a future head it gets treated as real rather than re-run.

Worth recording for whoever looks at this next: **this is the third distinct test** to hit the 180 s per-test cap on this PR's heads (after `heading_marks_survive_exploring_the_document`, twice, and once on master's own commits with `test_review_visual_stage_line_in_second_file`). Different test each time, same signature — one e2e test at exactly the cap while the other 9000-odd pass, usually with something flagged `slow` or `leaky` in the same run. That pattern is about the cap and the runners, not about any one test, and it belongs on master rather than in a feature branch.

---

PR #3096, opened 2026-08-27
fix(lsp): preserve Windows drive letter in workspace root_uri (#3067)

### LINKED ISSUE #3067: [Bug] LSP root_uri drops Windows drive letter (file:///Users/... instead of file:///C:/...) causing os error 3
### Prerequisites

- [x] I have searched existing issues to avoid duplicates (checked sinelaw/fresh #2923, #1071, and code-yeongyu/oh-my-openagent #6167)
- [x] I am using Fresh 0.4.10 (latest released at time of writing)
- [x] I have read the LSP docs (`docs/features/lsp.md`, "Workspace Root Detection")

### Bug Description

On Windows, when Fresh opens a file whose path lies under the user-profile directory (e.g. `C:\Users\<user>\...`, including `C:\Users\<user>\.config\...`), the LSP `initialize` request is sent with a **malformed `root_uri` whose Windows drive letter has been stripped**: `file:///Users/Lance/.config/opencode` instead of `file:///C:/Users/Lance/.config/opencode`. Any root-dependent LSP operation then fails with `The system cannot find the path specified. (os error 3)`.

Reproduced on Windows 11 with Fresh 0.4.10 and the Biome LSP configured for `jsonc`. The per-document `didOpen` URI is correct (`file:///C:/Users/...`), so the server still spawns and initializes; only the workspace `root_uri` is corrupted, which breaks workspace-root features (file watching / root resolution) and surfaces as `os error 3` in the Fresh log.

This is the same class of Windows `file://` URI bug as oh-my-openagent #6167 (drive-letter *case* mismatch) — here it is drive-letter *stripping*.

### Steps to Reproduce

1. On Windows, configure an LSP for a language (e.g. Biome for `jsonc`):

```json
{
  "lsp": {
    "jsonc": [
      { "command": "C:\\...\\biome.exe", "args": ["lsp-proxy"], "enabled": true, "auto_start": true }
    ]
  }
}
```

2. Open any file located under your user-profile directory, e.g. `C:\Users\Lance\.config\opencode\opencode.jsonc`.
3. Observe Fresh's LSP log at `%LOCALAPPDATA%\fresh\logs\fresh-*.log`.

### Expected Behavior

`root_uri` in the LSP `initialize` request should be a valid Windows `file://` URI: `file:///C:/Users/Lance/.config/opencode`.

### Actual Behavior

`root_uri` is `file:///Users/Lance/.config/opencode` (drive letter missing). Fresh logs:

```
ERROR fresh::services::lsp::async_handler: LSP (jsonc): The system cannot find the path specified. (os error 3)
```

The server still spawns (PID assigned) and `didOpen` uses the correct URI, but root-dependent operations fail.

### Error Logs

Fresh LSP log (`%LOCALAPPDATA%\fresh\logs\fresh-15380.log`):

```text
... Sending didOpen to LSP 'biome.exe' for: file:///C:/Users/Lance/.config/opencode/opencode.jsonc   <- CORRECT (drive present)
... Spawning async LSP server: C:\Users\Lance\AppData\Roaming\npm\...\biome.exe ["lsp-proxy"]
... Initializing async LSP server with root_uri: Some(Uri(Uri { scheme: Some("file"),
      authority: ..., path: "/Users/Lance/.config/opencode", ... }))   <- MALFORMED (drive stripped)
... LSP (jsonc) progress begin: Biome is scanning the project
... ERROR fresh::services::lsp::async_handler: LSP (jsonc): The system cannot find the path specified. (os error 3)
... LSP (jsonc): Server initialized with PID: 22908
```

### Configuration

```json
{
  "lsp": {
    "jsonc": [
      {
        "command": "C:\\Users\\Lance\\AppData\\Roaming\\npm\\node_modules\\@biomejs\\biome\\node_modules\\@biomejs\\cli-win32-x64\\biome.exe",
        "args": ["lsp-proxy"],
        "enabled": true,
        "auto_start": true
      }
    ]
  }
}
```

#### Root cause (confirmed by reading `crates/fresh-editor/src/services/lsp/manager.rs`)

The workspace root URI is built by `path_to_uri()` (≈ lines 203–289). It converts a `Path` to a `file://` URI **manually, without the `url` crate**, iterating `path.components()` and only handling `RootDir` and `Normal`:

```rust
fn path_to_uri(path: &Path) -> Option<Uri> {
    let abs = if path.is_absolute() { path.to_path_buf() } else { std::env::current_dir().ok()?.join(path) };
    // Percent-encode each path component for RFC 3986 compliance
    let encoded: String = abs.components().filter_map(|c| match c {
        std::path::Component::RootDir => None, // handled by leading '/' in Normal
        std::path::Component::Normal(s) => { /* percent-encode segment, push leading '/' */ ... }
        _ => None,                              // (catch-all) Component::Prefix (drive) dropped here
    }).collect();
    format!("file://{}", encoded).parse().ok()
}
```

Component analysis for `C:\Users\Lance\.config\opencode` on Windows:

| Component | `Component` kind | Handled? | Result |
|---|---|---|---|
| `C:` | `Prefix` | `_ => None` | **dropped** |
| `\` | `RootDir` | `None` | dropped |
| `Users`, `Lance`, `.config`, `opencode` | `Normal` | encoded + leading `/` | `/Users/Lance/.config/opencode` |

`format!("file://{}", encoded)` → `file:///Users/Lance/.config/opencode`. The drive letter is gone because `Component::Prefix` (which carries `C:` on Windows) is not handled — it falls into the catch-all `_ => None`.

Note: the per-document `didOpen` URI is correct, which implies document URIs are produced by a *different* (Windows-aware) code path. `root_uri` should use the same conversion so both are consistent.

#### Why this surfaces for profile-directory files

`path_to_uri` is applied to the resolved workspace root. For any absolute Windows path with a drive (e.g. `C:\Users\<user>\...`), the root path still carries the `C:` `Prefix`, which the function strips. The symptom appears as `os error 3` when Fresh or the LSP uses the broken `root_uri` for a root-dependent operation (workspace symbols, file watching, project scan, root resolution).

#### Related issues

- sinelaw/fresh #2923 — LSP never starts for `fresh ssh://` launched from a Windows host (POSIX remote path → `file://` URI conversion)
- sinelaw/fresh #1071 — Diagnostics Panel Windows Paths: file URIs not decoded properly on Windows
- code-yeongyu/oh-my-openagent #6167 — `file://` drive-letter *case* mismatch (same family; this issue is *stripping*)

#### Suggested fix (sketch, not a PR)

`path_to_uri` should preserve the Windows drive prefix. Two options:

1. **Handle `Component::Prefix` on Windows.** For `std::path::Component::Prefix(prefix)`, match `Prefix::Disk(d)` and emit `/<drive>:` (e.g. `/C:`) so the assembled URI becomes `file:///C:/Users/...`. Also cover `Prefix::VerbatimDisk` / UNC as needed.
2. **Use the `url` crate** (`Url::from_file_path(abs)`). The function's own comment says it avoids the `url` crate — that avoidance is the root cause. `Url::from_file_path` already produces correct, RFC-compliant Windows URIs (`file:///C:/...`) and handles percent-encoding and UNC. Reusing it removes the hand-rolled encoder entirely.

Either way, `root_uri` should go through the same Windows-aware conversion used for document `didOpen` URIs.

#### Evidence files / reproducers

- Fresh LSP log excerpt above (`%LOCALAPPDATA%\fresh\logs\fresh-15380.log`).
- Config in "Configuration". Repro is 100% on any LSP-enabled file opened under `C:\Users\<user>\...` on Windows with an LSP configured.

### Operating System

Windows (Windows 11)

### Fresh Version

0.4.10

56steve:
## What

Closes #3067 — on Windows, the LSP `initialize` request's `root_uri` dropped the drive letter (`file:///Users/…` instead of `file:///C:/Users/…`), so every root-dependent LSP operation failed with `The system cannot find the path specified. (os error 3)`.

## Root cause

`LspManager::path_to_uri` (`crates/fresh-editor/src/services/lsp/manager.rs`) hand-rolled its own `file://` encoder that matched only `Component::RootDir` and `Component::Normal`. On Windows a path like `C:\Users\Lance\.config\opencode` decomposes as:

| Component | kind | old encoder |
|---|---|---|
| `C:` | `Prefix` | **dropped** (catch-all `_ => None`) |
| `\` | `RootDir` | dropped |
| `Users`, `Lance`, … | `Normal` | encoded |

→ `file:///Users/Lance/.config/opencode`, drive gone.

Per-document `didOpen` URIs were already correct because they go through the shared, platform-aware `fresh_core::file_uri` converter (which handles `Prefix::Disk`/`VerbatimDisk`/UNC). The reporter diagnosed this exactly — credit to them for the excellent write-up.

## Fix

Route `path_to_uri` through that **same shared converter** (`fresh_core::file_uri::path_to_lsp_uri`) so the workspace root and document URIs are encoded identically, and delete the buggy duplicate encoder. The manager's relative-path → absolute behavior is preserved (the shared converter returns `None` for relative paths, so absolutization stays in `path_to_uri`).

This also removes a second, divergent `file://` encoder — the kind of duplication the contributing guide warns about ("when a bug recurs, consider centralizing").

## Tests

- **Windows-gated regression test** `test_path_to_uri_preserves_windows_drive_letter` — `C:\Users\Lance\.config\opencode` must encode to `file:///C:/Users/Lance/.config/opencode`. It fails against the old encoder and passes with the fix, and runs on the **windows-latest** CI runner (this repo's CI matrix already includes it).
- Existing Unix `test_path_to_uri_basic` / `test_path_to_uri_with_spaces` continue to pass unchanged (behavior parity).
- The delegate `fresh_core::file_uri` has its own extensive `#[cfg(windows)]` coverage for drive/UNC/percent-encoding, also CI-run on Windows.

## Verification note

I developed this on macOS, where `std::path` can't produce a Windows drive `Prefix`, so the drive-specific test is `#[cfg(windows)]` and is validated by the Windows CI runner rather than locally. Locally I confirmed: the fix builds, the shared converter's full test suite passes, and the Unix `path_to_uri` parity tests pass. `cargo fmt` ✓ · `cargo clippy -p fresh-editor --all-targets` ✓ (0 errors) · `cargo check --all-targets` ✓.


56steve:
Pushed a follow-up that fixes the windows-latest failure.

Quick summary of what happened: the new Windows test (test_path_to_uri_preserves_windows_drive_letter) passed, so the actual fix works. The two failures were the pre-existing tests, test_path_to_uri_basic and test_path_to_uri_with_spaces.

Those tests feed POSIX paths like /tmp/test. On Windows that path has no drive, so path_to_uri resolves it against the current directory and the encoded URI ends up carrying that drive (file:///D:/tmp/test). The old encoder stripped the drive, so file:///tmp/test used to match. In other words those two only stayed green on Windows because of the bug this PR fixes.

The fix: I gated both POSIX-shaped tests to cfg(unix), since that is the platform whose encoding they actually assert, and moved their space-encoding check into the cfg(windows) test so each OS still covers both the drive letter and percent-encoding.

Verified locally on macOS (Unix tests pass, fmt and clippy clean); the Windows drive test already passed in the previous run. Could you re-run CI when you get a chance? Thanks!

sinelaw (changes requested):

- **Fix is correct and well-motivated**: verified the bug still exists on current `master` (`path_to_uri` in `manager.rs` drops `Component::Prefix`), and that routing through `fresh_core::file_uri::path_to_lsp_uri` is the right call — it's the same shared, extensively-tested converter used for `didOpen` URIs, matching CONTRIBUTING.md's "when a bug recurs, consider centralizing" guidance. Issue #3067 is still open and unaddressed elsewhere, so no duplicate.
- `cargo fmt --check` and `cargo clippy -p fresh-editor --all-targets` are clean on this branch; no new warnings in the touched file.
- Interactively validated the LSP root_uri/document-URI pipeline still works end-to-end on Linux (opened a Python file, started `pylsp`, confirmed semantic highlighting and "Go to Definition" correctly jumped to the definition) — no regression from the refactor. Couldn't validate the Windows drive-letter case directly (no Windows host here), but it's covered by the new `#[cfg(windows)]` test and the delegate's existing Windows test suite.
- **Blocker: this branch is stale and now conflicts with master.** `8abfe0016` (already on master) touched the same function and made `path_to_uri` `pub`, removed `all_handles`/`all_handles_mut`, and replaced the inline `detect_language` with a re-export from `fresh_editor_core::language_detect`. A raw merge of this PR's diff would silently revert those. Please rebase onto current `master` and resolve keeping both changes (shared `file_uri` converter + `pub`/re-export/removed-methods from `8abfe0016`).

---

56steve:
Rebased onto current master (566ad6c14) and resolved the conflict in manager.rs, keeping both sides: master's `pub fn path_to_uri` signature from 8abfe0016, plus this PR's change that routes the body through the shared `fresh_core::file_uri::path_to_lsp_uri` converter. None of master's work on that function is reverted.

Confirmed the shared converter still lives at crates/fresh-core/src/file_uri.rs, so the delegation is valid on the new base.

Verified locally on macOS: cargo fmt --check clean, cargo clippy -p fresh-editor --all-targets clean on the touched file, and the Unix path_to_uri tests pass (the Windows drive-letter test compiles out here and runs on the Windows CI runner). Ready for another look when you get a chance.

sinelaw:
:fire:

Issue #2314, opened 2026-06-11
Configurable indentation rules ([languages.<id>.indent]) have no effect — all custom patterns are ignored


sinelaw:
**Version:** fresh 0.4.0 (built from `master` @ `1b5d7f8c8`)

The 0.4.0 headline feature *"Configurable indentation rules per language via `[languages.<id>.indent]`, including VS Code-style regex rules"* (CHANGELOG 0.4.0; documented in `docs/configuration/index.md` → "Customize Auto-Indentation") does not work. A custom language's `indent` block is parsed enough that the language is recognized (its name shows in the status bar), but **none of the five `indent` patterns take effect**. Auto-indentation falls back to Fresh's built-in heuristics regardless of what the user configures.

## Steps to reproduce
1. Create a project dir, e.g. `/tmp/indent-test/`, with `.fresh/config.json`:
   ```json
   {
     "languages": {
       "incend":  { "extensions": ["t1"], "tab_size": 4, "use_tabs": false,
                    "indent": { "increase_indent_pattern": "OPEN\\s*$" } },
       "decr":    { "extensions": ["t3"], "tab_size": 4, "use_tabs": false,
                    "indent": { "increase_indent_pattern": "OPEN\\s*$", "decrease_indent_pattern": "^\\s*CLOSE\\b" } },
       "indnext": { "extensions": ["t4"], "tab_size": 4, "use_tabs": false,
                    "indent": { "indent_next_line_pattern": "^\\s*HDR\\b" } },
       "dednext": { "extensions": ["t5"], "tab_size": 4, "use_tabs": false,
                    "indent": { "increase_indent_pattern": "COLON\\s*$", "dedent_next_line_pattern": "^\\s*RET\\b" } }
     }
   }
   ```
   (`OPEN`/`CLOSE`/`HDR` are deliberately non-bracket, non-colon tokens so they do **not** overlap Fresh's built-in heuristics — this isolates whether the *custom* rule fires.)
2. From that dir: `fresh --no-restore test.t1` (status bar confirms filetype `incend`, so the language entry loaded).
3. Type `foo OPEN`, press **Enter**, type `child`.

## Expected behavior
Per the docs, `increase_indent_pattern` matches the reference line, so the new line is one level deeper:
```
foo OPEN
    child
```
This mirrors VS Code's `indentationRules.increaseIndentPattern` (`language-configuration.json`), which Fresh's docs explicitly model.

## Actual behavior
[...]

PR #3097, opened 2026-08-27
test: deflake nine e2e tests — shared process state, and four waits that gated nothing

sinelaw:
Five independent causes behind the reported flaky tests, each fixed where it lives. All changes are in `tests/`; no production code is touched.

Four of the five are the same shape: **a wait that reads as a gate but isn't one.** A wait for "a prompt" satisfied by the prompt just confirmed; a wait for a toolbar that the toolbar precedes its content; a wait for quiescence that an unobserved round trip satisfies; and a pair of bare renders standing in for a 150 ms debounce. Each was replaced with a condition that can only become true once the thing being waited for has actually happened.

Tests fixed (9): `locale::test_locale_switch_updates_search_cancelled_message`, `locale::test_locale_switch_via_command_palette`, `locale::test_multiple_locales_can_be_loaded`, `auto_indent::test_config_increase_indent_pattern_applies`, `lsp::lsp_spawn_failure_writes_stub_log_for_view_log_popup`, `plugins::review_diff_ux_bugs::test_range_review_ignores_external_diff_and_format_settings`, `plugins::review_diff_ux_bugs::test_range_review_lays_out_every_file`, `markdown_compose_scrollbar_markers::heading_marks_survive_exploring_the_document`, `occurrence_highlight::test_zero_width_block_selection_keeps_word_highlight`.

## 1. Review Range opened on a half-drawn frame

`open_review_range_head` gated on `screen.contains("next hunk")` alone, on the reasoning that the toolbar's appearance implies a settled stream. It does not: `createBufferGroup` paints the toolbar near the *start* of `openReviewPanels`, the stream body is written by the `updateMagitDisplay` after it, and the "Generating Review Diff Stream..." status is only replaced by the `updateReviewStatus` after that. So the helper returned on a frame whose diff area was still blank.

Reported failure, `test_range_review_lays_out_every_file` — toolbar up, diff empty, status still generating:

```
 Tab focus  , . file  n next hunk  p prev hunk  v select  c comment
 1 2 unified/split  ↑↓ move in panel  Enter jump  Alt+o open file  / filter  ? help  e export  q close
 ───────────────────────────────────────────────────────────────────
                                    (empty)
 Trusted  Local  Generating Review Diff Stream...      [RO]  LF  UTF-8
```

`test_range_review_ignores_external_diff_and_format_settings` shares the helper. Every other review-diff test file already waits for the toolbar **and** the status to move on, so that condition now lives in one `wait_for_review_ready` that the worktree and range helpers both call.

Two smaller races in the same preamble go with it:

- `"Review Range"` is registered by the audit_mode plugin and Quick Open re-filters on input change only, so typing the name and pressing Enter blind could fire on whichever row was selected when the last keystroke landed. It now goes through `run_palette_command`, which waits for the row — the same discipline `open_review_diff` already used.
- The `wait_for_prompt` meant to wait for the range picker was satisfied by the palette prompt just confirmed — a wait that cannot fail and therefore gates nothing (CONTRIBUTING.md Code §16). It now waits for the picker's own label, after which the backspaces edit the picker's prefilled `HEAD` rather than the palette's query.

## 2. Two process globals clobbered by unrelated harness construction

`EditorTestHarness` writes two process-global registries while building an editor, both derived from the `Config` it was handed:

- the i18n locale — `i18n::init_with_config` → `rust_i18n::set_locale`;
- the user indentation rules — `Editor::for_test` → `reload_indent_overrides`, which **clears** `indent_rules::USER_RULES` before registering its own `[languages.<id>.indent]` blocks.

Cargo runs a binary's tests in parallel threads, so a harness built by any unrelated test overwrites both, and the failure lands on whoever was depending on them:

- the three `locale::` tests pin a non-English locale and then read an English screen. `lock_locale` serialized the locale tests against each other but never against harness construction — the TODO `harness.rs` already carried.
- `auto_indent::test_config_increase_indent_pattern_applies` registers an `increase_indent_pattern` under a config-only language id and finds the rule gone by the time it presses Enter, so the new line merely copies the previous indent: the exact symptom #2314 fixed, arriving for a reason unrelated to the fix. (`indent_rules`' own unit tests already keep a single owner of that global for the same reason.)

Both now go through one reader/writer pin in `tests/common/global_state.rs`, replacing `locale_lock.rs`:

- every harness construction takes the **read** side for the length of the construction;
- a test that needs one of those globals to hold still takes the **write** side for its whole body.

Readers do not exclude each other, so ordinary tests keep building editors in parallel; only the pinning tests serialize, and only against construction. The pin is re-entrant for its own thread, so a pinned test still builds harnesses of its own.

## 3. The LSP stub log was deleted after the editor could write it

`lsp_spawn_failure_writes_stub_log_for_view_log_popup` removed `lsp_log_path("pascal")` *after* building the harness. The bogus command is already in the `Config` that harness is built from, so construction is a point at which the editor can reach `force_spawn` and write the stub — and losing that race deletes the stub the wait is waiting for. That is not a failed assertion but a wait that can no longer resolve, i.e. a three-minute nextest kill whose screen dump says nothing about why. Deleting first cannot lose: the only writer afterwards is this test's own failed spawn.

## 4. A mark snapshot taken before the pre-scan had published

`heading_marks_survive_exploring_the_document` compares the marked scrollbar rows before and after paging through the document, so the "before" set has to be the complete one. It read that set after `wait_for_async_quiescence(8)` and got `[2, 3]` — the two headings in the opening viewport — against the eighteen the assertion expects.

Quiescence cannot be the gate. The complete set comes from `prescanHeadingMarkers`, which compose mode starts fire-and-forget (`void prescanHeadingMarkers(...)`): it reads the whole buffer over the async bridge and then publishes every heading in one whole-namespace `setScrollbarMarkers`. While that `getBufferText` round trip is outstanding the editor has no work to report for the tick, so the quiescence wait returns with only the per-batch `lines_changed` viewport marks published — and the test measured a partial set against a complete one and blamed the scroll. (The comment that stood there described the pre-scan as walking the document over several ticks; it does not, and that mis-model is what sent the previous fix to quiescence.)

Because the publish is atomic, the marks themselves are the gate: the viewport pass cannot reach ten on a document this long, so waiting for ten resolves only once the pre-scan has landed — the wait `every_heading_is_marked_without_scrolling` already uses for the same step. Quiescence stays after it, settling a per-batch republish on top of an already-whole set.

## 5. An assertion on a frame kept alive only by a 150 ms debounce

`test_zero_width_block_selection_keeps_word_highlight` failed on the Windows and macOS legs while passing on Linux. It was racing a timer.

The occurrence highlight debounces its target by 150 ms of wall clock (`ReferenceHighlightOverlay::update`, `Instant::now()`) and advances that clock only inside a render: the first render after a target change records the time and keeps the *previous* overlays; a later one applies the new ones once the delay has elapsed.

A block selection is a single cursor carrying a `block_anchor`, not a second cursor, so `Alt+Shift+Down` moves the primary cursor to line 2 column 7 — inside `"alpha"`, not `"beta"`. The target therefore changes, and the `"beta"` overlays the test read after `render(); render();` were stale: they survive only while those two renders land within 150 ms of each other. On a loaded runner they straddle the debounce, the highlight has already moved to the word actually under the cursor, and the assertion fails. The test was asserting a pre-debounce frame rather than any behaviour the editor promises.

It now waits for the settled state — after the move, `"alpha"`'s two occurrences carry the same background the `"beta"` pair had.

## Also

`test_multiple_locales_can_be_loaded` built ten full editors — ten plugin-runtime startups in one test body, the one locale test with real exposure to nextest's slow-timeout — to check a table of ten strings. It now switches locale on a single editor through the same `init_with_config` the harness calls and re-renders; the search labels are translated by `t!` inside the status bar's render, so the path under test is unchanged and the locale-code matching (`zh-CN`, `pt-BR`) is still covered.

The Spanish "search cancelled" assertion is now a wait rather than a sample off the single `render()` after Escape, whose effect lands through the async drain (CONTRIBUTING.md Testing §3).

## Known gaps, deliberately left

- **Locale read side.** A test that built its editor *before* a locale test takes the pin keeps rendering while the pin is held, and so renders in the pinned language. The pin narrows that window — such a test now blocks at its *next* construction instead of resetting the locale mid-pin — but cannot close it. Closing it properly means giving the locale tests a test binary of their own, or taking the locale off a process global; both are larger than these flakes call for. Written down in `global_state.rs`. The indent half has no such gap: `USER_RULES` is read by config id, and no test but the pinning one has a rule under its id.
- **`plugins::audit_mode.rs:2019`** opens a review with the same toolbar-only gate the range helper had — same latent failure, left alone as it was not in the reported set and drives its own palette flow.
- **Two neighbours of the compose test.** `comments_in_fenced_code_blocks_are_not_headings` and `only_top_level_headings_are_marked_by_default` assert an *absence* after quiescence. They cannot flake — the expected set is the one viewport mark either way — but they can pass vacuously, before the pre-scan they mean to be testing has run. Fixing that needs documents whose pre-scan contribution is distinguishable from the viewport pass's, which is a test redesign rather than a flake fix.
- **Scope of cause 2 in CI.** CI runs `cargo nextest`, one process per test, so the global-clobbering half can only bite `cargo test`. Under nextest the locks are uncontended and cost nothing. For the locale trio under nextest I could not identify a race beyond the ten-harness cost and the single-frame assertion; those two are what the locale changes address there.

## Verification

`cargo fmt --check` is clean and `cargo check -p fresh-editor --test e2e_tests` passes (one pre-existing dead-code warning for `expand_filters` in `orchestrator_plugin_api.rs`, unrelated).

Cause 5 was verified both ways locally, since it changes what the test asserts:

- with the fix in place, `test_zero_width_block_selection_keeps_word_highlight` passes in 0.54 s;
- with the production fix it guards temporarily removed (the `start_col == end_col => None` case in `overlays.rs`), the rewritten wait never resolves — the run was killed at the 420 s bound, which under nextest is the 180 s kill with periodic screen dumps naming the problem. So the rewrite is a real regression guard, not a vacuous one.

The other tests were not run locally; CI is the check that matters for flakiness, and re-runs of these nine are the real signal.

https://claude.ai/code/session_01RgtaEGLhLM8Xdnj3zk1oJV


sinelaw:
CI on `cb50e91` is red on one job — **test windows-latest** — and the failure is not this PR's. Recording why, and what I did about it.

**The failure:** `e2e::occurrence_highlight::test_zero_width_block_selection_keeps_word_highlight`, 1 failed out of 8877. The other nine jobs are green, including both other test matrix legs (`test ubuntu-latest`, `test macos-latest`) and `clippy` / `fmt`.

**Why it isn't this PR's:** the same test failed on `master` earlier today — [run 33041686050](https://github.com/sinelaw/fresh/actions/runs/33041686050), on the `test macos-latest` leg, `FAIL [0.891s] (5252/9266)`. Same test, different OS, no change of mine in the picture. This PR touches `tests/common/{global_state,harness,mod}.rs` and five e2e test files; `occurrence_highlight.rs` is not among them, and the one harness change that could reach it — a read-side `RwLock` acquisition around editor construction — is uncontended under `cargo nextest`, which runs each test in its own process. It cannot change behaviour, and the test was failing before it existed.

**What I did:** re-ran the failed job once, which is the single re-run I'll spend on it. No fix for this flake exists anywhere to port into the branch, so there is nothing to bring in.

**Proposed patch, not applied.** The failing step has the same shape as three of the four causes this PR fixes — a fixed number of frames standing in for a gate:

```rust
harness
    .send_key(KeyCode::Down, KeyModifiers::ALT | KeyModifiers::SHIFT)
    .unwrap();
harness.render().unwrap();
harness.render().unwrap();

assert_eq!(bg(&harness, col(1), line2), word_bg, "...");
```

Two bare `render()` calls are a frame count, not a wait, and the first half of this very test gates properly with `wait_until`. The assertion here is an *invariance* — the background must be unchanged — so there is no condition to wait for; the honest tool is to let the pipeline settle and then assert, i.e. replace the two renders with `harness.wait_for_async_quiescence(n)`.

I am proposing rather than pushing it for two reasons. It is outside what this PR was asked to do, and — more to the point — I have not diagnosed why this fails on Windows and macOS but not Linux. Without that, "settle first" is a plausible shape rather than a verified fix, and one of the lessons in this PR's own §4 is that a quiescence wait can itself return early when a fire-and-forget round trip is outstanding. It deserves its own diagnosis and its own change.

---

sinelaw:
Correcting my comment above: I've fixed that test here rather than leaving it, and `733b6b9` is on the branch. Deflaking is the point of this PR, so a flake in reach is in scope whichever commit introduced it — declining it because it also fails on `master` was the wrong call.

The cause turned out to be the same family as the rest of this PR, not anything platform-specific. The occurrence highlight debounces its target by 150 ms of wall clock (`ReferenceHighlightOverlay::update`, `Instant::now()`) and advances that clock only inside a render. A block selection is a *single* cursor carrying a `block_anchor`, so `Alt+Shift+Down` moves the primary cursor to line 2 column 7 — inside `"alpha"`, not `"beta"` — and the debounce target changes. The `"beta"` overlays the test read after `render(); render();` were therefore stale, surviving only while those two renders landed within 150 ms of each other. Windows and macOS straddle it; Linux didn't. The test was asserting a pre-debounce frame, not any behaviour the editor promises.

It now waits for the settled state: after the move, `"alpha"`'s two occurrences carry the same background the `"beta"` pair had.

Verified both ways locally, since the change alters what the test asserts:

- with the fix in place it passes in 0.54 s;
- with the guarded production fix temporarily removed (the `start_col == end_col => None` case in `overlays.rs`), the wait never resolves — killed at a 420 s bound, which under nextest is the 180 s kill with screen dumps naming the problem.

The earlier re-run of the Windows job was cancelled when this push superseded it, so it never returned a verdict; that is moot now. PR title and description are updated — nine tests, five causes.

---

PR #3111, opened 2026-08-30
fix(editor): deno runtime detection

### LINKED ISSUE #2981: bug: LSP is switched to deno when a deno.json(c) is found whether deno is available or not
I am on a musl-based void linux for arm64, so deno is currently not available. Would it be possible to allow switching between the deno-based lsp and the node based method?

atk:
Fixes #2981

After rummaging through the code, I realized that any deno.json(c) in the project would cause fresh to switch LSP to deno whether a runtime was available or not. This code checks if a runtime is in path and only then makes the switch.

### DIFF
--- a/crates/fresh-editor/src/app/window/mod.rs
+++ b/crates/fresh-editor/src/app/window/mod.rs
@@ -55,7 +55,8 @@ use crate::view::file_tree::FileTreeView;
 use crate::view::split::{SplitManager, SplitViewState};
 use fresh_core::{BufferId, WindowId};
 use std::collections::HashMap;
-use std::path::PathBuf;
+use std::env;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 /// A project-rooted unit of editor state.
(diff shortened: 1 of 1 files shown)


sinelaw:
Heads up so this isn't duplicated work: #3159 includes the runtime check from this PR, built on @atk's diagnosis in #2981 (the `deno.json(c)` detection in `configure_lsp_servers`).

Two differences from the version here:

- The probe goes through `LspManager::command_exists_via_authority` — the same authority-routed check the LSP status popup makes — rather than walking `PATH` directly, so a container or SSH authority is asked about its own `PATH` instead of the host's, and nothing branches on backend kind.
- It also fixes the messages, which named the wrong binary: the status bar said `TypeScript LSP server 'typescript-language-server' not found` and the dialog offered `npm install -g typescript-language-server` for what was really a missing `deno` — the reason "I did not find deno anywhere in the configuration" was such a reasonable thing to conclude.

@atk, the decision on how to land this is the maintainer's, not mine: #3159 can be rebased onto your commit if you'd prefer the authorship to sit here, or this PR closed in favour of it. Either is fine by me — thanks for tracking the cause down.

---

sinelaw:
I'll merge this one and add a follow-up to move this all out of the rust core and into a plugin

sinelaw:
🙏

PR #3140, opened 2026-09-02
Dragging a divider keeps up with the pointer

sinelaw:
Pulling a vertical divider — the border between two split panes, the file explorer's right edge, or the orchestrator dock's — was sluggish and lagged a second or more behind the mouse.

## Root cause

A terminal reports pointer travel as one event per cell crossed, in two kinds: `Moved` with no button down, and `Drag(button)` with one held. Only the first was ever collapsed to its latest report — in `TtyReader::push_coalesced` (which owns stdin on unix) and in `coalesce_mouse_moves` (the crossterm path elsewhere).

So a drag delivered every intermediate cell as its own event, and each one costs a full `Editor::relayout` (every window's panes placed, every visible PTY resized) plus a full repaint. A repaint costs **more** than the event loop's 16ms frame budget, so `last_render.elapsed() >= FRAME_DURATION` was always open and the editor spent a whole frame per intermediate cell rather than per frame — each frame drawing a divider position the pointer had already left.

A second, smaller cost sat on top: a grip captures the pointer for the whole drag, so its `Move` fires for reports that travel *along* the divider too. Those recomputed the ratio (or width) already in place and still ran the full layout funnel.

## The fix

Three commits:

1. **`input:`** motion coalesces whether or not a button is held. Only a run of the *same* report collapses — `Drag(Left)` never swallows a `Drag(Right)`, and a modifier change ends the run — and presses, releases and wheel notches are untouched, because each of those means something at the moment it happened. This is also what makes the backlog structurally impossible: a burst that lands while the editor is busy painting is drained as the one report still true.
2. **`chrome:`** `handle_separator_drag` and `handle_file_explorer_border_drag` compare before they write, the way `handle_dock_resize_drag` always has, and return without reflowing when nothing moved.
3. **`input:`** the Linux console mouse goes through the same queue. Review of the first two commits found the one pointer path that still escaped coalescing: `poll_with_gpm` reads GPM's own fd and returned the converted report straight out, so it never met `push_coalesced` — and `coalesce_mouse_moves` declines on unix *precisely because* the reader owns fd 0. GPM reports now enter via `TtyReader::queue_external`, and the loop drains every report GPM has ready rather than one per wakeup (one-per-poll would hand the flood back a poll at a time and coalesce nothing). Queued before stdin is drained, so the GPM-first precedence is unchanged.

## Measured

Driven by hand in tmux with SGR mouse reports in a 200x50 pane (debug build), dragging the file explorer's edge from column 60 to 130 at a realistic hand speed (~9ms/cell, 0.63s for the pull), sampling the divider from a separate thread so the sampler never paces the drag.

**Before** — the divider crawls, and is forty columns behind the hand when the hand stops:

```
59 → 63 → 65 → 69 → 71 → 75 → 77 → 81 → 83 → 87 → 89
                                     hand stopped at 129
```
then takes a further **0.67s** to arrive.

**After** — it follows the hand, eight columns behind at the stop:

```
59 → 65 → 71 → 77 → 81 → 87 → 93 → 99 → 105 → 109 → 115 → 121
                                     hand stopped at 129
```
arriving in **0.05s**.

Both make ~11 visible updates — that is the render rate, unchanged. What differs is what each update is worth: before, ~3 columns of stale backlog per frame; after, ~6 columns of current position.

Catch-up time after the hand stops, all three surfaces:

| Divider | Before | After |
|---|---|---|
| File explorer right edge | 0.67s | 0.05s |
| Split pane divider | 0.73s | 0.07s |
| Orchestrator dock right edge | 2.35s | 0.12s |

Text-selection drag was re-checked interactively the same way and still lands on the exact cell the drag ended on.

## Tests

Each fix has a reproducer that was confirmed to fail without it and pass with it:

- `services::tty_input::tests::a_drag_flood_collapses_to_its_latest_report` — 60 drag reports in one read batch must come out as one. Without the fix it yields all 60.
- `services::tty_input::tests::coalescing_keeps_presses_releases_and_distinct_drag_runs` — a press, two left drags, a Shift+left drag, two right drags and a release collapse into exactly the five events they mean.
- `services::tty_input::tests::an_externally_queued_motion_flood_collapses_too` — ten console drags queued between two stdin keystrokes come out as key, one collapsed drag at the last column, key; pins both the collapse and the interleaving order.
- `geometry_pass::a_divider_drag_that_moves_nothing_lays_out_nothing` — counts shell layout passes rather than timing them: four reports sliding along the divider cost zero, and the one that crosses a column costs exactly one.

**One limit worth stating:** the GPM fd loop in commit 3 cannot be exercised here — it needs a Linux VT with the `gpm` daemon and libgpm. The queueing half is covered by the test above; the fd loop is verified by `cargo check --all-targets` (`poll_with_gpm` is `#[cfg(target_os = "linux")]` and does compile on this target) and by review, not by execution. Worth a maintainer's eye on a real console.

`cargo fmt --check`, `cargo clippy --all-targets` (no new warnings in the touched files) and `cargo check --all-targets` are clean; the full `e2e_tests` suite passes locally.

https://claude.ai/code/session_019nsxE7kbWSMnDD3x2LNH3g


sinelaw:
`test macos-latest` is red on 6af23c4. It is a single timeout, and it is not this PR's:

```
Summary [1400.381s] 9462 tests run: 9461 passed (4 slow), 1 timed out, 162 skipped
TIMEOUT [180.204s] fresh-editor::all_tests
```

**Why it isn't this PR's.** The test is documented as failing exactly this way on a loaded macOS runner, in its own comment:

> Twelve pages of that timed out at nextest's 180s cap on a loaded macOS runner while taking under three seconds unloaded. Six still carries the property […]

So the page count was already cut 12 → 6 for this, and 6 still exceeds the cap when the runner is loaded enough.

It is also unreachable from the diff. The change has three parts, and none of them is on this test's path:

- `TtyReader::push_coalesced` — the test harness never constructs a `TtyReader`; it injects `MouseEvent`s straight into the editor.
- `coalesce_mouse_moves` — lives in `main.rs`, i.e. the binary, not the library the tests link.
- the `handle_separator_drag` / `handle_file_explorer_border_drag` early returns — only run inside a divider drag.

`crates/fresh-editor/tests/e2e/markdown_compose_scrollbar_markers.rs` contains no mouse, drag, or divider code at all — the sweep under test is `PageDown` keystrokes.

**Corroborating runs.** `test ubuntu-latest` is green on this same commit, as are `fmt`, `clippy`, `doc`, `schema`, `check (no plugins)`, `actionlint`, `installer` and `build-musl`. Locally the full `e2e_tests` suite passes 3615/0 on this branch, this test included.

**No fix ported.** None exists to port, and the two things that would make this check green from here — trimming the page sweep further or skipping the test — would weaken an unrelated test's coverage, so I have not done either. I have re-run the job once, which is the one re-run I will spend on it; if it is red again on the same test I will say so rather than re-running.

---

sinelaw:
Correction to the timing in my previous comment: it said "I have re-run the job once", but at that moment the re-run had actually been refused — `rerun-failed-jobs` returned `403 This workflow is already running`, because `test windows-latest` was still in progress. The claim was wrong when I made it.

`test windows-latest` has since completed **green**, and the re-run of `test macos-latest` is now genuinely queued. That is the one re-run I will spend on this; if it comes back red on the same test I will report it rather than re-running again.

Everything else on this head is green: `fmt`, `clippy`, `doc`, `schema`, `check (no plugins)`, `actionlint`, `installer`, `build-musl`, `test ubuntu-latest`, `test windows-latest`.

---

crates/fresh-editor/src/main.rs
```diff
@@ -6385,16 +6385,32 @@ fn poll_with_gpm(
...
     }
 
+    // Only coalesce motion.
+    let Some(run) = motion(&event) else {
+        return Ok((event, None));
+    };
+
     // On Unix, the TtyReader owns stdin and already coalesces motion floods as
     // it queues them; draining crossterm here would race the reader on fd 0
     // (and re-introduce crossterm's fragile mouse parsing). Nothing to do.
```
sinelaw: GPM-sourced drags escape both coalescers, so the fix doesn't reach the Linux console.

`poll_with_gpm` reads the GPM fd itself and returns `Ok(Some(InputEvent::Mouse(..)))` directly (main.rs ~6360-6366) — those events never pass through `TtyReader::push_coalesced`, because they never came from stdin. Then here, `raw_input_active()` is true (the `TtyReader` is alive and owns fd 0 even in the GPM configuration), so this early-return fires and the crossterm drain is skipped too.

Concrete scenario: on a Linux VT with `gpm` running, `gpm_to_crossterm` maps a button-held motion to `MouseEventKind::Drag` (services/gpm/convert.rs:44), one event per poll iteration. Pulling a split divider there still costs a full `Editor::relayout` + repaint per cell crossed, i.e. exactly the pre-PR behaviour the PR body says is now "structurally impossible". The `handle_separator_drag` / `handle_file_explorer_border_drag` guards still help for reports that slide *along* the divider, but the cross-column backlog is unchanged.

Not a regression — `Moved` was equally exempt before — but the new docstring ("the `TtyReader` … already coalesces motion floods as it queues them") now asserts a Unix-wide invariant that this path doesn't satisfy. Either coalesce the GPM event before returning it (compare against the previous GPM report the same way `push_coalesced` compares against `queue.back()`), or narrow the comment to say it holds only for stdin-sourced motion.

---

sinelaw (commented):
Reviewed the two commits hunk by hunk (input coalescing + the two chrome guards) and verified locally: `cargo check --all-targets` clean, the three new tests pass, and the full `e2e_tests` suite is green (3614/3615 — the one failure, `test_diagnostics_api_with_fake_lsp`, passes in isolation and is unrelated LSP-timing flake).

No correctness defects found in the diff. Specifically checked and satisfied:

- **Coalescing never merges two intents.** `push_coalesced` gates on `last.kind == m.kind && last.modifiers == m.modifiers` *inside* the `Moved | Drag(_)` arm, so `last` is necessarily motion too; presses, releases, wheel notches and a `Moved`/`Drag` boundary all break the run. `coalesce_mouse_moves` mirrors it by comparing every drained event against the run's `(kind, modifiers)` tuple.
- **Nothing downstream accumulates per-event deltas**, which is what would break under dropped intermediate reports. `handle_separator_drag` (start_ratio + delta-from-start), `handle_file_explorer_border_drag` (start_width + delta-from-start), `handle_dock_resize_drag` (absolute column), `handle_scrollbar_drag_relative` (drag_start_row/top_byte), `TabDragState::is_dragging` (start vs current) are all absolute-from-drag-start, so the last report of a run carries the full truth.
- **The two new equality guards can't wedge a drag or drop a needed reflow.** Both `SplitManager::set_ratio` and `Window::set_grouped_split_ratio` store `new_ratio.clamp(0.0, 1.0)` and `new_ratio` is already clamped to that range, so the `f32` `==` is exact rather than approximate; the min-pane-size guard is layout-time only and never writes back, so `current` can't drift from what was stored. Because `new_ratio` is recomputed from `drag_start_ratio` each time, a drag that clamps at an edge and comes back still differs from the stored value on the very next column. `ExplorerWidth` is `PartialEq` and compared variant-preservingly. The grip's `Move` still returns `claimed`, so a skipped drag repaints (just doesn't relayout) — the geometry test asserts exactly that.
- `Grip::Separator`'s release relayouts unconditionally, so the final position is never left to a skipped report.

One note left inline about a path the fix doesn't reach.

---

crates/fresh-editor/src/main.rs
sinelaw: Confirmed by reading the path, and fixed in 325be43 — I took the first of the two options rather than narrowing the comment, since console users have the same drag to do.

GPM reports now go in through `TtyReader::queue_external`, which is `push_coalesced` under another name, so console motion collapses on the same rule as terminal motion. Two things the fix needed beyond the obvious swap:

- **Drain the loop, not one report per wakeup.** Reading a single event per poll would hand the flood back a poll at a time and coalesce nothing, so the queue only helps if every ready report goes in. `read_event` blocks when nothing is pending (it's `Gpm_GetEvent`), so the fd is re-polled with a zero timeout before each further read.
- **Queueing instead of returning changes who serves the event.** The tail is now a single `reader.next_buffered().or_else(take_resize)`. GPM reports are queued *before* stdin is drained, so the GPM-first precedence this function has always had is preserved.

`an_externally_queued_motion_flood_collapses_too` covers the queueing half: ten console drags between two stdin keystrokes come out as key, one collapsed drag at the last column, key — so both the collapse and the interleaving order are pinned. The GPM fd loop itself I can't exercise here (it needs a Linux VT with the `gpm` daemon and libgpm), so that half is verified by `cargo check --all-targets` — `poll_with_gpm` is `#[cfg(target_os = "linux")]` and does compile on this target — and by review rather than by execution. Worth a maintainer's eye on a real console.

---

sinelaw:
Update on the macOS timeout, because on the new head (`325be43`) it is a **different test** than the one my earlier comment named — which makes this a pattern worth reporting rather than a repeat.

```
Summary [1413.747s] 9467 tests run: 9466 passed (4 slow), 1 timed out, 162 skipped
TIMEOUT [180.149s] e2e::markdown_compose_scrollbar_markers::every_heading_is_marked_without_scrolling
```

Previous head `6af23c4` timed out on `heading_marks_survive_exploring_the_document`; this head times out on `every_heading_is_marked_without_scrolling`. Two different tests, same file, two consecutive macOS runs. Both are compose-mode tests that block on the async plugin pre-scan via an intentionally unbounded `wait_until` (per CONTRIBUTING, the external nextest cap is the bound), over a 600-line generated document. That reads less like a flaky test and more like `markdown_compose_scrollbar_markers` as a whole sitting close to the 180s cap on macOS runners, with load deciding which member trips first. Worth a look independently of this PR — I have not touched it, since trimming or skipping it would weaken unrelated coverage.

**Not this PR's, and commit `325be43` in particular cannot reach macOS at all:**

- `poll_with_gpm` — the only function that commit changes behaviourally — is `#[cfg(target_os = "linux")]` (main.rs:6306). macOS builds the `#[cfg(not(any(target_os = "linux", windows)))]` event loop instead (main.rs:6027).
- `TtyReader::queue_external`, the method it adds, has exactly one production caller: inside that Linux-gated function. On macOS nothing calls it.
- The earlier two commits are unreachable from this test for the reasons already given: the harness never constructs a `TtyReader`, `coalesce_mouse_moves` is binary-only, the drag guards only run inside a divider drag — and this test performs no mouse input whatsoever.

`test ubuntu-latest` is green on this head, as are `fmt`, `clippy`, `doc`, `schema`, `check (no plugins)`, `actionlint`, `installer` and `build-musl`; `test windows-latest` is still running. The previous head went green on macOS after one re-run, with this test passing. I will re-run this one once Windows finishes and the workflow stops reporting itself as still running.

---

PR #3145, opened 2026-09-02
fix ssh bootstrap when target is windows

mommysgoodpuppy:
fresh ssh connection was hanging on windows. `plain sys.stdin.read` behaves differently per platform which was causing the bootstrap to hang.

the added test timeouts nextest without the fix


sinelaw:
`transport.rs:102` has the same one-liner with the same bug (the kubectl carrier) — please fix it there and call it from `configure_agent_carrier_ssh` instead of adding a second copy, and use `.decode()` rather than `.decode('utf-8')` (same default) so the no-single-quote assertion at `transport.rs:412` still holds.

---

sinelaw:
Thanks!

PR #3147, opened 2026-09-02
fix(welcome): a startup surface — opens once, as a tab, and touches nothing else

sinelaw:
The welcome screen decided when to appear from "is the workspace empty?", asked at two moments. Both readings were wrong, and every attempt to refine them added a concern the plugin had no business owning. Fixing the flicker and the stale layout underneath also turned up four host bugs — the last two in the renderer, and general.

The plugin now has **one concern**: `showOnStartup`. When Fresh launches, the page opens as a tab behind whatever is already open. It never takes the foreground on its own, never counts the other buffers, never reads `auto_create_empty_buffer_on_last_buffer_close`, never closes any buffer but its own on the reader's request, and never closes itself because a file opened or reopens because one closed. `Welcome` in the palette is the one path that brings it to the front.

## Behaviour

| | before | now |
|---|---|---|
| `fresh` in an empty dir | page opens, `[No Name]` seed closed | `[No Name]` keeps the pane and focus; `Welcome ×` beside it |
| same, empty-buffer setting **off** | page opens, seed closed | identical — the setting is the host's, not the plugin's |
| `fresh note.txt` / restored session | never appears again until you close everything and relaunch | `Welcome` tab beside the work; `note.txt` stays in front, never displaced |
| close the last buffer | page reopens over the empty pane | empty pane, as asked |
| open a file with the page open or behind | page silently deleted unless "engaged" | tab stays — the reader closes tabs |
| `showOnStartup: false` | — | nothing opens; palette still works |
| resize while the page is a background tab, then switch to it | stale layout until the next keystroke | correct layout, no input needed |

## Host changes

All four found by driving the real binary under tmux.

- **`widget_panel_width` sized a panel from its neighbour's compose width.** It read `vs.compose_width`, and `SplitViewState` derefs to whichever buffer is *active* in that split — so a panel that is not on screen was sized from another tab's answer. It now reads the panel's own buffer state.

- **`createVirtualBuffer` could not open a tab quietly.** Creation made the new buffer active unconditionally. `background: true` adds it to the tab bar and skips the one `set_active_buffer` call — no open-then-switch-back flicker. Also new: `highlightCurrentLine`, for a buffer whose rows are laid out by widgets and whose caret row means nothing.

- **The tab-switch `SlideIn` froze the incoming pane at its first frame.** It took its snapshot once, on its first apply, and shifted that for 260 ms; anything that changed under the slide (a panel catching up to its geometry ~100 ms in) was painted every frame and covered every frame. The snapshot is now retaken from the freshly painted frame on every apply.

- **No frame after the last effect retired.** The frame an effect finishes on is its own composite; the runner retired it, `is_active()` went false, and nothing asked for the frame that paints the true content. The runner now owes one settle frame (`take_settle_frame`), honoured by both frame loops (direct and server).

The design note (`docs/internal/welcome-screen-design.md`, §8 and findings 32–34) records the wrong turns — the seed rules, the step-aside rules, an idle-loop theory and a wrap-index theory — and how each was ruled out.

## Plugin changes

- `openWelcome(force)`: `background: !force`; nothing else about the workspace is consulted. `engaged`, `shown`, `seedWanted`, `isSeed`, `hasOtherBuffers` and the `after_file_open` handler are gone.
- `welcomeOnBufferClosed` only drops this plugin's handle on its own buffer.
- A background tab hears nothing about geometry (`viewport_changed` is per split, against the active buffer), so the page catches up on `buffer_activated`, one tick later, with a `flush` between the compose hint and the repaint.
- `highlightCurrentLine: false`.

## Tests

- `tests/e2e/plugins/welcome_screen.rs` — 12. The seed survives under either setting, the page survives a file open whether or not the reader has turned to it (both red against the previous plugin), `showOnStartup: false` opens nothing until asked, startup never changes the active buffer.
- `view::animation::tests` — `slide_in_retakes_the_incoming_snapshot_every_apply` and `a_finished_effect_owes_exactly_one_settle_frame`.
- `tests/e2e/virtual_buffer_repaint.rs` — the slide settles on the pane's current content, and a control-and-treatment test for `highlightCurrentLine`.

https://claude.ai/code/session_01DkWt9oBvQfnVJFsgBFooip

### DIFF
--- a/crates/fresh-editor/plugins/welcome_screen.ts
+++ b/crates/fresh-editor/plugins/welcome_screen.ts
@@ -1476,7 +1476,8 @@ function hasOtherBuffers(): boolean {
  *  command — which both overrides `showOnStartup` and brings the page to
  *  the front. Startup passes `false` and lets `hasOtherBuffers` decide
  *  the second half: an empty workspace lands on the page, a workspace
- *  with work in it gets a tab and keeps looking at the work. */
+ *  with work in it gets a tab and keeps looking at the work — the page
+ *  is never made active and then unmade. */
 async function openWelcome(force: boolean): Promise<void> {
   if (bufferId !== null) {
     editor.showBuffer(bufferId);
@@ -1485,11 +1486,6 @@ async function openWelcome(force: boolean): Promise<void> {
   if (opening) return;
   if (!force && !showOnStartup()) return;
   const foreground = force || !hasOtherBuffers();
-
-  // `createVirtualBuffer` tabs the new buffer into the focused pane and
-  // makes it active — there is no "open it quietly" option — so a
-  // background open is an open followed by putting the reader back.
-  const restoreTo = foreground ? 0 : editor.getActiveBufferId();
   opening = true;
   readActiveTheme();
   try {
@@ -1510,6 +1506,12 @@ async function openWelcome(force: boolean): Promise<void> {
       // lit its occurrences and the reader could see no reason why.
       showCursors: true,
       editingDisabled: true,
+      // Never take the view from a workspace that already has something
+      // in it. The page arrives as a tab, in one step: opening and then
+      // switching back is two visible switches, and the reader watches
+      // their file get displaced and returned for no reason they asked
+      // for.
+      background: !foreground,
     });
     bufferId = res.bufferId;
     // Nothing focused is this page's resting state: it opens as
@@ -1539,11 +1541,7 @@ async function openWelcome(force: boolean): Promise<void> {
     probeThemes();
     probeWorkspaces();
     render();
-    if (foreground) {
-      editor.showBuffer(bufferId);
-    } else if (restoreTo !== 0 && restoreTo !== bufferId) {
-      editor.showBuffer(restoreTo);
-    }
+    if (foreground) editor.showBuffer(bufferId);
     void probeRepoFiles();
     void probeGit();
   } catch (e) {
--- a/crates/fresh-editor/plugins/welcome_screen.ts
+++ b/crates/fresh-editor/plugins/welcome_screen.ts
@@ -22,8 +22,10 @@ const editor = getEditor();
 // ═════════════════════════════════════════════════════════════════════
 //   WELCOME SCREEN
 //
-//   The empty-workspace surface: a scrollable buffer that onboards
-//   three audiences on one page without overwhelming the simplest.
+//   The startup surface: a scrollable buffer that onboards three
+//   audiences on one page without overwhelming the simplest. It opens
+//   when Fresh launches with nothing to restore, and otherwise only on
+//   request — closing buffers never summons it.
 //   Design note: docs/internal/welcome-screen-design.md.
 //
 //   Structure is a ladder. The first viewport is a zero-anxiety zone
@@ -150,18 +152,6 @@ let opening = false;
  *  auto-opened screen nobody touched steps aside when a real file
  *  opens; one the reader engaged with is theirs to close. */
 let engaged = false;
-/** Set when the reader closes the page — by Escape, by the tab's `×`,
- *  by `Ctrl+W`. It answers one question only: *this* emptying of the
- *  workspace has been answered. Closing the screen must never be undone
- *  by the very `buffer_closed` event the close itself produced.
- *
- *  It is cleared the moment the workspace stops being empty, because
- *  the next time it empties is a new question. It used to hold for the
- *  rest of the session, which read as the screen appearing at random:
- *  close it once and it never came back, however many times you emptied
- *  the workspace afterwards. */
-let dismissed = false;
-
 const folded = new Set<string>();
 
 let finderQuery = "";
@@ -1445,7 +1435,7 @@ function probeWorkspaces(): void {
 
 editor.defineConfigBoolean("showOnStartup", {
   default: true,
-  description: "Open the welcome screen when Fresh starts with nothing to restore, and after the last buffer is closed.",
+  description: "Open the welcome screen when Fresh starts with nothing to restore.",
 });
 
 /** The footer toggle writes plugin global state, which persists across
@@ -1463,10 +1453,9 @@ function showOnStartup(): boolean {
 
 /** Is anything at all open besides this page?
  *
- *  Anything: a file, a terminal, an agent session, another plugin's
- *  panel. This is the *empty workspace* screen, and a workspace with a
- *  shell running in it is not empty — closing the last text buffer
- *  while a terminal is still open used to pop the page up over it.
+ *  Asked once, at startup: the page stands in for an empty workspace, so
+ *  a restored session with anything in it — a file, a terminal, an agent
+ *  session, another plugin's panel — is not one to stand in for.
  *
  *  Two buffers do not count: this page itself, and the host's empty
  *  untitled seed, which is the very thing the page stands in for. */
@@ -1485,8 +1474,7 @@ async function openWelcome(force: boolean): Promise<void> {
     return;
   }
   if (opening) return;
-  if (force) dismissed = false;
-  if (!force && (dismissed || !showOnStartup())) return;
+  if (!force && !showOnStartup()) return;
   opening = true;
   readActiveTheme();
   try {
@@ -1549,13 +1537,9 @@ async function openWelcome(force: boolean): Promise<void> {
   opening = false;
 }
 
-/** `dismiss` distinguishes the reader closing the page (stay away)
- *  from the page stepping aside for a file it had nothing to do with
- *  (stay available). */
-function closeWelcome(dismiss: boolean): void {
+function closeWelcome(): void {
   if (bufferId === null) return;
   const id = bufferId;
-  if (dismiss) dismissed = true;
   panel?.unmount();
   panel = null;
   bufferId = null;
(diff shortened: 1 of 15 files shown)


crates/fresh-editor/plugins/welcome_screen.ts
```diff
@@ -1479,14 +1472,24 @@ function hasOtherBuffers(): boolean {
...
+/** `force` is the reader asking for the page by name — the `Welcome`
+ *  command — which both overrides `showOnStartup` and brings the page to
+ *  the front. Startup passes `false` and lets `hasOtherBuffers` decide
+ *  the second half: an empty workspace lands on the page, a workspace
+ *  with work in it gets a tab and keeps looking at the work. */
 async function openWelcome(force: boolean): Promise<void> {
   if (bufferId !== null) {
     editor.showBuffer(bufferId);
     return;
   }
```
sinelaw, line 1464: **`force` no longer implies "engaged" once the startup tab exists.**

This early return brings the page forward but never touches `engaged` — `engaged = force` only runs on the creation path below. Before this PR that was mostly harmless, because the page could only already be open if it had just auto-opened on an empty workspace. Now startup *always* creates the buffer, so this is the path the `Welcome` command / **Help ▸ Welcome** takes for the rest of the session.

Concrete: `fresh note.txt` → Welcome opens as a background tab with `engaged === false`. The reader runs `Welcome` from the palette, the page comes to the front, they read it, then press `Ctrl+P` and open a file. `welcomeOnAfterFileOpen` sees `engaged === false` and calls `closeWelcome()` — the page the reader asked for *by name* is destroyed as if nobody had touched it.

```suggestion
  if (bufferId !== null) {
    // Asking for the page by name is engagement even when it is already
    // open — otherwise the startup tab stays "ambient" for the session
    // and the next file open closes what the reader just summoned.
    if (force) engaged = true;
    editor.showBuffer(bufferId);
    return;
  }
```


---

crates/fresh-editor/plugins/welcome_screen.ts
```diff
@@ -1572,33 +1571,29 @@ editor.registerCommand(
...
 registerHandler("welcomeOnAfterFileOpen", (_e: { buffer_id: number; path: string }) => {
-  // The workspace is not empty any more, so a previous "I closed it"
-  // no longer answers anything.
-  dismissed = false;
   // An auto-opened screen nobody touched was ambient — step aside. One
   // the reader engaged with is a document they are reading; leave it.
   if (bufferId === null || engaged) return;
-  closeWelcome(false);
+  closeWelcome();
 });
```
sinelaw: **The new background tab is destroyed by the reader's first file open.**

`engaged` is only ever set true by a keystroke or click *on the page*; the background startup open sets `engaged = force` → `false` (line 1537). So the "tab beside the work" this PR introduces survives exactly until the reader opens anything:

`fresh note.txt` → ready fires → Welcome tab created behind `note.txt`. Reader presses `Ctrl+P` and opens `src/main.rs` → `after_file_open` → `bufferId !== null && !engaged` → `closeWelcome()` → the Welcome tab silently disappears, without ever having been on screen. The only way back is the palette, and that hits the early-return path above which also leaves `engaged === false`, so it evaporates again on the next open.

The "step aside" rule was written for a page *occupying the foreground* — "an auto-opened screen nobody touched steps aside when a real file opens". A background tab has nothing to step aside from; closing it is not politeness, it is deleting a tab the reader did not ask to close. This wants to be conditional on the page actually holding the foreground (e.g. record `foreground` from `openWelcome` and only step aside when it is true, or treat "was ever shown" as engagement).

---

crates/fresh-editor/plugins/welcome_screen.ts
```diff
@@ -1479,14 +1472,24 @@ function hasOtherBuffers(): boolean {
...
   if (opening) return;
-  if (force) dismissed = false;
-  if (!force && (dismissed || !showOnStartup())) return;
+  if (!force && !showOnStartup()) return;
+  const foreground = force || !hasOtherBuffers();
+
+  // `createVirtualBuffer` tabs the new buffer into the focused pane and
+  // makes it active — there is no "open it quietly" option — so a
+  // background open is an open followed by putting the reader back.
+  const restoreTo = foreground ? 0 : editor.getActiveBufferId();
```
sinelaw: **A backgrounded page never hears about a resize, so its own spec goes stale — the host-side fix only covers half of it.**

`viewport_changed` is fired only for the split's *active* buffer: `render.rs` resolves the hook's `buffer_id` with `get_buffer_id(split_id)`, i.e. whatever tab is on screen. The plugin registers no `buffer_activated`-style handler, so for as long as the page sits behind a file `welcomeOnViewportChanged` never runs — `paneWidth` stays `0`, `layoutKey()` is never recomputed, and neither `applyComposeWidth()` nor `render()` is re-issued. (The design note's finding 32 says exactly this: "`viewport_changed` never fires for this buffer at all.")

Scenario: `fresh note.txt` in a 120-column terminal → Welcome opens behind it and hints `composeWidth = measure() + 1 ≈ 87`. The reader resizes the terminal to 80 columns. `relayout()` does re-render every registered panel, but it re-renders the *stored* spec against `widget_panel_width(welcome)`, which now returns `(Some(87), _)` → `86` for an 80-column pane. Every row is laid out 6 columns wider than the pane and wraps. Switching to the tab does not rescue it either: the split's `(top_byte, width, height)` tuple after the switch usually equals the one recorded for `note.txt` (both at `top_byte 0`, same pane size), so `changed` is false and no hook fires — the same "the stale key it records while hidden equals the key it computes when shown" the new e2e test's comment describes.

The `widget_panel_width` change fixes which `compose_width` is read; it can't fix a `compose_width` that was computed for a terminal that no longer exists. Worth either dropping the hint while the page is backgrounded, or having the host deliver `viewport_changed` (or an activation hook) to non-active tabs that own a widget panel.

---

docs/internal/welcome-screen-design.md
```diff
@@ -506,9 +536,9 @@ An ambient screen you ignored was ambient; a document you started reading is
 yours. This also means `fresh src/main.rs` never leaves a Welcome tab behind,
```
sinelaw: This sentence is now false, and contradicts the section this PR added ~40 lines above ("At startup, always — … as a tab beside the work when it produced anything") plus the new `startup_with_a_file_already_open_still_gets_a_welcome_tab` test.

CLI file arguments are drained *before* `fire_ready_hook()` (`main.rs`, `process_pending_file_opens()`), so with `fresh src/main.rs` the `after_file_open` for `main.rs` arrives while `bufferId === null` and is ignored — `ready` then opens the page as a background tab that is deliberately left behind. Leaving this claim in place is the one line a future reader would use to justify re-adding the ambient path.

---

sinelaw (commented):
Reviewed the diff for correctness. The core change is right: `welcomeOnBufferClosed` no longer reopening on an emptied workspace, and the `dismissed` flag going with it, are both clean — no remaining caller resets or reads it, and no path other than `ready` and the `Welcome` command reaches `openWelcome` any more. The `widget_runtime.rs` fix (`vs.buffer_state(buffer_id).and_then(|b| b.compose_width)` instead of the `SplitViewState` deref) is correct: `handle_set_layout_hints` writes the hint into that buffer's `keyed_states` entry, `viewport.width` is still the pane (`sync_viewport_to_content` resizes it to `content_rect`, keeping `compose_width` separate), and the `find` predicate already guarantees the `buffer_state` lookup is `Some`.

Four findings, all in the "startup now always opens a tab" behaviour that came in alongside the stated fix:

1. `openWelcome`'s early return doesn't set `engaged` for `force`, so the `Welcome` command on the startup tab leaves the page ambient.
2. The background startup tab is closed by the reader's first `after_file_open` — the tab the PR adds lasts until the next `Ctrl+P`.
3. A backgrounded page never receives `viewport_changed`, so a terminal resize leaves its `composeWidth` hint and spec sized for the old pane; the host-side width fix can't compensate for that.
4. A stale doc claim that now contradicts the PR's own new test.

Also worth flagging for the merge decision: the PR description says "Unchanged: the startup path", but the startup path is in fact substantially changed here (unconditional open, foreground/background split, the `[No Name]` reaping removed, plus the Rust width fix). Nothing wrong with the changes being here, but the description should describe them, and the e2e tests still haven't been run locally per the description.

---

crates/fresh-editor/plugins/welcome_screen.ts
sinelaw: Confirmed and fixed in ce455b069 — `if (force) engaged = true;` in the already-open path, as suggested.

The scenario is exactly right and it is now the common one: startup always creates the buffer, so every `Welcome` from the palette after the first takes this path. Covered by `asking_for_the_page_by_name_keeps_it_through_the_next_file_open`.

---

crates/fresh-editor/plugins/welcome_screen.ts
sinelaw: Confirmed and fixed in ce455b069. This one mattered most — the tab this PR exists to add was deleted by the reader's first `Ctrl+P`, before they had ever seen it.

Took the "was ever shown" route: a `shown` flag, set from `foreground` at open and set true on `buffer_activated` for this buffer. `welcomeOnAfterFileOpen` now requires `shown` as well as `!engaged`. Your framing is the one I put in the design note: "step aside" was written for a page occupying the pane, and a tab the reader has never seen has nothing to step aside from.

Covered by `a_welcome_tab_the_reader_has_not_seen_survives_their_first_file_open`.

---

crates/fresh-editor/plugins/welcome_screen.ts
sinelaw: Reproduced and partly fixed in ce455b069. Driving the real binary: `fresh note.txt` at 140 columns, resize the terminal to 70 while Welcome is the background tab, then switch to it — the wordmark wraps and the startup toggle spreads over four lines, exactly as described. Your reading of `previous_viewports` is right, and I confirmed the switch fires no hook.

Fixed the layout: the page catches up on `buffer_activated`. Two details that cost me a few builds and are now in the code comments —

- The repaint has to be deferred one tick. `getViewport()` *at* activation still reports the viewport this buffer had when it was last on screen (139), and only the following frame corrects it to 69. Reading it at activation re-applies the stale hint.
- `await editor.flush()` between `applyComposeWidth()` and `render()` is load bearing: `widget_panel_width` reads `compose_width` while it processes the update, so a repaint issued in the same breath as the hint is laid out against the previous one. Host trace confirms `raw=(Some(68), 69)` after the flush versus `(Some(89), 70)` before.

**Residual, and I am leaving it open deliberately.** The corrected content reaches the buffer but not the screen until the reader's next input — an idle editor draws no frame. Switching to the tab shows the old layout until any keystroke, then it is right. I tried `invalidate_layouts_for_buffer` on the virtual-content write (which is what the other plugin content path does) and it does not help: it marks the layout stale without asking for a draw. Waking a frame from the repaint path is a main-loop change, not a welcome-screen one, and widening this PR into the renderer is how the last two rounds went wrong. Recorded as finding 33 in `docs/internal/welcome-screen-design.md`, including your suggestion that the host deliver `viewport_changed` (or an activation hook) to non-active tabs owning a panel — I think that is the real fix and it wants its own change.

---

docs/internal/welcome-screen-design.md
sinelaw: Right, and fixed in ce455b069. The sentence was true of the old behaviour and is now the opposite of what the PR does.

Rewrote the ruling to name the third case explicitly — a page that has never been on screen is not in the way of anything and stays — and replaced the CLI claim with why `fresh src/main.rs` *does* leave a Welcome tab behind now, including the `process_pending_file_opens` ordering you point at. Also recorded that summoning the page by name counts as engagement, which is the sibling finding above.

---

crates/fresh-editor/plugins/welcome_screen.ts
sinelaw: Closing the loop on the residual — it is fixed at the root in 0ea45d72d, and it was not what I said it was.

Probing one frame end to end: the buffer held the new text, the content pass painted it correctly (toggle at column 36), and `animations.apply_all`, a few lines later in `Editor::render`, painted the old cells (57) back over the pane. `Ctrl+PageDown` starts a 260 ms `SlideIn`, which (a) snapshotted the incoming pane once on its first apply — before the catch-up you describe here had landed — and shifted that snapshot for the whole slide, and (b) on finishing left its own composite as the last frame, with `is_active()` already false so no frame followed. Both fixed in `view/animation.rs`: the snapshot is retaken from the freshly painted frame every apply, and the runner owes one settle frame after an effect retires, honoured by both frame loops.

Your catch-up fix from this thread stands; it was doing its job all along — the slide was covering it. The regression test is red without the `SlideIn` change and green with it. Design note findings 33–34 have the full trail, including the wrap-index detour that turned out to be wrong.

---

PR #3152, opened 2026-09-02
fix(terminal): stop losing the scrollback just above the top of the screen (#3151)

### LINKED ISSUE #3151: Terminal scroll-back permanently loses the lines just above the top of the screen
## Symptoms

With the built-in terminal, entering scroll-back and scrolling up shows a **gap**: a contiguous block of lines that were on screen moments earlier — the ones that had *just* scrolled off the top — is simply not there. Scrolling up jumps straight from an older line to a much newer one.

The loss is **permanent**. The lines are gone from the terminal's backing file (`~/.local/share/fresh/terminals/<workdir>/fresh-terminal-N.txt`), so they never come back, are missing from a restored session, and cannot be recovered by scrolling, resizing, or re-entering scroll-back.

There are two ways to hit it.

## Scenario A — output arrives while the scroll-back view is open

Most visible with programs that print in bursts with pauses between them (build tools, test runners, `claude`-style TUIs, any loop with a `sleep`), because with `terminal.jump_to_end_on_output = true` (the default) every burst that arrives while you are in scroll-back snaps the view back to live — and each of those transitions eats a block of history.

1. Open a terminal in Fresh (`Alt+`` ` `` for the dock terminal).
2. Run a command that prints in bursts with pauses:

   ```sh
   n=0; for b in $(seq 1 12); do for i in $(seq 1 40); do n=$((n+1)); echo LINE_$n; done; sleep 1.2; done
   ```

3. While it runs, press `Ctrl+Space` about once per second (~8 times) — i.e. keep dropping into scroll-back; the next burst snaps you back to live each time.
4. When the command finishes, press `Ctrl+Space` and scroll up through the scroll-back.

**Expected:** every `LINE_1` … `LINE_480` is reachable, in order.

**Actual:** whole blocks are missing. Checking the backing file directly:

```sh
$ grep -c LINE_ ~/.local/share/fresh/terminals/*/fresh-terminal-0.txt
396          # 480 were printed
```

Missing on one run:

```
286–312, 329–352, 374–392, 430–432, 462–472
```

Each missing block ends exactly where the live screen began at the moment the view snapped back to live.

Setting `terminal.jump_to_end_on_output = false` and pressing `Ctrl+Space` manually to leave scroll-back while output is flowing loses a block the same way.

## Scenario B — no scroll-back visit at all

Resuming a terminal that is **already live** loses the most recently scrolled-off lines, with the scroll-back view never opened.

1. Open a terminal.
2. `for i in $(seq 1 60); do echo L_$i; done`
3. Run **Focus Terminal** from the command palette (`Ctrl+P`). Switching away to another tab and back does the same thing.
4. Press `Ctrl+Space` and scroll up.

**Expected:** `L_1` … `L_60` all reachable.

**Actual:** the batch that most recently scrolled off is gone.

## Notes

- A terminal that is fully **idle** across a scroll-back visit loses nothing.
- Pure live output, with neither a scroll-back visit nor a terminal-focus action, is complete.

## Environment

- `fresh` @ `cf626fc` (master), debug build
- Linux, `TERM=xterm-256color`, reproduced under tmux

sinelaw:
Fixes #3151.

## Symptom

Terminal scroll-back opened on a **gap**: a block of lines that had just scrolled off the top of the screen was missing — and missing permanently. Gone from the backing file, absent from a restored session, unrecoverable by scrolling, resizing, or re-entering scroll-back. Programs that print in bursts (build tools, test runners, agent TUIs) lost a block on every burst.

Measured on the repro in the issue: **84 of 480 printed lines unreachable**, in five contiguous runs. With this branch: 0.

## Root cause

The backing file is `[streamed scrollback][temporary visible-screen tail]`, with `backing_file_history_end` as the boundary; returning to live mode truncates back to it. Three things kept that boundary from being the true end of the scrollback.

1. **The read loop recorded the boundary before flushing its own `BufWriter`.** `metadata()` reports what is on disk, so every recorded boundary was short by exactly the batch just written. `enter_terminal_mode` truncates unconditionally and is reached with the split *already live* (Focus Terminal, a tab switch, resuming after a copy), so this lost scrollback with no scroll-back view ever opened — scenario B in the issue.

2. **The read loop kept appending scrolled-off lines past a tail still in the file.** The truncation that ends the visit cuts at the scrollback end and took them along. Scenario A, which `terminal.jump_to_end_on_output` (on by default) triggers on every burst.

3. **Not every transcript handle was `O_APPEND`** (found in review). Only `BackingMode::Continue` appended; every terminal from `Open Terminal` is `BackingMode::Fresh`, opened `write(true).truncate(true)`. So the read loop wrote at an offset of its own while the UI thread appended through a separate handle — and once a scroll-back sync had written anything, the loop's next batch landed *on top of* it. Since `sync_terminal_to_buffer` and `sync_terminal_backing_files` both call `flush_new_scrollback` on their own handle, the bytes at risk included real scrollback already counted in `synced_history_lines` and therefore never re-emitted: the same permanent loss, by a third route. It also made the boundary meaningless whenever a tail was present, and made the two backing modes behave differently (CONTRIBUTING #13).

## The fix

All three are the same defect — writers disagreeing about where the scrollback ends — so the fix is to the boundary, not to each truncation:

- **Every transcript handle is `O_APPEND`.** No writer can land on another's bytes, and the file's length *is* the read loop's write position. `Fresh` still starts empty, via a separate immediately-dropped truncating handle (truncate and append cannot be combined in one open).
- **Flush before measuring**, so the recorded boundary includes the batch just written.
- **Track whether a tail is actually there** (`TerminalState::backing_file_has_tail`, set *before* the write so a part-written tail stays truncatable). Truncation happens only when there is a tail to remove, and each path that writes a tail removes the previous one first instead of stacking onto it.
- **When the read loop streams scrollback past a tail, it adopts that tail as scrollback.** It cannot truncate — a scroll-back view may be reading the file — so the alternative is deleting those lines for good. The cost is one duplicated screen per overlapping event, permanently, and is now documented on the field, at the branch, and in the module's architecture doc rather than implied.

## Tests

Two end-to-end tests, one per user-visible trigger. Both drive keys and assert only on rendered output — they page through the whole scroll-back view and require every printed line to appear.

- `test_scrollback_survives_output_during_a_scrollback_visit` — a split renders live output so the second batch can be waited for without a timer while the focused pane sits in scroll-back. Fails on `cf626fc` with 36 lines unreachable (10/10 runs, never 0).
- `test_focusing_an_already_live_terminal_keeps_its_scrollback` — no scroll-back visit at all. Fails on `cf626fc` with 36 lines unreachable.

Three unit tests pin the handle semantics. `transcript_handles_append_in_both_backing_modes` fails on the old open mode with `stream-1\nstream-2\n` instead of `stream-1\nui-tail\nstream-2\n` — the other writer's append overwritten in place.

Existing coverage is unchanged: 92 `services::terminal` unit tests and 103 terminal e2e tests green.

Also verified manually under tmux against the issue's exact steps: 84 lines lost before, 0 after (with 64 duplicated line occurrences across 8 overlapping visits — the documented adoption cost).

## Review

Reviewed adversarially; seven findings, six fixed in eb75900, one declined with reasoning and filed as #3153.

https://claude.ai/code/session_01FBzQhDe6BRUaBfxqst9qBf

### DIFF
--- a/crates/fresh-editor/src/app/terminal.rs
+++ b/crates/fresh-editor/src/app/terminal.rs
@@ -2039,14 +2039,25 @@ impl Editor {
                     .get(&terminal_id)
                 {
                     if let Some(handle) = self.active_window().terminal_manager.get(terminal_id) {
-                        if let Ok(state) = handle.state.lock() {
-                            let truncate_pos = state.backing_file_history_end();
-                            // Always truncate to remove appended visible screen
-                            // (even if truncate_pos is 0, meaning no scrollback yet)
-                            if let Err(e) =
-                                terminal_backing_fs().set_file_length(backing_path, truncate_pos)
-                            {
-                                tracing::warn!("Failed to truncate terminal backing file: {}", e);
+                        if let Ok(mut state) = handle.state.lock() {
+                            // Truncate only when a visible-screen tail is
+                            // actually there. Truncating unconditionally used
+                            // to be harmless because the tail was assumed to
+                            // be the only thing past the history end — but the
+                            // PTY read loop appends there too, and cutting at
+                            // a history end that predates its writes deleted
+                            // live scrollback for good (fresh#3151).
+                            if state.backing_file_has_tail() {
+                                let truncate_pos = state.backing_file_history_end();
+                                match terminal_backing_fs()
+                                    .set_file_length(backing_path, truncate_pos)
+                                {
+                                    Ok(()) => state.set_backing_file_has_tail(false),
+                                    Err(e) => tracing::warn!(
+                                        "Failed to truncate terminal backing file: {}",
+                                        e
+                                    ),
+                                }
                             }
                         }
                     }
@@ -2415,10 +2426,28 @@ impl Window {
         let mut grid_cols: Option<usize> = None;
         if let Some(handle) = self.terminal_manager.get(terminal_id) {
             if let Ok(mut state) = handle.state.lock() {
-                use std::io::BufWriter;
+                use std::io::{BufWriter, Write};
 
                 let (cols, _) = state.size();
                 grid_cols = Some(cols as usize);
+
+                // An earlier visit (or a session checkpoint) may have left its
+                // visible-screen tail in the file. It is rewritten below, so
+                // remove it first: leaving it in place would bake a stale
+                // screen into the scrollback prefix, and — since the tail is
+                // what the truncation on the way back to live mode cuts at —
+                // would also leave real scrollback stranded past that point
+                // (fresh#3151).
+                if state.backing_file_has_tail() {
+                    let history_end = state.backing_file_history_end();
+                    match terminal_backing_fs().set_file_length(&backing_file, history_end) {
+                        Ok(()) => state.set_backing_file_has_tail(false),
+                        Err(e) => tracing::error!(
+                            "Failed to drop stale terminal visible-screen tail: {}",
+                            e
+                        ),
+                    }
+                }
 
                 // Flush any scrollback that has scrolled off but isn't in the
                 // file yet — in particular the lines a resize spilled from the
@@ -2442,8 +2471,15 @@ impl Window {
                 // Open backing file in append mode to add visible screen
                 if let Ok(mut file) = terminal_backing_fs().open_file_for_append(&backing_file) {
                     let mut writer = BufWriter::new(&mut *file);
-                    match state.append_visible_screen(&mut writer) {
-                        Ok(head) => prepended = head,
+                    let appended = state.append_visible_screen(&mut writer);
+                    // Only claim the tail once its bytes have actually reached
+                    // the file: the flag is what tells the next writer there
+                    // is something there to cut back off.
+                    match appended.and_then(|head| writer.flush().map(|()| head)) {
+                        Ok(head) => {
+                            prepended = head;
+                            state.set_backing_file_has_tail(true);
+                        }
                         Err(e) => {
                             tracing::error!(
                                 "Failed to append visible screen to backing file: {}",
--- a/crates/fresh-editor/src/app/terminal.rs
+++ b/crates/fresh-editor/src/app/terminal.rs
@@ -2471,14 +2471,17 @@ impl Window {
                 // Open backing file in append mode to add visible screen
                 if let Ok(mut file) = terminal_backing_fs().open_file_for_append(&backing_file) {
                     let mut writer = BufWriter::new(&mut *file);
+                    // Claim the tail *before* writing it. `BufWriter` spills
+                    // to the file as soon as its buffer fills, so a failure
+                    // part-way through still leaves bytes past the history
+                    // end; the flag is what tells a later path to cut them
+                    // back off, so it has to err towards "something may be
+                    // there" rather than "the write succeeded".
+                    state.set_backing_file_has_tail(true);
                     let appended = state.append_visible_screen(&mut writer);
-                    // Only claim the tail once its bytes have actually reached
-                    // the file: the flag is what tells the next writer there
-                    // is something there to cut back off.
                     match appended.and_then(|head| writer.flush().map(|()| head)) {
                         Ok(head) => {
                             prepended = head;
-                            state.set_backing_file_has_tail(true);
                         }
                         Err(e) => {
                             tracing::error!(
--- a/crates/fresh-editor/src/app/workspace.rs
+++ b/crates/fresh-editor/src/app/workspace.rs
@@ -2134,7 +2134,7 @@ impl crate::app::window::Window {
     /// Sync this window's active terminal visible screens to their
     /// backing files (so the snapshot captures complete terminal state).
     pub(crate) fn sync_terminal_backing_files(&self) {
-        use std::io::BufWriter;
+        use std::io::{BufWriter, Write};
 
         let terminals_to_sync: Vec<_> = self
             .terminal_buffers
@@ -2150,6 +2150,25 @@ impl crate::app::window::Window {
         for (terminal_id, backing_path) in terminals_to_sync {
             if let Some(handle) = self.terminal_manager.get(terminal_id) {
                 if let Ok(mut state) = handle.state.lock() {
+                    // Drop a visible-screen tail an earlier checkpoint or
+                    // scroll-back visit left behind: this snapshot writes a
+                    // fresh one below, and scrollback appended past a tail is
+                    // cut away by the truncation that ends the next scroll-back
+                    // visit (fresh#3151).
+                    if state.backing_file_has_tail() {
+                        let history_end = state.backing_file_history_end();
+                        match crate::app::terminal::terminal_backing_fs()
+                            .set_file_length(&backing_path, history_end)
+                        {
+                            Ok(()) => state.set_backing_file_has_tail(false),
+                            Err(e) => tracing::warn!(
+                                "Failed to drop terminal {:?} visible-screen tail: {}",
+                                terminal_id,
+                                e
+                            ),
+                        }
+                    }
+
                     // Persist any scrolled-off lines not yet in the file (e.g.
                     // lines a resize spilled into history on a terminal that was
                     // never viewed before quitting) so a restored workspace keeps
@@ -2167,16 +2186,29 @@ impl crate::app::window::Window {
                         }
                     }
 
+                    // Record the scrollback end before the tail goes in, so
+                    // the next writer knows where to cut it back off.
+                    if let Ok(metadata) =
+                        crate::app::terminal::terminal_backing_fs().metadata(&backing_path)
+                    {
+                        state.set_backing_file_history_end(metadata.size);
+                    }
+
                     if let Ok(mut file) = crate::app::terminal::terminal_backing_fs()
                         .open_file_for_append(&backing_path)
                     {
                         let mut writer = BufWriter::new(&mut *file);
-                        if let Err(e) = state.append_visible_screen(&mut writer) {
-                            tracing::warn!(
+                        let appended = state.append_visible_screen(&mut writer);
+                        // Only claim the tail once its bytes have actually
+                        // reached the file: the flag is what tells the next
+                        // writer there is something there to cut back off.
+                        match appended.and_then(|_| writer.flush()) {
+                            Ok(()) => state.set_backing_file_has_tail(true),
+                            Err(e) => tracing::warn!(
                                 "Failed to sync terminal {:?} to backing file: {}",
                                 terminal_id,
                                 e
-                            );
+                            ),
                         }
                     }
                 }
--- a/crates/fresh-editor/src/app/workspace.rs
+++ b/crates/fresh-editor/src/app/workspace.rs
@@ -2198,12 +2198,13 @@ impl crate::app::window::Window {
                         .open_file_for_append(&backing_path)
                     {
                         let mut writer = BufWriter::new(&mut *file);
+                        // Claim the tail before writing it: a part-written
+                        // tail still leaves bytes past the history end, and
+                        // the flag is what gets them cut back off later.
+                        state.set_backing_file_has_tail(true);
                         let appended = state.append_visible_screen(&mut writer);
-                        // Only claim the tail once its bytes have actually
-                        // reached the file: the flag is what tells the next
-                        // writer there is something there to cut back off.
                         match appended.and_then(|_| writer.flush()) {
-                            Ok(()) => state.set_backing_file_has_tail(true),
+                            Ok(()) => {}
                             Err(e) => tracing::warn!(
                                 "Failed to sync terminal {:?} to backing file: {}",
                                 terminal_id,
--- a/crates/fresh-editor/src/services/terminal/manager.rs
+++ b/crates/fresh-editor/src/services/terminal/manager.rs
@@ -893,20 +893,40 @@ impl ReaderLoop {
 
         // Incrementally stream new scrollback lines to the backing file.
         if let Some(writer) = self.backing_writer.as_mut() {
+            let mut failed: Option<std::io::Error> = None;
+
             match state.flush_new_scrollback(writer) {
                 Ok(lines_written) => {
                     if lines_written > 0 {
-                        if let Ok(pos) = writer.get_ref().metadata() {
-                            state.set_backing_file_history_end(pos.len());
+                        // Flush *before* measuring. `metadata()` reports what
+                        // is on disk, so reading it while the lines just
+                        // written still sat in the `BufWriter` recorded a
+                        // history end short by exactly this batch — and the
+                        // next truncation then cut them away (fresh#3151).
+                        match writer.flush().and_then(|()| writer.get_ref().metadata()) {
+                            Ok(pos) => {
+                                state.set_backing_file_history_end(pos.len());
+                                // Scrollback now sits *after* whatever
+                                // temporary visible-screen tail was in the
+                                // file, so that tail can no longer be
+                                // truncated off without taking these lines
+                                // with it. Adopt it as scrollback instead: a
+                                // duplicated screen is bounded, losing the
+                                // lines is not — the same trade
+                                // `flush_new_scrollback` makes when the grid
+                                // overruns.
+                                state.set_backing_file_has_tail(false);
+                            }
+                            Err(e) => failed = Some(e),
                         }
-                        #[allow(clippy::let_underscore_must_use)]
-                        let _ = writer.flush();
                     }
                 }
-                Err(e) => {
-                    tracing::warn!("Terminal backing file write error: {}", e);
-                    self.backing_writer = None;
-                }
+                Err(e) => failed = Some(e),
+            }
+
+            if let Some(e) = failed {
+                tracing::warn!("Terminal backing file write error: {}", e);
+                self.backing_writer = None;
             }
         }
     }
--- a/crates/fresh-editor/src/services/terminal/manager.rs
+++ b/crates/fresh-editor/src/services/terminal/manager.rs
@@ -708,13 +708,35 @@ fn open_transcript_file(
     path: &std::path::Path,
     mode: BackingMode,
 ) -> Option<std::io::BufWriter<std::fs::File>> {
-    let mut options = std::fs::OpenOptions::new();
-    options.create(true);
-    match mode {
-        BackingMode::Continue => options.append(true),
-        BackingMode::Fresh => options.write(true).truncate(true),
-    };
-    options.open(path).ok().map(std::io::BufWriter::new)
+    // `Fresh` empties the transcript first, through a handle of its own — the
+    // one we hand back must be `O_APPEND` in *both* modes.
+    //
+    // The UI thread writes these same files through its own append handle: a
+    // scroll-back sync flushes pending scrollback and then the visible-screen
+    // tail (`sync_terminal_to_buffer`, `sync_terminal_backing_files`). A
+    // non-append handle keeps its own offset, which those writes leave behind
+    // — so this loop's next batch would land *on top of* them, overwriting
+    // scrollback that `flush_new_scrollback` has already counted as persisted
+    // and therefore never re-emits (fresh#3151). With `O_APPEND` every writer
+    // lands at the end, nothing is overwritten, and the file's length is by
+    // construction this loop's write position, which is what
+    // `backing_file_history_end` records.
+    if matches!(mode, BackingMode::Fresh) {
+        // `truncate` and `append` cannot be combined in one `open`, so the
+        // reset is a separate, immediately-dropped handle.
+        std::fs::OpenOptions::new()
+            .create(true)
+            .write(true)
+            .truncate(true)
+            .open(path)
+            .ok()?;
+    }
+    std::fs::OpenOptions::new()
+        .create(true)
+        .append(true)
+        .open(path)
+        .ok()
+        .map(std::io::BufWriter::new)
 }
 
 /// Wait-thread body: block on the child's exit and fire `TerminalExited` once.
@@ -893,8 +915,6 @@ impl ReaderLoop {
 
         // Incrementally stream new scrollback lines to the backing file.
         if let Some(writer) = self.backing_writer.as_mut() {
-            let mut failed: Option<std::io::Error> = None;
-
             match state.flush_new_scrollback(writer) {
                 Ok(lines_written) => {
                     if lines_written > 0 {
@@ -903,6 +923,16 @@ impl ReaderLoop {
                         // written still sat in the `BufWriter` recorded a
                         // history end short by exactly this batch — and the
                         // next truncation then cut them away (fresh#3151).
+                        //
+                        // Neither failure justifies dropping the writer: a
+                        // failed flush leaves the bytes in the `BufWriter` for
+                        // the next call to retry (dropping it would discard
+                        // lines `flush_new_scrollback` has already counted as
+                        // persisted), and a failed `metadata()` is not a write
+                        // failure at all. Both just leave the recorded end
+                        // where it was — stale-low, which the tail flag below
+                        // keeps harmless, since nothing truncates to it unless
+                        // a tail is known to be there.
                         match writer.flush().and_then(|()| writer.get_ref().metadata()) {
                             Ok(pos) => {
                                 state.set_backing_file_history_end(pos.len());
@@ -911,22 +941,28 @@ impl ReaderLoop {
                                 // file, so that tail can no longer be
                                 // truncated off without taking these lines
                                 // with it. Adopt it as scrollback instead: a
-                                // duplicated screen is bounded, losing the
-                                // lines is not — the same trade
+                                // duplicated screen is bounded per event,
+                                // losing the lines is not — the same trade
                                 // `flush_new_scrollback` makes when the grid
-                                // overruns.
+                                // overruns. The cost is real and worth
+                                // knowing: every visit-or-checkpoint that
+                                // overlaps output splices one more screen's
+                                // worth of already-seen rows into the
+                                // transcript, and nothing removes them.
                                 state.set_backing_file_has_tail(false);
                             }
-                            Err(e) => failed = Some(e),
+                            Err(e) => tracing::warn!(
+                                "Terminal backing file sync error (scrollback kept, \
+                                 history end not advanced): {}",
+                                e
+                            ),
                         }
                     }
                 }
-                Err(e) => failed = Some(e),
-            }
-
-            if let Some(e) = failed {
-                tracing::warn!("Terminal backing file write error: {}", e);
-                self.backing_writer = None;
+                Err(e) => {
+                    tracing::warn!("Terminal backing file write error: {}", e);
+                    self.backing_writer = None;
+                }
             }
         }
     }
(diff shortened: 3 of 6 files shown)


crates/fresh-editor/src/services/terminal/manager.rs
```diff
@@ -893,20 +893,40 @@ impl ReaderLoop {
...
+                                // Scrollback now sits *after* whatever
+                                // temporary visible-screen tail was in the
+                                // file, so that tail can no longer be
+                                // truncated off without taking these lines
+                                // with it. Adopt it as scrollback instead: a
+                                // duplicated screen is bounded, losing the
+                                // lines is not — the same trade
+                                // `flush_new_scrollback` makes when the grid
+                                // overruns.
+                                state.set_backing_file_has_tail(false);
```
sinelaw, line 952: **MEDIUM — CONFIRMED.** `metadata().len()` is not the read loop's write position, so the new boundary invariant (and this comment) is wrong for the common case.

`open_transcript_file` (manager.rs ~L710) opens the reader's handle as

```rust
BackingMode::Continue => options.append(true),
BackingMode::Fresh    => options.write(true).truncate(true),   // <- no O_APPEND
```

Every terminal opened by `Open Terminal` is `BackingMode::Fresh` (`spawn_terminal_session`), so the reader writes **at its own file offset**, while `sync_terminal_to_buffer` / `sync_terminal_backing_files` append through a separate `O_APPEND` handle. Once a tail has been appended, the reader's offset and the file length are different numbers.

Instrumented run of this PR's own `test_scrollback_survives_output_during_a_scrollback_visit` (I patched a `stream_position()` probe in next to the `metadata()` call, on this branch):

```
DIAG reader: lines=21 stream_pos=278 file_len=278
DIAG DIVERGENCE: reader pos=278 file_len=423     <- tail appended by the UI
DIAG reader: lines=2  stream_pos=288 file_len=423
```

So at the adoption point this code records `backing_file_history_end = 423` while the read loop has only written up to 288, and then clears the tail flag. Consequences:

1. Bytes `[288, 423)` are a leftover fragment of the previous visible screen, now declared to be streamed scrollback. Whenever the batch is shorter than the tail, that fragment (typically a bare prompt / partial screen) is spliced into the middle of the scrollback.
2. The comment is factually wrong for `Fresh` terminals: scrollback does **not** sit "after" the tail — it overwrites it in place. And "a duplicated screen" describes only `BackingMode::Continue` (session restore / respawn), where the handle really is `O_APPEND` and the adoption really does leave a full duplicate screen. The two variants behave differently here (CONTRIBUTING #13).
3. *(SPECULATIVE — I could not build a reproducer)* anything the **UI** appends past the reader's stale offset can be overwritten by the reader's next batch. `flush_new_scrollback` in `sync_terminal_to_buffer`/`sync_terminal_backing_files` advances `synced_history_lines`, so those lines are never re-emitted — i.e. exactly the "counted as persisted, gone from disk" loss this PR is fixing, via a different route. It is hard to hit today only because the tail's rows happen to be the same rows that scroll off next, so the overwrite is usually byte-identical.

Suggested fix: record `writer.get_mut().stream_position()` here instead of `metadata().len()`, or open the reader handle with `.append(true)` in both modes so "file length" and "reader position" are the same quantity by construction.

---

crates/fresh-editor/src/services/terminal/manager.rs
```diff
@@ -893,20 +893,40 @@ impl ReaderLoop {
...
-                    tracing::warn!("Terminal backing file write error: {}", e);
-                    self.backing_writer = None;
-                }
+                Err(e) => failed = Some(e),
+            }
+
+            if let Some(e) = failed {
+                tracing::warn!("Terminal backing file write error: {}", e);
+                self.backing_writer = None;
             }
```
sinelaw: **MEDIUM — CONFIRMED (by reading; behaviour change vs. the old code).** The restructure widened the "disable the backing writer" path, and that now loses lines that `synced_history_lines` has already counted as persisted.

Before:

```rust
if let Ok(pos) = writer.get_ref().metadata() { ... }
let _ = writer.flush();          // error ignored, writer KEPT
```

A failed flush left the bytes sitting in the `BufWriter` and the writer alive, so the next `process_output` retried them.

After: a `flush()` **or** `metadata()` error sets `failed`, which drops `self.backing_writer`. `BufWriter::drop` retries the flush once and then discards whatever is still buffered. Meanwhile `flush_new_scrollback` has already advanced `synced_history_lines` for those lines, so nothing ever re-emits them — one transient `ENOSPC`/`EINTR` turns into a permanent hole in the scrollback *and* permanently stops scrollback streaming for the rest of the session.

Two separate points:

* Failing `metadata()` is not a write failure. Treating it as one and tearing down the writer is exactly the catch-all recovery CONTRIBUTING #9 warns about — it should not be in the same arm as a real write error.
* If a flush error really does mean the writer is dead, the lines it swallowed should be un-counted (rewind `synced_history_lines`) rather than reported as persisted.

---

crates/fresh-editor/src/services/terminal/term.rs
```diff
@@ -466,6 +466,16 @@ pub struct TerminalState {
...
     backing_file_history_end: u64,
+    /// A temporary visible-screen tail (written by `append_visible_screen`)
+    /// currently sits in the backing file past `backing_file_history_end`.
+    ///
+    /// The tail is not scrollback: it is the exit frame the scroll-back view
+    /// anchors to, and it is rewritten on every visit. Anything appended
+    /// *after* it is therefore living on borrowed time — the truncation that
+    /// ends the visit cuts the file back to `backing_file_history_end` and
+    /// takes it along. So every writer checks this flag and removes the tail
+    /// before appending real scrollback (fresh#3151).
```
sinelaw: **LOW — CONFIRMED.** The doc comment states an invariant the patch does not implement.

"So every writer checks this flag and removes the tail before appending real scrollback" — the PTY read loop, which is *the* writer of real scrollback, does not remove the tail. It appends (or, for `Fresh` terminals, overwrites) past it and clears the flag (`manager.rs` L906-918). Only the three UI-thread paths remove it.

This matters because the sentence is the reader's mental model of the flag, and it is the opposite of the actual contract: the flag means "a tail is there *unless the read loop got there first*". Please state the adoption case here.

Related, and better fixed in the same PR: `services/terminal/mod.rs` is the module's architecture doc and still says

> 3. **Scrollback → Terminal** (`enter_terminal_mode`): Truncates backing file to `backing_file_history_end` (removes visible screen tail)

and, under "Backing File Structure", that the visible screen is "present only in scrollback mode". After this change the truncation is conditional and an adopted tail is a permanent part of the scrollback. That doc is what the PR description leans on; it should be updated too.

---

crates/fresh-editor/src/app/terminal.rs
```diff
@@ -2442,8 +2471,15 @@ impl Window {
...
-                    match state.append_visible_screen(&mut writer) {
-                        Ok(head) => prepended = head,
+                    let appended = state.append_visible_screen(&mut writer);
+                    // Only claim the tail once its bytes have actually reached
+                    // the file: the flag is what tells the next writer there
+                    // is something there to cut back off.
+                    match appended.and_then(|head| writer.flush().map(|()| head)) {
+                        Ok(head) => {
+                            prepended = head;
+                            state.set_backing_file_has_tail(true);
```
sinelaw: **LOW — CONFIRMED (reasoning; I/O-error path only).** The flag now lies about the file in the *unsafe* direction on a partial write.

`append_visible_screen` writes into a `BufWriter`, which spills to the file whenever its 8 KiB buffer fills, and its `Drop` flushes whatever is left. So on an error from either `append_visible_screen` or `writer.flush()`, some tail bytes can already be on disk past `backing_file_history_end` — but `backing_file_has_tail` stays `false`, so no later path truncates them. The partial screen is then baked into the scrollback prefix permanently.

Before this PR, `enter_terminal_mode` truncated unconditionally and cleaned that up. The safe direction for this flag is "set it whenever bytes may have been written"; the current code sets it only when everything succeeded. Either set the flag before the write, or truncate back to `history_end` in the `Err` arm.

Same shape in `workspace.rs` L2201-2212.

---

crates/fresh-editor/src/app/workspace.rs
```diff
@@ -2150,6 +2150,25 @@ impl crate::app::window::Window {
...
+                            .set_file_length(&backing_path, history_end)
+                        {
+                            Ok(()) => state.set_backing_file_has_tail(false),
+                            Err(e) => tracing::warn!(
+                                "Failed to drop terminal {:?} visible-screen tail: {}",
+                                terminal_id,
+                                e
+                            ),
+                        }
+                    }
```
sinelaw, line 2170: **LOW — SPECULATIVE (not reproduced).** This makes the session-checkpoint path *shrink* the backing file, which it never did before.

`save_workspace_for` → `sync_terminal_backing_files` runs on mid-session checkpoints (window switch, `checkpoint_window_workspace`, plugin session ops), and it can run while a split is parked in scroll-back with an `EditorState` loaded from this very file. Previously the checkpoint only appended, so the file only ever grew and any byte offset the loaded buffer held stayed valid. Now it truncates to `history_end` and writes a new tail of a possibly different length.

For a backing file above `editor.large_file_threshold_bytes` the buffer reads lazily from disk, so a shrink under a live view is the classic stale-reference case (CONTRIBUTING #8/#11). Worth either skipping the truncate when that terminal's buffer is currently in scroll-back, or refreshing the view through `sync_terminal_to_buffer` afterwards.

I did not manage to build a repro (it needs a >threshold backing file plus a checkpoint mid-visit), hence SPECULATIVE — but it is a new failure mode this hunk introduces, not a pre-existing one.

---

crates/fresh-editor/src/app/workspace.rs
```diff
@@ -2167,16 +2186,29 @@ impl crate::app::window::Window {
                         }
                     }
 
+                    // Record the scrollback end before the tail goes in, so
+                    // the next writer knows where to cut it back off.
+                    if let Ok(metadata) =
+                        crate::app::terminal::terminal_backing_fs().metadata(&backing_path)
+                    {
+                        state.set_backing_file_history_end(metadata.size);
+                    }
```
sinelaw, line 2195: **LOW — CONFIRMED (design cost, not a defect on this line).** Setting the boundary here is the right call for save/restore — the restore path re-seeds `backing_file_history_end` from `metadata()` on a fresh `TerminalState` (`manager.rs` L416-424) and never reads the persisted value, so nothing downstream is disturbed — but it does newly arm the adoption path for *live* terminals.

Concretely: a checkpoint on a busy, live (non-scroll-back) terminal now leaves `has_tail = true` with a boundary behind it. The next PTY batch hits the adoption branch and the checkpoint's visible screen becomes permanent scrollback. Under `BackingMode::Continue` (a restored or respawned terminal, where the read loop's handle really is `O_APPEND`) that is one full duplicated screen per checkpoint, and nothing ever removes them — so the commit message's "a duplicated screen is bounded" holds per event but not per session: N checkpoints/visits with concurrent output ⇒ N duplicated screens interleaved into the transcript, each of which the user also sees a second time when those rows scroll off for real.

Not necessarily a blocker, but it should be an explicit, documented trade rather than an aside in a comment — and it argues for the read loop clearing the tail itself (it holds the state lock and a handle to the file) rather than adopting it.

---

crates/fresh-editor/tests/e2e/terminal.rs
```diff
@@ -5220,3 +5220,232 @@ fn test_tab_drag_split_resizes_terminal() {
...
+        harness
+            .send_key(KeyCode::PageDown, KeyModifiers::NONE)
+            .unwrap();
+        harness.render().unwrap();
+        let after = harness.screen_to_string();
+        seen.push_str(&after);
+        if after == before {
+            break;
+        }
+        pages_left -= 1;
```
sinelaw, line 5374: **LOW — test-quality notes.** I verified both tests are real reproducers, so this is not a "prove it" objection:

* On `cf626fc` with only the test file applied: `test_focusing_an_already_live_terminal_keeps_its_scrollback` fails with 36 unreachable lines; `test_scrollback_survives_output_during_a_scrollback_visit` fails 10/10 runs (5-36 unreachable, never 0).
* On this branch both pass 12/12 runs. No fixed timers; `wait_until` only. Good.

Two caveats on the helper:

1. **The sweep counts the temporary tail as "reachable".** Reading the history back re-enters scroll-back, which re-appends the current visible screen; a marker present *only* in that tail therefore scores as reachable even though the next truncation would remove it. Harmless for these two cases (the lost lines are older than the last screenful), but the helper cannot distinguish "durably in scrollback" from "in the volatile tail", so please don't reuse it for a case where that is the distinction under test.
2. **`if after == before { break }` is a content-equality stop, not an end-of-buffer stop.** Two consecutive identical pages (a run of blank rows after a `clear`, a repeated TUI frame) terminate the sweep early. That direction is fail-safe (it reports false *missing* markers rather than passing vacuously), but it makes the helper brittle for less dense content. A viewport-didn't-move check would be sturdier.

**Uncovered paths** worth naming: the third tail writer, `sync_terminal_backing_files` (checkpoint), has no test — nothing drives `save_workspace` on a live terminal and then asserts the scrollback survives — and the `BackingMode::Continue` variant of the adoption (restored/respawned terminal, `O_APPEND` handle, real duplication) is untested. The two tests cover exactly the two triggers the author already knew about.

---

sinelaw (commented):
## Verdict

The bug is real and the two reproducers are real. I confirmed on `cf626fc` (worktree at the parent, this PR's test file applied, everything else untouched) that `test_focusing_an_already_live_terminal_keeps_its_scrollback` loses 36 lines and `test_scrollback_survives_output_during_a_scrollback_visit` loses 5-36 lines on 10/10 runs; both pass 12/12 on this branch, and `services::terminal` unit tests plus the `e2e::terminal`/`terminal_close`/`terminal_scrollback_isolation`/`remote_terminal_backing_local` suites (89 tests) are green. The "flush before measuring" half of the diagnosis is unambiguously correct and is the actual root cause of the second, worse trigger.

**Where I disagree with the design.**

`backing_file_has_tail` is a *remembered* fact about a file four actors mutate through three different handles. The alternative the PR does not consider is to make the boundary **derivable** instead: the tail is written by exactly one function and is always the file's suffix, so the writer that owns it can record its start offset and any reader can re-derive "is there a tail" as `len(file) > history_end`. That removes the whole class of flag-vs-disk desync, and does not require moving the tail out of the file (which would be the properly structural fix, and is out of scope for a bug fix). I am not asking for that here — the boolean is a defensible bug-fix-sized change — but I do think the invariant it is meant to protect is not actually held, for the reason in the top comment:

**The read loop's handle is not `O_APPEND` for `BackingMode::Fresh`, so `metadata().len()` ≠ the read loop's write position.** I instrumented this branch and, in the PR's own scroll-back-visit test, the loop records `history_end = 423` while it has written only up to `288`. The new invariant "history end = file length after flush" is therefore false in the common case, the "scrollback now sits *after* the tail / a duplicated screen" comment describes only the `Continue` variant, and the two backing modes now diverge in behaviour (CONTRIBUTING #13). Fixing it is a one-liner (`stream_position()`, or open with `.append(true)` in both modes) and it makes the rest of the reasoning in this PR true rather than approximately true.

**On the adoption trade.** Trading bounded duplication for no loss is the right call *given* that the read loop cannot safely truncate under a loaded scroll-back buffer. But "bounded" is per event, not per session: for a restored/respawned terminal each visit-with-output and each checkpoint-with-output splices another full screen in, and nothing ever removes them. Please say that in the comment, and consider having the read loop clear the tail itself (it holds both the state lock and the file handle) instead.

**Findings, ranked**

| # | Sev | File | Status |
|---|-----|------|--------|
| 1 | Medium | `manager.rs:901-918` — `metadata().len()` is not the reader's write position; boundary and comment wrong for `Fresh` terminals | CONFIRMED (loss consequence SPECULATIVE) |
| 2 | Medium | `manager.rs:924-930` — flush/metadata error now kills the writer and discards buffered lines already counted by `synced_history_lines` | CONFIRMED |
| 3 | Low | `term.rs:472-477` — doc comment claims every writer removes the tail; the read loop does not. `services/terminal/mod.rs` architecture doc also stale | CONFIRMED |
| 4 | Low | `terminal.rs:2474-2481` / `workspace.rs:2201-2212` — partially written tail leaves the flag `false`, so nothing ever truncates it | CONFIRMED |
| 5 | Low | `workspace.rs:2158-2170` — checkpoint now shrinks the backing file, possibly under a loaded scroll-back buffer | SPECULATIVE |
| 6 | Low | `workspace.rs:2189-2195` — checkpoints now arm adoption on live terminals; duplication accumulates over a session | CONFIRMED (design cost) |
| 7 | Low | `tests/e2e/terminal.rs:5343-5363` — sweep counts the volatile tail as reachable; content-equality stop condition; checkpoint path untested | CONFIRMED |

Nothing here is a blocker for the loss the PR fixes; #1 and #2 are the two I'd want addressed before merge.

---

crates/fresh-editor/src/services/terminal/manager.rs
sinelaw: Fixed in eb75900 — and you undersold point 3. It is not speculative; it reproduces directly.

I took the `.append(true)`-in-both-modes option rather than `stream_position()`, because the overwrite is the real defect and only appending removes it. `stream_position()` would have made the *boundary* honest while leaving the read loop still writing on top of the UI's bytes.

A unit test pins it (`transcript_handles_append_in_both_backing_modes`): open a transcript handle, write through it, let a second handle append behind its back (what a scroll-back sync does), write again. With the old open mode:

```
left:  "stream-1\nstream-2\n"
right: "stream-1\nui-tail\nstream-2\n"
```

`ui-tail` is gone — overwritten in place. And since `sync_terminal_to_buffer` / `sync_terminal_backing_files` call `flush_new_scrollback` on their own handle, the bytes at risk are not only the tail but real scrollback already counted in `synced_history_lines` and therefore never re-emitted. That is the same permanent loss this PR exists to fix, reached by a different route.

`Fresh` still starts empty, via a separate immediately-dropped truncating handle — `truncate` and `append` cannot be combined in one `open`. Two more tests cover that (`fresh_backing_mode_empties_the_transcript_first`, `continue_backing_mode_keeps_the_existing_transcript`).

With every handle `O_APPEND`, `metadata().len()` **is** the loop's write position by construction, so the boundary invariant and the comment are now true in both modes, and the two modes no longer diverge (CONTRIBUTING #13).

---

crates/fresh-editor/src/services/terminal/manager.rs
sinelaw: Fixed in eb75900. You're right on both halves, and the first is the important one: I widened a teardown path while claiming to only reorder a measurement.

Only a real write error from `flush_new_scrollback` drops the writer again. A `flush()` failure keeps it, so the bytes stay in the `BufWriter` for the next `process_output` to retry — the old behaviour — and a `metadata()` failure is now handled separately, since it is not a write failure at all (CONTRIBUTING #9). Both just leave `backing_file_history_end` where it was: stale-low, which the tail flag keeps harmless, because nothing truncates to it unless a tail is known to be there.

On your second point — rewinding `synced_history_lines` when a flush error really is terminal — I've deliberately not done it here. The counter is advanced per logical line inside `flush_new_scrollback` and a partial `BufWriter` flush can split mid-line, so an honest rewind needs the flush to report how far it got; guessing would risk re-emitting or skipping. That hole predates this PR (the old code had the same "counted, never written" gap on a dying writer, it just kept the writer around to paper over it), and unwinding it properly is a bigger change than this fix should carry. Worth its own issue if you want it tracked.

---

crates/fresh-editor/src/services/terminal/term.rs
sinelaw: Fixed in eb75900. The sentence was aspirational — it described the design I started with (the read loop truncating the tail itself), which I abandoned when it broke session-restore, and I never rewrote the comment to match what shipped.

The field doc now spells out both writers and reads as you put it — a tail is there *unless the read loop got there first*:

- UI-thread paths (scroll-back entry, session checkpoint) rewrite the tail, so they truncate the old one off first.
- The read loop cannot truncate (a scroll-back view may be reading the file), so when it streams scrollback while the flag is set it *adopts* the tail as scrollback and clears the flag — one duplicated screen per overlap, permanently.

`services/terminal/mod.rs` is updated too: item 3 now says the truncation is conditional, the "Backing File Structure" bullet no longer claims the tail is scroll-back-only, and there's a new "Who owns the tail" section covering the three writers, the state lock, the `O_APPEND` rule from the thread above, and the adoption trade with its cost stated.

---

crates/fresh-editor/src/app/terminal.rs
sinelaw: Fixed in eb75900, taking the first of your two options: the flag is now set *before* the write, in both `sync_terminal_to_buffer` and `sync_terminal_backing_files`.

I'd had it backwards — I set it on success to make the flag mean "the tail is on disk", when what it has to mean is "bytes may be past the history end". Truncating in the `Err` arm would work too, but it can itself fail, and then nothing has recorded that there is something to clean up; setting the flag first leaves the file recoverable by the next path that truncates either way.

`backing_file_history_end` is already recorded before the append in both places, so the truncation target is right at the moment the flag goes up.

---

crates/fresh-editor/src/app/workspace.rs
sinelaw: Not changing this one, and I want to be explicit that it's a judgement call rather than a dismissal.

The shrink-under-a-lazily-loaded-buffer hazard is real, but it isn't new with this hunk: `enter_terminal_mode` has always truncated this exact file to this exact offset while a scroll-back `EditorState` was loaded from it, and `sync_terminal_to_buffer` (this PR) does the same — it just reloads immediately after. What the checkpoint adds is a truncate with no reload, and the tail it writes back is the *same current screen*, so the divergence is a length difference at the very end of the file, not a shifted body.

Both mitigations you propose cost more than that's worth right now: skipping the truncate on a scroll-back split means threading per-split view state into a save path (and a checkpoint on an idle terminal would then stack tails without bound, which is the failure the truncate exists to prevent); calling `sync_terminal_to_buffer` from the checkpoint path puts a buffer reload inside workspace save.

So I'd rather leave the correct file structure in and fix the view-refresh properly if it ever reproduces, than trade a definite structural improvement for an unreproduced rendering glitch. I've filed it as a follow-up — sinelaw/fresh#3153 — with your reasoning and the conditions needed to hit it, so it isn't lost.

If you'd rather it be handled in this PR, say so and I'll add the refresh.

---

crates/fresh-editor/src/app/workspace.rs
sinelaw: Documented in eb75900 rather than redesigned, and you're right that "bounded" was doing unearned work in the commit message.

The cost is now stated where someone will meet it — on the `backing_file_has_tail` field, in the adoption branch itself, and in the new "Who owns the tail" section of `services/terminal/mod.rs` — as: one duplicated screen per visit-or-checkpoint that overlaps output, permanently, with nothing removing it. Per event, not per session.

Note the `O_APPEND` fix in the other thread makes this *uniform* rather than worse: previously `Fresh` terminals overwrote the tail in place (which is why the duplication looked bounded — it was partly data loss instead), and only `Continue` accumulated. Now both accumulate, honestly. Measured on the issue's own repro — 480 lines, 8 scroll-back visits overlapping output — that is 64 duplicated line occurrences, ~8 per event, against 84 lines permanently lost before the fix.

On having the read loop clear the tail itself: I tried exactly that first. It holds the state lock and a handle, so mechanically it works — but it deletes a session checkpoint's screen capture the moment any output arrives after the save, and `test_session_restore_terminal_scrollback` / `test_scrollback_captured_after_session_restore` both fail on it (the marker lives only in the tail). Trading "sometimes lose the last screen from a restored session" for "never duplicate a screen" is the worse side of that deal, so adoption stayed. If you'd rather have it the other way, that's your call and I'll switch it.

---

crates/fresh-editor/tests/e2e/terminal.rs
sinelaw: Both caveats are now written on the helper in eb75900, in your terms: it counts the volatile tail as reachable (so it can't be reused where "durably in scrollback" vs "in the tail" is the distinction under test), and its stop condition is two identical pages rather than end-of-buffer — fail-safe, but brittle on sparse content. I left the mechanism alone; a viewport-didn't-move check would mean reaching into view state, which rule 2 pushes these tests away from.

On the uncovered paths — the checkpoint writer now has coverage from a different angle than you suggested. `sync_terminal_backing_files` is exercised by `test_session_restore_terminal_scrollback` and `test_scrollback_captured_after_session_restore`, and those were the two tests that caught my first (wrong) design, where the read loop truncated the tail; they're what forced adoption. What's still missing is a test that drives `save_workspace` on a *live* terminal with concurrent output and asserts no loss, which is the checkpoint's own arm of the adoption path.

I've not added it here because I couldn't make it fail on `cf626fc` — the checkpoint reaches truncation only through the same `enter_terminal_mode` the second test already drives, so it would be a passing test guarding a path already covered, which CONTRIBUTING #1 and #16 both argue against. Same for the `Continue` variant: post-`O_APPEND` both modes share one code path, so a restore-mode duplication test would assert the trade, not a regression.

Naming them as gaps rather than pretending otherwise: neither the checkpoint's own arm nor `BackingMode::Continue` adoption has a dedicated test.

---

sinelaw:
All seven threads answered and resolved; six fixed in eb75900, one (#5) declined with reasoning and filed as #3153.

Answering the design objection in your summary, since it's the one thing I'm keeping against your preference.

You're right that `backing_file_has_tail` remembers what could be derived — and after the `O_APPEND` fix your derivation is not just cleaner, it's *exact*: `backing_file_history_end` is now the read loop's write position by construction, so `len(file) > history_end` is precisely "a tail is there". It would also have made finding #4 unrepresentable rather than something I had to reorder two call sites to fix. On elegance you win outright.

I'm still keeping the boolean, for one reason: **the two disagree about which way to fail.**

The flag only ever says "truncate" when a writer recorded that it wrote a tail. If `backing_file_history_end` is ever wrong — stale-low from a path nobody thought about — the flag makes that a no-op. Derivation would see `len(file) > history_end`, conclude "tail", and truncate real scrollback away. That is exactly the failure mode of this issue, re-armed, and it fires on the *unknown* case rather than the known one.

That is not hypothetical. `reset_sync_state` sets `backing_file_history_end = 0` while leaving the file on disk intact; it is reachable (`term.rs:2211`). Under derivation, the next truncation after it would take the whole transcript to zero bytes. Today the flag is cleared alongside it and nothing happens.

So: derivation is right when the offset is trustworthy, and the offset is only trustworthy because of invariants this same PR just finished repairing. I'd rather the safety gate be independent of the thing it's guarding, at the cost of one redundant bool. Happy to be overruled — if you'd prefer the derived form, it's a small change and I'll make it.

CI is green on eb75900 and the issue's own tmux repro now loses 0 of 480 lines (84 before), with 64 duplicated line occurrences across 8 overlapping visits — the adoption cost, now documented rather than implied.

---

sinelaw:
**CI: `test windows-latest` failed on eb75900 — not this PR's, standing down and re-running once.**

One failure out of 9101 tests: `semantic::scroll_edges::a_multi_line_notch_walks_the_view_a_line_at_a_time`. Every other check is green, including `test ubuntu-latest`, `test macos-latest`, `clippy`, `fmt`, `doc`, `schema` and the musl build.

Why it isn't this PR's:

- The diff touches no view, viewport or scroll code — only `services/terminal/{manager,term,mod}.rs`, `app/{terminal,workspace}.rs` and `tests/e2e/terminal.rs`. There is no path from a terminal backing-file handle to mouse-wheel smooth scrolling.
- `test windows-latest` passed on this PR's previous commit (ba23db1), and the second commit added the `O_APPEND` change plus docs and terminal tests.
- The whole `semantic::scroll_edges` suite (10 tests) passes locally.

The failure is a wall-clock sensitivity in the test itself. The walk is driven by `pending.last_step.elapsed()` against `SMOOTH_SCROLL_LINE` (`app/mouse_input.rs:502`), and the test asserts an *exact* viewport position on the single frame `mouse_scroll_down` renders. On a loaded runner — this one took 33 minutes — more than one line comes due before that frame and the equality fails. CONTRIBUTING #3 rules this shape out.

No fix exists to port, so I'm not widening this PR with one. Filed as #3155 with a proposed patch (assert a range on the first frame, keep the semantic wait).

Re-running the failed job once now, as the single confirming re-run.

---

Issue #2930, opened 2026-08-07
switching splits regreshon


FreekyFrank:
Thanks for your good work.
Sorry, the keyboard shortcuts for switching from one split to the other and back are not working.
So <Alt>] and <Alt>[ have stopped working in version 0.4.6 and 0.47. I cannot comment any further back because I am still finding my way around.

sinelaw:
Reproduced on master (5c0d61c), debug build, Linux, plain tmux 200x50 (no kitty keyboard protocol), default config.

Steps to reproduce:
1. Open a file, create a split: Alt+V (View menu) → "Split Vertical". The menu lists "Focus Next Split  Alt+]" / "Focus Previous Split  Alt+[".
2. Press Alt+] (a legacy terminal transmits this as `ESC ]`).
3. Type any letters.
4. Press Alt+[ (`ESC [`), then type a letter.

Observed:
- Step 2: focus does not move. Steps 2–3: every subsequent keystroke is swallowed (nothing inserted in either pane) until a BEL/ST byte arrives.
- Step 4: focus does not move, and the next typed key is misparsed as a CSI key (typing `A` moved the cursor up — parsed as `ESC [ A` = Up).
- Control: running "Focus Next Split" from the View menu works, so only the key chords are broken.

Root cause: `crates/fresh-input-parser/src/lib.rs:386` — `feed_escape` treats `ESC ]` as an OSC introducer (entering `StringSeq`, discarding bytes until ST/BEL) and `ESC [` as a CSI introducer. Terminals without the kitty keyboard protocol transmit Alt+] / Alt+[ as exactly those sequences, so the chords can never reach the keymap. The test `osc_replies_are_swallowed_not_emitted_as_text` (fresh-input-parser/src/tests.rs:1219) describes emitting Alt+] as the "pre-fix behaviour" — the OSC-swallowing fix that shipped in 0.4.6 introduced this regression. A lone `ESC ]`/`ESC [` with no follow-up bytes needs to resolve to Alt+]/Alt+[.

---
_This comment is autogenerated: an automated reproduction attempt driven interactively in tmux against a debug build._

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #1580, opened 2026-04-15
Issues in "Search and Replace in Project" feature


sinelaw:
From @mandolyte - https://github.com/mandolyte/fresh-ideas-testing/blob/main/Project-search-replace.md

sinelaw:
Summary of **fixed** issues:

[x] **Undo Failure:** Pressing `Ctrl+Z` fails to undo project-wide search and replace changes.
   -> Can't easily be done for the entire project, but I did fix it for open buffers. Also you get a prompt now warning you about undo limitations.

[x] **Stale Match List:** The list of matches does not refresh after replacements are made, leaving old matches visible on the screen.
[x] **UI Duplication:** Opening the search-replace split erroneously duplicates it in the other panel.
[x] **Repeat Execution Error:** Because the match list doesn't clear, pressing the execution shortcut (`Alt+Ret`) multiple times incorrectly re-applies the changes over and over.
[x] **Improper State Persistence (Random File):** Quitting (`Ctrl+Q`) and restarting the application with a search panel open incorrectly remembers the split and populates it with a random file.
[x] **Buffer Mix-up on Restart:** If file tabs are added to an active search-replace split, quitting and restarting the application causes those tabs to load with incorrect content. 

Not fixed:

[ ] **Broken Expand/Collapse:** The chevron ("v" shape) icon fails to expand or reveal the files that contain matches.
   **Note**: it's collapsible by pressing enter, mouse click is missing

Other issues I found (also not fixed):

[ ] Replace sometimes leaves some files un-replaced and they stubbornly remain even if trying again and again

[ ] Can't scroll horizontally in the search panel, if a line is long there's no way to see the rest of it (should work with left/right arrow keys and shift+mouse wheel)

[ ] Can't use Alt+] shortcut to switch to the other split pane (file buffer) when focused on search and replace, see gif:

<img width="1920" height="1144" alt="Image" src="https://github.com/user-attachments/assets/785d9019-2e32-438e-addf-a4ef7826db6b" />

sinelaw:
Labeling as deferred for now: this links to an external document listing multiple issues. It would be easier to act on if split into individually-scoped issues.

---
_Generated by [Claude Code](https://claude.ai/code)_

sinelaw:
Status sweep on master (5c0d61c), debug build, Linux, tmux 200x50. Setup: git project, 10 case-varied `needle` matches across 5 files (one match at ~col 257 of a 290-char line); Alt+A panel.

[...]

Issue #1128, opened 2026-02-26
Key binding issue when using *


michelpado:
Hello,

I am not able to configure a keybinding using the "*" key (either on normal pad or numpad).
I have tried several naming but not working. 
Please note that, when I use a "normal" key (A, Z, E, etc), it works fine !

My attempts in config.json:
```
"keybindings": [
    {
      "key": "*",
      "modifiers": ["ctrl"],
      "action": "duplicate_line"
    },
    {
      "key": "asterisk",
      "modifiers": ["ctrl"],
      "action": "duplicate_line"
    },
    {
      "key": "kp_multiply",
      "modifiers": ["ctrl"],
      "action": "duplicate_line"
    }
]
```

I am using:

- fresh v0.2.9
- on GNOME Terminal 3.52.0
- on linux mint 22.3

Thanks for your help !

sinelaw:
`"*"` works if you don't use modifier - I think the issue is terminal not passing Ctrl+* correctly.

Open Help... Debug Keyboard Events and type your ctrl+* combo, in my linux it only sees the * without the ctrl....

michelpado:
I confirm that, when using Debug Keyboard Events, I only see the * and not the ctrl.
It seems to be true for all numpad inputs.

But please note that, when I do "ctrl+_any_key_from_main_keyboard_", it works fine. 

So, it seems to be related to modifiers & numpad input  (same behavior with alt, shift)

sinelaw:
Deferred: this isn't fixable purely in fresh — GNOME Terminal doesn't support the Kitty keyboard protocol, so `Ctrl+*` (and other modifier+numpad combos) are stripped of their modifiers before fresh ever sees them; resolving this requires either using a terminal that supports the Kitty keyboard protocol (kitty, alacritty, ghostty, wezterm) or design discussion around terminal-specific workarounds.

---
_Generated by [Claude Code](https://claude.ai/code)_

michelpado:
Thanks for your answer. I confirm that with the exact same environment but with Ghostty, it works perfectly

sinelaw:
Reproduced on master (5c0d61c), debug build, Linux, tmux 200x50, config `{"key": "*", "modifiers": ["ctrl"], "action": "duplicate_line"}`.

Steps to reproduce:
1. Send a bare `*` byte (what a legacy-encoding terminal delivers for Ctrl+*): `*` is inserted as text; the binding does not fire — the reported symptom.
[...]

PR #3154, opened 2026-09-02
input: bracketed paste into buffer-mounted panels, search rows windowed onto their match, symbolic key names

### LINKED ISSUE #1960: Unable to paste value for search in "Search and Replace in Project"

sinelaw:
## Try it

Everything below is copy-pasteable and non-destructive — it uses a scratch `HOME`, so your own config is untouched. The before/after output is captured from real runs of `master` (`cf626fc`) and this branch (`1d46cb9`), same fixture, same terminal size, not written from memory.

<details open>
<summary><b>Setup</b></summary>

```sh
cargo build                     # -> target/debug/fresh
FRESH=$PWD/target/debug/fresh

mkdir -p /tmp/srdemo && cd /tmp/srdemo && git init -q
printf 'alpha needle beta\n' > plain.txt
python3 - <<'EOF'
# match at column 263 of a 296-column line — past what the panel can show
open("long.txt", "w").write("start " + "x"*256 + "needle" + "y"*28 + "\n")
# same, behind a multibyte prefix (200 CJK chars = 600 bytes)
open("cjk.txt",  "w").write("漢"*200 + "x"*900 + "needle" + "y"*900 + "\n")
EOF

export DEMO_HOME=/tmp/srdemo-home
mkdir -p "$DEMO_HOME/.config/fresh"
cat > "$DEMO_HOME/.config/fresh/config.json" <<'JSON'
{ "keybindings": [ { "key": "kp_multiply", "action": "select_all" } ] }
JSON

HOME="$DEMO_HOME" "$FRESH" --no-restore
```
</details>

### 1. Terminal paste into Search & Replace — #1960

**Steps:** `Alt+A` to open the panel (the Search field has focus), then paste from the terminal — `Ctrl+Shift+V`, middle-click, or right-click → Paste. (Scripted: `tmux paste-buffer -p`.)

| | |
|---|---|
| **before** | Field stays empty. Status bar: `Editing disabled in this buffer` |
| **after** | `Search: [needle              ]  (3 matches / 3 files)`, status `Found 3 matches` |

`Tab` to the Replace field and paste again — same result. Two controls that should be unaffected: pasting into a normal buffer, and into the `Ctrl+F` prompt.

### 2. A match far along a long line is visible — #1580

**Steps:** with the panel open, type `needle`. Look at the `long.txt` and `cjk.txt` rows.

**before** — the match is nowhere on screen; the row shows the head of the line and truncates:
```
[v] cjk.txt:1   - 漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢…
[v] long.txt:1  - start xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx…
[v] plain.txt:1 - alpha needle beta
```

**after** — the window slides to bring the match in, `…` marking what was cut:
```
[v] cjk.txt:1   - …xxxxxxxxxxxxxxxxxxxxxxxneedleyyyyyyyyyyyyyyyyyyyyyyyyyy
[v] long.txt:1  - …xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxneedleyyyyyyyyyyyyyyy
[v] plain.txt:1 - alpha needle beta
```

`cjk.txt` is the one worth checking: the match position arrives as a *byte* column while the row is sliced in UTF-16 units, so an ASCII-only test passes while this case is broken. A short row like `plain.txt` must stay exactly as it was — the window only moves when it has to.

### 3. A symbolic key name actually binds — #1128

**Steps:** the setup config binds `kp_multiply` (the X11 keysym spelling) to `select_all`. Open a file and press `*`.

| | |
|---|---|
| **before** | A literal `*` is typed into the buffer; the binding never fires. `[⚠ 1]` sits in the status bar — the entry was dropped |
| **after** | The buffer is selected (cursor jumps to the end, whitespace shown as `·`); nothing is inserted, and no `⚠` badge |

The reported symptom of #1128, `Ctrl+*`, is **not** fixed and can't be — see that section below.

### 4. Two things that must NOT change

- **Paste must not leak under an overlay.** With the panel focused, press `Alt+F` to open the File menu, then terminal-paste. The Search field must be unchanged. (An earlier revision of this PR appended the text to the field *behind* the menu; that's what the third commit fixes.)
- **`Alt+]` still switches splits** from the panel. It was broken when #1580 was filed, but #2930's fix already landed on master — there is no change here, and it's listed so you can confirm nothing regressed.

---

Three input-routing fixes, plus four follow-up commits fixing defects that review found in the first three. Every issue was re-reproduced on master before any code was written, driving the editor interactively in tmux with real keystrokes.

## #1960 — bracketed paste is not routed to the Search & Replace panel

**Reproduced.** `Alt+A`, then `tmux paste-buffer -p`: nothing lands in the search field and the status bar reads `Editing disabled in this buffer`. Same in the Replace field. `Ctrl+V` works in both.

A bracketed paste arrives as a single `Event::Paste`, so it never passes through the widget key routing that `Action::Paste` (`Ctrl+V`) uses to reach a focused `Text` widget. `paste_bracketed_into_focused_panel` handled the Settings modal, floating panels and docks — the S&R panel is none of those (it is a widget panel rendered into a read-only widget buffer in a split), so the paste fell through to `paste_text` and hit that buffer's read-only gate.

It now routes to the same widget target the key path uses, `focused_text_widget_panel_for_buffer` on the active buffer, **gated on who owns the keyboard**. My first attempt guarded only on `prompt.is_some()`, which was the wrong question: with the File menu open over the panel, a terminal paste was appended to the search field *behind the menu* and the search re-ran — in mirror image the bug this function exists to prevent. It now asks the overlay stack, so a menu, prompt, modal, context menu or key-capturing popup all decline by declaring themselves rather than by being named in a list here.

Two deliberate non-changes, so the claim is not broader than the code: the file explorer sits inside the editor's own layer and is declined explicitly rather than silently redirected — `Action::Paste` sends it to `file_explorer_paste`, and wiring a bracketed paste there is a separate change. So the two paste paths agree on the *widget target*, not on every arm.

### 1. A match far along a long line was never visible — **fixed**

**Reproduced.** A match at column 257 of a 290-column line rendered as `long.txt:1 - start xxxx…xx..`, with the matched text nowhere on screen; `Right`, `End` and Shift+WheelDown changed nothing.

Two truncations discarded it: the plugin's 512-char context cap sliced from the start of the line, and the display-width truncation then cut the head of what was left. Both are now anchored on the match, which slides the window right when the match falls past the row's width, marking each elided end with `…`. A line with several matches yields one row per match and each anchors on its own.

**Corrections to earlier revisions of this description**, since review disproved two of its claims:

- It said the window "carries half a cap of slack" and was therefore safe against `match.column` being a byte column while the slicing counts UTF-16 units. Wrong: a CJK prefix inflates the column 3x, so ~128 such characters ahead of a match put the window past it entirely, and the row then rendered the text *after* the match. The ASCII-only test suite passed throughout.
- The replacement — converting the column — then reintroduced the unbounded per-row work `CONTEXT_HARD_CAP` exists to prevent (14.9 ms per flush on a 50 KB minified line, against 0.83 ms for the native search that replaced it).

The anchor is now a hit **located in the line itself**, with the byte column used only as an upper bound to search back from — sound because a byte offset is never smaller than the UTF-16 index of the same position. That also removes a dependence review caught: `resolveMatchTarget` rewrites `line`/`column` after a jump or replace and leaves `context` at its grep-time value, so the two can describe different strings. No locatable hit means no anchor, and the row falls back to the head of the line rather than presenting a mid-line slice as one.

Also correcting a performance claim: the per-codepoint work downstream is bounded as before, but locating the match adds a native scan per long row. Roughly double, not unchanged.

**Open question for the maintainer — this does not add horizontal scrolling.** Making the results tree pan sideways is a design decision I did not want to make unilaterally:

- `Left` / `Right` are already the Tree's collapse/expand (`lateral` in `widgets/kinds/tree.rs`), which is what the issue asks to rebind. Some other chord would have to own panning, in a mode whose bindings are already dense.
- No widget kind has a horizontal offset today — `PaintedWindow::offset` counts rows only — so Shift+wheel has nothing to move.
- The plugin truncates each row to the panel width before the host sees it, so a host-side h-offset would scroll onto padding. Real panning means the plugin sends full-width rows and the host slices, which changes the row-building contract.

Auto-windowing removes the reason most people reach for horizontal scroll — seeing the match — so it seemed worth landing on its own. Happy to follow up with panning given a preference on the keys.

### 2. `Alt+]` does not switch splits from the panel — **no longer reproduces; not fixed here**

Same root cause as #2930 (`ESC ]` swallowed as an OSC introducer), and #2930's fix is on master. Re-checked on `cf626fc` in a plain tmux with no kitty keyboard protocol:

- `Alt+]` from the S&R panel switches to the editor split (`Switched to next split`), and again back.
- The keybinding editor's record mode captures it: `Record Key: Alt+]`, filtering to the two `Alt+]` bindings — the repro comment's evidence was that it recorded *nothing*.

No code change for this one.

## #1128 — symbolic key names in config

The reported symptom (`Ctrl+*` not firing) is not fixable in Fresh — a legacy-encoding terminal strips the modifier before Fresh sees the key — and is untouched here.

Of the actionable secondary finding, **half is already on master**: an unparseable key name is no longer silent. `warn_invalid_key` logs at load and that lights the status bar's `[⚠ N]` badge.

Missing was the other half — the names. `"key": "asterisk"` (what the issue was configured with) parsed as nothing, and so did `kp_multiply`, `backslash`, `quotedbl` and the rest. `parse_key` now accepts the X11 keysym spellings for ASCII punctuation, alongside the single-character name, which stays canonical. Keypad aliases resolve to the characters their key transmits, since a terminal reports the numeric keypad through its ASCII byte — so binding `kp_multiply` also binds the main-keyboard `*`; they are not distinguishable at this layer.

A separate quiet drop one line below that table is fixed too: the single-character arm measured *bytes*, so `"é"` was two bytes and fell through to `None`.

Verified in tmux: `{"key": "asterisk", "modifiers": ["ctrl"], …}` fires on the CSI-u encoding `ESC[42;5u` (note: that is CTRL only — a kitty-protocol terminal on a US layout would send CTRL|SHIFT, which this binding would not match; unverified either way).

Not done, and arguably should be: `docs/features/keybinding-editor.md` still documents `key` as `"s" / "Enter" / "F1" / "Up"` with no list of the accepted names, and the keypad coverage stops at the arithmetic keys (`kp_enter`, `kp_0`…`kp_9` are not there). Say the word and I'll add both.

## Tests

Each new test verified to fail on master, or without its specific fix, and pass with it:

- `test_search_replace_bracketed_paste_reaches_fields` — `Event::Paste` into both panel fields; fails on master with the `Editing disabled` screen.
- `test_search_replace_bracketed_paste_declines_under_an_open_menu` — fails against the `prompt.is_some()` gate with `LEAKED` in the field behind the menu.
- `test_search_replace_long_line_match_is_visible` — a match past the row width, one past the 512-char cap, two matches on one line rendering differently, and a short match still unwindowed.
- `test_search_replace_multibyte_prefix_match_is_visible` — CJK and non-BMP prefixes.
- `test_search_replace_unlocatable_pattern_renders_line_head` — drives the no-hit fallback deterministically with `(?i)…`, a regex the Rust engine accepts and JavaScript rejects.
- `test_parse_key_symbolic_names`, `test_symbolic_key_name_binds_from_config` — the names parse and a config entry using one registers.
- `test_unknown_key_name_entry_is_dropped_but_load_continues` moves to a name nothing claims, since the two it used now bind.

Two of these first passed against the bug they targeted — the multibyte one because every fixture was ASCII, the head-fallback one because filler alone cannot distinguish "the head of the line" from "a slice of the filler". Both were caught by running each new test against the unfixed code, which is the only reason they are load-bearing now.

Checks: `cargo fmt --all --check`, `cargo clippy --all-targets` (no warnings in the changed files; the one `sort_by_key` note in `overlay.rs` is on master too), `cargo build`, `check-types.sh search_replace.ts` (adds no errors; the four in `lib/widgets.ts` are on master too), and the `e2e::search_replace` (51), `paste`/`menu_bar`/`markdown_compose` (281) and `overlay`/`explorer`/`keybinding` (308) suites.

Closes #1960

https://claude.ai/code/session_01FMy6GBN9N5HLMGe8oUvUQX

### DIFF
--- a/crates/fresh-editor/plugins/search_replace.ts
+++ b/crates/fresh-editor/plugins/search_replace.ts
@@ -273,6 +273,38 @@ function truncate(s: string, maxLen: number): string {
   return result + "...";
 }
 
+/** Matches a string containing any non-ASCII scalar. Non-global so it
+ *  carries no `lastIndex` state between calls. */
+const NON_ASCII_RE = /[^\x00-\x7F]/;
+
+/** Convert a 0-based UTF-8 *byte* offset into a line to the UTF-16 index
+ *  the same line is sliced by.
+ *
+ *  `SearchMatch.column` is documented as a byte column
+ *  (`fresh-editor-core/src/model/filesystem.rs`), while every `slice`
+ *  here counts UTF-16 units. On an ASCII line the two coincide, which is
+ *  why the mismatch hid: it only bites when the text before the match is
+ *  multibyte, and then it overstates the position by up to 3x (CJK) —
+ *  enough to slide the window clean past the match it was meant to
+ *  centre on.
+ *
+ *  The ASCII fast path is a native regex scan with early exit, so the
+ *  common case (source code, minified assets) costs no per-codepoint
+ *  work. The slow path walks only as far as the match. */
+function byteColumnToIndex(line: string, byteCol: number): number {
+  if (byteCol <= 0) return 0;
+  if (!NON_ASCII_RE.test(line)) return Math.min(byteCol, line.length);
+  let bytes = 0;
+  let index = 0;
+  for (const ch of line) {
+    if (bytes >= byteCol) break;
+    const cp = ch.codePointAt(0)!;
+    bytes += cp < 0x80 ? 1 : cp < 0x800 ? 2 : cp < 0x10000 ? 3 : 4;
+    index += ch.length;
+  }
+  return index;
+}
+
 /** Codepoint index and length of the `pattern` hit in `text` nearest to
  *  `preferAt`, or `null` when the pattern is empty, invalid, or doesn't
  *  occur. A line with several matches yields one row per match, so the
@@ -791,16 +823,19 @@ function renderFlatItemEntry(item: FlatItem, W: number): TextPropertyEntry {
   // of a 290-char line rendered as `start xxxx…`, with the matched
   // text nowhere on screen).
   //
-  // `match.column` is a 1-based *byte* column; the slice indices here
-  // are UTF-16 units. They coincide for the ASCII-dominated case, and
-  // a multibyte prefix only pushes the window right — which is why the
-  // window keeps half a cap of slack on the left and `windowAroundMatch`
-  // re-locates the match in the sliced text rather than trusting the
-  // column arithmetic (falling back to head-truncation, the old
-  // behaviour, when the match isn't in the window after all). The
-  // window is still one cap wide, so the per-codepoint work downstream
-  // is bounded exactly as it was.
-  const matchCol = Math.max(0, (result.match.column || 1) - 1);
+  // `match.column` is a 1-based *byte* column and every slice here counts
+  // UTF-16 units, so it is converted before use. Half a cap of slack does
+  // NOT absorb the difference — a CJK prefix inflates the byte column 3x,
+  // so ~128 such characters ahead of the match are enough to put the
+  // window past it entirely; `windowAroundMatch` then finds no match and
+  // falls back to truncating a window centred on the wrong part of the
+  // line, which is worse than the head-truncation it replaced.
+  //
+  // `windowAroundMatch` still re-locates the match in the sliced text
+  // rather than trusting this arithmetic, and still degrades to
+  // head-truncation if it isn't there. The window stays one cap wide, so
+  // the per-codepoint work downstream is bounded as it was.
+  const matchCol = byteColumnToIndex(rawCtx, Math.max(0, (result.match.column || 1) - 1));
   const capStart = rawCtx.length > CONTEXT_HARD_CAP
     ? Math.max(0, Math.min(matchCol - CONTEXT_HARD_CAP / 2, rawCtx.length - CONTEXT_HARD_CAP))
     : 0;
--- a/crates/fresh-editor/plugins/search_replace.ts
+++ b/crates/fresh-editor/plugins/search_replace.ts
@@ -273,36 +273,71 @@ function truncate(s: string, maxLen: number): string {
   return result + "...";
 }
 
-/** Matches a string containing any non-ASCII scalar. Non-global so it
- *  carries no `lastIndex` state between calls. */
-const NON_ASCII_RE = /[^\x00-\x7F]/;
-
-/** Convert a 0-based UTF-8 *byte* offset into a line to the UTF-16 index
- *  the same line is sliced by.
+/** UTF-16 index of the `pattern` hit in `line` nearest to (at or before)
+ *  `byteHint`, or -1 when the pattern is empty, invalid, or absent.
  *
- *  `SearchMatch.column` is documented as a byte column
- *  (`fresh-editor-core/src/model/filesystem.rs`), while every `slice`
- *  here counts UTF-16 units. On an ASCII line the two coincide, which is
- *  why the mismatch hid: it only bites when the text before the match is
- *  multibyte, and then it overstates the position by up to 3x (CJK) —
- *  enough to slide the window clean past the match it was meant to
- *  centre on.
+ *  This is how a row learns where its own match sits, and it deliberately
+ *  does NOT convert `match.column` to a UTF-16 index. Three reasons:
  *
- *  The ASCII fast path is a native regex scan with early exit, so the
- *  common case (source code, minified assets) costs no per-codepoint
- *  work. The slow path walks only as far as the match. */
-function byteColumnToIndex(line: string, byteCol: number): number {
-  if (byteCol <= 0) return 0;
-  if (!NON_ASCII_RE.test(line)) return Math.min(byteCol, line.length);
-  let bytes = 0;
-  let index = 0;
-  for (const ch of line) {
-    if (bytes >= byteCol) break;
-    const cp = ch.codePointAt(0)!;
-    bytes += cp < 0x80 ? 1 : cp < 0x800 ? 2 : cp < 0x10000 ? 3 : 4;
-    index += ch.length;
+ *  1. A conversion has to walk the line codepoint by codepoint, and on the
+ *     very files `CONTEXT_HARD_CAP` was written for — 5 000-50 000 char
+ *     minified bundles — that walk is tens of thousands of QuickJS
+ *     iterations per row, per re-render. One `©` anywhere on the line puts
+ *     every row on it onto that path. Native string search is two to three
+ *     orders of magnitude cheaper and is all this needs.
+ *  2. `column` and `context` are not guaranteed to describe the same
+ *     string: `resolveMatchTarget` rewrites `line`/`column` from the live
+ *     interval marker after a jump or replace and leaves `context` at its
+ *     grep-time value. Searching `context` itself cannot drift from it.
+ *  3. A byte offset is never smaller than the UTF-16 index of the same
+ *     position, so `column` is a soundupper bound to search back from —
+ *     which is what makes it useful as a hint without being trusted as a
+ *     position. Picking the hit at or before it is what keeps several
+ *     matches on one line anchored to their own rows.
+ *
+ *  Returns -1 rather than guessing when the pattern isn't on the line at
+ *  all — the stale-row case, where the tree renders the previous search's
+ *  results against the pattern currently being typed. The caller reads
+ *  that as "no anchor" and falls back to the head of the line. */
+function matchIndexNearByteColumn(
+  line: string,
+  pattern: string,
+  isRegex: boolean,
+  caseSensitive: boolean,
+  byteHint: number,
+): number {
+  if (!pattern) return -1;
+  // The hint is a byte offset; clamping to the UTF-16 length keeps it a
+  // valid search position and costs nothing when they coincide.
+  const hint = Math.max(0, Math.min(byteHint, line.length));
+  try {
+    if (!isRegex) {
+      const hay = caseSensitive ? line : line.toLowerCase();
+      const needle = caseSensitive ? pattern : pattern.toLowerCase();
+      const before = hay.lastIndexOf(needle, hint);
+      // Nothing at or before the hint: the byte column overstated the
+      // position past the last hit (a heavily multibyte prefix), so take
+      // the first hit on the line instead of reporting none.
+      return before >= 0 ? before : hay.indexOf(needle);
+    }
+    const re = new RegExp(pattern, caseSensitive ? "g" : "gi");
+    let best = -1;
+    let first = -1;
+    let m;
+    while ((m = re.exec(line)) !== null) {
+      if (m[0].length === 0) {
+        re.lastIndex++;
+        continue;
+      }
+      if (first < 0) first = m.index;
+      if (m.index > hint) break;
+      best = m.index;
+    }
+    return best >= 0 ? best : first;
+  } catch (_e) {
+    // Invalid regex mid-typing: no anchor, same as no hit.
+    return -1;
   }
-  return index;
 }
 
 /** Codepoint index and length of the `pattern` hit in `text` nearest to
@@ -347,7 +382,16 @@ function matchSpanNear(
  *  exactly `truncate`, so ordinary rows are unchanged.
  *
  *  `matchAt` is where this row's own match is expected to sit in
- *  `text` — see `matchSpanNear`. */
+ *  `text` — see `matchSpanNear`.
+ *
+ *  `alreadySliced` says whether `text` is itself a mid-line window. It
+ *  governs the fallbacks, not the happy path: head-truncating a slice
+ *  that does not start at the beginning of the line renders it as
+ *  though it did, with a trailing "..." implying nothing was cut on the
+ *  left. The caller now anchors at 0 whenever it cannot locate the
+ *  match, so this should not arise — but a marker costs one column and
+ *  a row that silently lies about where it starts is worse than a row
+ *  that is one character narrower. */
 function windowAroundMatch(
   text: string,
   budget: number,
@@ -355,17 +399,25 @@ function windowAroundMatch(
   isRegex: boolean,
   caseSensitive: boolean,
   matchAt: number,
+  alreadySliced: boolean,
 ): string {
-  if (charLen(text) <= budget) return text;
+  const ELLIPSIS = "…";
+  // Head-truncation, honest about a left-hand elision when there is one.
+  const head = (t: string) =>
+    alreadySliced
+      ? ELLIPSIS + truncate(t, Math.max(1, budget - 1))
+      : truncate(t, budget);
+  // `head` (not a bare prepend) so the marker cannot push the row one
+  // column over `budget` when the text already fills it exactly.
+  if (charLen(text) <= budget) return alreadySliced ? head(text) : text;
   const span = matchSpanNear(text, pattern, isRegex, caseSensitive, matchAt);
   // No locatable match (empty/invalid pattern, or a hit that fell
   // outside the capped window): keep the historical head-truncation.
-  if (!span) return truncate(text, budget);
+  if (!span) return head(text);
   // The match is already visible in the head — `truncate`'s own "..."
   // marker covers the elided tail.
-  if (span.start + span.length <= budget - 3) return truncate(text, budget);
+  if (span.start + span.length <= budget - 3) return head(text);
 
-  const ELLIPSIS = "…";
   // Codepoint array: slicing it can never split a surrogate pair, and
   // the input is capped well under a couple of thousand codepoints.
   const cps = Array.from(text);
@@ -376,7 +428,7 @@ function windowAroundMatch(
   // Nothing is elided on the left, so there is no window to slide: the
   // match starts inside the first third and is simply wider than the
   // row. Head-truncation is all this width allows.
-  if (start === 0) return truncate(text, budget);
+  if (start === 0) return head(text);
   // One column goes to the leading `…`.
   const room = budget - 1;
   if (start + room >= cps.length) {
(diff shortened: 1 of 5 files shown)


sinelaw:
`test macos-latest` is red on `3c78a5b`. I don't believe it's this PR's, and here is the reasoning rather than an assertion:

**The failure.** One test of 9504, by timeout, not assertion:

```
TIMEOUT [180.379s] (4748/9504) fresh-editor::all_tests
  e2e::markdown_compose_scrollbar_markers::every_heading_is_marked_without_scrolling
Summary [1322.280s] 9504 tests run: 9503 passed, 1 timed out, 162 skipped
```

**Why it isn't this diff.** The diff touches `app/clipboard.rs`, `app/overlay.rs`, `input/keybindings.rs`, `plugins/search_replace.ts` and `tests/e2e/search_replace.rs`. That test:

- copies only the `markdown_compose` plugin into its temp plugin dir, so `search_replace.ts` is never loaded in it;
- drives no paste, no menu and no keybinding config, so the clipboard and keybinding changes are not on its path;
- and the two functions added to `overlay.rs` are new, called only from the paste route — no existing behaviour was changed there.

**And the same test passed on this same commit.** `test ubuntu-latest` is green on `3c78a5b`, running that test. It also passes locally in under 5s (whole file: 6 tests, 4.86s), and `test macos-latest` was green on this PR's previous head `739e0f4`.

The test waits on an unbounded `wait_until` for a pre-scan that crosses the async plugin bridge, so a starved bridge on a loaded runner shows up exactly like this — a 180s timeout on a test that otherwise takes milliseconds, with everything around it passing.

**What I've done.** Tried to re-run the failed job; GitHub refuses while the run is still going (Windows is in flight), so I've scheduled the retry. That is the one re-run — if macOS fails the same way again I'll treat it as real and root-cause it rather than re-running further. I have not touched the test, and won't: skipping or loosening it to get green is not on the table.

---

sinelaw:
The re-run passed — `test macos-latest` is green on `3c78a5b`, and all 11 checks are now green. So the timeout was the starved async wait it looked like, not a defect, and the analysis above stands as the record of why it was re-run rather than "fixed".

That was the one re-run. Nothing further on this unless something new goes red.

---

crates/fresh-editor/plugins/search_replace.ts
```diff
@@ -273,6 +273,118 @@ function truncate(s: string, maxLen: number): string {
...
+ *  multibyte, and then it overstates the position by up to 3x (CJK) —
+ *  enough to slide the window clean past the match it was meant to
+ *  centre on.
+ *
+ *  The ASCII fast path is a native regex scan with early exit, so the
+ *  common case (source code, minified assets) costs no per-codepoint
+ *  work. The slow path walks only as far as the match. */
+function byteColumnToIndex(line: string, byteCol: number): number {
+  if (byteCol <= 0) return 0;
+  if (!NON_ASCII_RE.test(line)) return Math.min(byteCol, line.length);
```
sinelaw: `byteColumnToIndex` reintroduces the unbounded per-row work that `CONTEXT_HARD_CAP` exists to prevent.

Both of its paths run over the **whole** `rawCtx`, before any capping:

* `NON_ASCII_RE.test(line)` scans the full line (native, cheap-ish, but O(N) per row).
* When it hits, the `for (const ch of line)` walk runs `byteCol` iterations — up to the match column, which on the very files the cap was written for is tens of thousands.

The cap's own comment names the case: "Minified CSS / JSON / single-line generated files routinely have match context strings 5 000-50 000 chars long … at 5 000 chars and 50 items per flush that adds up to several hundred ms of JS work per pump iteration, blocking Tab and other queued requests." The ASCII fast path does not cover it — one `©`, one curly quote, one emoji anywhere in a bundled JS/CSS line flips every row on that line onto the slow path.

Concrete scenario: grep a 50 KB minified single-line bundle containing a single non-ASCII byte, with matches near the end of the line. Each of the ~50 rows in a flush walks ~50 000 codepoints in QuickJS with a `codePointAt` per step — exactly the several-hundred-ms stall the hard cap was added to avoid, and it is paid on every re-render of the results tree, not once.

Bounding it is cheap: only the last `CONTEXT_HARD_CAP` bytes/chars of the line can ever matter, so the byte→UTF-16 walk can start from `Math.max(0, byteCol - CONTEXT_HARD_CAP)` region rather than from index 0, or the whole conversion can be skipped when `rawCtx.length <= CONTEXT_HARD_CAP` (where `capStart` is 0 anyway and `matchCol` is only used as a `preferAt` hint).

---

crates/fresh-editor/plugins/search_replace.ts
```diff
@@ -273,6 +273,118 @@ function truncate(s: string, maxLen: number): string {
...
+  pattern: string,
+  isRegex: boolean,
+  caseSensitive: boolean,
+  matchAt: number,
+): string {
+  if (charLen(text) <= budget) return text;
+  const span = matchSpanNear(text, pattern, isRegex, caseSensitive, matchAt);
+  // No locatable match (empty/invalid pattern, or a hit that fell
+  // outside the capped window): keep the historical head-truncation.
+  if (!span) return truncate(text, budget);
```
sinelaw: The comment says this "keeps the historical head-truncation", but by the time `windowAroundMatch` runs, `text` may already be a **mid-line** window (`capStart > 0`), so this renders a slice from the middle of the line with no marker on the left — it reads as the start of the line when it isn't. That is a new failure mode, not the historical behaviour.

Easiest way to hit it, no exotic input needed: the results tree deliberately renders stale rows while the pattern is being edited ("render the stale results … until the next search completes" in `buildMatchListSpec`), and `renderFlatItemEntry` windows those stale rows with `panel.searchPattern` — the *new*, still-being-typed pattern. So with results from `NEEDLE` on a 5 000-char minified line, typing one more character gives `NEEDLEX`, `matchSpanNear` finds nothing in the capped window, and every long row flips to `truncate(capped, budget)` — characters ~1744-1843 of the line presented as though they were its head, complete with a trailing `...` implying nothing was cut on the left. On master those rows showed the actual head of the line. It re-renders that way on every keystroke.

Two other ways in: an anchored/`$`-style regex that can't re-match inside the window, and a `match.column` that no longer lines up with `context` (see the separate note on `resolveMatchTarget`).

Fix is small — tell `windowAroundMatch` whether its input was already sliced (or have the caller prepend the marker), so this arm returns `ELLIPSIS + truncate(text, budget - 1)` when `capStart > 0`.

---

crates/fresh-editor/plugins/search_replace.ts
```diff
@@ -704,9 +816,34 @@ function renderFlatItemEntry(item: FlatItem, W: number): TextPropertyEntry {
...
+  // so ~128 such characters ahead of the match are enough to put the
+  // window past it entirely; `windowAroundMatch` then finds no match and
+  // falls back to truncating a window centred on the wrong part of the
+  // line, which is worse than the head-truncation it replaced.
+  //
+  // `windowAroundMatch` still re-locates the match in the sliced text
+  // rather than trusting this arithmetic, and still degrades to
+  // head-truncation if it isn't there. The window stays one cap wide, so
+  // the per-codepoint work downstream is bounded as it was.
+  const matchCol = byteColumnToIndex(rawCtx, Math.max(0, (result.match.column || 1) - 1));
```
sinelaw: `result.match.column` and `result.match.context` are not guaranteed to describe the same string, and this is the first code to depend on them agreeing.

`resolveMatchTarget` (~line 2246) rewrites the match position from the live interval marker on every jump/replace:

```ts
result.match.line = pos.line;
result.match.column = pos.column;
```

`result.match.context` is left at its grep-time value. So after `Enter` on a result (or `openResultLocation` via the nav commands) and then an edit earlier on that same line in the source split, `column` is the live column while `context` is the old line text.

For lines ≤ 512 chars this is harmless — `capStart` is 0 and `matchAt` is only a `preferAt` tiebreaker. For a line longer than the cap it isn't: `capStart = matchCol - 256` now anchors the 512-char window on a position that doesn't exist in `context`, so the real match falls outside the window entirely, `matchSpanNear` returns `null`, and the row renders an unrelated middle slice of the line (and, per the note on line 363, without a left-hand elision marker). That is worse than the head truncation this replaced, and it persists until the next search re-runs.

Cheapest guard: verify the window actually contains a hit before trusting `matchCol`, and fall back to `capStart = 0` when it doesn't.

---

sinelaw (commented):
Reviewed the full diff. The Rust halves both look correct to me:

- **Bracketed paste routing** — `editor_base_owns_keyboard()` is the right gate. I checked every chrome component's `layers()`: `Menu`, `Prompt`, `ContextMenu`, all four modals and `FloatingModal` set `owns_keyboard: true` while present/focused, `Dock`/`Popup` only while focused/capturing, and `Base` is the only one whose kind is `Editor` — so the new arm is reached exactly when the editor pane owns the keyboard. `file_explorer.rs` contributes no layer, which is why the explicit `KeyContext::FileExplorer` decline is needed rather than redundant. The widget target matches `Action::Paste` in `action_dispatch.rs` (`focused_text_widget_panel_for_buffer`, same CRLF normalisation), and the `None` case still falls through to `paste_text` as on master. No issues found.
- **`parse_key`** — the new arms sit above both the single-character arm and the `f<N>` arm, none of the new names start with `f`, no other key-name table exists in the tree (only one `"pagedown" =>`), `key_name` is used for logging only in `insert_binding_with_equivalents`, and `key_code_to_config_name` writes `Char(c)` as the single character, so the round-trip is unaffected. The `chars().count() == 1` fix is right and can't reach the `s[1..]` slice with a non-boundary index. No issues found.

Three findings, all in `search_replace.ts`, inline. The first is a performance regression against the invariant `CONTEXT_HARD_CAP` was added to hold; the other two are cases where the match-anchored window renders a mid-line slice as though it were the head of the line.

I did check the pieces the description flags as risky and they hold up: `byteColumnToIndex` converts correctly (verified against `SearchMatch { column }` in `filesystem.rs` and `buffer/search.rs` — byte column within the line, `context` the whole line, consistent in the chunked paths too); the window arithmetic in `windowAroundMatch` is exact at both ends (`1 + (room-1) + 1 == budget`, and `1 + room == budget`), the branch condition has no off-by-one, and the match start is provably inside the window in both branches. The `capStart` slice can split a surrogate pair, but the resulting lone surrogate is always in the region the window elides, so it never renders — not worth changing.

---

crates/fresh-editor/plugins/search_replace.ts
sinelaw: You're right, and the conversion is gone rather than bounded — `1d46cb9`.

I went with a native search instead of the bounded walk, because it also answers your other two findings: the anchor is now a hit located in `rawCtx` itself (`matchIndexNearByteColumn`), with the byte column used only as an upper bound to search back from. That's sound because a byte offset is never smaller than the UTF-16 index of the same position, and it keeps several matches on one line anchored to their own rows — which was the only thing the exact position was buying.

Measured your scenario (50 KB line, one `©`, matches near the end, 50 rows per flush):

| | per flush |
|---|---|
| old byte→UTF-16 walk | 14.89 ms |
| native search, case-insensitive | 0.83 ms |
| native search, case-sensitive | 0.01 ms |

That's V8, where the walk is JIT-compiled — under QuickJS the gap is wider, since the walk is interpreted and `lastIndexOf` is native in both. The case-insensitive path is the slower one because it allocates a lowercased copy of the line per row; I left it, since 0.83 ms is far inside the budget the cap was defending, but a one-entry memo keyed on the line would collapse it (consecutive rows on a huge line share `context`) if you'd rather it were tighter.

---

crates/fresh-editor/plugins/search_replace.ts
sinelaw: Confirmed and fixed in `1d46cb9`. I checked `resolveMatchTarget` before acting: lines 2255-2256 rewrite `result.match.line` and `.column`, and `.context` is untouched — so the two genuinely can describe different strings, and this was the first code to assume otherwise.

Rather than guarding after the fact, the anchor no longer comes from `column` at all: it's an index found by searching `context` itself, so it cannot point at a position that string doesn't have. Drift in `column` now only weakens the *hint* that picks between several hits on a line, and can't misplace the window.

The `capStart = 0` fallback you suggested is still there for the case where the pattern isn't on the line at all — `anchor >= 0 && rawCtx.length > CONTEXT_HARD_CAP` — so a row with nothing to anchor to renders the head of the line.

---

crates/fresh-editor/plugins/search_replace.ts
sinelaw: Agreed — "keeps the historical head-truncation" was wrong once `capStart > 0`, and the stale-render path reaches it on every keystroke. Fixed both ways in `1d46cb9`:

- **At the source.** With no locatable hit the anchor collapses to 0, so `capStart` is 0 and the row really is the head of the line — the historical behaviour the comment claimed. Your `NEEDLE` → `NEEDLEX` case lands here.
- **As a backstop.** `windowAroundMatch` now takes `alreadySliced` and every fallback goes through a `head()` helper that emits `… + truncate(text, budget - 1)` when the input was already a mid-line window. A row that is one column narrower beats a row that lies about where it starts.

Both fallbacks stay within `budget`; I checked the width arithmetic across the branches (including the `charLen(text) <= budget` early return, where a bare prepend would have overflowed by one).

For the test I didn't want to race the stale-render window, so it drives the same state deterministically: regex mode with `(?i)NEEDLE_MARKER`, which the Rust engine accepts and JavaScript rejects — the search finds the matches, the plugin's `new RegExp` throws, and the renderer has no locatable hit.

Worth recording that the first version of that test passed against the bug. The fixture was filler plus a match, and "the head of the line" and "a slice of the filler" both start with `a` — so the assertion couldn't tell them apart. The fixture now carries a distinct `HEAD_MARKER`, and I re-checked it fails against the previous commit.

---

Issue #3149, opened 2026-09-02
locale from config.json is ignored in daemon (--server) mode


kirinriki7777-sys:
## Summary

`locale` set in `~/.config/fresh/config.json` has no effect when the editor runs as a daemon (`--server`). The documented way to choose a UI language ([`docs/i18n.md`](https://github.com/sinelaw/fresh/blob/master/docs/i18n.md)) only takes effect on the standalone path.

## Environment

- fresh 0.4.10 (AppImage, x86_64 Linux)
- `LANG=en_US.UTF-8`, no `LC_ALL` / `LC_MESSAGES` set

## Steps to reproduce

1. Put this in `~/.config/fresh/config.json`:

   ```json
   { "locale": "ja" }
   ```

2. Start a named daemon and attach to it:

   ```
   fresh --cmd daemon new mysession
   ```

3. The UI stays in English.

`fresh --cmd config show` does report `"locale": "ja"`, so the file is read — it just never reaches the i18n layer on this path. Changing the value to `"fr"` and re-running `config show` reflects the change, confirming the config itself is live.

Both of these do work: `fresh --locale ja …` and `LC_MESSAGES=ja_JP.UTF-8 fresh …`.

## Where it goes wrong

Line numbers are from tag `v0.4.10`.

`config.locale` reaches i18n in exactly one place, inside `initialize_app()`:

```rust
// crates/fresh-editor/src/main.rs:1924
let locale_override = args.locale.as_deref().or(config.locale.as_option());
fresh::i18n::init_with_config(locale_override);
```

Daemon mode returns before reaching it:

```rust
// crates/fresh-editor/src/main.rs:5085
if args.server {
    return Some(run_server_command(args, None));
}
```

`run_server_command()` does load the config (`Config::load_with_layers` → `editor_config`), but never calls `i18n::init_with_config`, so `config.locale` is dropped on this path.

The only i18n init that runs for a daemon is the pre-clap one:

```rust
// crates/fresh-editor/src/main.rs:5418
fresh::i18n::init_with_config(pre_clap_locale_override().as_deref());
```

and `pre_clap_locale_override()` scans raw argv for `--locale` only — it does not consult the config. With no `--locale`, `init_with_config(None)` falls through to `detect_locale()`, which reads `LC_ALL` / `LC_MESSAGES` / `LANG`. So the daemon's UI language follows the environment and ignores the config file.

[...]

Issue #3134, opened 2026-09-01
Deletion of temporary files for STDIN should be immediately on exit


Korkman:
fresh 0.4.10

Reproduce:
```bash
echo test | fresh -
```

... will create `/tmp/fresh-stdin-*.tmp`. Exit and it remains. Cleanup is part of [cleanup_legacy_tmp_logs()](https://github.com/sinelaw/fresh/blob/a7dca75c04ebc57c99a9786051729b55fdd72a2f/crates/fresh-editor/src/services/log_dirs.rs#L125) and thus only after 24h.

Not great when ingesting large decompressed logs 😉

I suggest deleting the file on exit.

sinelaw:
Reproduced on 0.4.10 (release tarball, x86_64-unknown-linux-musl), Linux, tmux 200x50, default config.

Steps:
1. In tmux: `echo hello-from-stdin | fresh -`
2. While the editor is open: `ls -la /tmp/fresh-stdin-*` → `-rw-r--r-- 17 /tmp/fresh-stdin-2271.tmp`
3. Ctrl+Q to quit (process exits, tmux session ends)
4. `ls -la /tmp/fresh-stdin-*` → **the file is still there**, unchanged

Confirms the report: nothing removes the file at exit, so it waits for the 24h `cleanup_legacy_tmp_logs()` sweep.

---
_Reproduced by an automated agent driving the editor interactively in tmux against the 0.4.10 release build._

---
_Generated by [Claude Code](https://claude.ai/code)_

sinelaw:
Re-checked on both builds — **this looks fixed on master** (`fbddf641`), by e920a745 (the spool file is never named, so there is no cleanup path to miss).

| build | `echo hello \| fresh -` → `/tmp/fresh-stdin-*` |
|---|---|
| 0.4.10 release (musl, via `install.sh`) | `/tmp/fresh-stdin-4533.tmp` while running, **still there after `Ctrl+Q`** |
| master `fbddf641` (debug) | no named file at any point, nothing left after quit |

Probably closable once 0.4.11 ships.

---
_Verified by an automated agent driving both builds interactively in tmux._

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3157, opened 2026-09-02
Three process-lifecycle fixes: the daemon's locale (#3149), the stdin spool file (#3134), and the signal handler that crashed on the way out

sinelaw:
## What each fix does

- **#3149** — Daemon mode returned *before* the one place that applied `config.locale`, and the spawn dropped the client's flags: `run_server_command` now initialises i18n from the config it already loaded, and forwards `--locale` and `--config` to the daemon it starts.
- **#3134** — The spool file stayed *named* so the buffer could page it back in, which made deleting it the process's job: it is now unlinked at creation and read through the retained descriptor, so the kernel reclaims it however the process dies.
- **Signal handler** (found while testing the above) — `SIGINT`/`SIGTERM` ran the whole diagnostic dump *inside the handler* — allocating, locking, logging, calling into plugins — and deadlocked or aborted: the handler now only writes the signal to a pipe, a waiting thread does the work, and a watchdog guarantees the exit.

## Before and after

`locale` rows assume `~/.config/fresh/config.json` has `{ "locale": "ja" }` and no `LC_ALL`/`LC_MESSAGES`/`LANG`.

| Steps | Before | After |
|---|---|---|
| `fresh calc.py` | `ファイル 編集 表示 選択 移動` | unchanged |
| `fresh --cmd daemon new s1` | **`File Edit View Selection Go`** | `ファイル 編集 表示 選択 移動` |
| `fresh --locale ja --cmd daemon new s2` (no `locale` in config) | **English** — flag dropped at the spawn | Japanese; `ps` shows `… --locale ja` |
| `fresh --config alt.json --cmd daemon new s3` | **English** — daemon ignored `alt.json` entirely | reads `alt.json` |

`echo hello-from-stdin \| fresh -`, then `ls /tmp/fresh-stdin-*`:

| Steps | Before | After |
|---|---|---|
| While the editor is still running | **`/tmp/fresh-stdin-<pid>.tmp` is there** | nothing — it never has a name |
| `Ctrl+Q` / `kill -TERM` / SIGHUP / `kill -INT` | **still there**, until the 24-hour sweep | nothing |
| `kill -9`, a segfault, a power cut | **still there** | nothing |
| Launched from a non-tty, so it aborts before the editor starts | **still there** | nothing |

`kill -TERM` on a running editor, ten runs on a loaded machine:

| Steps | Before | After |
|---|---|---|
| `kill -TERM <pid>` | **`SIGABRT`, `SIGSEGV`, or a clean exit, depending where the signal landed** | exit 130 every time, under 50ms |
| `kill -TERM` twice | second was **ignored** | second takes the default action — an immediate escape |

Both issues were reproduced on master (`cf626fc`) first by driving the editor in tmux with real keystrokes, and re-verified the same way.

## Notes on the two designs

**#3134.** The first draft of this PR kept the file named and deleted it on exit, which took an exit guard, two signal sweeps and a panic hook — and still leaked on `SIGKILL`. Review was right that this was four mechanisms for one invariant. The premise was wrong: the buffer needs to *read* the file, not to have a name for it. `services::stdin_spool` wraps the filesystem so the spool's opaque key (`fresh-stdin://<pid>`) resolves to the retained handle, with **positional** reads (`pread`/`seek_read`) so overlapping chunk loads share no cursor. That also avoids `/proc/self/fd` (needs procfs, absent on macOS/Windows) and macOS's `/dev/fd/N`, which `man 4 fd` documents as `dup(N)` — a shared offset, exactly the race positional reads avoid. Deleted with it: `services::temp_files`, the `CleanupOnDrop` guard, the SIGHUP handler, the signal-safe terminal restore, the panic-hook sweep.

**Signal handler.** Termination no longer depends on the report working: a watchdog ends the process with `_exit` on a deadline; a second signal restores the default disposition and re-raises; and if the pipes cannot be created, no handler is installed and the signals keep their prior behaviour. Two bugs the rewrite only revealed by being run: the sweep `tgkill`ed its own reporting thread (so it asked itself for a backtrace mid-allocation and deadlocked on glibc's arena lock) and its watchdog; and the collect loop spun on `yield_now`, starving the threads it was waiting for.

## One deliberate behaviour change — worth your call

**The per-thread backtrace sweep is now off by default**, behind `FRESH_SIGNAL_BACKTRACES=1`.

Capturing a backtrace allocates, and nothing here can change that — it needs a frame-walking primitive `std` does not expose, and glibc's `execinfo` is absent on musl, a target this ships on. Measured on a piped-stdin session under load: **with the sweep on, ~7 SIGTERMs in 10 cost a full watchdog deadline** (all still exited 130 — bounded, not fatal), plus ~2000 log lines on every ordinary Ctrl+C; **with it off, 10 of 10 exited in under 50ms.** The JS execution state — the cheap half, and the half that names what the editor was doing — is still reported on every terminating signal. One line to revert if you'd rather have it always on.

## Tests

The bugs live in the process, not the library, so these drive the real binary. `tests/common/pty.rs` allocates a pty and hands the child the slave as its controlling terminal (`fresh -` needs a pipe on fd 0 *and* a `/dev/tty`; the daemon client refuses to attach without a tty), reading back through `vt100`. Waiting is a blocking read on the pty — no sleeps or deadlines, per CONTRIBUTING's "semantic waiting".

- `daemon_locale_from_config.rs` — config file, forwarded `--locale`, forwarded `--config`, and an English control. The `--config` case is observed through the locale because that is what shows on screen.
- `stdin_spool_lifetime.rs` — one test per row above, each also asserting `$TMPDIR` is empty **while the editor runs**, plus `sigkill_leaves_nothing_behind`, which no cleanup handler could ever have passed.
- `spool_handle_probe.rs` — the platform property the design rests on. Green on Linux and macOS.
- `signal_termination.rs` — the signal is still reported; the sweep no longer runs unless asked for; it still works when asked for.
- Unit tests for the relay (a real signal, through a real handler, onto a waiting thread), the sweep's exclusion list, the spool and its decorator, and both platforms' daemon argv builders.

Each was checked against the unfixed code: reverting the i18n init and `--locale` forwarding fails two locale tests and not the control; reverting `--config` fails that one alone; leaving the spool named fails every `stdin_spool_lifetime` test at the while-running assertion; forcing the sweep back on fails `a_terminating_signal_is_reported_without_sweeping_every_thread` and not its companion.

`cargo fmt --check`, `clippy --all-features --all-targets`, `check --no-default-features --features runtime --all-targets`, and `check --all-targets --features gui` are clean.

## Left alone deliberately

The 24-hour `cleanup_legacy_tmp_logs` sweep of `fresh-stdin-*.tmp` is dead weight for new versions but stays — it is the only thing that clears files left by an older `fresh` on the same machine.

One portability fix worth naming: the pty helper originally typed the `ptsname_r` buffer as `[i8; N]`, which does not compile on aarch64/arm Linux where `c_char` is unsigned — a shipped target (AppImage, AUR) that x86_64-only CI never builds tests for.

https://claude.ai/code/session_01AiVTz9AGhjKKFJzvi3BfZk


crates/fresh-editor/src/services/signal_handler.rs
```diff
@@ -68,6 +74,51 @@ pub fn install_signal_handlers() {
...
+        SigSet::empty(),
+    );
+
+    // SAFETY: installing a signal handler; the handler itself only calls
+    // async-signal-safe `raise` plus the cleanup sweep, which is the same
+    // latitude the termination handler above already takes.
+    unsafe {
+        if let Err(e) = sigaction(Signal::SIGHUP, &action) {
+            tracing::error!("Failed to set SIGHUP handler: {}", e);
+        }
```
sinelaw: **The handler is installed unconditionally, which clobbers an inherited `SIG_IGN` and turns `nohup` into "dies on SIGHUP".**

`sigaction` here replaces whatever disposition the process inherited. When SIGHUP was inherited as `SIG_IGN` — exactly what `nohup fresh …` (and a number of launchers/supervisors) sets up — the process used to be immune to SIGHUP. After this change it gets `hangup_handler`, and because `SA_RESETHAND` resets to `SIG_DFL` (not back to `SIG_IGN`), the `raise` on line 103 terminates it.

Concrete: `nohup fresh -a mysession &`, then close the terminal. Before: survives (SIGHUP ignored). After: killed by SIGHUP. The PR description's claim that "the only added behaviour is the cleanup" holds for the default disposition but not for an inherited ignore.

The usual guard is to read the current disposition first and skip installing when it is already `SIG_IGN` (a query `sigaction`, or install-then-restore if the old handler came back as `SigIgn`).

---

crates/fresh-editor/src/services/signal_handler.rs
```diff
@@ -68,6 +74,51 @@ pub fn install_signal_handlers() {
...
+    extern "C" fn hangup_handler(_: libc::c_int) {
+        // The terminal is already gone, so there is nothing to restore and
+        // nowhere to print; just drop the scratch files.
+        crate::services::temp_files::cleanup_all();
+        // SAFETY: `raise` is async-signal-safe. SIGHUP is blocked for the
+        // duration of the handler, so this leaves it pending and it is
+        // delivered — to the default action, thanks to `SA_RESETHAND` — as
+        // soon as the handler returns.
+        unsafe { libc::raise(libc::SIGHUP) };
+    }
```
sinelaw: **`cleanup_all()` inside the handler is async-signal-unsafe, and if it blocks the `raise` below never happens — SIGHUP stops being fatal.**

`cleanup_all` takes `REGISTERED` (a `std::sync::Mutex`) and, for every path it removes, calls `tracing::debug!`. `tracing_setup::build_subscriber` adds `Level::DEBUG` as the base directive, so that `debug!` is *enabled* in a normal editor run and really goes through `EnvFilter` (`RwLock`) and the layer stack.

Concrete: `zcat big.log | fresh -` is running (so the registry is non-empty and the `debug!` on removal actually fires); the user closes the terminal window. SIGHUP is delivered to whichever thread has it unblocked, and if that thread was already inside the tracing subscriber — the editor logs continuously — the handler blocks on a non-reentrant lock. It never returns, so the pending SIGHUP is never unblocked and never delivered, and instead of dying the process becomes an orphan spinning on a pty nobody owns. With the previous default disposition, SIGHUP always killed it immediately.

The SIGTERM/SIGINT handler takes the same latitude, but it ends in `process::exit(130)` — reaching the exit is the whole point there, and a hang there is equally new for SIGHUP. Restricting the sweep to `unlink(2)` (async-signal-safe) with no locking and no `tracing` would keep the property the doc comment promises.

---

crates/fresh-editor/src/services/signal_handler.rs
```diff
@@ -68,6 +74,51 @@ pub fn install_signal_handlers() {
...
+/// so the `raise` below terminates the process *by SIGHUP*, exactly as it
+/// would have with no handler installed. The only added behaviour is the
+/// cleanup.
+#[cfg(unix)]
+fn install_hangup_handler() {
+    use nix::sys::signal::{sigaction, SaFlags, SigAction, SigHandler, SigSet, Signal};
+
+    extern "C" fn hangup_handler(_: libc::c_int) {
+        // The terminal is already gone, so there is nothing to restore and
+        // nowhere to print; just drop the scratch files.
```
sinelaw: **"The terminal is already gone" is only true for one of the two ways SIGHUP arrives.**

SIGHUP is also delivered by an explicit `kill -HUP <pid>`, by a session leader exiting while the pty survives, and by `tmux`/`screen` teardown paths that leave the pane alive. In those cases the terminal is very much still there, and the process now exits from this handler (via the re-raise) with raw mode and the alternate screen still set — the user is left with a wrecked shell that needs `reset`.

That was true before this PR too (no handler at all), so it is not a regression, but this is now the one place that knows a SIGHUP is being handled, and the fix is a line: call `terminal_modes::emergency_cleanup()` (already used from the panic hook) before `cleanup_all`, unconditionally — it is a no-op when the tty is gone.

---

crates/fresh-editor/tests/stdin_temp_file_cleanup.rs
```diff
@@ -0,0 +1,227 @@
...
+fn isolated_fresh(home: &Path) -> Command {
+    let mut cmd = Command::new(env!("CARGO_BIN_EXE_fresh"));
+    cmd.current_dir(home)
+        .env("TMPDIR", home)
+        .env("HOME", home)
+        .env("XDG_CONFIG_HOME", home.join("config"))
+        .env("XDG_DATA_HOME", home.join("data"))
+        .env("XDG_STATE_HOME", home.join("state"))
+        .env("XDG_CACHE_HOME", home.join("cache"))
+        .env("TERM", "xterm-256color");
```
sinelaw: **This helper isolates `HOME`/`XDG_*`/`TMPDIR` but not the locale, so every test in this file hangs on a non-English machine.**

`running_editor_with_piped_stdin` waits on `screen.contains("File") && screen.contains("Help")`. `LANG`/`LC_ALL`/`LC_MESSAGES` are inherited here (unlike `daemon_locale_from_config.rs`, which deliberately sets `LANG=C.UTF-8` and `env_remove`s the other two), and `i18n::init_with_config(None)` falls back to `fresh_i18n::detect_locale()`, which reads exactly those three.

Concrete: a developer with `LANG=ja_JP.UTF-8` (or `de`, `fr`, …) runs `cargo test`. The editor draws `ファイル … ヘルプ`, the predicate never matches, and `fresh` is interactive so it never exits — `wait_for_screen` blocks on `read` forever. Under plain `cargo test` that is an indefinite hang; under nextest it is a 180s `terminate-after` kill reported as a failure. All four tests in this file go through this helper (the early-failure one only via `isolated_fresh`, so it is unaffected, but the other three are not).

Same three lines as the sibling file fix it:

```suggestion
        .env("XDG_CACHE_HOME", home.join("cache"))
        .env("TERM", "xterm-256color")
        // The screen predicates below are English. Pin the language the same
        // way `daemon_locale_from_config.rs` does, so a developer with
        // `LANG=ja_JP.UTF-8` does not get a menu bar these tests can never
        // match — and an interactive editor that never exits means the wait
        // never ends.
        .env("LANG", "C.UTF-8")
        .env_remove("LC_ALL")
        .env_remove("LC_MESSAGES");
```

---

sinelaw (commented):
Read both fixes end to end against the surrounding code.

The locale half looks correct to me: `resolve_locale_override` is called in `run_server_command` before `EditorServer::new`, and `Editor::new` is where `MenuConfig::translated()` is materialised (`app/editor_init.rs:646`), so the re-init genuinely lands before the strings are built; `i18n::init_with_config` → `fresh_i18n::set_locale` is idempotent, so calling it a second time after the pre-clap init is fine. Both `spawn_server_detached` call sites (`run_attach`, `run_open_files_command`) and the one `run_attach` caller were updated, and the argument order at the `run_open_files_command` call site matches the new signature.

The temp-file half: guard placement in `real_main` checks out — I walked the `process::exit` sites reachable from it and they are all CLI subcommands that return long before `start_stdin_streaming`, and there is no `exec`/`abort` path. The `#[cfg(panic = "abort")]` narrowing is right (`min-size` is the only aborting profile), and `open_stdin_buffer` clears the file path, so nothing persisted references the spool file after deletion.

Four things below — three on the new SIGHUP handler, one on test isolation. The `nohup` one is a behaviour regression rather than a nit.

---

crates/fresh-editor/src/services/signal_handler.rs
sinelaw: Confirmed and fixed in ccbdf52 — this was a real regression, and my "the only added behaviour is the cleanup" was wrong exactly where you say.

The disposition is now queried first (`libc::sigaction` with a null `act`, which is a pure query) and an inherited `SIG_IGN` is left untouched, with a `debug!` noting why.

There is a regression test for it: `sighup_ignored_by_the_parent_stays_ignored` starts the editor with a `pre_exec` that does `signal(SIGHUP, SIG_IGN)` — what `nohup` does to its child — sends it a SIGHUP, and asserts it is still alive by quitting it with Ctrl+Q and checking it exited successfully rather than by the signal. Verified against the unfixed code: with the guard removed the test fails, killed by the SIGHUP it should have ignored.

I left SIGINT/SIGTERM installing unconditionally, since that predates this PR and changing it would be a separate behaviour decision — happy to do it here if you'd rather.

---

crates/fresh-editor/src/services/signal_handler.rs
sinelaw: Agreed, and fixed in ccbdf52. The orphan-instead-of-dying outcome is worse than the leak this whole module exists to fix, so the handler now takes no locks at all.

`temp_files::register` additionally writes each path into a small lock-free table of leaked, NUL-terminated C strings, and `cleanup_from_signal()` sweeps it with a `swap` plus `unlink(2)` per entry — no mutex, no `tracing`, no allocation. The strings are deliberately never freed (`free` is not async-signal-safe and the process is on its way out), and taking each slot with a `swap` makes a second or concurrent call a no-op rather than a double unlink. `cleanup_all` keeps the `Mutex` + `tracing` version for the normal exit paths, where locking is fine.

Both termination handlers now use the signal-safe sweep, not just SIGHUP — you're right that a hang before `exit(130)` is equally new there, and it was my `cleanup_all` call that introduced it.

---

crates/fresh-editor/src/services/signal_handler.rs
sinelaw: You're right about the premise — `kill -HUP`, a session leader exiting while the pty survives, and tmux/screen teardown all leave a live terminal — and the doc comment claiming otherwise is now fixed.

I couldn't take the suggested line as written, though: `emergency_cleanup()` goes through crossterm, which allocates and locks, so calling it here is exactly what your other comment rules out. Doing both as proposed would have traded a wrecked shell for the orphan-process hang.

So ccbdf52 implements the intent in a signal-safe way. `terminal_modes::emergency_cleanup_from_signal()` writes the same teardown bytes with a single `write(2)`, and restores cooked mode with `tcgetattr`/`tcsetattr` — all three are on POSIX's async-signal-safe list. It deliberately does not read crossterm's saved termios (that would take the lock we're avoiding); it re-enables the flags a shell needs, `ICANON | ECHO | ISIG | IEXTEN` and `OPOST`, which is what makes the terminal usable again without a `reset`. Best-effort throughout, so it's a no-op when the tty really is gone.

One thing to keep an eye on: the byte sequence is a literal, because the `sequences` constants can't be concatenated in a `const`. I've noted in the doc comment that it mirrors `terminal_teardown_sequences()` in `server::capture_backend` and the two need to stay in step. If you'd prefer that expressed as something the compiler enforces, say the word.

---

crates/fresh-editor/tests/stdin_temp_file_cleanup.rs
sinelaw: Applied verbatim in ccbdf52 — thanks, that was a genuine hole in the test and the failure mode is as bad as you describe, since the editor is interactive and never exits on its own.

The one thing I'd add to the diagnosis: it also means these tests were only ever exercised in an English environment, so a locale-dependent regression in the spool-file path would have shown up as a hang rather than a failure. Pinning the locale makes them deterministic in both directions.

The new `sighup_ignored_by_the_parent_stays_ignored` test goes through the same helper, so it picks this up too.

---

sinelaw:
`test windows-latest` went red on ccbdf52. I don't believe it belongs to this PR, and here is the reasoning rather than an assertion.

**What failed.** 9108 of 9109 passed; one timed out:

```
TIMEOUT [180.017s] (5196/9109) fresh-editor::all_tests
  e2e::plugins::review_diff_line_staging::test_review_visual_discard_status_is_localized
```

**Why it isn't this PR's.** On Windows the whole of ccbdf52 compiles down to doc-comment edits plus one `drop(paths)` — an explicit drop of a lock guard that was going out of scope on the next line anyway. Everything else in it is `#[cfg(unix)]` (`temp_files::signal_safe`, `cleanup_from_signal`, `emergency_cleanup_from_signal`, the SIGHUP handler), and the test file it touches is `#[cfg(target_os = "linux")]`. So the Windows binary is behaviourally the same one that passed this job on 88a3679 and again on ff49a4c.

The timed-out test is also nowhere near the diff: it drives the review-diff plugin over a real git repo, and neither this PR's locale path nor its temp-file path is reachable from it.

**What it looks like instead.** A timeout, not an assertion — and the test waits with `wait_for_screen_contains`, which is indefinite by design. Its own comment anticipates this exact failure mode ("instead of hanging to nextest's external timeout"). The whole Windows leg took 2016s, so a slow runner plus an indefinite wait on git and the plugin runtime is the obvious candidate. That makes it a pre-existing fragility on this runner rather than something ccbdf52 introduced, and there is no fix to port: nothing has changed in that test's path.

**What I've done.** Re-run the failed job once, which is the one re-run I'll spend on this. If it comes back green, that confirms the reading above. If it fails again — the same test or another — I'll treat it as real and root-cause it rather than re-running again.

I have deliberately not touched the test. Making an indefinite wait bounded, or marking it slow, would be a change to someone else's test to get my PR green, which is the wrong trade; if it turns out to be genuinely flaky on Windows it deserves its own issue.

---

Issue #3125, opened 2026-09-01
[Bug] Inconsistent multi cursor input behavior


X-Ryl669:
Let's say you're editing C++ code and want to type `a.open();` (with multi cursor, let's say on 2 different lines).
When you're typing the opening parenthesis `(`, fresh appends the closing parenthesis (*good*), then you type the closing parenthesis (not required, but it should work), fresh doesn't do anything (it should, at least move all cursors to the right, *after the closing parenthesis*). If you then type `;`, it's inserted between the parenthesis, not after.

Also, when you have multicursor active (let's you are on a word and press Ctrl + D to select a duplicate), then click somewhere else with your mouse, the last cursor is moved to where you click, but it doesn't undo the multicursor mode. So if you don't observe it (the other cursors can be out of the current viewport), and start typing, you're breaking code at random places. When clicking anywhere with a multicursor mode already defined (not while defining with Alt+Shift selection), it should cancel the multicursor mode. Also, typing is very slow in that case (I guess my file is big, but it's not a reason why it should be *that* slow).

sinelaw:
Both halves reproduced on 0.4.10 (release tarball, x86_64-musl), Linux, tmux 100x15, default config.

### 1. Typing `)` over an auto-closed `)` is ignored when multiple cursors are active

`mc.cpp` = two lines, each just `a`. Put the cursor at end of line 1, then `Ctrl+Alt+Down` (status bar confirms `2 cursors`). Type `.open(`, then `)`, then `;`.

| | 1 cursor (control) | 2 cursors |
|---|---|---|
| after `.open(` | `a.open()` — Col 8 | `a.open()` ×2 — Col 8 |
| after `)` | `a.open()` — **Col 9** (skipped over) | `a.open()` ×2 — **Col 8** (nothing happened) |
| after `;` | `a.open();` | **`a.open(;)`** ×2 |

So the type-over exists and works with one cursor, and is simply not applied when more than one cursor is live — which is what puts the `;` inside the parens.

### 2. A mouse click does not clear the existing multi-cursor

`mc2.txt`:
```
alpha one
alpha two
alpha three
```
`Ctrl+D` ×2 → status `2 cursors` (`alpha` selected on lines 1 and 2). Now click on line 3 with the mouse (SGR `ESC[<0;12;5M` / `…m`).

Status becomes `Ln 3, Col 6` — but still **`2 cursors`**. Typing a single `Z` then gives:

```
[...]

PR #3158, opened 2026-09-02
Multi-cursor: a keystroke that only moves cursors is no longer dropped

sinelaw:
Fixes the type-over half of #3125. Everything this PR does **not** fix is now filed separately: #3165, #3166, #3167. Nothing here touches indentation.

> **Corrections to the first revision of this description.** An adversarial review of the first commit found three claims of mine that were wrong. They are corrected in place below, and called out here so the record is not quietly rewritten:
> 1. I claimed mixed insert + skip-over lists were "unaffected… verified separately". **False** — I had only tested one of the two orderings. The other is still broken (now #3166).
> 2. I claimed the diff "drops a per-keystroke `format!` on the single-cursor path". **False** — master already had that `format!` inside the multi-cursor branch. The diff removed a `String` clone, nothing more. (This wrong sentence is in commit `d4c677e`'s message; I have not rewritten pushed history to erase it.)
> 3. I claimed #1425 "no longer reproduces". **Too broad** — the reported shape is fixed, but a blank line at end-of-file still reproduces it (now #3165).

## #3125 — typing over an auto-closed `)` was ignored with multiple cursors

Reproduced first, exactly as the issue's repro comment describes (debug build, driven interactively in tmux with real keystrokes, asserting on the saved file rather than only the screen):

| | 1 cursor | 2 cursors, before | 2 cursors, after |
|---|---|---|---|
| after `.open(` | `a.open()` — Col 8 | `a.open()` ×2 — Col 8 | `a.open()` ×2 — Col 8 |
| after `)` | `a.open()` — Col 9 | `a.open()` ×2 — **Col 8, nothing happened** | `a.open()` ×2 — Col 9 |
| after `;` | `a.open();` | **`a.open(;)`** ×2 | `a.open();` ×2 |

### Cause

Not in the auto-close logic. `insert_char_events` emits the correct skip-over for *every* cursor, and its multi-cursor unit tests already covered that — which is why this was invisible below the app layer.

The events were discarded one level up. `apply_events_as_bulk_edit` returned `None` for an event list holding no `Insert`/`Delete`, leaving each caller to notice and apply the moves itself:

```rust
if !has_buffer_mods {
    // No buffer modifications - use regular Batch
    return None;
}
```

A skip-over's entire effect *is* cursor movement, so with two cursors the list was two `MoveCursor` events, the bulk path declined it, and nothing applied them. With one cursor a different branch applied the event directly, which is why the single-cursor case always worked.

### Fix

Nine call sites reach that function; **two** remembered to handle the cursor-only case. So the same bug was live behind `DeleteBackward`: backspace in virtual space has nothing to delete yet and emits only `MoveCursor`, and with two cursors past the end of a line it did nothing at all, while one cursor worked. Verified by hand and now covered by a test.

The fix therefore went into the shared function rather than the call site: it applies a cursor-only list as one `Batch` and returns it, so every caller logs it through the path it already had. The per-site splits in `handle_insert_char_editor` and `apply_action_as_events` are deleted — behaviour unchanged for the latter, since a `Batch` is what it was already building. Net effect is fewer branches than master, and CONTRIBUTING rule 12 ("when a bug recurs, consider centralizing") points the same way.

### Verification

- `e2e/multicursor.rs`: regression test driving real key events (`End`, `Ctrl+Alt+Down`, characters), asserting buffer *and* screen. Fails on master, passes here. Asserts `2 cursors` in the status bar first so it cannot pass vacuously.
- `e2e/virtual_space.rs`: regression test for the backspace variant. Fails on the first commit of this branch, passes on the second.
- Local suites green: `indent` (161), `indent_dedent` (32), `multicursor` (59), `virtual_space` (28), `popup` (130), `paste` (97), `auto_indent` (78), `bracket` (65), `smart_editing` (47), `undo_redo` (33), `auto_close` (27). Unit tests: `fresh-editor --lib` 2524, `fresh-editor-core --lib` 1294.
- `cargo fmt` clean; `cargo clippy --all-targets` 0 errors, no new warnings.
- Checked by hand and unchanged: dirty flag, undo/redo stepping, LSP `didChange` (a cursor-only Batch is not `modifies_buffer()`, so none is sent), preview promotion, marker replay.

## What this PR does not fix

Each verified by hand on this branch's head and filed with symptom + steps:

- **#3166 — mixed insert + skip-over.** When one cursor inserts and another steps over an existing `)` at a *higher* offset, the skipping cursor still lands one byte short, giving `a(;)` — the same symptom as #3125 from the other end. Reproduces identically on master; **not** a regression from this PR.

  I attempted the fix and backed it out. The applier has to know which coordinate space each `MoveCursor` is in: skip-over emits pre-edit coordinates, indent emits post-edit. "Cursor owns no edit" looks like the discriminator but is not — indent does not attribute an `Insert` to each cursor, so that guess fixes skip-over and **breaks multi-line-selection indent** (`test_multicursor_indent_with_selections` fails, the cursor overshooting by one indent width). Separating them properly means carrying the coordinate space on the event rather than inferring it, which is a design change beyond this PR.

- **#3165 — blank line at end of file auto-indents to the previous block.** Split out of #1425 (now closed): the shape reported there is fixed, but a blank line with nothing non-empty after it still lands the cursor at `Ln 6, Col 9` and saves `        x`. Left alone because the same path deliberately indents a blank line at the end of an *unclosed block* (`test_indent_after_empty_line_incomplete_syntax` asserts `Some(4)` for that shape in Rust), which is wanted for code — so making plain text differ at end-of-buffer is a product call.

- **#3167 — status bar shows a stale line number after `Ctrl+Alt+Down`.** Pre-existing; the column is right and the line number catches up on the next real cursor move. Unrelated to this fix, noticed while testing it.

- **The second half of #3125** — whether a mouse click should collapse an existing multi-cursor — is a behaviour decision left alone deliberately, so **#3125 stays open** for it. It is real: after `Ctrl+D` ×2 a click moves the cursor but the status bar still reads `2 cursors`, and one keystroke then edits two places.

https://claude.ai/code/session_01VkQhXSRFAHBvzSxeeaHGwy


crates/fresh-editor/src/app/event_apply.rs
```diff
@@ -264,8 +264,28 @@ impl Editor {
...
+            //
+            // Returning `None` and leaving this to the caller is what made
+            // typing `)` over an auto-closed `)` do nothing at all with more
+            // than one cursor: only two of this function's callers remembered
+            // to special-case it, so the rest dropped the keystroke outright
+            // (#3125).
+            if events.is_empty() {
+                return None;
+            }
+            let batch = Event::Batch {
```
sinelaw, line 300: The function's contract changed here but its doc comment (lines 241-253) still says otherwise:

```
/// Apply multiple Insert/Delete events efficiently using bulk edit optimization.
/// # Arguments
/// * `events` - Vec of Insert/Delete events (sorted by position descending ...)
/// # Returns
/// The BulkEdit event that was applied, for tracking purposes
```

After this change the `Some(_)` can be either an `Event::BulkEdit` or an `Event::Batch`, and the accepted input is no longer "Insert/Delete events". The eleven existing call sites all just forward the value to `active_event_log_mut().append(...)`, so nothing breaks today — but they also all bind it as `bulk_edit`, which reinforces the stale contract. A future caller that reasonably trusts the doc and does something like `if let Some(Event::BulkEdit { edits, .. }) = self.apply_events_as_bulk_edit(...)` would silently drop the cursor-only case — exactly the class of bug this PR is fixing one level up.

Worth updating the doc (and ideally renaming the local bindings) to say that a cursor-only list is applied and returned as a `Batch`.

---

sinelaw (commented):
Reviewed for correctness (diff read hunk-by-hunk, all 12 call sites of `apply_events_as_bulk_edit` audited, `cargo check --tests` clean, new tests + the multicursor / virtual_space / undo_redo / auto_close / bracket / smart_editing / auto_indent / indent / paste / popup / clipboard / replace / search / cursor / comment / completion suites all green locally). No correctness bugs found. One low-severity note inline about the now-stale doc contract on the changed function.

---

crates/fresh-editor/src/app/event_apply.rs
sinelaw: Agreed, and fixed in 458d114 — both parts.

The doc now names the two shapes of input and says what each returns, and states outright not to match on the payload, citing exactly the failure mode you describe:

> `Some(Event::BulkEdit)` for the first shape, `Some(Event::Batch)` for the second, and `None` only for an empty list. **Match on the payload at your peril:** every caller today just forwards it to `active_event_log_mut().append(...)`, and a caller that pattern-matched `Some(Event::BulkEdit { .. })` would silently drop the cursor-only case — which is the bug this function's own contract used to invite (#3125).

The bindings are renamed to `applied` at all eleven call sites (`search_ops`, `action_dispatch`, `popup_actions`, `text_ops`, `clipboard`, `input_helpers`). Two `bulk_edit` names deliberately stay: the local inside the bulk path itself, which really is a `BulkEdit`, and the unrelated binding in `lsp_requests.rs:3195` that comes from a different function.

`cargo fmt` and `clippy --all-targets` clean; re-ran `multicursor`, `virtual_space`, `clipboard`, `paste`, `popup`, `search`, `replace`, `undo_redo`, `auto_close`, `indent` and `--lib` (2524) — all green.

---

Issue #1572, opened 2026-04-15
Inlay hint position after closing brace is weird - moves with newlines and whitespace


sinelaw:
<img width="1920" height="1144" alt="Image" src="https://github.com/user-attachments/assets/e5a0f753-40d6-4b94-b4e7-f669338c2504" />

PR #3159, opened 2026-09-02
fix(lsp): silent request timeouts, and inlay-hint drift (#2197, #722)

### LINKED ISSUE #722: LSP inlay hints sometimes rendered in wrong place
See "Duration" in the example below

Image

### LINKED ISSUE #2197: Pyright LSP: all request-based features (hover, definition, completions) timeout after 30s while LSP shows "ready"
## Steps to reproduce

1. Install pyright: `pip install pyright` (tested with v1.1.408, `pyright-langserver` available on PATH)
2. Create `~/.config/fresh/config.json`:
   ```json
   {
     "lsp": {
       "python": {
         "command": "pyright-langserver",
         "args": ["--stdio"],
         "enabled": true
       }
     }
   }
   ```
3. Create a small Python project:
   ```
   mkdir /tmp/py_test && cd /tmp/py_test
   cat > main.py << 'EOF'
   def add(a: int, b: int) -> int:
       return a + b

   class Calculator:
       def compute(self, x: int, y: int) -> int:
           return add(x, y)

   calc = Calculator()
   print(calc.compute(3, 4))
   EOF
   ```
4. Launch Fresh from the project directory: `fresh --no-restore main.py`
5. Wait for status bar to show `LSP (on)` and `LSP (python) ready`
6. Place cursor on `add` at line 6 (inside `Calculator.compute`)
7. Press `F12` (Go to Definition) — wait 30 seconds
8. Press `Alt+K` (Hover) — wait 30 seconds
9. Press `Ctrl+Space` (Show Completions) at `calc.` — wait 30 seconds

## Expected behavior

- `F12` should jump the cursor to the definition of `add` at line 1
- `Alt+K` should show a hover popup with the function signature `(a: int, b: int) -> int`
- `Ctrl+Space` should show a completion popup listing `compute` and `__init__`

These are standard LSP features that pyright provides. The `pyright` CLI confirms it can analyze the file (`pyright main.py` → `0 errors, 0 warnings, 0 informations`).

## Actual behavior

All LSP request-based features **silently fail** after a 30-second timeout:

- No cursor jump for Go to Definition
- No hover popup (nothing shown, no error message)
- No completion popup
- Status bar continues to show `LSP (on)` with no indication of failure

The Fresh log (`~/.local/state/fresh/logs/warnings-<PID>.log`) shows the timeouts:
```
WARN fresh::services::lsp::async_handler: LSP request 'textDocument/definition' (lsp_id=1) on 'pyright-langserver' (python) timed out after 30s
WARN fresh::services::lsp::async_handler: LSP request 'textDocument/hover' (lsp_id=2) on 'pyright-langserver' (python) timed out after 30s
WARN fresh::services::lsp::async_handler: LSP request 'textDocument/completion' (lsp_id=3) on 'pyright-langserver' (python) timed out after 30s
```

The `fresh-<PID>.log` confirms the initialize handshake succeeded:
```
INFO fresh::services::lsp::manager: LSP initialization started for python (pyright-langserver)
INFO fresh::services::lsp::async_handler: Async LSP server initialized successfully
INFO fresh::app::async_dispatch: LSP server 'pyright-langserver' initialized for language: python
```

So the server starts and initialize succeeds, but all subsequent requests (hover, definition, completion, signatureHelp) time out.

The diagnostics panel shows "0 items" for both Current File and All Files — pyright is not publishing diagnostics either, despite the `[⚠ N]` counter in the status bar incrementing (which appears to count LSP timeout warnings rather than code diagnostics).

## Additional observations

- `pyright-langserver` is a Python wrapper that spawns a Node.js process (`node langserver.index.js -- --stdio`). The Python process (PID X) and Node process (PID Y) both run during the session.
- Both processes are in "sleeping" state, not consuming CPU, suggesting they are waiting for input rather than stuck in processing.
- The `LSP initialize result: position_encoding=None` log entry may be relevant — if Fresh assumes UTF-16 positions while pyright expects a different encoding (or vice versa), the server may silently discard all requests.

## Workaround

None known. `pylsp` (python-lsp-server) is the default configured Python LSP and may work if installed, but was not tested.

## Environment

- Fresh: 0.3.10 (built from `claude/awesome-clarke-VmLci`, 2026-06-02)
- pyright: 1.1.408 (`pyright-langserver` from `pip install pyright`)
- Platform: Linux (tmux terminal, 220×50)
- Python: 3.11 (via uv pyright environment)

## Search queries used (no open duplicates found)

- `repo:sinelaw/fresh pyright timeout LSP`
- `repo:sinelaw/fresh LSP hover definition timeout`
- `repo:sinelaw/fresh lsp requests timeout pyright langserver`
- `repo:sinelaw/fresh pyright langserver not responding`

---
*Filed by tui-agent-auto-bug (automated TUI QA agent, Run #17, 2026-06-02)*

sinelaw:
Fixes #2197. Fixes #722.

### #2197 — pyright 1.1.408 never answers

1. `npm i pyright@1.1.408`, configure `pyright-langserver --stdio` for Python, open a file with `def add(a, b): …` and a call `add(1, 2)`; the status bar reads `LSP (python) ready`.
2. Put the cursor on `add` in the call, press `F12`. Then `Alt+K`.

| | before | after |
|---|---|---|
| `F12` | nothing for 30s, then `No definition found`; status bar still `LSP (python) ready` | jumps to `def add` immediately |
| `Alt+K` | nothing for 30s, then `No hover information available`; still `ready` | hover popup with the signature |

The server is unblocked by a `workspace/didChangeConfiguration` pushed right after `initialized`, which VS Code, Neovim and Helix all send (details below). For a server that stays silent anyway — a stub that answers `initialize` and nothing else reproduces it — the timeouts are no longer invisible:

| | before | after |
|---|---|---|
| after the first 30s | `No definition found`; status bar still `ready` | `LSP (python): 'textDocument/definition' timed out after 30s` |
| after the second | `No hover information available`; still `ready` | pill turns `LSP (stuck)` (warning colours); `Alt+K` popup shows `◍ pyright-langserver (not responding)` with Restart/Stop |
| first answered request afterwards | — | pill back to `LSP (on)`, popup row back to `(ready)` |

### #722 — inlay hint jumps to the next line on Enter

1. Rust project with rust-analyzer and inlay hints on; a line `let d = Duration::from_secs(5);` shows the hint `: Duration` after `d`.
2. Put the cursor immediately after `d` (on the hint's anchor) and press Enter.

```
before, for ~1s until the next inlayHint refresh:     after, first frame:
  5 │     let d                                           5 │     let d : Duration
  6 │     : Duration  = Duration::from_secs(5);           6 │      = Duration::from_secs(5);
```

The "after" frame is identical to what the refresh produces for the split buffer, which the regression test asserts directly.

> **The Deno half (#2981) has been removed from this PR.** #3111 took that fix and #3168 moved the detection into a plugin; both are on master. `configure_lsp_servers` here is master's version verbatim.

## #2197 — what changed

**The hang is pyright 1.1.408's.** With the client advertising `workspace.workspaceFolders`, its `initialized` handler skipped `updateSettingsForAllWorkspaces()`, so the workspace created during `initialize` never resolved its `isInitialized` promise and every request that awaits it waits forever ([microsoft/pyright#11239](https://github.com/microsoft/pyright/pull/11239), shipped in 1.1.409; 1.1.408 was the only affected npm release, Jan 8 – Apr 23, 2026). Other editors never saw it because their clients push a `workspace/didChangeConfiguration` right after `initialized` — VS Code via `synchronize.configurationSection`, Neovim when `settings` is set, Helix when `config` is set — and pyright's handler for that notification happens to call the same `updateSettingsForAllWorkspaces()`. Fresh sent no such notification and hit the bug head-on.

**Compatibility push.** Fresh now sends `workspace/didChangeConfiguration` once, right after `initialized`, with the same object it answers `workspace/configuration` from (the server's `initialization_options`, or `{}`), so a server sees one configuration whichever way it asks. Not required by the spec, but servers are written against the clients that do it. Verified against the real pyright 1.1.408 in tmux: `F12` jumps, `Alt+K` hovers.

**Visibility.** What was Fresh's bug is that a silent server looked healthy. Requests are cancelled after 30s, which is correct, but the user saw only an empty result, indistinguishable from a symbol that genuinely has no definition, with `ready` on the status bar throughout. Three surfaces now:

1. **An expired request the user made is named** on the status bar. Deliberately *not* the requests the editor makes on its own — inlay hints, diagnostics, semantic tokens, folds, and (per review) completion and signature help, which fire on ordinary typing. Against a wedged server those would claim the status bar once per keystroke, 30s later. They still count towards the streak; the indicator is their surface.
2. **An empty result that follows an expiry of the same request explains itself** — `LSP (python): no answer — 'textDocument/definition' timed out after 30s` in place of `No definition found`. Keyed by method, consumed on use, and dropped when the server answers again, so it explains exactly one empty result and never a later, genuine one. Universal servers are matched by scope, since their records are labelled `"universal"` rather than by language.
3. **Two expiries in a row with nothing answered in between** put the server in a new `LspServerStatus::Unresponsive`: the `LSP (stuck)` pill (same fixed width, nothing else on the bar shifts) and the `(not responding)` popup row. The first answered request clears it back to `Running`, and both transitions re-render an open popup. A server reporting `$/progress` keeps its spinner, so an indexing rust-analyzer or jdtls is never called stuck. `Unresponsive` also counts as "was running" for the crash path, so a wedged server that then dies still gets its diagnostics cleared and its auto-restart.

A request the editor cancels itself no longer counts towards that streak: `handle_cancel_request` withdrew the request from `active_requests` but left the pending entry in place, so a server that ignores `$/cancelRequest` parked the requester for the full 30s and the expiry looked exactly like a hang — and the editor cancels superseded completion, definition and semantic-token requests on ordinary cursor movement. Cancelling now drops the pending entry, which also resolves the superseded requester immediately rather than 30s later.

**Open question:** `UNRESPONSIVE_TIMEOUT_STREAK = 2` and the "expired" phrasing are judgement calls, and something louder — a popup, or offering a restart when a server goes quiet — is a product decision I left alone. Also open, from review: classifying requests by method name is a proxy for what we actually want to know (did the user ask for this?), and threading that from the call site would be the real fix.

## #722 — what changed

Hints are byte-offset markers, and the marker was carried along by text inserted at the hint's own anchor. The inlay-hint path now asks for a left-gravity marker, so an insertion at the anchor stays behind the hint. Gravity is an explicit `MarkerGravity` argument rather than something inferred from `BeforeChar`/`AfterChar`: the two are not the same question. The async-paste placeholder is also a `BeforeChar` entry, but its marker *is* the eventual insertion point, so it must keep travelling with text typed at it. `AfterChar` hints keep right gravity, which the #1572 end-of-line anchoring relies on.

Gravity travels with the decoration into the wrap index too. `IndexDecorations::shift_for_edit` keeps its own copy of these anchors and shifted them all with right gravity by contract, so without this the index would have placed the hint on the next line while the renderer drew it on this one — moving the drift from the renderer into the coordinate service (row counts, cursor visual columns, scrollbar extent, h-scroll) for the same ~1s window.

Scope note: this fixes the Enter-at-the-anchor case from the repro. Undo of that split still shows a brief one-column shift before the refresh lands — the marker survives the undo correctly, but the renderer's end-of-line hint padding differs by a space; a smaller artifact of the same class, not addressed here.

## Tests

* `issue_2197_lsp_timeout_visible.rs` — the settings push follows `initialized`, precedes the first `didOpen`, and happens exactly once (logging fake server); a user-invoked timeout reaches the status bar while background and auto-triggered ones do not; an unresponsive server is not rendered as ready, recovers, and is still treated as a crash if it dies while quiet; an empty result after a timeout says the request expired, but only for the method that expired, only once, and not after the server has answered again.
* `issue_722_inlay_hint_split_drift.rs` — Enter at a hint anchor keeps the hint on its line and already matches the refreshed rendering; a parameter hint stays in front of text typed at its anchor. Both fail without the fix. `view::virtual_text` pins the gravity contract, `view::wrap_index` pins that the index shifts each anchor by its own gravity, and the pre-existing marker-tracking test in `state.rs` is untouched from master because the paste anchor's behaviour is untouched.
* `services::lsp::async_handler` — unit tests for the streak transitions (first expiry, the crossing, reset by an answer, recovery reported once), which had no coverage.
* `app::lsp_status::tests` — `LSP (stuck)` for an unresponsive server, error still outranks it, pill width unchanged.

Both issues were reproduced on master and re-verified after the fix by driving the editor interactively in tmux; the existing LSP handshake and ordering tests pass with the extra notification. `cargo fmt --check` and `cargo clippy -p fresh-editor --all-targets` are clean on the touched files. Note for anyone reproducing locally: plain `cargo test` is not equivalent here — `wait_until` has no timeout and the orchestrator/SSH tests share process-global env, so they need nextest's process-per-test isolation.

Two new user-facing strings (the timeout messages) are not routed through `t!`, matching the LSP status messages beside them — including the `"LSP ({}) ready"` line this issue is about. The locale files are flat with a strict key-parity test, so localizing them means adding them to all fifteen locales; happy to do that if you'd rather they were translated from the start.

---

https://claude.ai/code/session_01SHsuKyMdSZ7soeFzDziRNw


crates/fresh-editor/src/app/lsp_requests.rs
```diff
@@ -1500,6 +1506,16 @@ impl Editor {
             // Use the hint text as-is - spacing is handled during rendering
             let display_text = text;
 
+            // Left gravity: a hint rendered *before* its anchor belongs in
+            // front of anything typed at that anchor. Without it, pressing
+            // Enter with the cursor on a hint's anchor dragged the hint onto
+            // the next line until the debounced refresh arrived (#722).
+            // `AfterChar` hints — the end-of-line anchoring from #1572 —
+            // keep the default right gravity so they travel with their glyph.
+            let gravity = match position {
```
sinelaw, line 1515: **Left-gravity hint markers desync `WrapIndex`'s decoration snapshot.**

`IndexDecorations::shift_for_edit` (`view/wrap_index.rs:145`) mirrors marker movement for the wrap index, and its contract is explicitly *right gravity for everything*:

> **Insertion** is *right gravity*: a position sitting exactly at the insertion point is pushed forward. Every decoration anchor in the editor is right-gravity, because that is all `MarkerList::create` makes.

`IndexDecorations::inline_hints` is populated from these very virtual-text markers (`state.rs:742` → `resolve_inline_hints`), so after this change the two disagree for exactly the edit this PR is about.

Scenario (the #722 repro itself): the `: Duration` hint is anchored at byte N; the cursor is on N; Enter inserts 1 byte at N.
* live marker (now left gravity) stays at N → renderer draws the hint on line 5 ✅ (what the new tests assert)
* `shift_for_edit` applies `p >= start → p + inserted`, moving the snapshot anchor to N+1 → the index believes the hint sits on line 6

`damage_bytes` then repairs lines against that shifted snapshot, so the index attributes the hint's ~12 display columns to the wrong line. That feeds row counts, cursor visual-column math, scrollbar extent and h-scroll until the next decoration version bump (the debounced `inlayHint` refresh ~1s later) forces a diff — i.e. the same "wrong until the server answers again" window the PR set out to close, just moved from the renderer to the coordinate service. The doc on `shift_for_edit` calls out this exact failure class and names LSP inlay hints as the source that "stayed wrong indefinitely".

The new e2e tests don't catch it because they use short, unwrapped lines and compare rendered rows, not index coordinates.

Fix direction: `IndexDecorations` needs to know a hint's gravity (carry it on `InlineHint`, or shift `inline_hints` with `p > start` instead of `p >= start`), otherwise the two models have to stay on the same gravity.

---

crates/fresh-editor/src/app/async_messages.rs
```diff
@@ -1185,9 +1185,20 @@ impl Editor {
         // that language so the server can start providing diagnostics, etc.
         // without waiting for the next user edit.
         if status == LspServerStatus::Running {
-            let was_already_running = old_status
-                .as_ref()
-                .is_some_and(|s| matches!(s, LspServerStatus::Running));
+            // `Unresponsive` counts as already-running: the server never
+            // stopped, it just went quiet, so recovering from it must not
+            // re-`didOpen` every buffer.
+            let was_already_running = old_status.as_ref().is_some_and(|s| {
```
sinelaw, line 1191: **`Unresponsive` was added to the `Running` predicate here but not to the crash predicate below, so a stuck server that then dies loses auto-restart.**

A few lines down (unchanged, ~line 1230):

```rust
if status == LspServerStatus::Error {
    let was_running = old_status
        .as_ref()
        .map(|s| matches!(s, LspServerStatus::Running | LspServerStatus::Initializing))
        .unwrap_or(false);
    if was_running {
        self.clear_diagnostics_for_server(&server_name_ref);
        ... lsp.handle_server_crash(...)   // auto-restart + status message
    }
}
```

`Unresponsive` is now a state a live server passes through, and it's the state most likely to precede a death — pyright wedged, user waits, process is OOM-killed or the stdout reader errors. Sequence: `Running` → 2 expiries → `Unresponsive` → read error → `Error`. `old_status` is `Unresponsive`, so `was_running` is `false` and the editor:

* leaves the dead server's diagnostics on screen indefinitely,
* skips `handle_server_crash`, i.e. no auto-restart and no status message.

Before this PR the same sequence had `old_status == Running` and did all three. Adding `| LspServerStatus::Unresponsive` to `was_running` restores it.

---

crates/fresh-editor/src/app/async_dispatch.rs
```diff
@@ -661,26 +698,102 @@ impl Editor {
...
+            (language, method),
+            crate::app::window::LspRequestTimeoutRecord {
+                at: std::time::Instant::now(),
+                timeout,
+                consecutive,
+            },
+        );
+
+        // Keep an open status popup in step with the new server state.
+        self.refresh_lsp_status_popup_if_open();
```
sinelaw: **This refresh runs *before* the `Unresponsive` status lands, and the recovery transition refreshes nothing at all — so the servers popup shows the wrong row.**

`note_request_timed_out` sends `LspRequestTimeout` and then `LspStatusUpdate{Unresponsive}` on the same channel, drained in that order. So on the crossing expiry:

1. `handle_lsp_request_timeout` → `refresh_lsp_status_popup_if_open()` rebuilds the popup while `lsp_server_statuses` still says `Running` → row renders `● pyright-langserver (ready)`.
2. `handle_lsp_status_update(Unresponsive)` is applied next, and that handler only calls `refresh_lsp_status_popup_if_open()` for `Error | Shutdown`.

`build_and_show_lsp_status_popup` is a snapshot, not re-rendered per frame, so an already-open popup keeps saying `(ready)` while the status bar reads `LSP (stuck)` — until the *next* expiry 30s later.

The recovery side is worse: `note_request_answered` sends only `LspStatusUpdate{Running}`, which hits neither refresh path, so an open popup stays on `◍ pyright-langserver (not responding)` indefinitely after the server starts answering again. The new `(not responding)` row in `popup_dialogs.rs` is precisely what this PR added, and the e2e tests only assert on the status-bar indicator, so neither case is covered.

Simplest fix: refresh the popup in `handle_lsp_status_update` for `Unresponsive` and `Running` too (or drop the refresh here and do it entirely from the status handler).

---

crates/fresh-editor/src/app/async_dispatch.rs
```diff
@@ -19,6 +19,27 @@ use super::Editor;
...
 const ASYNC_MESSAGE_FRAME_BUDGET: std::time::Duration = std::time::Duration::from_millis(50);
 
+/// Whether an LSP request was made by the editor rather than asked for by
+/// the user.
+///
+/// Background traffic is debounced and re-issued on every edit, so a stuck
+/// server produces a steady stream of expiries from it; user-invoked
+/// requests are one per keypress and are exactly what the user is waiting
+/// on. Only the latter are worth the status bar (issue #2197).
+fn is_background_lsp_request(method: &str) -> bool {
```
sinelaw, line 29: **Auto-triggered `completion` and `signatureHelp` are editor-initiated but classified as user-invoked, so they produce exactly the status-bar flood this function exists to prevent.**

The doc says "user-invoked requests are one per keypress", but two of the biggest request sources fire without the user asking for anything:

* `input_helpers.rs:279` — `if c == '(' || c == ',' { self.request_signature_help(); }` on every such character typed.
* `lsp_requests.rs:790-810` — `maybe_trigger_completion` fires `request_completion()` immediately on any LSP trigger char, and on a `quick_suggestions_delay_ms` timer for every word character.

Against a wedged server (the #2197 scenario), typing a 20-character line issues ~20 completion requests; 30s later each one expires and each overwrites `status_message` with `LSP (rust): 'textDocument/completion' timed out after 30s`. The status bar is then unusable for as long as the server stays stuck — the exact behaviour the background-request carve-out was written to avoid, and worse than the inlay-hint case because completions are one-per-keystroke.

`textDocument/completion` and `textDocument/signatureHelp` need to be in the background set (or the classification needs to come from the call site, since an explicitly-invoked Ctrl+Space genuinely is user-invoked). Note also that `textDocument/documentSymbol` is currently listed as background even though the symbol picker is user-invoked — the two mistakes point in opposite directions, which suggests method name alone is the wrong key.

---

crates/fresh-editor/src/app/lsp_requests.rs
```diff
@@ -1509,12 +1525,44 @@ impl Editor {
...
+    /// A request that expires after its timeout reaches the feature as
+    /// "no result", so "No definition found" and "No hover information
+    /// available" were reported for a server that simply never replied
+    /// (issue #2197). When *this* request's method expired for this
+    /// buffer's language a moment ago, say that instead.
+    ///
+    /// Deliberately narrow: it matches on the method the feature actually
+    /// asked for, and consumes the record, so one expiry explains one empty
+    /// result and never a later, genuinely empty answer.
+    fn expired_request_message(&mut self, method: &str) -> Option<String> {
```
sinelaw, line 1547: **Key mismatch for universal servers: the lookup uses the *buffer's* language, but the record is keyed by the server's registration label, which is `"universal"`.**

`LspHandle::spawn` sets `language_label = scope.label()` (`async_handler.rs:4740`), and `LanguageScope::label()` returns `"universal"` for an empty scope. Every message from such a server — including the new `AsyncMessage::LspRequestTimeout` and `LspStatusUpdate{Unresponsive}` — carries `language: "universal"`.

So for a globally-configured server (`universal_configs`) serving a `.rs` buffer:

* `expired_request_message("textDocument/definition")` builds key `("rust", …)`, the record is stored under `("universal", …)` → miss, and the user still sees `No definition found` for a request that simply never got an answer.
* the `retain(|(lang, _), _| lang != &recovered_language)` clear in `handle_lsp_status_update` never matches either, so those `("universal", method)` records accumulate for the life of the window.

Keying the record by the language the *request* was made for (or matching on "this language or a universal server that accepts it", the way `Window::is_lsp_server_ready` already does via `server_scope(...).accepts(language)`) would make the feature work for universal servers.

---

crates/fresh-editor/plugins/lsp_help.ts
```diff
@@ -605,7 +697,12 @@ editor.on("lsp_server_error", (data) => {
 });
 
 editor.on("lsp_status_clicked", (data) => {
-  const help = byLanguage.get(data.language);
+  // Answer with the entry that took this language's last failure, so the
+  // popup documents the binary that actually failed rather than whatever
+  // the language is normally served by (issue #2981).
+  const entryId = lastFailureByLanguage.get(data.language);
```
sinelaw: **Regression: the help popup is now unreachable from a sibling language that the same entry covers.**

The old code was `byLanguage.get(data.language)` + `failures.get(help.id)`, and `failures` is keyed by *entry* id — so a failure recorded for `typescript` was reachable by clicking the indicator from a `.js` buffer, because both languages map to the same entry.

`lastFailureByLanguage` is keyed by the language string the failure arrived with, and `lsp_server_error` only sets the one language (`lastFailureByLanguage.set(data.language, help.id)`). Concretely, in the #2981 Deno project: `deno lsp` fails to spawn for `typescript`, the user switches to a `.js` file in the same project, the LSP indicator is still red (the warning domain scans all languages), they click it — `lastFailureByLanguage.get("javascript")` is `undefined`, the handler returns early, and no install-help popup appears. Before this change it did.

Since the entry already declares `languages` / `disables`, populating `lastFailureByLanguage` for every language `help` covers (or falling back to `helpFor(data.language, …)`'s entry when there's no exact-language record but `failures` has one for it) restores the old reach.

---

sinelaw (commented):
Careful read of the whole diff. Six findings, all posted inline. The Deno detection change (`deno_lsp_override` + the bounded authority probe) and the `LspError { server_command }` plumbing look correct to me; the `handle_cancel_request` fix (dropping the pending entry) is right and the request-id gating on the editor side means the now-immediate empty result is discarded as an outdated response, so that one is safe.

The two I'd treat as blocking are the wrap-index gravity desync (#722 fix moves the drift from the renderer into `WrapIndex`'s decoration snapshot) and the missing `Unresponsive` case in the crash predicate in `handle_lsp_status_update` (a stuck server that then dies no longer auto-restarts and leaves stale diagnostics on screen).

---

crates/fresh-editor/src/app/lsp_requests.rs
sinelaw: Confirmed and fixed in 0d1a743 — you're right that this moved the drift rather than removing it, and the `shift_for_edit` doc predicted exactly this failure class.

Gravity now travels with the decoration rather than being assumed: `VirtualText` records the gravity its marker was created with, `InlineHint` carries it through `resolve_inline_hints`, and `shift_for_edit` applies each anchor's own rule — a left-gravity hint sitting exactly at the insertion point stays, everything else keeps the documented right-gravity behaviour. I took that over `p > start` for `inline_hints` because `AfterChar` hints are genuinely right-gravity (that's what #1572 relies on), so one blanket rule for the list would have been wrong in the other direction.

`shift_for_edit`'s doc is updated to state the exception, and there's a unit test in `wrap_index` asserting a left- and a right-gravity hint anchored at the same byte diverge under an insertion at that byte, and both collapse to the deletion start under a deletion covering it.

---

crates/fresh-editor/src/app/async_messages.rs
sinelaw: Confirmed and fixed in 0d1a743 — `Unresponsive` is now in the crash predicate alongside `Running | Initializing`, with a comment on why it belongs there (it is the state most likely to precede a death, not a state a dead server should be excused by).

Added an e2e test that walks `Running → Unresponsive → Error` and asserts the crash path still fires.

---

crates/fresh-editor/src/app/async_dispatch.rs
sinelaw: Fixed in 0d1a743, taking the second option: the refresh is gone from the timeout handler (with a comment saying why — it runs before the status update it accompanies), and `handle_lsp_status_update` now refreshes for `Unresponsive` and `Running` too, guarded on the status actually changing so an already-running server does not re-render the popup on every status message.

---

crates/fresh-editor/src/app/async_dispatch.rs
sinelaw: Right on both counts — I checked `maybe_trigger_completion` and the `'(' | ','` signature-help trigger, and completion is the worst case precisely because it is one per keystroke. Both are in the background set as of 0d1a743, with a comment recording why they read as user-invoked but are not.

On `documentSymbol` and "method name alone is the wrong key": agreed in principle, and I left it in the background set rather than swapping one misclassification for another. Its automatic uses are debounced per document, so an expiry storm from it is bounded in a way completion's is not, and the failure mode of getting it wrong in this direction is a missing status line rather than an unusable status bar — whereas moving it out would flood on every automatic refresh. Threading the classification from the call site is the real fix; it needs the flag to reach `send_request_with_timeout`, which is a wider change than I wanted to make inside this PR. Happy to do it here if you'd prefer it done properly now rather than left as a follow-up.

---

crates/fresh-editor/src/app/lsp_requests.rs
sinelaw: Confirmed and fixed in 0d1a743 by the second route you suggested. The record now carries the `server_name` it came from, and `expired_request_message` accepts a record whose label is the buffer's language *or* whose server's scope accepts that language — the same `server_scope(...).accepts(language)` test `is_lsp_server_ready` uses. The recovery clear in `handle_lsp_status_update` matches on the server as well as the language, so `("universal", method)` records no longer accumulate for the life of the window.

---

crates/fresh-editor/plugins/lsp_help.ts
sinelaw: Confirmed and fixed in 0d1a743 — `lsp_server_error` now records the failure under every language the entry covers, so clicking the indicator from a `.js` buffer in a Deno project reaches the same help it did before. The entry's own `languages` is the source, so it stays correct for the multi-language entries (c/cpp, the TypeScript family) without a second lookup path.

---

Issue #3126, opened 2026-09-01
the git status not update if changed in background


asukaminato0721:
when I use codex to edit a project, and in another terminal tab, I open fresh to check the diff.

but fresh won't refresh to get the newest status.

sinelaw:
Reproduced on 0.4.10 (release tarball, x86_64-musl), Linux, tmux 110x30.

Steps:
1. Git repo with a clean tree (`git status --porcelain` empty).
2. `fresh --no-restore calc.py` → `Ctrl+P` → **Review Diff**. Panel shows `No changes to review.`
3. From **another** shell, while the panel stays open: `printf '\n# external change\n' >> calc.py` (`git status --porcelain` now `M calc.py`).
4. Wait and watch the panel.

Observed:

```
t+5s    No changes to review.
t+15s   No changes to review.
t+30s   No changes to review.
```

Pressing `n` (next hunk) inside the panel does not refresh it either. The panel is still empty while git has had a modification for half a minute.

Re-running `Ctrl+P` → **Review Diff** does pick the change up — `Review Diff — 1 files, +2 / -0` — but it does so by opening a **second** panel: the tab bar becomes `calc.py | *Review Diff* 1 | *Review Diff* 2`. So there is no refresh path for an open panel, only "open another one", and repeating it accumulates duplicate panels.

Separately: the File Explorer's status column marks untracked files (`U`) but shows nothing for a modified tracked file, whether the edit came from outside or from inside the editor — so it is not a second view of this state you can fall back on.

---
_Reproduced by an automated agent driving the editor interactively in tmux against the 0.4.10 release build._

---
_Generated by [Claude Code](https://claude.ai/code)_

sinelaw:
Re-checked on both builds — **this looks fixed on master** (`fbddf641`), by e0022f57 (the panel's watch gained a `git status` poll, and re-running focuses the existing panel instead of stacking a new one).

Same setup both times: clean repo, `fresh --no-restore calc.py`, `Ctrl+P` → **Review Diff**, then `printf '\n# external change\n' >> calc.py` from another shell.

| build | panel after the external edit |
|---|---|
| 0.4.10 release (musl, via `install.sh`) | `No changes to review.` still, 14s later |
| master `fbddf641` (debug) | refreshes on its own to `Review Diff — 1 files, +2 / -0`, `▾ UNSTAGED (1)` / `▾ calc.py +2 / -0` |

Probably closable once 0.4.11 ships.

---
[...]

Issue #3104, opened 2026-08-29
Review diff panel lacks syntax highlighting


tinchoz49:
## Description

The review diff panel does not provide syntax highlighting for the code being reviewed. All code appears in plain text without color coding for keywords, strings, comments, etc.

This makes it harder to read and understand diffs, especially for larger changes.

## Steps to Reproduce

1. Open fresh editor
2. Make some changes to a file
3. Open the review diff view (e.g., `*Review main..HEAD*`)
4. Observe that the code in the diff panel has no syntax highlighting

## Expected Behavior

The review diff panel should apply syntax highlighting based on the file type (TypeScript, Python, etc.), similar to how the main editor view does.

## Actual Behavior

All code in the review diff panel is displayed without syntax highlighting - just plain text with diff markers (red/green backgrounds).

## Environment

- Fresh version: v0.4.10
- OS: macOS
- File type tested: .tsx files
- theme: dracula

## Additional Context

See attached screenshot showing the lack of syntax highlighting in the review diff panel.

<img width="1203" height="905" alt="Image" src="https://github.com/user-attachments/assets/ca789872-27f9-465c-8cf2-17636d752216" />

asukaminato0721:
https://github.com/sinelaw/fresh/issues/2871

sinelaw:
Confirmed on 0.4.10 (release tarball, x86_64-musl), Linux, tmux 140x30, default theme.

Steps:
1. In a git repo, append to `calc.py`:
   ```python
   def sub(a, b):
       return a - b
   ```
2. `fresh --no-restore calc.py` → `Ctrl+P` → **Review Diff**.
3. Compare the fg colours of the added lines in the panel against the same lines in the editor buffer.

| line | in the **Review Diff** panel | in the **editor** |
|---|---|---|
| `def sub(a, b):` | `108`, `40`, `42`, `231` | `51` (`def`), `226` (`sub`), `6` (parens), `251` (`,` `:`), `231` |
| `return a - b` | `108`, `40`, `231` | `51` (`return`), `231`, `244` |

The panel paints added lines in flat add-greens plus plain text — no keyword, function-name or bracket colours anywhere — while the identical text one tab over gets the full Python theme. Same for the context rows, which carry no function/keyword colour either.

---
_Reproduced by an automated agent driving the editor interactively in tmux against the 0.4.10 release build._

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2873, opened 2026-07-31
fix git log panel support highlight #2871

asukaminato0721:
fix #2871
fix #3104

Git Log read the parent commit and current commit file content, highlight old/new version.
map @@ hunk line number back to diff, avoid in version parser pollution.
keep diff background and theme change.
if normal .diff, Git blob > 2 MiB, use original highlight.

PR #3160, opened 2026-09-02
review-diff: discard to HEAD with an honest report, and a panel that follows the working tree (#2318, #3126)

### LINKED ISSUE #2318: Review Diff: file-level Discard (`D`) reports "Discarded" but leaves staged changes intact for a fully-staged file

sinelaw:
One commit, two Review Diff bugs. Each was reproduced on `master` first by driving the editor interactively in tmux, and re-checked the same way after the fix; each has regression tests that fail (or hang) without its part of the change.

Syntax highlighting of the diff body (#3104) was part of this PR earlier and has been split out: it is being redone as a host-side design in a separate PR, so this one carries no highlighting code and no plugin API change.

## #2318 — `D` announced a discard it had not performed

With a fully staged modification (`git status --porcelain` = `M  calc.py`, working tree clean), `D` → confirm printed `Discarded: calc.py` while the change sat untouched in the index, on disk, and in the panel. The discard ran `git checkout -- <path>`, which rewrites the working tree *from the index* — a no-op when they already agree — and the success message was printed either way.

| | master | this branch |
|---|---|---|
| status bar | `Discarded: calc.py` | `Discarded: calc.py` |
| `git status --porcelain` | `M  calc.py` | *(empty)* |
| `git diff --cached` | still shows the change | *(empty)* |
| `calc.py` on disk | still `# CHANGED` | back to the committed content |
| the panel | still lists the STAGED hunk | `No changes to review.` |

* **The action does what its dialog promises.** "Discard changes in `<file>`" / "This cannot be undone" describes a discard of the file, so the file goes back to HEAD — index *and* working tree — via `git restore --source=HEAD --staged --worktree`. That also covers two cases `git checkout HEAD -- <path>` cannot: a staged add (not in HEAD at all; the prompt says **Delete** for it) and a rename (one change, two pathspecs). When the discard also takes the *other* section's changes, the prompt says so.
* **The report is earned, not assumed.** The outcome is checked against `git status` for the affected paths before anything is said. `Discarded` means git agrees the file is clean; a failed command, a change that survived, or a status that could not be read is reported as `Discard failed: <file> — <git's own message>`.

## #3126 — an open panel never refreshed, and the workaround leaked tabs

Open Review Diff on a clean tree, then modify a tracked file from another shell: the panel still read `No changes to review.` at t+30s, and `n` did not refresh it. Re-running the command did pick the change up — by opening a *second* panel (`calc.py | *Review Diff* 1 | *Review Diff* 2`), and repeating piled up more.

The panel had a watch, but it was opt-in and keyed off the editor's own save events, so it could only see edits made in Fresh — and a review of the working tree is a view of state nothing in the editor owns. And the plugin tracks one review session, so a second `openReviewPanels` left the first group orphaned with nothing driving it.

* The watch gains a `git status` poll (`--no-optional-locks`, so a timer never fights the user's own git for the index lock) whose interval adapts to how long a tick takes. A tick hashes what came back — file list, sections, and the text of every hunk line — and rebuilds **only when it differs**, so a quiet repo costs the reader nothing. Ticks yield while a line selection, the filter field, a discard confirmation, a comment prompt, another refresh, or a side-by-side composite is open; cancelling the discard dialog leaves the watch running; refreshes are serialised.
* **The watch is on by default**; `W` turns it off. (`test_review_watch_toggle_status` updated: the first `W` now turns it *off*.)
* Running **Review Diff** with a worktree review already open refreshes and focuses that panel instead of stacking another. A range or stash review already open is closed first.
* **A refresh nobody asked for never steals focus.** Every refresh-driven focus call first asks the host which buffer is active and stands down unless it is one of the review's, so saving a file in another tab refreshes the review behind the reader and leaves them where they are.

## Open question (for the maintainer, not guessed at)

`D` on a file listed under **UNSTAGED** now also drops that file's staged part, because the discard is file-level and goes to HEAD. VS Code splits this instead: discard means the working tree, and staged changes are unstaged first. If that split is preferred here, the honest-reporting layer is independent of the choice — only the git command changes, and the tests for it stay as they are.

## Verification

* Every change driven interactively in tmux (`tmux send-keys` / `capture-pane -p -e`) against a real build, not `--cmd script`; the #2318 assertions are on `git status --porcelain`, `git diff --cached` and the file on disk, not on the status bar.
* Eight new e2e tests in `tests/e2e/plugins/audit_mode.rs`, each confirmed to fail (or hang) with its fix removed and the test kept:
  * `test_issue2318_discarding_a_fully_staged_file_reverts_it_to_head`, `test_issue2318_a_discard_that_fails_is_reported_as_failed`, `test_issue2318_discarding_a_staged_add_says_delete`, `test_issue2318_discarding_an_added_then_edited_file_says_delete`
  * `test_issue3126_open_panel_picks_up_a_change_made_outside_the_editor`, `test_issue3126_reopening_review_diff_does_not_stack_a_second_panel`, `test_issue3126_cancelling_a_discard_leaves_the_watch_running`, `test_issue3126_saving_in_another_tab_does_not_steal_focus`
* `cargo fmt --all -- --check`, `plugins/check-types.sh audit_mode.ts` (only the pre-existing `lib/widgets.ts` errors), and the `audit_mode` / `review_diff` e2e suites green (130 tests) on current `master`.
* The seven new `audit_mode.i18n.json` keys are translated in all 15 locales.

Fixes #2318, #3126.

https://claude.ai/code/session_017uXJ7dBKoq6BeqyvQPZ8NE


sinelaw:
Pushed `ab5d04a`, which fixes defects an independent review found in the three commits above. Recording them here because two of them undercut claims this PR's description makes, and one would have hung the editor.

- **The syntax cache could not converge on a large review.** An entry is one file's *side*, so a review holds two per file, against a fixed 512-entry ceiling that `clear()`ed wholesale on reaching it — and the render following a batch of arrivals re-requests every miss. Above ~256 files that is an unbounded loop of full relayouts and synchronous parses. The ceiling now scales with the changeset, eviction is oldest-first, and a bounded round count backstops it.
- **A span covering several rows coloured only the first row.** Parsing each side as one document is the point — and then the mapping back onto rows dropped everything past the opening line. It only shows on tree-sitter grammars, because the TextMate backend splits per line already; TypeScript is tree-sitter-backed, which is the file type #3104 was filed against. My original test used Python and passed while the reported case was still broken. Fixed, with a `.ts` test over a block comment.
- **A fresh review showed no colour at all** — `openReviewPanels` bumped the data revision after its first render, so that render's highlight answers came back tagged stale and were dropped. (Self-inflicted while fixing the above; caught by re-driving the editor rather than trusting the diff.)
- **Escape on the discard dialog stopped the watch for the rest of the session** — the flag the poll stands down on was cleared only on *confirm*. That is #3126's own symptom, reintroduced by #3126's fix.
- **The discard prompt warned about the wrong side.** "Discards the staged changes too" matched the row the cursor was already on, so a staged-only file was warned about its own staged changes, and standing on the STAGED row of a file changed on both sides said nothing about the unstaged edits — the ones that exist nowhere else. It now names the *other* section. And a staged **add** is deleted by `git restore --source=HEAD`, not rolled back, so the prompt now says "Delete" rather than "Discard changes in".

Also measured and partly addressed: building a highlight engine costs ~94 ms against ~0.3 ms to parse with it (small TypeScript document, debug build), and it was being paid per file per side on the editor thread — the frame-budget guard named it in the logs. Engines are now cached per grammar per catalog generation and invalidated before each use, so construction is paid once per language per session instead of once per file. Moving the parse off the editor thread entirely is the deeper fix and is **not** done here: `HighlightEngine` is `!Send` (the TextMate engine holds `Rc` and oniguruma pointers), so it would have to be built and used inside the off-loop task rather than moved there. Happy to do that as a follow-up if you'd prefer it in this PR.

Three regression tests added, each confirmed to fail (or hang) without its fix: `test_issue3104_multi_line_constructs_colour_every_row`, `test_issue3126_cancelling_a_discard_leaves_the_watch_running`, `test_issue2318_discarding_a_staged_add_says_delete`. The `audit_mode` / `review_diff` suites are green at 98 passed.

One review point I did **not** act on, for the record: the reviewer would have rebased the host API onto the open #2873 rather than duplicating it. The two designs agree (same command, response and `TsTextHighlightSpan` shape), but this branch is off `master` and cannot depend on an unmerged PR — whoever lands second drops their Rust half and keeps the plugin-side consumer.

---

crates/fresh-editor/plugins/audit_mode.ts
```diff
@@ -3994,21 +4363,129 @@ function rememberPendingHunkAnchor(hunkId: string | null) {
...
     pendingDiscardFile = f;
-    rememberPendingHunkAnchor(null);
-    const action = f.category === 'untracked' ? "Delete" : "Discard changes in";
+    // A file the repo has never had — untracked, or added and staged — is
+    // not restored by this, it is removed. `git restore --source=HEAD` on
+    // a path HEAD does not carry deletes it, so the prompt has to say
+    // "Delete": "discard changes in" describes an edit surviving as a
+    // file, which is not what happens.
+    const removesFile = f.category === 'untracked'
+        || (f.category === 'staged' && f.status === 'A');
```
sinelaw: `removesFile` misses the `AM` case, so the dialog can still promise something the action does not do.

`parseGitStatusPorcelain` emits **two** entries for `AM` (staged add, then edited in the worktree): `{status:'A', category:'staged'}` and `{status:'M', category:'unstaged'}`. Press `D` on the **UNSTAGED** row and `f.category === 'unstaged'`, `f.status === 'M'`, so `removesFile` is false and the prompt reads *"Discard changes in new.py?"* with the description *"Discards the staged changes too — back to HEAD"*.

But `discardFileToHead` then runs `git restore --source=HEAD --staged --worktree -- new.py`, and since `new.py` is not in HEAD at all, git **deletes** it from the index and from disk. The prompt describes an edit surviving as a file; the file is gone — the same dialog/action mismatch #2318 is about, just on the other row.

Repro: `touch new.py; git add new.py; echo x >> new.py` → open Review Diff → cursor on the UNSTAGED `new.py` → `D` → Enter.

Deriving `removesFile` from the path rather than from the row would cover it, e.g. `state.files.some(o => o.path === f.path && o.category === 'staged' && o.status === 'A')`, or an explicit `git cat-file -e HEAD:<path>`.

---

crates/fresh-editor/plugins/audit_mode.ts
```diff
@@ -4079,25 +4549,90 @@ function setReviewConfirmation(text: string): void {
...
-    // A rebuild supersedes whatever the last action confirmed: `r` and the
-    // watch-driven refreshes should land on the summary, not on a stale
-    // "Lines discarded" from several actions ago.
-    reviewConfirmation = null;
+async function refreshMagitData(opts?: { onlyIfChanged?: boolean }): Promise<boolean> {
+    refreshInFlight = true;
+    try {
+        return await refreshMagitDataInner(opts);
+    } finally {
+        refreshInFlight = false;
```
sinelaw: `refreshInFlight` only guards one direction, and the shared `finally` clears it too early.

`refreshInFlight` is read in exactly one place — `review_watch_poll` (L6338). Every other entry point calls `refreshMagitData` unconditionally: `r` (L3436), stage/unstage/discard-hunk (L3961, L4311–L4345), the save/revert watch (L6385) and the discard confirmation (L7142, L7171). So the poll yields to them, but they never yield to the poll.

Concrete failure: on a repo where `git status` + two `git diff`s take longer than a keystroke, the poll tick is mid-`await` when the user presses `s`. Two `refreshMagitDataInner` runs now overlap. Both assign `state.files`/`state.hunks` wholesale and both call `updateMagitDisplay()` + `restoreCursorAfterRebuild()`, so:

* the poll's snapshot (taken *before* the stage) can land last and repaint the panel with pre-stage data, which then sticks until the next change — and it also writes its own `lastDataSignature`, so the poll will not correct it;
* the first run to finish consumes the single `pendingHunkAnchor`, leaving the second to fall through to `jumpToGlobalHunk(0)`.

And because both share this `try/finally`, whichever finishes first sets `refreshInFlight = false` while the other is still running, so the next tick can start a third overlapping refresh. A counter (`refreshDepth++/--`) or serializing every caller on one promise would hold the invariant the doc comment states.

---

crates/fresh-editor/plugins/audit_mode.ts
```diff
@@ -5733,9 +6323,55 @@ function review_toggle_watch() {
...
+    // from under them: a line selection, the filter field, an open comment,
+    // or the discard confirmation.
+    if (state.lineSelection !== null || filterEditing
+        || pendingDiscardFile !== null || pendingCommentInfo !== null) return;
+    watchPollInFlight = true;
+    try {
+        // Keep the reader where they were: the rebuild's cursor restore
+        // uses this anchor, so an external change three files away does
+        // not scroll the panel back to the top.
+        rememberPendingHunkAnchor(getHunkAtDiffCursor()?.id ?? null);
```
sinelaw: When the cursor is not on a hunk row, this anchor makes the poll jump the reader to hunk 0.

`getHunkAtDiffCursor()` returns null for every row that is not a diff body row — the summary rows at the top of the stream, a section header (`STAGED`/`UNSTAGED`), a file header, the `(no diff available)` filler. `rememberPendingHunkAnchor(null)` then stores `{hunkId: null, section: cur?.gitStatus ?? null → null, row}`, and `restoreCursorAfterRebuild` skips both the id branch and the section branch and lands on `if (state.hunks.length > 0) jumpToGlobalHunk(0)`.

So: park the cursor on a file header while reading the changeset overview, let a formatter or agent touch any file in another terminal, and two seconds later the panel scrolls to the first hunk of the review. That is the "yank the reader out from under themselves" behaviour this function's own comment says it is avoiding, and it now happens by default.

`anchor.row` is already recorded but never used by `restoreCursorAfterRebuild`. Either return early when the cursor was not on a hunk (leave `pendingHunkAnchor` null so the restore is a no-op) or use `anchor.row` for that case.

---

crates/fresh-editor/plugins/audit_mode.ts
```diff
@@ -5733,9 +6323,55 @@ function review_toggle_watch() {
...
+async function review_watch_poll(): Promise<void> {
+    if (!reviewWatchEnabled || state.groupId === null) return;
+    // Range reviews are ref-to-ref; the working tree doesn't affect them.
+    if (state.mode === 'range') return;
+    if (watchPollInFlight || refreshInFlight) return;
+    // Anything the reader is in the middle of that a rebuild would yank out
+    // from under them: a line selection, the filter field, an open comment,
+    // or the discard confirmation.
+    if (state.lineSelection !== null || filterEditing
+        || pendingDiscardFile !== null || pendingCommentInfo !== null) return;
```
sinelaw, line 6074: The skip list misses the one thing a rebuild disturbs most: an open composite.

If the reader is in side-by-side / drill-down (`state.reviewLayout === 'side-by-side'`, `state.centerComposite` set), a poll-triggered `refreshMagitDataInner` does three things in order: `state.dataRevision++` (L4640), `discardParkedComposite()` (L4641), then `updateMagitDisplay()` → `renderCenter()` → `buildCenterComposite()`.

`compositeSignature` embeds `state.dataRevision` (L5225), so the reuse path at L5325 can never match after a refresh — and the parked composite was just discarded anyway. The composite is therefore torn down and rebuilt from two fresh `git show` calls plus a re-alignment, dropping the reader at `focusHunkIdx = 0` of the file they were reading. Scenario: reading file A in side-by-side while an agent/formatter writes unrelated file B; two seconds later the pane blinks and the position is lost.

`review_relayout_diff` already refuses to run in exactly this situation (`if (state.centerComposite !== null) return;`, L3326) for the same "must not pull the reader out of the composite they just opened" reason. The poll fires at a moment nobody asked for too, so it wants the same guard — or `compositeSignature` needs to stop keying off `dataRevision` so an unrelated file's change can reuse the composite.

---

crates/fresh-editor/plugins/audit_mode.ts
```diff
@@ -819,7 +819,16 @@ interface GitStatusResult {
...
 async function getGitStatus(): Promise<GitStatusResult> {
     const cwd = gitCwd();
-    const result = await editor.spawnProcess("git", ["status", "--porcelain", "-z", "-uall"], cwd);
+    // `--no-optional-locks`: plain `git status` refreshes the index and takes
+    // `.git/index.lock` to write it back. This runs on a timer while a review
+    // is open (#3126), so without it the panel would periodically lose a
+    // `git commit` or `git add` in the user's own terminal the race for that
+    // lock.
+    const result = await editor.spawnProcess(
+        "git",
```
sinelaw, line 828: `--no-optional-locks` covers one of the three git calls the tick makes, so the lock race it is meant to close is still open.

Every poll tick runs `getGitStatus()` **and** `fetchDiffsForFiles()`, and the latter issues `git diff --cached` (L900) and `git diff` (L908) through `diffArgs`, neither of which carries the flag. Plain `git diff` also refreshes the index and writes it back under `.git/index.lock` — that is precisely why `--no-optional-locks` / `GIT_OPTIONAL_LOCKS=0` exists — so the panel still grabs the optional index lock twice every 2 s, and a user's concurrent `git add`/`git commit` can still lose the race the comment above describes.

`fileIsClean` (L4400-ish) is on the same footing, though it runs only on a discard.

Simplest fix is to put `--no-optional-locks` in `diffArgs` alongside the other format-pinning flags (it is a top-level git option, so it has to go before the sub-command), or set `GIT_OPTIONAL_LOCKS=0` in the environment for all of the panel's git invocations.

---

crates/fresh-editor/plugins/audit_mode.ts
```diff
@@ -5713,18 +6280,41 @@ editor.defineMode(REVIEW_FILTER_MODE, [
...
+// It starts on, because a stale diff is not a neutral default — it is a
+// panel quietly disagreeing with the repository. `W` turns both channels
+// off for anyone who would rather not pay for the poll.
+let reviewWatchEnabled = true;
 let reviewWatchGen = 0;
 const WATCH_DEBOUNCE_MS = 200;
 
+/** How often an open worktree review re-reads git looking for outside changes. */
+const WATCH_POLL_MS = 2000;
+let watchPollTimer: number | null = null;
```
sinelaw, line 6021: A 2 s poll that is on by default costs one process **per untracked file**, not the "three `git` invocations" the header comment budgets for.

`fetchDiffsForFiles` batches staged and unstaged into one `git diff` each, but untracked files are done one at a time: `for (const f of untrackedFiles) { await editor.spawnProcess("git", [... "--no-index", "/dev/null", f.path]) }` (L915–L920). `git status -uall` enumerates every untracked file individually, so a repo with an un-ignored `node_modules`, `target/`, `dist/`, or a fresh clone before the first `.gitignore` produces hundreds to thousands of untracked entries — and each tick spawns that many `git diff --no-index` processes, sequentially.

`watchPollInFlight` only prevents *overlap*; it does not throttle. Once a tick takes longer than `WATCH_POLL_MS`, the next one starts as soon as it finishes, so the editor sits in a permanent process-spawn loop for as long as the review is open, with the watch on by default and nothing on screen saying why the machine is busy.

Worth either capping/omitting untracked bodies on poll ticks (the file list alone already tells the signature something changed), or backing the interval off to some multiple of the last tick's measured duration.

---

crates/fresh-editor/src/app/plugin_commands.rs
```diff
@@ -1442,6 +1442,102 @@ impl Editor {
...
+            .find_by_path(path_buf, first_line)
+            .map(|entry| entry.language_id.clone())
+        else {
+            return Vec::new();
+        };
+        let key = (self.grammar_registry.catalog_gen(), language_id);
+        if !self.text_highlight_engines.contains_key(&key) {
+            let engine =
+                HighlightEngine::for_file(path_buf, first_line, self.grammar_registry.as_ref());
+            self.text_highlight_engines.insert(key.clone(), engine);
```
sinelaw: `text_highlight_engines` is insert-only — nothing ever removes an entry.

Because the key is `(catalog_gen, language_id)`, a registry reload does not replace the old entry, it adds a *new* one beside it: every previous catalog generation's engines stay resident for the life of the process. Nothing evicts by language either, so a session that reviews a polyglot repo and reloads config a handful of times accumulates `generations × languages` `HighlightEngine`s — each one the ~94 ms of tree-sitter construction this doc comment quantifies, plus its grammar/query state.

The generation is already in the key, so the cheap fix is to drop the stale generations at insert time:

```suggestion
        if !self.text_highlight_engines.contains_key(&key) {
            // A reload retires every engine built against the old catalog.
            self.text_highlight_engines.retain(|(gen, _), _| *gen == key.0);
            let engine =
                HighlightEngine::for_file(path_buf, first_line, self.grammar_registry.as_ref());
            self.text_highlight_engines.insert(key.clone(), engine);
```

---

sinelaw (commented):
Careful pass over the whole diff. The three fixes look sound in their core mechanics — I traced the row↔side index lockstep in `buildDiffLines`/`diffSideLines` (including the `\ No newline` skip, the word-diff pair path and empty lines) and it holds; `splitSpansByLine`'s multi-line clipping and the `contentToEntriesHighlighted` whole-entry offsets are both correct against `TextPropertyManager::from_entries`; the syntax cache's capacity-vs-spin reasoning checks out; and the discard's verify-against-git step is a real improvement.

Seven findings below, mostly clustered around the new always-on poll (#3126) rather than the highlighting.

The one I would fix before merge is the discard prompt: `removesFile` only looks at the row under the cursor, so `D` on the UNSTAGED row of an `AM` file still says "Discard changes in X" and then deletes X.

---

crates/fresh-editor/plugins/audit_mode.ts
sinelaw: Fixed in b7551ac. Confirmed your repro against real git first — `AM new.py` + `git restore --source=HEAD --staged --worktree` does delete the file, exit 0.

`removesFile` now asks the path, not the row: `state.files.some(o => o.path === f.path && o.category === 'staged' && (o.status === 'A' || o.status === 'C'))`, plus the untracked case. `C` is in there for the same reason — a staged copy's destination is not in HEAD either, so it is deleted too.

Covered by `test_issue2318_discarding_an_added_then_edited_file_says_delete`, which walks to the UNSTAGED row of an `AM` file and asserts the prompt says Delete; it fails on the previous commit.

---

crates/fresh-editor/plugins/audit_mode.ts
sinelaw: Right on both counts, and the one-directional guard was the weaker half — fixed in b7551ac by serializing rather than by making the flag accurate.

Every caller now chains onto the last (`refreshQueue = run.then(...)`), so only one `refreshMagitDataInner` is ever in progress and the "poll's pre-stage snapshot lands last and sticks" path cannot happen regardless of which side started first. `refreshPending` is a counter of callers, so a run finishing no longer clears the flag out from under one queued behind it, and the poll consults that counter as before.

The rejection handler on the chain is deliberate: a predecessor that throws must not cancel the refreshes queued behind it.

---

crates/fresh-editor/plugins/audit_mode.ts
sinelaw: Fixed in b7551ac, taking the first of your two options: the poll only sets an anchor when there is actually a hunk under the cursor, so a header/summary/filler row leaves `pendingHunkAnchor` null and `restoreCursorAfterRebuild` is a no-op rather than falling through to `jumpToGlobalHunk(0)`.

I went with the early return over `anchor.row` because the row number is not stable across the rebuild that just happened — a change in a file above the cursor shifts every row below it, so restoring by row would move the reader too, just less obviously. Leaving the buffer cursor where it is keeps it on the same byte, which is the closest thing to "don't touch it" available here.

---

crates/fresh-editor/plugins/audit_mode.ts
sinelaw: Fixed in b7551ac with the guard you point at — `if (state.centerComposite !== null) return;`, matching `review_relayout_diff` for the reason you give.

I did not take the `compositeSignature` option. Dropping `dataRevision` from it would let a composite be reused after a refresh that genuinely changed *that* file, which trades a rare annoyance for a wrong diff on screen; the signature is doing its job.

The cost is that a reader sitting in side-by-side stops getting live updates until they leave it or refresh — which I think is the right side of the trade, since the composite is a deliberate "I am reading this one file" gesture. Verified in tmux: an external write to an unrelated file while in side-by-side leaves the pane untouched, and the change is there on returning to the unified stream.

---

crates/fresh-editor/plugins/audit_mode.ts
sinelaw: Fixed in b7551ac — `--no-optional-locks` moved into `diffArgs`, in front of the sub-command alongside the `-c` overrides, so every diff-producing call in the panel carries it (`git diff`, `git diff --cached`, the untracked `--no-index` ones, `git stash show`). Also added to `fileIsClean`'s status call.

Putting it at the one call site was the wrong altitude for exactly the reason you name: the flag belongs wherever the panel shells out on a timer, not on the one command I happened to be editing.

---

crates/fresh-editor/plugins/audit_mode.ts
sinelaw: Fixed in b7551ac, taking your second option — the interval is now derived from what the last tick actually cost rather than being constant.

`setInterval` is gone; each tick re-arms itself with `max(WATCH_POLL_MS, spent * 4)`, capped at 60s. So the duty cycle is bounded at ~25% of wall-clock regardless of repo shape: a normal repo (a few ms a tick) keeps the 2s cadence, and the un-ignored `node_modules` case degrades to refreshing rarely instead of spawning continuously. The re-arm is in a `finally`, so a tick that throws or returns early still owes the next one a slot.

I did not cap or omit untracked bodies. The file list alone would catch an untracked file appearing or disappearing, but not an edit to one that is already listed — and "an agent is writing files in another terminal" is exactly the case #3126 is about, so I would rather poll a pathological repo slowly than be blind to a class of change in every repo.

Verified in tmux across three successive external changes that the chain keeps re-arming — that was the regression I was most worried about introducing here.

---

crates/fresh-editor/src/app/plugin_commands.rs
sinelaw: Applied your suggestion in b7551ac, with the comment reworded to say what the `retain` is for ("a catalog reload does not replace an entry, it adds one beside it").

---

sinelaw:
**Do not merge yet — the #3104 highlighting regresses large range reviews badly.** Measured, not estimated: release builds of `master` (`cf626fc`) and this head (`08d8bc4`), driven in tmux against this repo's own `Review Diff: Range` on `HEAD~100..HEAD` (463 files, +84,726 / −30,126, 1,961 hunks), same machine, back to back, CPU otherwise idle. Latency = key sent → first frame that differs.

| | master | this branch |
|---|---|---|
| open: Enter → panel populated | 13.3 s | **30.1 s** |
| PageDown ×5 | 30–49 ms | 29–32 ms |
| `n` (next hunk) #1 | 87 ms | **>15 s (timed out)** |
| `n` #2 | 87 ms | **9.7 s** |
| `n` #3–5 | 34–87 ms | 11–31 ms |
| End / PageUp / Home | 29–32 ms | 30–34 ms |
| RSS after the run | 797 MB | **1,964 MB** |
| CPU time | 14 s | 65 s |

Where it goes, from the editor-thread budget log on the branch run:

* **`SetPanelContent` 9,863 ms** (the `n` stall) and **2,849 ms** (the open) — mounting the stream with a syntax overlay on every token of every row. Master mounts ~2 overlays per row (gutter dim, word-diff); this multiplies that by the tokens per line across the whole changeset, and every one goes through `TextPropertyManager::from_entries` into the overlay store. That is also the extra 1.2 GB.
* **`RequestTextHighlights` × 27, 26 over the 50 ms soft budget, largest 2,515 ms** — a single file parsed synchronously on the editor thread; the per-grammar engine cache helped construction but not parse time.

Plain scrolling is unaffected because it doesn't remount; anything that rebuilds the stream (`n` paging a collapsed file in, a refresh) pays the whole cost again.

What I think the fix is, and it is a rework of the plugin side rather than a tweak: stop embedding syntax overlays in the mounted entries at all. Keep the per-side span cache, but paint overlays only for the rows in and around the viewport via `addOverlay` on `viewport_changed` (the plugin already tracks the viewport for the sticky header), clearing and repainting a namespace on scroll — a few hundred overlays per frame, bounded regardless of changeset size, and it also means only files whose rows are on screen get parsed. That is what CONTRIBUTING §2 asks for and I did not do it. Until that lands, this PR's #3104 half should not merge; the #2318 and #3126 halves are independent of it.

---

sinelaw:
The performance regression reported above is fixed in `b64724a`, and the do-not-merge no longer applies.

**What changed.** The #3104 design was reworked so its cost scales with the screen instead of the changeset. Rows no longer carry per-token inline overlays through the mount; each code row carries a small reference (hunk, side, line, where its text starts) and a painter clears and repaints an overlay namespace for the rows in and around the viewport on every viewport change, mount or fold. The unit of parsing is one side of one hunk, requested only when a row of it is on screen, cached by content. An arrival repaints the viewport; it never rebuilds the stream. Folds are respected: the window is counted in rows the reader can see, so a collapsed file between two visible ones does not eat it.

**Same measurement as above** (release build, `Review Diff: Range` over `HEAD~100..HEAD` in this repository, 463 files / 1961 hunks, driven in tmux), master and the new commit back to back on a warm cache:

| | master `cf626fc` | `08d8bc4` (previous head) | `b64724a` (now) |
|---|---|---|---|
| open → first hunk row | 8.0s (13.3s cold) | 30.1s | 8.8s (12.0s cold) |
| `n` next hunk, ×5 | 23–99ms | >15s, 9.7s, … | 24–75ms |
| PageDown / End / PageUp | ~25ms | ~25ms | ~25ms |
| RSS after the run | 802MB | 1964MB | 824MB |
| CPU time after the run | 9s | 65s | 10s |
| handlers over 500ms | 1 (`SetPanelContent` of the initial mount) | many | 1 (the same one) |

Colours verified on the release build in tmux: YAML keys/strings on the first screen, and after `Ctrl+End` the shell rows at the bottom of the stream are coloured too; a word-diff span inside a coloured row keeps its own foreground.

**Tests.** Two new e2e tests, each confirmed to hang with its mechanism removed from the plugin: `test_issue3104_syntax_colours_follow_the_viewport` (Ctrl+End onto rows the first paint never reached) and `test_issue3104_collapsing_a_file_paints_the_one_below_it` (mouse-collapse a 300-row file; the file that scrolls up into view must be coloured). The two earlier #3104 tests now wait for the colour rather than asserting on the frame after the text arrives. `audit_mode::` (54) and `review_diff` (80) suites green; fmt, clippy and `check-types.sh` clean.

---

Issue #3085, opened 2026-08-26
Bracket highlighting: matchting bracket highlight should use a color not present in the rainbow effect


Korkman:
fresh v0.4.10

Shifting the rainbow color for highlighting the matching bracket is a bad idea. It often matches another color close by:

<img width="384" height="132" alt="Image" src="https://github.com/user-attachments/assets/e7f868bb-bcec-486a-85cc-e2ff86f5d5f9" />

Either reserve a color not present in the rainbow list (white?) or maybe inverse the cell for a very visible effect.

Issue #3084, opened 2026-08-26
Bracket highlighting: first page does not highlight matching brackets when rainbow is enabled


Korkman:
fresh v0.4.10

Example: https://github.com/sinelaw/fresh/blob/master/crates/fresh-editor/keymaps/default.json

With Rainbow Brackets enabled, while on the first page, matching brackets won't be highlighted. Scroll a few pages down until the second level color alternates, then highlighting will work.

sinelaw:
Reproduced on 0.4.10 (release tarball, x86_64-musl), Linux, tmux 100x20, default theme, `highlight_matching_brackets: true`.

Steps:
1. `t.json`: `{ "a": [1, 2, 3], "b": { "c": 4 } }`
2. `fresh --no-restore t.json` — cursor lands at 1:1, on the opening `{`.
3. Read the fg colour of every bracket, then move the cursor off the bracket and read again.

```
rainbow_brackets: false   cursor ON  '{' :  {=15  [=251 ]=251 {=251 }=251  }=15    <- pair changes
                          cursor OFF     :  {=251 [=251 ]=251 {=251 }=251  }=251

rainbow_brackets: true    cursor ON  '{' :  {=6   [=2   ]=2   {=2   }=2    }=6
                          cursor OFF     :  {=6   [=2   ]=2   {=2   }=2    }=6     <- identical
```

With rainbow on, moving the cursor onto a bracket produces **no change at all** — the rainbow depth colour is what you see either way.

One correction to the title: it is not limited to the first page. With a 62-line file I checked cursor-on-bracket at 1:1, after `Down`/`Up`, and after `PageDown`+`PageUp`+`Ctrl+Home` (a forced repaint) — `{` stays `6` in every case. A short same-line pair behaves the same: `[`/`]` on line 2 stay `2`/`2` whether or not the cursor sits on the `[`.

---
_Reproduced by an automated agent driving the editor interactively in tmux against the 0.4.10 release build._

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #3083, opened 2026-08-26
Bracket highlighting: default theme without rainbow style does not highlight matching brackets


Korkman:
fresh v0.4.10

As an example, edit `crates/fresh-editor/keymaps/default.json`.

With rainbow coloring turned off, the default theme won't show a different color for the matching bracket highlight, so it's invisible.

sinelaw:
Partly reproduced on 0.4.10 (release tarball, x86_64-musl), Linux, tmux 100x20, default theme, `rainbow_brackets: false`, `highlight_matching_brackets: true`.

Steps:
1. `t.json`: `{ "a": [1, 2, 3], "b": { "c": 4 } }`
2. `fresh --no-restore t.json` (cursor at 1:1, on the opening `{`), read each bracket's fg colour; then move the cursor off and read again.

```
cursor ON  '{' :  {=15   [=251 ]=251 {=251 }=251  }=15
cursor OFF     :  {=251  [=251 ]=251 {=251 }=251  }=251
```

The highlight **is** applied — the pair goes `251` → `15` — so it is not absent. But `15` is bright white and `251` is `#C6C6C6`: a small brightness step on the glyph only, with no background change and no bold, which is easy to miss entirely. That looks like the same complaint as #3085.

Worth separating from #3084: there the pair does not change colour **at all** when rainbow is on.

---
_Reproduced by an automated agent driving the editor interactively in tmux against the 0.4.10 release build._

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3161, opened 2026-09-02
fix(render): tab markers, indentation guides, block rectangles, angle brackets, and an explorer scrollbar

### LINKED ISSUE #2859: scroll bar in explorer is not visible

### LINKED ISSUE #3077: Visible whitespace should not show spaces when there are none
fresh version 0.4.10

With both enabled:
```json
"editor": {
    "whitespace_tabs_leading": true,
    "whitespace_spaces_leading": true,
}
```

This is the tab width being displayed as tab + spaces, which is wrong IMHO.
Image

(I need to see where I have to fix mixed indenting)

### LINKED ISSUE #3079: Indentation guide masks visible tab whitespace
fresh editor v0.4.10

Currently the visible tab indent character will be masked by the indentation guide character. I suggest, when tab size >= 2, to move the tab symbol one column to the right when the indentation guide would otherwise mask it.

Image

### LINKED ISSUE #3090: Bracket highlighting: Rainbow coloring treats operators as angle brackets
fresh v0.4.10

Image

Image

For the languages I know of, that is wrong (it is somewhat correct for HTML, but nesting angular brackets isn't a thing).

### LINKED ISSUE #3148: Tabs desynchronize block selection
fresh 0.4.10

Tab whitespace desynchronizes the block selection (alt-shift-arrows) background from the cursor. The cursor is correct and represents the copied boundary. Unrelated to whitespace visibility.

Image

Image

sinelaw:
## Summary

Five rendering defects, each reproduced on master first and each verified by
reading the pane cell by cell — foreground, background and column — against a
real editor driven with real keystrokes in tmux, which is how they were
measured on the issues.

| | before | after |
|---|---|---|
| #3077 | `\t` → `→···` | `→   ` |
| #3079 | `\t\t` → `▏···▏···` | `▏→  ▏→  ` |
| #3148 | highlight ends 3 cells left of the cursor | highlight ends *at* the cursor |
| #3090 | `)` of `(a < b` coloured `3`, next line's `(` `126` | `)` `2`, next line's `(` `6` |
| #2859 | no bar at any offset | thumb + track in the panel's last column |

---

### #3077 — a tab's padding is not spaces

With `whitespace_spaces_leading: true` every column of an expanded tab drew
the space dot, so a line indented with one tab rendered `→···` — three dots
for columns holding no spaces — and `\t` + four real spaces rendered
`→·······`, which made exactly the tab/space mix the reporter turned the
markers on to find unreadable.

The view pipeline maps every column of an expansion back to the tab's single
source byte, so a space that shares its byte with the cell before it is
padding rather than a space. Those columns now draw nothing, under the
leading/inner/trailing settings and inside a selection alike (the
`whitespace_in_selection` path dotted them too).

`\t    ` now reads `→   ····`: one marker per whitespace *character*.

### #3079 — a guide no longer swallows the tab marker

With `indentation_guide: "all"` every `→` was replaced by `▏`, at every
indent level, so while guides were on there was no way to tell tab
indentation from space indentation.

They share the expansion now: the guide keeps the tab stop it belongs to and
the marker moves one column right, onto the expansion's second column — the
reporter's suggestion. Mixed indentation reads directly:

```
▏→  x = 1        <- one tab
▏···▏···y = 2    <- eight spaces
▏→  ▏···z = 3    <- a tab and four spaces
```

### #3148 — a block rectangle is painted where the cursor is

The rectangle's columns are byte offsets within a line (`cursor.position -
line_start`, which is also what a block copy indexes by), but the painter
tested them against each cell's index into the *view* row — and a view row
has the tab already expanded. With one leading tab and `tab_size: 4` the
highlight covered display columns 1–4 while the cursor stood at 7.

Cells convert their source byte into a column in their **logical** line
before the block test, so a wrapped continuation row resolves against its
line rather than against the row it starts. A tab is then wholly in the
rectangle or wholly out of it, which is what copying it takes. Lines with no
tab are unchanged, and the one rope lookup per row is paid only when a block
rect is on screen.

### #3090 — angle brackets are brackets only in markup

The rainbow classifier counted `<` and `>` in every language, and they are
not merely tinted — they push a nesting level. In C the `<` of `if (a < b)`
was never closed, so the `)` lost its opening bracket's colour and *every*
bracket on the following lines shifted with it. One comparison operator
recoloured the whole file below it.

The pair set is the language's answer now — `Language::angle_brackets_are_delimiters()`,
true for HTML and templ, where a `<` opens a tag and the pair is
unambiguous. `BracketHighlightSettings::with_language` resolves it once per
frame at the one place that knows which buffer is being drawn, and it covers
all three consumers: the cursor's pair highlight, the depth scan, and the
viewport colorization (whose cache keys on it, so a language switch on the
same buffer re-derives).

### #2859 — the explorer draws a scrollbar when the tree overflows

The panel gave no sign that a tree was longer than its body, at any offset,
and had no bar code at all: the editor's bar is a split-pane feature and the
reusable widget was wired only into the settings UI.

The tree is deliberately not a `viewport` — `FileTreeView` windows its own
rows (sticky ancestors, the search filter) and owns the offset — so the bar
is drawn from the model's three numbers, carried on the description as
`Explorer::scroll` and present only when the tree overflows. Geometry is
`ScrollbarState::thumb_geometry`, the same one the settings panel and the
editor use, so the thumb sits where the rest of the editor would put it.

The bar takes the panel's last inner column rather than floating over the
rows: layout pushes a row's trailing git-status slot flush to the right edge,
so an overlay bar would sit on top of those markers. Rows are measured one
column narrower by the same layout that answers a press, so nothing
re-derives a column.

---

## Tests

11 new tests. Each asserts the cells actually measured on master before the
fix — the `→···`, the `▏···`, the off-by-`tab_size - 1` rectangle, the `3`
and `126` bracket colours, the absent bar — so each of them fails there:

- `issue_3077_tab_padding_markers` — the three indents of the reporter's
  file, cell by cell, plus the in-selection path.
- `issue_3079_guide_and_tab_marker` — guide + marker per level, markers
  unmoved with guides off, and a tab line that no longer renders like a space
  line.
- `issue_3148_block_selection_tabs` — the reported gesture, asserting the
  painted rectangle's right edge *is* the cursor's cell, with the
  no-tab control.
- `issue_3090_angle_brackets_not_rainbow` — the `)` shares its `(`'s colour,
  a `<` on one line does not shift the next line's brackets, and HTML keeps
  its tag delimiters. Plus unit tests on the depth scan.
- `issue_2859_explorer_scrollbar` — an overflowing tree draws a thumb on a
  track, the thumb follows a wheel scroll to the track's end, and a tree that
  fits draws neither. Plus description-level tests in `shell::file_explorer`.

Two existing unit tests changed, both in `split_rendering`:
`indentation_guide_render_for_tabs` and
`indentation_guide_active_mode_supports_tabs` asserted the masking — a
tab-indented line rendering `▏   child`, with no marker anywhere — which is
the behaviour #3079 reports. They now assert `▏→  child`.

`cargo fmt --check` and `cargo clippy --all-features --all-targets` are
clean, and `fresh-editor`'s 2543 unit tests plus the integration slices over
every surface these changes touch pass locally. The full `all_tests` binary
runs single-process and outlived the sandbox this was written in, so it was
run in slices there; CI's `nextest` run on all three platforms is the check
on the whole of it, and it is green.

## Open questions

1. **#3079, a one-column tab expansion.** A tab landing one column short of
   its tab stop has nowhere to shift its marker to. This gives that column to
   the marker (the tab is the fact that would otherwise be unrecoverable;
   the guide's column is still marked on every other row) — but the opposite
   choice is just as defensible, and the reporter's suggestion was silent on
   it.
2. **#3090, generics.** A C++ template parameter list, and a Rust generic
   argument list, now go uncoloured rather than mis-coloured: a comparison
   and a type-argument list cannot be told apart without parsing the
   expression. A tighter rule could opt those in per byte, from the grammar's
   own `PunctuationBracket` category — precise where the grammar scopes them,
   silent where it does not, and it would not cover the cursor-pair's
   backward scan, which runs outside the highlighted range. Left out
   deliberately; happy to add it.
3. **#2859, dragging.** The bar is drawn, not draggable: the wheel already
   scrolls the panel, and drag needs a gesture, an app-side fact and drag
   state that the issue ("scroll bar ... is not visible") does not ask for.
   Also note **#3071**, an open PR against this issue from before the
   explorer became a description; this does not build on it, and whichever
   lands first, the other needs rework.

Out of scope, deliberately: #3083 / #3084 / #3085 (matching-bracket
highlight visibility under rainbow), which need a decision on the visual
treatment.

fix #3077
fix #3079
fix #3148
fix #3090
fix #2859

---

https://claude.ai/code/session_01MMaEZuR7QdjFfiahH86HsM


crates/fresh-languages/src/lib.rs
```diff
@@ -479,6 +479,22 @@ impl Language {
...
+    /// Only markup qualifies. In HTML (and templ, which is HTML with Go in
+    /// it) a `<` opens a tag and the matching `>` closes it, so the pair is
+    /// unambiguous and nests exactly as far as the tag does. Everywhere else
+    /// the same two characters are overwhelmingly operators — `if (a < b)` —
+    /// and the cases where they are delimiters (`template <typename T>`,
+    /// `Vec<T>`) cannot be told from a comparison without parsing the
+    /// expression. Counting them as brackets there gave one comparison
+    /// operator the power to recolour every bracket below it: the `<` pushed
+    /// a nesting level nothing popped (issue #3090).
+    pub fn angle_brackets_are_delimiters(&self) -> bool {
```
sinelaw, line 494: The new language rule reaches the highlighter but not the *other* consumer of the same bracket table: `crates/fresh-editor/src/app/text_ops.rs:392` still hardcodes

```rust
const BRACKET_PAIRS: &[(char, char)] = &[('(', ')'), ('[', ']'), ('{', '}'), ('<', '>')];
```

plus the `'<' => Some(('<', '>', true))` / `'>' => …` arms just below, for every language. So after this PR the two surfaces disagree, and the jump command still reproduces #3090's exact failure mode.

Concrete: in a C file, `if (a < b) {`, put the cursor on `b` and run go-to-matching-bracket. The backward enclosing-bracket scan (text_ops.rs:425-437) walks the pairs including `<`/`>`, hits the `<` of `a < b` with `depths[3] == 0`, decides that comparison operator is the enclosing opening bracket, and jumps forward to the next `>` anywhere in the file — or reports `bracket_none` if there isn't one — instead of jumping to the `)` that actually encloses the cursor.

Gating that table on `angle_brackets_are_delimiters()` (the buffer's `highlighter.language()` is reachable from `active_state`) would keep the two surfaces telling the same story.

---

crates/fresh-editor/src/view/bracket_highlight_overlay.rs
```diff
@@ -158,6 +198,9 @@ struct ColorizationCache {
     viewport: Range<usize>,
     colors: [Color; 6],
     skip_ranges: Vec<Range<usize>>,
+    /// Whether `<`/`>` counted as brackets — a language switch on the same
+    /// buffer changes the overlays this pass would produce.
+    angle_brackets: bool,
```
sinelaw, line 214: The colorization cache gets an `angle_brackets` key so a language switch on the same buffer re-derives — good — but the *matching-pair* half of `update()` has no equivalent, and it short-circuits on cursor position alone:

```rust
if self.last_cursor_pos == Some(cursor_position) && !colors_changed {
    return updated;      // line 288
}
```

`angle_brackets` is not part of that condition and is not stored anywhere for the matching path, so a change to it with a stationary cursor is invisible.

Concrete: open an `.html` file, park the cursor on the `<` of a tag (the pair overlay paints `<`…`>`), then change the buffer's language to something else via the language dialog. `update_colorization` re-runs and drops the angle-bracket rainbow, but `last_cursor_pos` still equals `cursor_position`, so the stale `<`/`>` pair overlay survives until the cursor moves. The reverse (switch *to* HTML with the cursor already sitting on a `<`) leaves the pair unhighlighted until the caret moves.

Tracking the last `angle_brackets` alongside `last_cursor_pos` — or folding it into that guard the way `colors_changed` is — closes it.

---

crates/fresh-editor/src/view/ui/split_rendering/orchestration/selection_sweep.rs
```diff
@@ -109,17 +109,31 @@ impl<'a> SelectionActiveSet<'a> {
     /// (ANSI / virtual cells) — those still get block-rect checks
```
sinelaw: This paragraph is now false, and it is the paragraph a reader of `contains` will believe. It still says a byte-less cell "still get[s] block-rect checks but no linear-range coverage, matching the existing logic", but the new body gives such a cell `line_column == None` (`block_line_column` returns `None` when `byte_pos` is `None`) and `is_some_and` then makes `block` unconditionally `false`.

That is a real behaviour change, not only a doc drift: cells that map to no source byte — soft-wrap indent padding, fold placeholders, plugin-injected inline text — used to be covered by a block rect whose column span reached them and now never are. It is probably the change you want (a rectangle can't cover a cell that isn't in the file), but it deserves to be stated rather than contradicted, otherwise the next person debugging a missing block highlight over injected text reads this comment and looks in the wrong place.

---

crates/fresh-editor/src/app/render.rs
```diff
@@ -3799,7 +3810,17 @@ impl Editor {
                 )
             })
             .collect();
-        fe::Body::Rows(rows)
+        // The bar's three numbers, in tree rows: what the tree holds, what the
+        // panel shows, and where the window sits. `None` when the whole tree
+        // fits, which is the whole of "should there be a bar" (issue #2859).
+        let scroll = (viewport_rows > 0 && display.len() > viewport_rows).then(|| fe::Scroll {
+            offset: view
+                .get_scroll_offset()
```
sinelaw: The offset is clamped to `display.len() - 1`, but `ScrollbarState::thumb_geometry` computes its own `max_scroll = total_items - visible_items` = `display.len() - viewport_rows` and saturates the ratio at 1 there. The model's ceiling is higher: `FileTreeView::max_scroll_offset` (view.rs:378-394) deliberately returns something *larger* than `visible.len() - viewport_height` when sticky ancestors are pinned, because the pinned parents eat rows of the window — and `scroll_file_explorer_view` clamps the wheel to exactly that value.

So in any tree deep enough to pin ancestors, the last `sticky.len()` rows of scrolling happen with the thumb already parked at the bottom of the track: the bar says "you are at the end" while the wheel still moves the list. The `an_overflowing_tree_draws_a_scrollbar` test doesn't catch it because its fixture never produces sticky rows.

Feeding `ScrollbarState` a `window` derived the same way the model derives it — or clamping `offset` to `view.max_scroll_offset()` and passing `total`/`window` that make that offset land at the track's end — keeps the thumb honest.

---

sinelaw (commented):
Read every hunk against the surrounding files (view pipeline's tab expansion, the selection sweep, the file-tree model, the scrollbar widget). The five fixes themselves hold up under the cases I could construct:

- **#3077** — `is_tab_padding` keys off the shared source byte, and the pipeline only ever maps two cells to one byte for tab expansions and `<XX>` escapes (whose glyphs are never a space), so the test is sound. Tabs are never split across a soft wrap (`Break` tokens sit at token boundaries), so the `display_char_idx == 0` guard can't misfire on a continuation row.
- **#3079** — I walked the guide/marker handoff for guides at the tab stop, guides *inside* an expansion (`indent_size < tab_size`), a one-column expansion, adjacent tabs, and active mode. No case draws the marker twice or loses both glyphs.
- **#3148** — the new unit matches how `block_rect` builds its columns, and it also fixes wrapped continuation rows, which the old per-row `byte_index` got wrong.

Four things I'd want addressed, in rough order of how much they bite. The first is the one I'd call a bug in shipped behaviour.

---

crates/fresh-languages/src/lib.rs
sinelaw: Confirmed and fixed in d72caf4 — this was the real bug of the four.

Rather than gate a second table, `text_ops` now asks the highlighter's: `bracket_pairs()` and `get_bracket_pair()` are `pub(crate)`, and the local `const BRACKET_PAIRS` and the eight match arms are gone. One table, one language question, so the two surfaces cannot drift again.

Your repro is now a test — `the_jump_command_agrees_about_what_a_bracket_is`: caret on the `b` of `int g(void) { if (a < b) { return 2; } }` in a `.c` file, Ctrl+], assert the caret lands on the `)`. I checked it fails without the change by reverting `text_ops.rs` alone and re-running: it fails exactly as you described (the caret does not move, because there is no `>` in the file). Its sibling `the_jump_command_still_pairs_markup_delimiters` pins the other half — Ctrl+] on the `<` of `<span>` in HTML still jumps to the `>` — so the fix is the language gate, not a removal.

---

crates/fresh-editor/src/view/bracket_highlight_overlay.rs
sinelaw: Right — the guard was asymmetric with the cache. Fixed in d72caf4 by folding it in the way `colors_changed` is: a `last_angle_brackets` field beside `last_cursor_pos`, part of the short-circuit, and cleared with it in `clear()`, `invalidate()` and the `!settings.matching` early-out (so re-enabling matching recomputes rather than trusting a stale pair).

Covered by `a_language_switch_redraws_the_pair_under_a_stationary_cursor`: caret parked on the `<` of `<span>`, `update` called with `angle_brackets: true` (a pair is painted), then again with `false` and the *same* cursor — the overlays must be retracted — then back to `true`, repainted. Both of your directions, neither one waiting on the caret.

---

crates/fresh-editor/src/app/render.rs
sinelaw: Good catch — the two ceilings were different numbers and I fed the widget the wrong one. Fixed in d72caf4 with your second suggestion.

`Scroll` now carries `max_offset` (the model's `max_scroll_offset()`, clamped to `total - 1`) alongside `total` and `rows`, and the bar builds `ScrollbarState::new(total, total - max_offset, offset.min(max_offset))` — a `visible` chosen so the widget's own `total_items - visible_items` ceiling *is* the model's. With ancestors pinned that is fewer rows than the panel shows, which also makes the thumb slightly shorter: fewer of the tree's rows are reachable as ordinary ones, so that reads correctly too. (`window` became `rows` to stop the two senses sharing a name.)

The fixture gap you spotted is closed by asserting on the numbers rather than on a tree deep enough to pin: `the_thumb_reaches_the_end_only_at_the_models_last_offset` builds a `Scroll` with `total: 40`, `rows: 8`, `max_offset: 34` (two ancestors pinned) and asserts the thumb is *not* at the track's end at offset 32 — the naive ceiling — and *is* at 34.

---

crates/fresh-editor/src/view/ui/split_rendering/orchestration/selection_sweep.rs
sinelaw: Agreed on both counts — it is the change I want, and leaving the old sentence there would have sent the next reader to the wrong place. Rewritten in d72caf4: the paragraph now says a byte-less cell is in no selection of either kind, and a second one names what changed and why — soft-wrap indent padding, fold placeholders and plugin-injected inline text used to be swept in by their *view-row* index, and a rectangle cannot cover a cell that is not in the file, which is also what a block copy takes.

---

sinelaw:
`test ubuntu-latest` went red on d72caf4, and I don't believe it is this PR's:

```
Summary [1474.021s] 9562 tests run: 9561 passed (2 slow), 1 timed out, 160 skipped
TIMEOUT [ 180.162s] (7280/9562) fresh-editor::all_tests remote_channel_timeout_tests::test_multiple_reconnections
```

One test, and it **timed out** rather than failing an assertion. `test_multiple_reconnections` spawns one-shot remote agents and loops three rounds of *deliberately* waiting for a request timeout, then replacing the transport — so its wall clock is a function of how loaded the runner is, and 180s is nextest's cap.

Why it isn't this PR's:

- The diff touches the cell pass, the selection sweep, the bracket tables, the explorer description and the `Language` enum. Nothing in `services/remote`, and no path from any of it to a remote channel's reconnection.
- The same test passed on this PR's previous head (174f978) forty minutes earlier, in the run where all 11 checks were green — the remote-channel code is byte-identical between the two commits.
- It passes locally on this head: `remote_channel_timeout_tests` is 10 tests in 8.59s, `test_multiple_reconnections` among them.

So there is no fix to port — nothing has broken that a change of mine or anyone else's would repair. Making the test robust (giving those three rounds a budget that doesn't ride the runner's load) is a real improvement, but it is a different change to an unrelated test and I'd rather not widen this PR to carry it; happy to open it separately if you want.

Re-running the failed job once, per the usual rule that one re-run is the confirmation, not the fix. If it comes back red on the same test I'll treat that as real and dig in.

---

Issue #2843, opened 2026-07-28
Single-line files: highlighting falls out of JSON string state, and wheel scroll collapses to a 100 KB chunk boundary


sinelaw:
Two separate defects found while investigating #2838, both triggered by a file that is one very long line. Both are present on master and are independent of the highlight-perf work in #2841.

Reproduction file: the `fp4_poll.json` attached to #2838 — 451,855 bytes, zero newlines (one 441 KiB line), valid JSON.

## 1. Highlighting turns uniform after the first few wrapped rows

Past roughly 43 KB into the line, the highlighter emits one enormous `Comment` span instead of string/structure spans, so the rest of the view renders in the comment colour.

Spans returned for viewport windows on the same file:

| viewport | spans | first span |
|---|---|---|
| 0..4,000 | 81 | `String@1..8`, `Number@18..24`, … |
| 50,000..54,000 | 1 | `Comment@43059..62192` |
| 200,000..204,000 | 1 | `Comment@195575..212192` |
| 430,000..434,000 | 118 | `String@430001..430013`, … |

Byte 43,059 is exactly the `//` in `src=\"https://forum.fairphone.com/...`. The parse state has already fallen out of the enclosing JSON string, so `//` inside a URL opens a line comment — and because the file is a single line, that comment runs for ~19 KB until something re-synchronises.

Notes from investigation:
- Not a size threshold on its own: synthetic JSON with a single string padded to 8/16/32/48/64/96/128 KB containing a URL keeps the correct `String` category at every size. The real file's content (escaped quotes, HTML, unicode escapes) is needed to trigger it.
- Leading hypothesis: syntect's regex engine gives up on the string rule when matching tens of KB in one `parse_line` call (backtracking/stack limit), after which the parser continues outside the string context. Needs confirmation by instrumenting `parse_line_into_spans` in `crates/fresh-editor/src/primitives/highlight_engine.rs`.
- A naive mitigation (cap the bytes fed to syntect per call) is not sufficient alone: a fragment starting mid-string parsed from a line-start state would still be coloured wrong. The fragment's opening state has to be carried across chunks.

## 2. Wheel scroll after a scrollbar click collapses to a 100 KB boundary

Steps: open the file, click the vertical scrollbar to jump partway down, then scroll the wheel.

Observed viewport state:

```
initial                 top_byte=0       offset=0
after scrollbar click   top_byte=0       offset=935
after wheel-down #1     top_byte=100000  offset=3
[...]

PR #3162, opened 2026-09-02
Opening a huge file does work bounded by the screen, not by the file

### LINKED ISSUE #1806: Editor becomes unresponsive when opening large, single-line data files.
Very slow movements and open for one-liner huge json file, the whole thing almost hangs. something like asciinema-player.min.js

### LINKED ISSUE #3142: Trying to open a very large binary or zip file will freeze or crash
it's quite easy to freeze fresh by clicking a large zip file in file explorer. maybe should be fixed by making opening a file async

sinelaw:
Fixes #3142. Fixes #1806.

## The problem

Two reported bugs, one cause: opening a file did work proportional to the **file**, where the screen only ever needed work proportional to the **viewport**.

**#3142 — a large binary file or archive froze the editor.** It was read into memory whole and indexed before the editor could redraw anything. A 2.1 GB zip took no input for ~11 s, drew nothing for 14 s, and needed ~1.2× its own size resident; on a machine with less RAM than that, the reporter got an OOM instead of a pause. Arrowing onto the file in the explorer previewed it through the same path, so the freeze arrived before you pressed Enter.

**#1806 — a file that is one enormous line was slow to open, then impossible to move in.** Deciding where the viewport could sit meant laying out whole logical lines, so a 53 MB single-line JSON was wrapped in full — hundreds of thousands of rows — to answer a question bounded by the ~30 rows on screen. It took 19 s to first render, against 0.24 s for the identical bytes split across lines.

The second half of #1806 was worse than the slowness and is what the fix is really about: **once open, that file could not be scrolled at all.** `Down` did nothing — the view stayed on the first screenful however long you held the key, so everything past the first screen was unreachable by keyboard, and paging stopped dead ~100 KB in with the cursor stranded behind the view.

## The fix

Three changes, all the same principle — do the work the screen needs, not the work the file implies:

1. **Large binary files load lazily**, the way large text files already did. Nothing about a read-only buffer needs an eager read.
2. **A line's height is answered from as much of it as the screen needs.** Asking "does this line fill the viewport?" no longer means wrapping the whole line.
3. **A file too large to index records where its top row *starts*, rather than how many rows down a line it is.** Moving the view is then a walk over the rows it crosses, which costs the same 40 MB into a file as 40 bytes in.

The third is the structural one. It is why the scrolling works at all: a row *count* into a line cannot be taken past the first chunk the reader returns, so it saturated — and every earlier attempt to fix scrolling just moved the wall (2,861 characters, then 40,000, then byte 99,944, then a cycle around byte 100,000). Nothing else changes for ordinary files, which keep the existing row-numbered path untouched.

A fourth fix came out of walking the finished thing row by row: on the file's **last** row, `Down` threw the cursor back to byte 100,000 and dragged the view with it, so the walk looped into the middle of the file instead of stopping at the end. The wrap-aware intercept declines at a genuine buffer boundary — which is also how it says "not mine" — so the motion fell through to the generic handler, and that handler asks the line iterator for "the next line". `next_line` yields an over-long line in `MAX_LINE_BYTES` pieces, so on a one-line file the next "line" is the next read *piece*: a byte behind the cursor. A genuine next line starts after the cursor; anything else is the same line continuing, and means there is no line below.

## Decisions and trade-offs

**Not "make the open async",** which both issue threads propose. For #3142 that moves a 2.5 GB allocation to another thread: the freeze goes, the OOM stays. For #1806 the cost is not the open at all — the file is already lazily loaded today — it is the first render. With the unbounded read gone there is no large synchronous read left to move; if the open path should be async anyway, that is now a much smaller change.

**Scope was narrowed after review.** The new addressing initially also applied to any wrapped file over 5,000 lines. Review showed that put the viewport and the renderer in different coordinate systems on ordinary files, reachable by pasting twenty lines into a 4,990-line file. It now applies only where no index can exist, stated as one condition that cannot drift from the renderer's.

**Backward movement is sound, not guaranteed exact.** Walking up has no resume point, so it is reconstructed from a bounded look-back. Every press moves one row in both directions and neither stalls, and in every run measured the round trip was exact — including a full walk to the last row and back, which returned to byte 0 — but that is not something the design guarantees. Guaranteeing it needs the viewport or the index to remember where rows started — deliberately not added here.

**Rows injected by plugins are not modelled.** Collapsed folds are handled exactly; conceals, soft breaks and virtual lines are not, because reproducing that stage would duplicate the renderer, and excluding such buffers would reintroduce the two-coordinate problem above. Of the plugins that use them, only `flash` is realistically reachable on a file this large.

**One threshold has a user-visible effect:** past 64 KB into a single line, a continuation row draws flush instead of hanging-indented. Finding the line's start is unbounded work on such a file, which is the cost being removed; the effect is cosmetic, and an indent that deep is discarded as unusable anyway.

**No refusal threshold was added.** With the eager read gone, the memory argument for refusing to open a file goes with it, and picking a size limit is a product call rather than a bug fix.

## Behaviour change worth a decision

A large binary file now opens in large-file mode, so the gutter shows byte offsets and the status bar reads `Byte N` rather than line numbers — no line index is built, which is what makes the open free. That is the existing treatment for any lazily-loaded file. If line numbers there matter, the background line-scan machinery already exists to fill them in; say the word.

## Measurements

Driven in tmux with real keystrokes against `origin/master`, debug build, Linux. The single-line file is a 53 MB JSON array on one line; the control is the same bytes with every comma replaced by a newline.

| | master | this branch |
| --- | --- | --- |
| 2.1 GB zip: editor responds after opening | t+14.2 s, ~11 s accepting no input | **t+3.0 s**, no dead time |
| 2.1 GB zip: resident memory | 2184 MB | **118 MB** |
| 53 MB single-line: first render | 19.0 s | **0.46 s** |
| the same bytes, multi-line control | 0.24 s | 0.24 s |
| `Down` / `PageDown` in that file | 1.39 s / 0.99 s | **0.32 s / 0.30 s** |
| a file of three 17 MB lines: open | 6.27 s | **0.54 s** |

Scrolling was checked exhaustively rather than by sampling: an 11.16 MB single-line file walked from its **first** visual row to its **last**, one `Down` press at a time — 19,076 presses at 600×50, release build, each verified to land on exactly the byte one 585-byte row further on. No stall, no skip, no cycle, and a flat 128–134 ms per press whether the walk was at row 800 or row 17,500, which is the depth-independence claim measured directly. Walking back up returned to byte 0. On master, `Down` never leaves the first screenful.

That walk is what turned up the last-row bug above, at press 19,075.

**The tests assert on work, not on the clock,** so they mean the same on a loaded CI runner as on an idle laptop. Bytes of the file resident after opening it:

| | master | this branch |
| --- | --- | --- |
| 19 MB single-line | **19,688,891** — the file, exactly | 4,222,395 |
| the same bytes, multi-line control | 1,076,667 | 1,076,667 |
| 64 MB binary | **67,108,864** — all of it | 1,114,112 |

The scrolling tests drive arrow and page keys and assert only on what is rendered (CONTRIBUTING §2). Every test here was watched to fail without its fix (§1) — the earlier ones ported onto master, the later ones by reverting the fix in place.

CI green on all eleven checks, including the three platform test jobs.

https://claude.ai/code/session_01SMWCiwwS5MjkC7pwsQd9rX


crates/fresh-editor/src/view/viewport.rs
```diff
@@ -368,6 +368,111 @@ impl Viewport {
...
+    fn row_budget_bytes(cap: usize, wrap_config: &WrapConfig) -> usize {
+        let width = wrap_config
+            .grid_cols
+            .unwrap_or_else(|| {
+                wrap_config
+                    .first_line_width
+                    .saturating_add(wrap_config.gutter_width)
+            })
+            .max(1);
+        cap.saturating_mul(width)
```
sinelaw, line 408: The budget is `cap * width * 8`, and two of the four call sites pass a `cap` that grows with `top_view_line_offset`, not with the screen — so the "bounded by the viewport" property is lost exactly once you are scrolled *into* a huge line.

`apply_visual_scroll_limit` (line 1123) passes `viewport_height + top_offset`, and `scroll_down_visual` (line 1008) passes `top_view_line_offset() + rows_remaining + 1`. `set_top_view_line_offset` has no clamp, so on the 53 MB single-line JSON the offset climbs into the hundreds of thousands as you page down.

Concrete: 120 columns, 40-row viewport, after paging ~750 screens into the line `top_offset ≈ 30_000`. Then `budget ≈ 30_040 * 128 * 8 ≈ 30 MB`. Every wheel tick / PageDown now (a) joins ~300 `MAX_LINE_BYTES` pieces into a 30 MB `String`, (b) hands it to `count_visual_rows_for_text`, which `to_string()`s it again and materializes ~250 000 `WrapMachine` rows, and (c) throws the result away — the saturating branch deliberately passes `cache: None`. Master read the same line but got a `wrap_row_cache` hit for the wrap itself, so in this regime the change is a net regression rather than a win.

Note the cliff is only avoided at the far end: once `budget >= line_len` the read comes back `complete` and *is* cached again, so the pathological band is the middle of the line.

Worth either capping the byte budget independently of `cap` (the question "does this line have ≥ N rows" only needs a prefix; you can walk it incrementally instead of materializing `cap * width * 8` bytes at once), or caching the saturating floor under a separate key so repeated scroll steps at the same depth don't re-run the wrap machine.

---

crates/fresh-editor/src/view/viewport.rs
```diff
@@ -368,6 +368,111 @@ impl Viewport {
...
+            // word wrap is greedy left to right, so appending text never
+            // removes a wrap — which makes "the prefix already fills `cap`
+            // rows" a sound answer for the whole line. Deliberately not
+            // cached: the count is a floor, not the line's true height.
+            //
+            // One row of slack, because the prefix is cut at a byte offset:
+            // a multi-byte character split across the cut comes back as
+            // replacement characters, which can be a column or two wider than
+            // what it replaced and so add a row the real line does not have.
+            let prefix_end = line_start + content.len();
```
sinelaw: `prefix_end` should be `iter.current_position()`, not `line_start + content.len()`.

`next_line` builds each piece with `String::from_utf8_lossy`, so for content that is not valid UTF-8 every invalid byte becomes a 3-byte U+FFFD and `content.len() > (bytes actually consumed)`. `prefix_end` then points *past* the real end of the prefix, and `count_visual_rows_for_line` slices `virtual_lines` / `soft_breaks` with `partition_point(|p| *p < line_end)` — so plugin virtual lines and soft breaks that live beyond the prefix get counted into it (`extra_virtual_rows` is added verbatim to the returned count).

That inflates `rows` in the direction that makes `rows > cap` fire when the real line may not reach `cap`. The callers then trust it: `scroll_down_visual` line 1074 does `set_top_view_line_offset(rows_remaining)` on the strength of it, which parks the viewport past the line's real last row.

`iter.current_position()` is the exact byte the budgeted read stopped at and is already available here, so this is a one-line fix that also makes the value agree with the `line_end` used on the complete path below.

---

crates/fresh-editor/src/view/viewport.rs
```diff
@@ -368,6 +368,111 @@ impl Viewport {
...
+        let (line_start, mut content, complete) = iter.next_logical_line_budgeted(budget)?;
+
+        if !complete {
+            // The prefix is a lower bound on the whole line's row count —
+            // word wrap is greedy left to right, so appending text never
+            // removes a wrap — which makes "the prefix already fills `cap`
+            // rows" a sound answer for the whole line. Deliberately not
+            // cached: the count is a floor, not the line's true height.
+            //
+            // One row of slack, because the prefix is cut at a byte offset:
```
sinelaw: The "one row of slack" argument assumes the prefix has exactly one lossy cut. It can have many.

`next_logical_line_budgeted` sets `max_line_bytes = saved_max.min(max_bytes)`, i.e. `min(MAX_LINE_BYTES, budget)`. As soon as `budget > MAX_LINE_BYTES` (100 KB) — which happens whenever `cap > ~104` at 120 columns, so on any of the offset-scaled caps discussed on `row_budget_bytes` — the prefix is assembled by the join loop from `ceil(budget / 100_000)` separate `next_line` pieces, and **each** piece is independently `from_utf8_lossy`-decoded. Every 100 KB boundary that lands mid-character contributes its own replacement characters.

So the inflation is up to ~3 columns *per boundary*, not per prefix. With `budget = 8 MB` on a multi-byte line (CJK, or a UTF-8 JSON with non-ASCII strings) that is ~80 boundaries, potentially ~240 spurious columns — 2 rows at 120 columns, more on a narrow pane. Then `rows > cap` can be true for a line whose real row count is below `cap`, and `scroll_down_visual` / `apply_visual_scroll_limit` will place `top_view_line_offset` past the line's real bottom.

Either bound `budget` to `MAX_LINE_BYTES` so there is only ever one cut (which also fixes the cost problem above), or make the slack proportional to the number of pieces joined.

---

crates/fresh-editor/src/view/viewport.rs
```diff
@@ -2595,6 +2702,120 @@ mod tests {
...
+        let content = format!("{}\n", "word ".repeat(3_200_000));
+        let mut buffer = Buffer::from_str_test(&content);
+        let mut vp = Viewport::new(80, 24);
+        vp.line_wrap_enabled = true;
+
+        let started = std::time::Instant::now();
+        vp.set_top_byte_with_limit(&mut buffer, &[], &[], 0);
+        assert_eq!(vp.top_byte(), 0);
+        assert!(
+            started.elapsed() < std::time::Duration::from_secs(2),
```
sinelaw: A wall-clock bound in a unit test will be flaky on a loaded CI box. This one builds a 16 MB `String` via `"word ".repeat(3_200_000)`, feeds it through `Buffer::from_str_test`, and then asserts the clamp finishes in under 2 s — in a debug build, on a runner that may be sharing a core with other jobs. Note it also measures only `set_top_byte_with_limit`, but the 16 MB allocation and buffer construction ahead of it can themselves contend for memory bandwidth.

The behavioural half of the test (`assert_eq!(vp.top_byte(), 0)`) is the part that would actually catch a regression to a whole-line wrap only indirectly. Consider asserting on bytes read instead — e.g. drive `next_line_visual_rows_capped` directly and check `iter.current_position()`, the way `capped_row_count_saturates_on_a_huge_line` does — so the test fails for the right reason and never for scheduler noise.

---

crates/fresh-editor/tests/e2e/large_file_open_bounded.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    let started = Instant::now();
+    harness.open_file(&path).unwrap();
+    let elapsed = started.elapsed();
+    let growth_mb = current_rss_kb().saturating_sub(baseline) / 1024;
+
+    eprintln!("opened {FILE_MB} MB binary in {elapsed:?}, RSS +{growth_mb} MB");
+
+    assert!(
+        growth_mb < MAX_GROWTH_MB,
```
sinelaw: Asserting on process-wide `VmRSS` makes this test sensitive to anything else running in the same process. `cargo nextest` runs each test in its own process, so that part is fine, but the 16 MB headroom is still measured against a baseline taken after a single `render()` — the open itself legitimately allocates (syntax/grammar detection, the initial chunk loads the first render pulls in, tokenizer arenas), and glibc's arenas can also touch new pages without any of it being the file. A 64 MB file with a 16 MB budget leaves ~25% margin against a metric the test does not control.

Since the property under test is "the file is not read into memory", the direct assertion is the one `test_load_large_binary_file_is_lazy` already uses — `!buffer.buffers[0].is_loaded()` / `line_count() == None`. If the e2e coverage is wanted, consider reaching into the harness's buffer for the same check and keeping RSS as an `eprintln!` diagnostic rather than an assertion.

Same for the `elapsed < 5s` bound just below, and for the `control_time * 10` ratio in `single_line_open_is_not_penalised_against_multi_line` — on a shared runner a fast control turns a 10x ratio into noise, which is presumably why the `max(2s)` floor is there; it just moves the flake threshold rather than removing it.

---

sinelaw (commented):
Reviewed the diff for correctness. The overall shape holds up: I walked each of the four converted scroll-math sites and convinced myself the saturating count is consumed safely — in every case a saturating return (`line_end == None`) is immediately followed by a `return`, so the "iterator parked mid-line" contract is never violated, and the `cap` each site passes is one more than the largest value that could change its decision, so `rows > cap` always implies the real line clears the bar the caller needs. `next_logical_line` delegating to the budgeted version with `usize::MAX` is behaviour-preserving (the tightening `min` collapses to the saved cap and the extra loop condition can't fire), and the binary branch in `load_large_file_internal` really does land in the same state the old eager path produced (`from_bytes_raw` already set `is_binary`, and `BufferFormat::new(line_ending, encoding)` matches the old `set_default_encoding`). Chunk loads for `BufferData::Unloaded` are raw `read_range`s with no encoding conversion, so a binary buffer still round-trips byte-for-byte.

Five things worth a look, inline. The first three are about the read budget rather than the scroll logic:

1. `row_budget_bytes` — the budget scales with `top_view_line_offset`, so the viewport bound is lost once you scroll deep into a huge line, and in that band the uncached saturating count is slower than master's cache hit.
2. `prefix_end` should be `iter.current_position()`; `content.len()` overshoots for lossily-decoded bytes and pulls extra virtual lines / soft breaks into the prefix's count.
3. The "one row of slack" reasoning assumes a single lossy cut, but a budget above `MAX_LINE_BYTES` produces one cut per 100 KB piece.

4 and 5 are test robustness (wall-clock and RSS assertions).

---

crates/fresh-editor/src/view/viewport.rs
sinelaw: Right, and worse than "not bounded by the viewport" — as you say, in that band it was slower than master, because master at least got the cache hit. Fixed in 2a33321 by clamping the budget to `MAX_LINE_BYTES` so it no longer scales with `cap` at all.

I went with the ceiling rather than the incremental walk or a second cache key because it also settles the multi-cut problem you raise on line 455, and because the fallback is already the right one: a `cap` too large for 100 KB to settle comes back `complete`, and the complete path *is* cached. So the deep-scroll band is exactly master's behaviour — full read, cached wrap — instead of worse than it, and the open path (where `cap` is `viewport_height` and the budget lands around 20–80 KB) keeps the win untouched.

`row_budget_never_exceeds_one_read_piece` pins the ceiling against the offset-scaled caps specifically.

The incremental walk is the better answer if the deep-scroll band ever needs to be *faster* than master rather than merely no worse — but that wants the byte offset behind `top_view_line_offset`, which is the coordinate rework the module's own `anchor` doc comment is about, and I'd rather not start it inside a bug fix.

---

crates/fresh-editor/src/view/viewport.rs
sinelaw: Correct, and it errs in the dangerous direction — inflating `rows` so the saturating branch fires on a line that never reaches `cap`. Fixed in 2a33321: `prefix_end` is now `iter.current_position()`, which as you note also makes it agree with the `line_end` the complete path uses two lines down.

---

crates/fresh-editor/src/view/viewport.rs
sinelaw: Took the first option — `budget` is clamped to `MAX_LINE_BYTES` in 2a33321, so the join loop can never run: `next_line` returns at most `max_line_bytes == budget` bytes in one lossily-decoded piece, and `content.len() >= max_bytes` ends the loop immediately. One cut, which is what the row of slack covers.

Making the slack proportional to the piece count would have kept the argument sound but not the cost, and the cost is the same root cause — so one ceiling for both.

---

crates/fresh-editor/src/view/viewport.rs
sinelaw: Agreed, and the clock was measuring the wrong thing anyway. Replaced in 2a33321 with an assertion on work: a line's full row count is only ever computed through `wrap_row_cache`, so an empty cache after the clamp *is* the statement that the whole line was never wrapped. Deterministic, and it fails on master (I ported it there and watched it go red).

I didn't drive `next_line_visual_rows_capped` directly for this one, because then it would be testing the same thing `capped_row_count_saturates_on_a_huge_line` already tests rather than the clamp; the cache assertion keeps the clamp itself in the frame. `clamping_ordinary_lines_does_count_them` is its control, so an empty cache can't pass by the clamp never having run. The line is 1 MB now instead of 16, since size no longer buys the test anything.

---

crates/fresh-editor/tests/e2e/large_file_open_bounded.rs
sinelaw: Taken, for all three assertions — 2576180. `Buffer::resident_bytes` sums the loaded string buffers, and both e2e tests now assert on that instead of on RSS or the clock; timings stay as `eprintln!` diagnostics.

It turned out to be a sharper instrument as well as a steadier one, because the numbers are exact:

| | master | this branch |
| --- | --- | --- |
| 19 MB single-line, resident after open | **19,688,891** | 4,222,395 |
| 19 MB multi-line control | 1,076,667 | 1,076,667 |
| 64 MB binary, resident after open | **67,108,864** | 1,114,112 |

Master reads the single-line file's every byte to wrap it — 19,688,891 is the file's exact length — while the identical bytes in multi-line shape cost 1 MB. That is the bug stated in bytes rather than in seconds, and it removes the ratio question with it: the assertion is now `10x the control` on a figure that is byte-identical across runs (I ran it three times to check), not on one the scheduler moves.

`resident_bytes` is a new public accessor. It is a plain read-only sum and it is the only way to observe the lazy path's central property from outside the buffer module, so I think it earns its place beyond the tests — but say the word if you'd rather it were `#[cfg(test)]` or `pub(crate)` with the e2e coverage dropped.

---

sinelaw:
`test macos-latest` is red on `2576180`: one test **timed out** at nextest's 180 s limit, everything else passed.

```
Summary [1079.655s] 9520 tests run: 9519 passed (1 slow), 1 timed out, 162 skipped
TIMEOUT [ 180.070s] (4824/9520) fresh-editor::all_tests
```

That test is not one I'd dismiss on sight — compose soft breaks flow into `count_visual_rows_for_line`, which this PR changes — so I chased it rather than re-running first. It does not look like this PR's:

- **It costs 2.8 s, not 180 s.** On the commit before the review fixes it ran in 2.825 s; on `2576180` it runs in 2.831 s. Same machine, same test, no measurable change from the fixes.
- **Including under CI's own feature set.** My earlier sweep used default features, so I re-ran it the way CI does — `cargo nextest run --all-features` — on this head: **3.287 s**, 55× inside the timeout. All six tests in that module pass.
- **It was green on macOS on the previous head** (`313cc7f`, all 11 checks). The only production change since is in `2a33321`, and neither half of it can reach this test: the budget clamp only ever *lowers* the bytes read, and it only engages on logical lines longer than 4 KB, which this fixture doesn't have; `prefix_end` only moves on the same >4 KB path. (`2576180` is tests plus `resident_bytes`, which nothing in production calls.)

A 64× gap between a 2.8 s local run and a 180 s timeout is the shape of a stalled runner rather than of slow code, and the run's own numbers agree: this macOS job spent 1079 s on a suite that takes ~1205 s here across all of `fresh-editor` — it was not uniformly slow, one test stopped.

So I'm re-running the macOS job once to confirm, which is the one re-run I'll spend on it. `rerun-failed-jobs` is refused while `test ubuntu-latest` and `test windows-latest` are still in flight (`403 This workflow is already running`), so I'll trigger it as soon as the run finishes and report the result here. If it fails a second time I'll treat it as real and root-cause it properly rather than re-running again.

---

sinelaw:
Re-run came back green, so that was the runner:

```
Summary [1224.846s] 9520 tests run: 9520 passed (1 slow), 162 skipped
```

Same commit, same job, and the test that timed out passed with the rest. Note the run took *longer* overall than the failed one (1224 s vs 1079 s) while passing 9520 rather than 9519 — the first attempt wasn't slow, it stopped.

**#3162 is green on all 11 checks** on `2576180`, and mergeable. That was the one re-run; I won't spend another.

---

crates/fresh-editor/src/view/viewport.rs
```diff
@@ -797,6 +941,18 @@ impl Viewport {
             return;
         }
 
+        // Addressed by byte: scrolling reads the rows it moves over and nothing
+        // else. The walk below counts from the logical line's start, which on a
+        // one-line file is every row above the viewport.
+        if crate::view::row_walk::addresses_rows_by_byte(buffer, self.line_wrap_enabled) {
```
sinelaw, line 860: **The anchored branch silently drops folds, soft breaks and virtual lines.**

This returns before `hidden_ranges`, `soft_breaks` or `virtual_lines` are looked at even once, and `row_walk::row_starts_from` passes `&[]` as `build_base_tokens`' `fold_skip`. So for every buffer `addresses_rows_by_byte` claims, scrolling counts rows the renderer does not draw. `scroll_down_visual`'s mirror at line 1045 has the same hole.

The gate is not "one enormous line" — `line_count() > MAX_WRAP_SCROLLBAR_LINES` is **5,000 lines**, so this is now the ordinary path for any wrapped file over 5,000 lines. `tests/e2e/frame_once_per_pane.rs` already builds exactly that shape ("6,000 lines: past MAX_WRAP_SCROLLBAR_LINES (5,000), under the byte ceiling").

Concrete failure: 6,000-line file, `editor.line_wrap = true`, a 500-line region collapsed two rows below the top. One wheel tick asks for 3 rows; `row_start_after` walks 3 *unfolded* rows and lands inside the collapsed body.
* Through `scroll_split_by_lines` the fold-snap in `window/mod.rs` then moves the top to `range.end_line + 1` — a 500-line jump for a 3-row tick.
* Through `scroll_all_splits_by` (scroll sync, `window/mod.rs:4623`) and `scroll_overlay_preview_by_lines` (`window/mod.rs:2202`) there is no snap at all, so `top_byte` is left *inside* a collapsed fold; the render's `fold_skip` jumps past it, so the drawn top no longer matches `top_byte` and the next tick compounds.

The path this replaces handled all three (`visible_row_start` / `skip_hidden_forward` on the way down, `soft_breaks`/`virtual_lines` through `count_visual_rows_for_line`). The replacement takes the same four arguments and uses none of them, so the compiler cannot flag it.

Compose mode is affected the same way: `addresses_rows_by_byte` ignores `view_mode`, and `row_walk` applies no soft breaks, conceals or inline hints, so a >5,000-line markdown_compose buffer scrolls by rows that the frame never draws.

---

crates/fresh-editor/src/view/row_walk.rs
```diff
@@ -0,0 +1,400 @@
...
+
+/// Whether this buffer's visual rows are addressed by byte — true exactly when
+/// no wrap index can cover it, so `ViewAnchor::byte` is the first visible row's
+/// own start rather than a line start.
+///
+/// Must agree with the indexability test in `reconcile::place_pane`, or a
+/// viewport would be placed in one coordinate and rendered in the other. Keyed
+/// on the buffer, not on a frame, so the coordinate cannot change between a
+/// scroll and the render of it.
+pub fn addresses_rows_by_byte(buffer: &Buffer, line_wrap_enabled: bool) -> bool {
```
sinelaw, line 327: **"Must agree with the indexability test in `reconcile::place_pane`" — it does not agree today.**

`place_pane` is:

```rust
let indexable = !state.buffer.is_large_file();
let within_bounds = indexable
    && buffer_len <= MAX_WRAP_SCROLLBAR_BYTES
    && state.buffer.line_count().is_some_and(|lc| lc <= MAX_WRAP_SCROLLBAR_LINES);
let has_index = if within_bounds || (indexable && state.wrap_indices.get(&geometry).is_some()) { … }
```

`!within_bounds` does match this function's disjunction. The second arm — `wrap_indices.get(&geometry).is_some()` — has **no counterpart here**, and it is sticky: `WrapIndices::entry` (wrap_index.rs:2822) inserts on miss and only evicts by LRU, so once an index exists for a geometry it survives the buffer crossing a ceiling, and `place_pane` re-`ensure_built`s it every frame thereafter.

Reachable with one paste. Open a 4,990-line file with wrap on — inside the ceilings, index built. Paste 20 lines. Now `line_count() == 5010`:

* `place_pane`: `within_bounds` false, cached index present → `has_index = true` → `ensure_visible_in_rows` + `row_pass_owns_placement = true`. That function reads the top as
  `index.line_first_row(buffer.get_line_number(self.top_byte())) as usize + self.top_view_line_offset()` (viewport.rs:1477) — a **line start plus a row offset**.
* `addresses_rows_by_byte`: `true`, so `scroll_up_visual` / `scroll_down_visual` write `top_byte` as a **mid-line row start with `top_view_line_offset = 0`**. Those are reached from `scroll_all_splits_by` and `scroll_overlay_preview_by_lines`, neither of which consults the index.

Written in one coordinate, read in the other. `line_first_row` collapses the mid-line top back to the first row of its logical line, so the next frame snaps the view to the top of that line — and on a file whose 5,001st line is long, that is a visible jump every time the splits scroll-sync.

Two predicates asserted equal in a comment is the failure mode this whole PR is arguing against. Export one function and have `place_pane` call it (including the cached-index term), rather than restating it.

Secondary: this is keyed on `line_wrap_enabled` and `place_pane`'s is not, so the two also disagree whenever wrap is off. That happens to be harmless today, which is precisely why the divergence above went unnoticed.

---

crates/fresh-editor/src/view/row_walk.rs
```diff
@@ -0,0 +1,400 @@
...
+        },
+        on_continuation: true,
+        ansi_in_escape: false,
+        chars_in_row: 0,
+    }
+}
+
+/// Width of the hanging indent a continuation row resumes at. Read bounded:
+/// whitespace past this is content, not indentation.
+fn leading_indent_width(buffer: &mut Buffer, line_start: usize) -> usize {
```
sinelaw: **This is a second implementation of `WrapMachine::measure_indent`, and it disagrees with it.**

`measure_indent` (wrap_machine.rs:368-392) differs in two ways:

1. **The clamp is missing.** It ends with `if self.line_indent + MIN_CONTINUATION_CONTENT_WIDTH > eff { self.line_indent = 0; }` (wrap_machine.rs:390) — the clamp the machine's own comment calls "the clamp that also guarantees the split loop makes progress". `leading_indent_width` has no `eff` at all, so it cannot apply it.
2. **Tabs.** Flat 4 columns here; the machine uses `visual_layout::tab_expansion_width(col) = TAB_WIDTH - col % TAB_WIDTH`. `"  \t"` measures 6 here and 4 there.

This is not cosmetic: `line_indent` is a *row-width* input. `feed_word_text` computes `fresh_capacity = eff - line_indent`, and `emit_break` re-emits an indent token of that width and sets `col = line_indent`. The carry decides where rows break.

Failure inside a single walk. `row_starts_up_to` chunks at `CHUNK_ROWS = 64`:
* chunk 1 starts at `from`, which `walk_start_before` prefers to make a line start → `carry_at` returns `RowCarry::default()` → `WrapMachine::resume` sets `measuring_indent = hanging_indent && !on_continuation = true` → the machine measures the indent itself, **with** the clamp;
* chunk 2 starts at `chunk1.last()`, mid-line → `carry_at` → `line_indent` **unclamped**.

In a 40-column split (`eff ≈ 33`) a line indented 24 columns gives chunk 1 rows of 33 columns and chunk 2 rows of 9. Every row start `row_start_before` returns past row 64 of that line is a row start of no grid the renderer draws, so the "sub-row grid shift" the module documents as invisible is actually a factor-of-three change in row width.

The same carry feeds `render_buffer.rs`'s `BuildAnchor`, so it is also directly visible: scroll *into* the middle of such a line and its continuation rows are drawn with a 24-column hanging indent; park the top at the line's start and they are drawn with none. The line reflows as you scroll through it.

`wrap_indent` defaults to `true`, so `hanging: true` is the normal case; only the split has to be narrow (or the indent deep) for `indent + 10 > eff`.

Fix: get the carry from the machine rather than re-deriving it — run `WrapMachine` from the line start across the indent and take `carry()` — or, at minimum, thread `rule.available_width()` in and apply the same clamp and `tab_expansion_width` here.

---

crates/fresh-editor/src/view/viewport.rs
```diff
@@ -885,6 +1041,16 @@ impl Viewport {
             return;
         }
 
+        // Mirror of `scroll_up_visual`'s anchored branch.
+        if crate::view::row_walk::addresses_rows_by_byte(buffer, self.line_wrap_enabled) {
+            let rule = self.wrap_rule(buffer);
+            let top = self.top_byte();
+            let new_top = crate::view::row_walk::row_start_after(buffer, top, rule, visual_rows);
```
sinelaw: **The end-of-buffer clamp is gone on this branch.**

Every exit of the path below ends in `set_top_byte_with_limit(...)` or `apply_visual_scroll_limit(...)`; this one calls `set_top_byte` raw. `row_start_after` → `row_starts_from` stops only when `build_base_tokens` stops producing rows, i.e. at the buffer's last row start, so nothing bounds the top at "last full screen".

Failure: 6,000-line wrapped file, 40-row viewport, hold the wheel down at the end of the file. Today the top clamps so the final row rests at the bottom; with this branch the top keeps advancing until the buffer's last row is the *first* drawn row and 39 rows are `~`. Scrolling back up then has to walk the whole way.

`ensure_visible_anchored` (line 2318) has the same gap on its far-below branch: `row_start_before(cursor_row_start, last_row - margin, rule)` is written straight to `set_top_byte` with no end clamp, so putting the cursor near EOF leaves a partly empty screen where the non-anchored path would have filled it.

---

crates/fresh-editor/src/view/viewport.rs
```diff
@@ -576,47 +688,79 @@ impl Viewport {
...
-        // behavior for those lines.
-        let touches_soft_break = soft_breaks
-            .iter()
-            .any(|(p, _)| *p >= line_start && *p < line_end);
-        let touches_virtual = virtual_lines
-            .iter()
-            .any(|p| *p >= line_start && *p <= line_end);
-        if touches_soft_break || touches_virtual {
-            return self.top_byte();
+            let seg_bytes = Self::drawn_row_source_bytes(text, piece_start, &wrap_config);
```
sinelaw: **"The pieces are contiguous, so walk them" — the bytes are contiguous, the wrap is not.**

`drawn_row_source_bytes` → `wrap_segment_source_bytes` → `WrapMachine::run(tokens, rule)`, i.e. `run_from(.., RowCarry::default())`. So each `MAX_LINE_BYTES` piece is wrapped as if it began a fresh row at column 0, and — because `resume` sets `measuring_indent = hanging_indent && !on_continuation`, and `wrap_indent` defaults true — it also re-measures a hanging indent from whatever mid-line content the piece happens to start with. The renderer wraps the logical line as one continuous run.

Failure: the 441 KB logical line from the #2843 note two lines above, at 73 columns. The first 100 KB piece ends 63 columns into row 1,370; the renderer fills that row with the next piece's first ten characters, this loop starts a fresh row instead. `rows_before` gains one row per piece (5 pieces → ~4 spurious rows) and the byte returned for a given `offset` slips by up to a full row width each time.

That drift is *against a counter that does not have it*: `offset` is `top_view_line_offset`, produced by `count_visual_rows_for_line`, which wraps the whole joined line. So the two disagree by exactly the piece-boundary error, and PageUp/PageDown land the cursor on a row that is not the one at the top of the viewport — the symptom this rewrite exists to remove, just at 1/1000 the magnitude of the clamp-to-first-piece bug it replaces.

`row_walk` already solves this correctly by threading `RowCarry` across reads; this loop should carry state between pieces the same way (or call into `row_walk::row_starts_from`) rather than restarting the machine per piece.

---

crates/fresh-editor/src/view/row_walk.rs
```diff
@@ -0,0 +1,400 @@
...
+use crate::model::buffer::Buffer;
+use crate::view::ui::split_rendering::base_tokens::build_base_tokens;
+use crate::view::wrap_machine::{RowCarry, WrapMachine, WrapRule};
+
+/// Bytes read per row asked for, before slack.
+///
+/// A row holds at most `available_width` characters and a character is at most
+/// four bytes, so this cannot cut a row short. Over-reading a little is far
+/// cheaper than a second pass.
+const BYTES_PER_ROW_ESTIMATE: usize = 4;
```
sinelaw: **"A row holds at most `available_width` characters … so this cannot cut a row short" is false, and a short read is indistinguishable from EOF.**

A row holds at most `available_width` **columns**. Characters can be zero-width — combining marks, variation selectors, ZWJ — so the bytes a row consumes have no upper bound expressible in `available_width`. Four bytes per row-column is a guess, not the invariant the comment claims.

When the guess is short, `build_base_tokens` stops on `chars_seen >= budget` and `row_starts_from` returns fewer starts than asked for. Nothing distinguishes that from "the buffer ended", and every caller reads it as the latter:

* `row_start_after` returns a byte fewer than `rows` rows down, so a wheel tick / PageDown under-moves. Unlike `row_start_before` it has **no retry loop**, so it never recovers — the same scroll under-moves every time.
* `rows_between`'s `Some(row) if starts.len() <= max_rows` arm reports a bounded answer for a walk that was budget-truncated rather than buffer-truncated, which is exactly the case its sibling arm is written to reject.
* `row_start_before`'s `starts.len().checked_sub(rows_back + 1)` treats the shortfall as "we reached the top" once the back-off is exhausted.

Concretely: a line of ZWJ family emoji (`👨‍👩‍👧‍👦` = 25 bytes / 2 columns = 12.5 bytes per column), or Devanagari/Thai with two combining marks per base (≈5 bytes/column). `row_start_after(top, rule, 40)` at width 73 budgets `41 * 73 * 4 + 1024 = 12,996` bytes to cover 2,993 columns; at 12.5 bytes/column it covers ~1,040 columns, so a 40-row PageDown moves 14 rows.

`row_starts_from` should report whether it stopped on the budget or on the buffer, and `row_start_after` should retry on the former the way `row_start_before` does.

---

crates/fresh-editor/src/view/row_walk.rs
```diff
@@ -0,0 +1,400 @@
...
+                resumed, expected,
+                "a walk from row {start_row} (byte {}) diverged from the run",
+                whole[start_row]
+            );
+        }
+    }
+
+    /// Walking costs the rows asked for, not the distance from the line start.
+    #[test]
+    fn a_walk_reads_the_rows_asked_for_wherever_it_starts() {
```
sinelaw: **This test does not test its own thesis.**

The name and doc claim a cost property — "Walking costs the rows asked for, not the distance from the line start" — but nothing here measures cost. The assertions are `near_top.len() == 8` and `far_down[0] == deep`; both hold for an implementation that re-wraps all 60,000 values from byte 0 on every call, which is the bug the module exists to fix. It would pass unchanged on the code before this PR if that code merely returned the right answers slowly.

The rest of the PR gets this right and shows how: `capped_row_count_saturates_on_a_huge_line` asserts `iter.current_position() < 100_000`, and the e2e tests assert on `Buffer::resident_bytes()`, which this PR adds precisely so work is observable. Either assert on `resident_bytes()` before/after the deep walk, or on the iterator's read position, or rename the test to what it actually checks (`a_walk_starts_on_the_row_it_was_given`).

---

crates/fresh-editor/src/view/viewport.rs
```diff
@@ -576,47 +688,79 @@ impl Viewport {
...
-            )
-        };
-        let idx = self
-            .top_view_line_offset()
-            .min(seg_bytes.len().saturating_sub(1));
-        seg_bytes.get(idx).copied().unwrap_or(self.top_byte())
+    /// Source byte of each row `text` is drawn as, given it starts at
+    /// `text_start`.
+    ///
+    /// Chopped at `MAX_SAFE_LINE_WIDTH` before wrapping, because that is what
```
sinelaw: **Three comments describe a `MAX_SAFE_LINE_WIDTH` chop that none of this code performs.**

* Here: "Chopped at `MAX_SAFE_LINE_WIDTH` before wrapping" — the body goes straight to `grid_segment_source_bytes` / `wrap_segment_source_bytes` at the wrap width. Nothing is chopped.
* `compute_line_layout` (line 2255): "`WrapRule::Chop` owns that bound now" — its body is `Self::wrap_one_run(line_text, wrap_config)`, a bare forwarder, and `Chop` is never constructed anywhere in this file.
* `wrap_one_run` (line 2269): "each `MAX_SAFE_LINE_WIDTH` piece of a line that is not [short enough]" — nothing in this file ever splits a line into such pieces, so the second half of that sentence describes a caller that does not exist.

The chop that does exist is `view_data::effective_wrap_width`, which returns `MAX_SAFE_LINE_WIDTH` as the **`Word` rule's `content_width`** — word-boundary-preferring and reduced by the gutter — not `WrapRule::Chop`. So the comments are wrong about the mechanism as well as about the location.

Net effect of the split: two functions where there was one, with the outer one adding nothing, plus three comments a future reader will trust. Collapse `compute_line_layout`/`wrap_one_run` back together and drop the chop language, or make the chop real here.

(To be clear about what I checked: I could not find a live path that lost its width bound when the forced `Break` was removed from `build_base_tokens`. `build_view_data` still wraps at `effective_wrap_width`, the wrap index still uses `Chop`, `build_base_tokens_for_hook` has no non-test callers, and `line_wrap_cache::compute_line_layout` — the one path that materialises `ViewLine`s with no wrap transform when `line_wrap_enabled` is false — has no callers at all. The premise holds; only the comments don't.)

---

crates/fresh-editor/src/view/viewport.rs
```diff
@@ -1933,6 +2110,15 @@ impl Viewport {
 
         self.load_data_around_cursor(buffer, cursor.position, viewport_lines);
 
+        // No index covers this buffer, so placement is a bounded walk from the
+        // anchor. Horizontal handling below still applies.
```
sinelaw: "Horizontal handling below still applies" — it doesn't. The call is followed by `return`, so the tail of `ensure_visible_with_virtual` never runs. For a wrapped buffer that tail is `self.left_column = 0;` (line ~2205), which is now skipped for exactly the buffers this branch claims: toggle wrap on with a non-zero `left_column` on a >5,000-line buffer and nothing clears it.

Also note `ensure_visible_anchored` unconditionally does `set_top_view_line_offset(0)` before it decides anything (line 2327). A viewport that arrives here with a non-zero offset — e.g. one that `clamp_top_byte_wrapped` or the new `scroll_to_cursor_wrapped` branch parked mid-line while the buffer was still under the ceilings, before an edit pushed it over — silently loses that many rows of scroll position without `top_byte` moving. Either fall through to the horizontal block instead of returning, or say that the branch owns the whole placement including the reset.

---

crates/fresh-editor/src/view/line_wrap_cache.rs
```diff
@@ -403,6 +403,42 @@ fn layout_for_plain_text_under(
...
+/// visual column it sits at.
+///
+/// The byte-oriented counterpart of [`char_position_in_layout`], which walks
+/// *drawn* characters: a wrap breaking on a space consumes it, so that count
+/// falls one behind the source per wrapped row. Callers pass a byte offset,
+/// which is a character index only for single-byte text that never wraps on
+/// whitespace — deep in one enormous line the drift is the difference between
+/// the viewport agreeing with what was drawn and not (issue #1806). Rows carry
+/// real byte offsets, so ask them.
+pub fn byte_position_in_layout(layout: &[ViewLine], byte_in_line: usize) -> (usize, usize) {
```
sinelaw, line 401: The new doc block was inserted *between* `char_position_in_layout`'s existing doc comment and its signature, so the two paragraphs ending "If `layout` is empty, returns `(0, 0)`. If the position is past the end of the last row, returns the last row with the last visual column of that row." now document `byte_position_in_layout`, and `char_position_in_layout` — still live, still called from `wrap_index.rs` and `transforms.rs` — is left with none.

Worth checking the inherited sentence is even true of the new function: on a past-the-end offset `char_position_in_layout` returns `last.visual_width().saturating_sub(1)`, while this returns `row.visual_col_at_char(last_source_char)`. Those differ whenever the last row ends in an injected (source-`None`) token or a wide glyph, which is the case the callers you switched over hit at EOL.

---

sinelaw (commented):
Skeptical review focused on the four stated premises plus a correctness pass. Ten inline comments, most severe first.

**Premise-by-premise**

1. *"Buffers with no wrap index should be addressed by byte."* The second coordinate is real and the two placements do **not** agree today: `row_walk::addresses_rows_by_byte` omits `place_pane`'s cached-index arm (`wrap_indices.get(&geometry).is_some()`), which is sticky across a buffer crossing a ceiling. Reachable by pasting 20 lines into a 4,990-line file. See the comment on `addresses_rows_by_byte`.
2. *"Removing the forced `Break` is safe because `Chop` bounds row width."* **This one holds.** I traced every consumer of `build_base_tokens`: `build_view_data` still wraps at `effective_wrap_width` (= `MAX_SAFE_LINE_WIDTH` with wrap off), the wrap index and the scrollbar go through `WrapRule::Chop`, `build_base_tokens_for_hook` has no non-test callers, and `line_wrap_cache::compute_line_layout` — the only path that materialises `ViewLine`s with no wrap transform — has no callers at all. No live path lost its bound. What did not survive is the *documentation* of the change; see the comment on `drawn_row_source_bytes`.
3. *"The backward walk is sound rather than exact."* The admitted shift is not the only error: `carry_at`/`leading_indent_width` re-implement `WrapMachine::measure_indent` without its `MIN_CONTINUATION_CONTENT_WIDTH` clamp or its tab-stop arithmetic, so a single `row_starts_up_to` call can change row width by 24 columns at its 64-row chunk boundary. Termination itself is fine — `row_starts_up_to` advances or returns, and the back-off is bounded — but `row_start_after` has no retry where `row_start_before` does, and a budget-truncated walk is reported as a buffer-truncated one.
4. *"The saturating count is safe because word wrap is greedy."* Largely holds. The hanging-indent counterexample I tried to build does not fire, because `row_budget_bytes` floors at 4 KB and a 4 KB whitespace prefix is clamped to indent 0 either way; the U+FFFD widening is covered by the one row of slack in `rows > cap`. The "exactly one such cut" claim is weaker than stated — `next_logical_line_budgeted`'s join loop can pull more than one piece — but the extra columns stay inside the slack.
5. *Lazy binary open.* No downstream `is_binary()` consumer assumes a loaded buffer; binary buffers are forced read-only in `finalize_file_buffer`, save is gated on `!is_binary()`, and `build_base_tokens_binary` reads a bounded range. `BufferFormat::new(line_ending, encoding)` is equivalent to the removed `set_default_encoding`. No finding.

**Note on verification:** the checkout I was given is mid-rebase (`.git/rebase-merge` present, `CHANGELOG.md` conflicted, `HEAD` on master), so I reviewed against `git archive 6085f83` rather than the working tree and did not run `cargo test` — building that tree would have produced a mixture of the two revisions. Everything above is from reading the PR revision's sources.

---

crates/fresh-editor/src/view/viewport.rs
sinelaw: Leaving this one open — it is narrowed, not closed, and the other nine are resolved.

The gate is now `line_wrap && buffer.is_large_file()` (0db44c4), so the 5,000-line term you identified is gone and with it every case you named: the 6,000-line file, `frame_once_per_pane.rs`'s shape, the `markdown_compose` buffer, and the scroll-sync paths through `scroll_all_splits_by` / `scroll_overlay_preview_by_lines` on ordinary files. Those all keep the row-numbered path, which handles folds as it always did.

What remains is your finding on a genuinely large file: the anchored branch still takes `hidden_ranges`, `soft_breaks` and `virtual_lines` and uses none of them, and `row_starts_from` still passes `&[]` as `fold_skip`. Folds are not gated on file size — `folds.resolved_ranges()` in `reconcile.rs` will happily return ranges for a lazily-loaded buffer — so a folded region in a >10 MB file scrolls by rows the renderer does not draw. Narrower than what you found, and the same bug.

I did not fix it here because the obvious fix is the mistake this PR just made once. Excluding folded buffers from the anchored path means a second predicate — one that the render path at `render_buffer.rs:277` would also have to agree with, and that path does not have `hidden_ranges` in scope. Adding a predicate that must be kept in sync with `place_pane`'s is exactly what your other comment correctly objected to, and I would rather not do it twice in one PR on a case that needs its own reproducer first.

So: worth a follow-up issue rather than a rushed commit, unless you would rather I gate on it now and thread the fold state through the render path properly.

---

Issue #1971, opened 2026-05-14
Orchestrator: Open / New — multiple rendering and focus bugs in the picker dialog


sinelaw:
## Summary

Interactive testing of `Orchestrator: Open` and `Orchestrator: New Session` (built from `d217fdc` on `master`, run inside a 220×60 tmux pane) surfaced several rendering and focus bugs in the picker dialog and the new-session form. The preview *rectangle* is stable across ↑/↓ navigation, but the surrounding chrome and several content paths are not.

## Repro

1. Build `fresh` from current master.
2. Open it on a small git repo with a few files.
3. Run `Orchestrator: New Session` twice to create two extra sessions (one normal, one with a very long name like `gamma-very-long-session-name-for-truncation-test`). Add a vertical split + a real `Open Terminal` (with some output) to the base session.
4. Run `Orchestrator: Open`. Navigate with ↑/↓. Try `Tab`, type into the filter, type a filter that matches nothing.

## Bugs in the picker / preview pane

1. **Sessions list pane is dramatically shorter than the preview pane.** The two `labeledSection`s inside the right-hand `row()` should match in height (see the wireframe in `docs/internal/orchestrator-open-dialog-and-lifecycle.md`), but the left pane's bottom border closes immediately after the last item (~4 rows for 3 sessions) while the right preview pane continues for ~20 rows. The `Sessions` list widget seems to be sized to its item count rather than the height the preview reserved via `embedRows`.

2. **No selection indicator in the session list.** ↑/↓ correctly swaps the preview, but the list itself shows zero cue for which row is highlighted — no `▸`, no inverse video, nothing. With three rows and no marker, the only way to tell the selection is to read the preview pane title (`╭─ [N] name`). `setSelectedIndex("sessions", …)` is being called, so the `list` widget's selection styling isn't reaching the painted output (possibly the "ACT"/"RUN" styling on the row is overriding it).

3. **Tab / button focus has no visible indicator.** Pressing `Tab` should move focus to `Stop` / `Archive` / `Delete`. The screen is byte-for-byte identical before and after Tab. Same problem on the new-session form: `Cancel` and `Create Session` never visibly change state regardless of Tab/S-Tab/Right/Left, so you cannot tell whether `Enter` will commit or cancel without trial and error. (I lost a session that way.)

[...]

PR #3163, opened 2026-09-02
fix(orchestrator): the folder popup keeps the keyboard, and three on-disk names that were not names

### LINKED ISSUE #3137: Orchestrator "Move to Folder…" popup does not own the keyboard: Down/Up fall through to the dock, switch the live workspace, and dismiss the popup
**Version:** fresh 0.4.10 (built from `origin/master` @ `7bcc8ff6c`, the retained-mode series), Linux, tmux 220×50.

The **Orchestrator: Move to Folder…** chooser popup (`● Top level` / `New Folder…` / existing folders, footer `↑↓ choose  Enter select  Esc cancel`) ignores its own keys: pressing `↓` acts on the **dock's workspace list behind it** — the dock selection moves to the next workspace, which **live-switches the visible workspace** — and the popup is dismissed without any choice being made. Keyboard-only folder filing is impossible.

## Steps to reproduce
1. Create two orchestrator workspaces of one project (e.g. **Orchestrator: New Workspace** twice), so the dock lists at least `ws-A` and `ws-B`.
2. Visit `ws-A`. Command Palette (`Ctrl+P`) → **Orchestrator: Move to Folder…**.
3. The popup opens over the dock: `● Top level` / `New Folder…`, footer `↑↓ choose  Enter select  Esc cancel`.
4. Press `↓` (intending to highlight `New Folder…`).

## Expected behavior
Per the popup's own footer (`↑↓ choose`), the highlight moves to `New Folder…`; `Enter` then opens the New Folder dialog. This worked before: #2694 (fixed via PR #2696) explicitly lists the "Move to Folder…" dropdown among the dock dropdowns that "all support ↑/↓".

## Actual behavior
The popup closes, no folder action happens, and the **dock selection has moved one row down — switching the editor to `ws-B`**. The user asked to file a workspace into a folder and instead got teleported to a different workspace. Reproduced 3/3 (each attempt ended with the popup gone and the next workspace visited; ANSI capture confirms the `48;5;25` selection highlight lands on the dock row, and the popup's rows never highlight).

The severity is compounded by the workspace switch: with `↓`+`Enter` muscle memory the `Enter` then acts in the *newly switched-to* workspace.

Adjacent surfaces behave correctly (verified in the same session):
- The folder header's **F2 menu** (`Rename… / New Subfolder… / Delete Folder`) takes `↑`/`↓`/`Enter` fine with the dock focused.
- The **New Folder** name dialog takes typed text, `Ctrl+U`, and its buttons normally.
- **Mouse click** on the popup's `New Folder…` row works — the only way through.

## Workaround
Use the mouse: click `New Folder…` (or a folder row) in the popup.

---
*Search queries used (no open duplicate): `Move to Folder popup arrow keys switch workspace instead of choosing`, `orchestrator dock folder popup keyboard focus dismissed down arrow`, `dock dropdown keyboard navigation regression`. Related: #2694 (closed — established that this dropdown supported ↑/↓; this is a regression of that surface), #2189 (dock focus/input-routing papercuts, different symptoms).*

*Filed by tui-agent-auto-bug (automated TUI QA agent, Run #61, 2026-09-02).*

sinelaw:
Fixes two orchestrator faults, both reproduced from the issues' own repro comments before anything was written, and both verified afterwards by driving the editor interactively in tmux with real keystrokes (never `--cmd script`, which would bypass the input routing under test).

Closes #3137. Addresses the two remaining items of #1971.

---

# How to validate

> ## ⚠️ Which build shows the "before" behaviour
>
> **#3137 is a regression that is in no release.** The first bad commit is `8aba3b95a`, which landed *after* `v0.4.10` — the latest tag — so on **every released version the arrow keys already work correctly in the Move-to-Folder popup**. Testing an older build and finding it fine is the expected result, not a failed repro.
>
> To see it fail you must build the PR's base:
>
> ```bash
> git checkout cf626fc06 && cargo build --bin fresh     # "before": broken
> git checkout claude/fix-orchestrator && cargo build --bin fresh   # "after": fixed
> ```
>
> The bisect's own boundary, if you want to confirm it directly: `867a0eb3b` is the last good commit, `8aba3b95a` the first bad one.
>
> **The three #1971 faults below are not regressions** — they reproduce on master `cf626fc` and I did not bisect how far back they go, so an older build may well show them too. Only #3137 is release-clean.

Four separate faults, each with its own repro. All were measured on master `cf626fc` before the fix and again after, in a 200×50 tmux pane with a debug build.

> **Reading the dock's selection.** The highlight is background `48;5;25`, so
> `tmux capture-pane -p -e | grep -n '48;5;25'` prints the selected row number,
> and the tab title on line 2 names the live workspace.

---

### 1. The "Move to Folder…" popup does not own the keyboard (#3137)

**Requires a build at or after `8aba3b95a` — see the box above. No release has this bug.**

```bash
mkdir /tmp/rt3 && cd /tmp/rt3 && git init && touch a.txt && git add -A && git commit -m init
fresh --no-restore
```

1. `Ctrl+P` → **Orchestrator: New Workspace** → Tab to `[ Create Workspace ]` → Enter. **Twice**, giving dock rows `rt3-1` and `rt3-2`.
2. `Alt+O` to focus the dock, then `↑` so the highlight sits on **`rt3-1`** — the middle of the list. *(This matters: if the live workspace is the last row, `↓` has nowhere to go and you see the dismissal without the workspace switch.)*
3. `Ctrl+P` → **Orchestrator: Move to Folder…**. The popup opens (`● Top level` / `New Folder…`, footer `↑↓ choose  Enter select  Esc cancel`) with the highlight on its first row.
4. Press `↓`.

| | before `↓` | **before fix** (`cf626fc`) | **after fix** |
|---|---|---|---|
| popup | open | **gone** | open |
| highlight row | 5 (in popup) | **7 (dock)** | 6 (in popup) |
| live workspace | `rt3-1` | **`rt3-2`** | `rt3-1` |

After the fix all four keys drive the popup and nothing leaks to the dock: `↑`/`↓` move within it, `Enter` opens the New Folder dialog, `Esc` closes it and returns the dock highlight to `rt3-1`.

*Either route reaches the same popup.* The dock's row context menu (`F2` / right-click) is its own anchored panel, but its **Move to Folder…** option closes that panel and opens this same inline dropdown — `ctx-move` calls the same `openDockMenu(...)` the palette command does, deliberately, *"so it shares the keyboard-navigable menu path and the folder list"*. So the bug reproduces through `F2 → Move to Folder…` as well; the palette is just the shorter route to it.

---

### 2. Terminal storage outgrows NAME_MAX (#1971)

Needs a repo root deep enough that the *doubly* encoded path passes 255 bytes — the orchestrator encodes the repo root into the workspace path, then encodes that path again to name the terminal directory.

```bash
R=/tmp/d-one/d-two/d-three/d-four/d-five/d-six/d-seven/work/repo   # ~160 chars
mkdir -p $R && cd $R && git init && touch a.txt && git add -A && git commit -m init
fresh --no-restore --log-file /tmp/f.log
```

1. `Ctrl+P` → **Orchestrator: New Workspace**, set **Workspace Name** to `gamma-very-long-session-name-for-truncation-test`, create it.
2. `grep -c 'File name too long' /tmp/f.log` and `ls ~/.local/share/fresh/terminals/`.

**Before:**
```
WARN fresh::app::terminal: Failed to create terminal directory: File name too long (os error 36)
WARN fresh::app::terminal: Failed to create terminal backing file: File name too long (os error 36)
```
`~/.local/share/fresh/terminals/` is empty — the workspace has nowhere to write its scrollback.

**After:** no warning, and the directory exists at exactly 255 bytes with its backing files inside:
```
…deep-directory-seven%5Fwork%5Frepo_gamma-very-long-ses~24aec60fedffac2d/
    fresh-terminal-eph-0-1788465026018338657.log
    fresh-terminal-eph-0-1788465026018338657.txt
```

**Also check nothing was re-keyed.** A workspace whose encoding is *under* 255 must keep the exact name it had before this PR — that is what the cap being `NAME_MAX` buys. A shallower repo (say `/tmp/rt3`) should produce a byte-identical terminal directory name before and after.

---

### 3. The archived directory is named after a display string (#1971)

```bash
mkdir /tmp/rt && cd /tmp/rt && git init && touch a.txt && git add -A && git commit -m init
fresh --no-restore
```

1. Create a workspace named `good` and let its terminal settle (the tab title becomes `bash — root@vm: ~/…`).
2. `Ctrl+P` → **Orchestrator: Open** → select `good` → Tab to `[ Archive ]` → Enter → `[ Confirm Archive ]` → Enter.
3. `ls ~/.local/share/fresh/orchestrator/tmp_rt/.archived/`

**Before** — the row's *display label* names the directory, spaces, middle dot, em dash and all, and the `/`s inside it are read by `git worktree move` as further directory levels:
```
.archived/gamma-very-long-session-name-for-truncation-test · bash — root@vm: ~/
```

**After:**
```
.archived/good
```

The manifest still records the display label, and `unarchiveWorkspace` resolves from the recorded `root`, so unarchiving is unaffected. Archiving a second workspace of the same name lands at `.archived/good-2` instead of failing on an occupied directory.

---

### 4. Archive freezes the picker (#1971)

The frozen card needs a workspace that has **no window yet**, and those only come from a **remote** create — a local one gets a real positive window id from `createPreparingWindow`, so it will not reproduce this. An SSH workspace pointed at a host that does not exist is the easiest way in.

```bash
fresh --no-restore --log-file /tmp/f.log
```

1. Create one ordinary local workspace (`good`).
2. `Ctrl+P` → **Orchestrator: New Workspace** → `Shift+Tab` to the **Run in** row → `→` to select **SSH** → Tab to the host field → type `nonexistent-host-xyz` → create. It fails to connect and leaves a `!` placeholder row in the dock.
3. `Ctrl+P` → **Orchestrator: Open**, then `↓`/`↑` onto that `ssh:nonexistent-host-xyz` row.

**Before** — the log fills with the error from the issue and the picker stops responding; further `↓` does nothing and the selection stays put:
```
ERROR updateFloatingWidget: invalid spec: invalid value: integer `-1000000`, expected u32
```
Reaching this through **Archive → Confirm Archive** is what the issue reported: the archive completes on disk (worktree moved to `.archived/`, `archived.json` written) while the picker stays on "Archiving… / Waiting for git…" and only recovers on the next keypress.

**After:** `grep -c 'invalid spec' /tmp/f.log` is `0`, the placeholder row renders as blank embed rows, the picker keeps taking arrow keys, and Confirm Archive completes without a stuck card.

---

# What was wrong, and why

### Bisected

`git bisect` over `v0.4.10..cf626fc`, run against the interactive repro, lands on:

> **`8aba3b95a` — "Fix the CI failures and hangs, and drop the web's plugin-panel view"** (parent `867a0eb3b` verified good)

Two of the bisect's automated steps had to be re-checked by hand: at `f886c915f` the scenario could not be set up at all (the palette never opened, so the script's "no popup" reading was not the symptom), and `867a0eb3b` — the first bad commit's parent — measures cleanly good. That commit is where `focus_panel_widget_in_tree` was introduced, which is precisely the mechanism below.

Since `8aba3b95a` is contained in no tag, the regression has never shipped: it exists only on master.

### Root cause

Opening an in-panel dropdown is **two plugin writes** — a spec carrying the option rows, then `setFocusKey` onto the first of them — and both are applied before the next frame describes either.

`focus_panel_widget_in_tree` is the imperative half of the focus mirror: it moves the tree's focus to the widget the registry has just been told holds the panel's. It could only move focus to an element that *exists*, so it asked the tree for a `menu-pick:` row the previous spec had no row for, found nothing, and returned. A probe confirmed it on the running editor:

```
focus_panel_widget_in_tree: PROBE widget=menu-pick:move:root holds=true found=false
```

Nothing re-tried it, so the tree kept the focus the dropdown was meant to take — and the mirror's *other* direction wrote that disagreement back. The stale holder's `FocusGained` fact re-pointed the registry at the session list, and the router (which reads the registry to decide whether a dropdown owns the arrow keys) sent `↓` to the list:

```
set_panel_focus_and_notify: firing `focus` widget_event  old=menu-pick:move:root new=sessions
dispatch_floating_widget_key: decision code=Down focus_key=Some("sessions") outcome=SmartKey("Down")
```

Because `ctx-move` funnels into the same `openDockMenu`, the dock's three inline dropdowns — "New Task… ▾", the project scope picker and Move-to-Folder — all share this path, and all of them depend on that focus landing.

### Fix

Hold the move, and replay it immediately after the frame that builds the element. The gain it raises is queued *behind* the stale one, so `apply_settled_shell_messages` settles on the dropdown — which is what that path's existing ordering rule already says should happen. One replay, never a standing request: the frame that follows the write is built from the very spec the write was aimed at, so a widget still absent there is not coming. A request the registry no longer names has been overtaken by a newer focus decision (a Tab, a click, another `setFocusKey`) and is dropped without acting.

**Regression test:** `a_focus_onto_a_widget_the_next_frame_builds_lands_on_that_frame` (`app/widget_runtime.rs`), at the same seam as the existing `a_focus_the_host_decides_moves_the_trees_focus_too`. Confirmed to fail without the fix.

---

## #1971 — the two remaining items

Both still reproduce at HEAD; the rest of the issue's enumerated bugs are fixed, as the last status sweep found.

### 1. Terminal storage outgrew NAME_MAX

`encode_path_for_filename` encodes a path whole, so the name grows with the directory's depth, and the orchestrator nests one encoded path inside another: a workspace lives under `orchestrator/<slug of the repo root>/<name>`, and naming *its* terminal directory encodes that path a second time.

An encoding that no longer fits is now folded to `<head>~<hash>`. The head keeps the tail of the path readable in a listing; the FNV-1a digest — spelled out rather than reached for, because it names files that outlive the process — keeps two paths sharing a 238-character prefix apart. It runs over the *encoding*, which is the encoder's normalised form, so a folded name still maps doubled separators, a leading one, and either separator onto one filename the way an unfolded one does. A literal `~` can mark the fold because the encoder can never emit one (a real tilde becomes `%7E`), which is what lets `decode_filename_to_path` decline a folded name instead of handing back a path that was never encoded.

**The cap is `NAME_MAX` itself, and that is the point.** Folding is a rename, so any name that folds stops finding what the previous release wrote under it — a workspace snapshot, the recovery scope, terminal scrollback, per-workspace config, a running daemon's socket. Only names no caller could have created at all may move, which means the filesystem's limit and not a byte less. Leaving headroom for a suffix would be wrong in both directions: several callers append nothing and join the encoding as a whole directory name (so a 250-byte encoding is live today), and the longest suffix is not small either — `workspace_path_for` builds `<encoded>.<stable_id>.json`, about 28 bytes. A name that overflows *with* a suffix but fits without one stays a caller-specific limit, exactly as it was before this cap existed.

### 2. The archived directory was named after a display string

A workspace's `label` is a *display* string — `workspaceDisplayName` composes it from the workspace name and the terminal's live tab title — so filing one put spaces, a middle dot, an em dash and the `/`s of a path into a directory name, and `git worktree move` read those slashes as further directory levels to nest the graveyard under.

The basename of the worktree is the name git already gave the directory, so it is a valid single component by construction and the archived copy keeps the name it had while it was live. The manifest still records the display label, and `unarchiveWorkspace` resolves the directory from the recorded `root`, so nothing downstream moves. A repeat archive of the same name takes the next free `-2`, `-3` rather than failing on a directory that is already there.

### 3. Archive hung the picker — the same `invalid spec` the issue logged

`windowEmbed`'s `windowId` is a `u32` on the host, and serde fails the **whole spec** on a negative one — so `updateFloatingWidget` dropped it and the panel kept painting the last spec that parsed. After Confirm Archive that was the "Archiving… / Waiting for git…" card, still up long after the worktree had moved to `.archived/` and the manifest had been written.

The negative ids are the orchestrator's placeholders for a workspace that has no window yet (a remote create; `nextPendingId` counts down from `-1_000_000`, which is where the issue's `-1000002` comes from). "No window" is a state the widget already has a value for, and it is `0` — its own doc comment says an unknown id renders placeholder blanks — so the helper normalises to it and the placeholder renders as the blanks it is meant to be.

**Regression test:** `a_placeholder_window_embed_does_not_freeze_its_panel` (`tests/e2e/`) — mounts a panel carrying a placeholder embed and marker V1, replaces it with V2, and requires V2 on screen. Checked both ways: with the fix it passes in half a second; with the normalisation removed the mount spec never crosses the boundary at all, the host logs `invalid spec`, V1 never renders, and the test hangs on its first wait — which is the freeze itself.

---

## Open questions (deliberately not decided here)

- **The `_` / `%5F` mixing in one slug component** (#1971's first complaint) is *deliberate and unambiguous*: `_` is a path separator and `%5F` a literal underscore, which is what keeps `a_b/c` and `a/b/c` apart. Changing the scheme would rename every existing on-disk path — per-workspace configs, workspace snapshots, recovery state, terminal directories — and needs a migration, so this PR fixes only the overflow, which is the actual failure. Worth its own issue if the readability is judged to be worth the migration.
- **What the picker should show for a workspace that has no window yet.** It currently offers Visit / Stop / Archive / Delete against a placeholder row. The freeze is fixed, but which of those a placeholder should offer is a product decision.
- **A spec that fails to deserialise silently freezes its panel.** The host logs `invalid spec` and drops the update; the plugin gets no error back and the user sees a stuck panel. The plugin-side fault is fixed, but the host's failure mode is worth hardening on its own.

---

## Review round

Six findings, all real, fixed in `766f171`. The load-bearing one: the cap started at 200, so every encoding in **201..=255** — created without complaint today — would have folded on first run and orphaned what it named. The cap is now `NAME_MAX`. The others: the digest ran over the raw path and so dropped the encoder's own normalisation; the archive fallback reached for `s.hostLabel`, another display string, undoing the fix it sat in; the `windowId` guard closed only the low end of `u32`; the `windows.json` migration would have *persisted* an unresolvable path once the decoder began declining folded names; and the #3137 test re-spelled `render`'s sequence in a helper, so deleting the replay from `render` left it passing — `render` and the test now call one `Editor::lay_out_shell_tree`, and deleting the replay from it fails the test (checked both ways).

One correction to a review example rather than a fix: a **trailing** separator was never normalised — it becomes a trailing `_` that nothing strips, so `/a/b/` and `/a/b` have always named different files here. The test asserts the three equivalences that do hold and documents that one as pre-existing.

## Checks

`cargo fmt --check`, `cargo clippy --all-targets` (no new warnings on the changed files), `cargo test -p fresh-editor-core --lib` (1299 passed), `cargo test -p fresh-editor --lib` (2525 passed), both regression tests re-run after the review round, the two interactive repros re-verified in tmux against the raised cap, and the plugin type-checker — which reports the same 17 pre-existing errors on `orchestrator.ts` / `lib/widgets.ts` before and after, and no new ones.

---

https://claude.ai/code/session_01GuDYEaFrM5HFjjGEnNYwia

### DIFF
--- a/crates/fresh-editor-core/src/path_encode.rs
+++ b/crates/fresh-editor-core/src/path_encode.rs
@@ -6,6 +6,73 @@
 
 use std::path::Path;
 
+/// The longest filename [`encode_path_for_filename`] will return.
+///
+/// `NAME_MAX` is 255 bytes on Linux and macOS, and **every caller appends to
+/// this** — `.json` for a workspace snapshot, `-3.txt` and `-3.log` for a
+/// terminal's scrollback — so the budget is not the whole of `NAME_MAX`. 200
+/// leaves each of them room.
+const MAX_ENCODED_LEN: usize = 200;
+
+/// Marks a name [`fold_to_max_len`] shortened.
+///
+/// A literal `~` can carry that meaning because the encoder can never emit
+/// one: `~` is not in the pass-through set, so an actual tilde in a path
+/// comes out as `%7E`. Its presence therefore means exactly one thing, which
+/// is what lets `decode_filename_to_path` decline a folded name instead of
+/// handing back a path that was never encoded.
+const FOLD_MARK: char = '~';
+
+/// FNV-1a over the path, spelled out rather than reached for.
+///
+/// The hash names a file that outlives the process, so it has to give the
+/// same answer next week and in the next release. `DefaultHasher` documents
+/// that it does not promise that across Rust versions; twelve lines here do.
+fn fold_hash(bytes: &[u8]) -> u64 {
+    let mut h: u64 = 0xcbf2_9ce4_8422_2325;
+    for b in bytes {
+        h ^= *b as u64;
+        h = h.wrapping_mul(0x0000_0100_0000_01b3);
+    }
+    h
+}
+
+/// Shorten an over-long encoding to `<head>~<hash>`, keeping it unique.
+///
+/// **Why an encoding can outgrow a filename at all.** A path is encoded whole
+/// — every separator becomes an `_` — so the name grows with the depth of the
+/// directory, and the orchestrator nests one encoded path inside another: a
+/// workspace lives under `orchestrator/<slug of the repo root>/<name>`, and
+/// naming *its* terminal directory encodes that path a second time. Two
+/// encodings of one path, plus a long workspace name, is how a real repo
+/// checked out under a long root reached `File name too long (os error 36)`
+/// and left the workspace with nowhere to write its terminals (#1971).
+///
+/// The head keeps the tail of the path readable in a directory listing; the
+/// hash is what keeps two paths sharing a 183-character prefix apart. It runs
+/// over the *path*, not the truncated head, so the digest cannot be truncated
+/// with it.
+fn fold_to_max_len(encoded: String, path_str: &str) -> String {
+    if encoded.len() <= MAX_ENCODED_LEN {
+        return encoded;
+    }
+    // 1 mark + 16 hex digits.
+    const SUFFIX_LEN: usize = 17;
+    let keep = MAX_ENCODED_LEN - SUFFIX_LEN;
+    // The encoder emits ASCII and nothing else — every other byte is
+    // percent-escaped — so cutting on a byte is cutting on a character. It
+    // can still land *inside* a `%XX` triple, which would leave a stray `%`
+    // or a half-escape that decodes as text, so back off to the escape's
+    // start.
+    let mut head = &encoded[..keep];
+    if head.ends_with('%') {
+        head = &head[..head.len() - 1];
+    } else if head.len() >= 2 && head.as_bytes()[head.len() - 2] == b'%' {
+        head = &head[..head.len() - 2];
+    }
+    format!("{head}{FOLD_MARK}{:016x}", fold_hash(path_str.as_bytes()))
+}
+
 /// Encode a path into a filesystem-safe filename using percent encoding
 ///
 /// Keeps alphanumeric chars, `-`, `.`, `_` as-is.
@@ -13,6 +80,11 @@ use std::path::Path;
 /// Percent-encodes other special characters as %XX.
 ///
 /// Example: `/home/user/my project` -> `home_user_my%20project`
+///
+/// A path whose encoding would exceed [`MAX_ENCODED_LEN`] is folded to
+/// `<head>~<hash>` — see [`fold_to_max_len`]. That form is deliberately not
+/// reversible, and `workspace::decode_filename_to_path` declines it rather
+/// than inventing a path.
 pub fn encode_path_for_filename(path: &Path) -> String {
     let path_str = path.to_string_lossy();
     let mut result = String::with_capacity(path_str.len() * 2);
@@ -57,5 +129,55 @@ pub fn encode_path_for_filename(path: &Path) -> String {
         final_result = "root".to_string();
     }
 
-    final_result
+    fold_to_max_len(final_result, &path_str)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn a_path_that_fits_is_encoded_whole() {
+        let encoded = encode_path_for_filename(Path::new("/home/user/project"));
+        assert_eq!(encoded, "home_user_project");
+        assert!(!encoded.contains(FOLD_MARK));
+    }
+
+    /// The orchestrator nests one encoded path inside another, so a workspace
+    /// under a long root ran past `NAME_MAX` and its terminal directory could
+    /// not be created at all — `File name too long (os error 36)` (#1971).
+    #[test]
+    fn an_over_long_path_folds_under_the_cap_and_stays_unique() {
+        let deep = format!("/{}/workspace", "a_very_long_directory_name/".repeat(20));
+        let sibling = format!("{deep}-two");
+        let a = encode_path_for_filename(Path::new(&deep));
+        let b = encode_path_for_filename(Path::new(&sibling));
+
+        assert!(a.len() <= MAX_ENCODED_LEN, "{} bytes", a.len());
+        assert!(b.len() <= MAX_ENCODED_LEN, "{} bytes", b.len());
+        // Long enough to have been folded, and marked as such.
+        assert!(a.contains(FOLD_MARK));
+        // Two paths sharing every one of the kept characters still differ,
+        // because the hash runs over the path and not over the head.
+        assert_ne!(a, b, "a shared prefix must not collapse two workspaces");
+        assert_eq!(a, encode_path_for_filename(Path::new(&deep)), "stable");
+    }
+
+    /// The cut is taken on the encoded form, where a `%XX` escape is three
+    /// characters that mean one byte. Landing inside one would leave a stray
+    /// `%` that reads back as literal text.
+    #[test]
+    fn folding_never_cuts_a_percent_escape_in_half() {
+        // Spaces encode to `%20`, so escapes land at every offset in turn.
+        for pad in 0..8 {
+            let path = format!("/{}{} x", "b".repeat(pad), " x".repeat(200));
+            let encoded = encode_path_for_filename(Path::new(&path));
+            let head = encoded.split(FOLD_MARK).next().unwrap();
+            assert!(!head.ends_with('%'), "{head}");
+            assert!(
+                !(head.len() >= 2 && head.as_bytes()[head.len() - 2] == b'%'),
+                "{head}"
+            );
+        }
+    }
 }
--- a/crates/fresh-editor-core/src/path_encode.rs
+++ b/crates/fresh-editor-core/src/path_encode.rs
@@ -6,13 +6,28 @@
 
 use std::path::Path;
 
-/// The longest filename [`encode_path_for_filename`] will return.
+/// The longest filename [`encode_path_for_filename`] will return: `NAME_MAX`
+/// itself.
 ///
-/// `NAME_MAX` is 255 bytes on Linux and macOS, and **every caller appends to
-/// this** — `.json` for a workspace snapshot, `-3.txt` and `-3.log` for a
-/// terminal's scrollback — so the budget is not the whole of `NAME_MAX`. 200
-/// leaves each of them room.
-const MAX_ENCODED_LEN: usize = 200;
+/// **The cap is what decides which existing names get re-keyed, so it is the
+/// largest one that changes nothing.** This value is not a budget for the
+/// callers that append to the encoding — it deliberately is not. Folding is a
+/// rename, and every place the encoding is a *key* (a workspace snapshot, the
+/// recovery scope, a terminal's scrollback directory, a daemon's socket) would
+/// stop finding what the previous release wrote. So the only names that may
+/// fold are the ones no caller could have created at all, which means the
+/// filesystem's own limit and not a byte less.
+///
+/// Leaving headroom for a suffix would have been wrong in both directions.
+/// Several callers append nothing — `terminal_dir_for`, `working_data_dir_for`,
+/// `project_state_dir` and the recovery scope each join the encoding as a whole
+/// directory name — so a 250-byte encoding is live today and must not move.
+/// And the longest suffix is not a small constant either: `workspace_path_for`
+/// builds `<encoded>.<stable_id>.json`, and a `ws-<hex>-<hex>` id makes that
+/// roughly 28 bytes. A name that overflows *with* a suffix but fits without one
+/// is a caller-specific limit, and it is no better or worse here than it was
+/// before this cap existed.
+const MAX_ENCODED_LEN: usize = 255;
 
 /// Marks a name [`fold_to_max_len`] shortened.
 ///
@@ -49,10 +64,21 @@ fn fold_hash(bytes: &[u8]) -> u64 {
 /// and left the workspace with nowhere to write its terminals (#1971).
 ///
 /// The head keeps the tail of the path readable in a directory listing; the
-/// hash is what keeps two paths sharing a 183-character prefix apart. It runs
-/// over the *path*, not the truncated head, so the digest cannot be truncated
-/// with it.
-fn fold_to_max_len(encoded: String, path_str: &str) -> String {
+/// hash is what keeps two paths sharing a 238-character prefix apart. It runs
+/// over the whole encoding, not the truncated head, so the digest cannot be
+/// truncated with it.
+///
+/// **Over the encoding rather than the path, because the encoding is the
+/// normalised form.** `encode_path_for_filename` maps several spellings of one
+/// path onto one name: both separators become `_`, runs of them collapse, and a
+/// leading one is stripped, so `/a/b`, `/a//b`, `/a/b/` and `a/b` already share
+/// a filename. Digesting the raw path would have broken that for exactly the
+/// long paths that fold — same head, different hash, two directories for one
+/// place — and neither `terminal_dir_for` nor `working_data_dir_for`
+/// canonicalises what it is given (`spawn_terminal_session_impl` passes a
+/// caller-supplied `cwd` straight through), so a trailing separator would have
+/// been enough to strand a workspace's scrollback beside itself.
+fn fold_to_max_len(encoded: String) -> String {
     if encoded.len() <= MAX_ENCODED_LEN {
         return encoded;
     }
@@ -70,7 +96,7 @@ fn fold_to_max_len(encoded: String, path_str: &str) -> String {
     } else if head.len() >= 2 && head.as_bytes()[head.len() - 2] == b'%' {
         head = &head[..head.len() - 2];
     }
-    format!("{head}{FOLD_MARK}{:016x}", fold_hash(path_str.as_bytes()))
+    format!("{head}{FOLD_MARK}{:016x}", fold_hash(encoded.as_bytes()))
 }
 
 /// Encode a path into a filesystem-safe filename using percent encoding
@@ -129,7 +155,7 @@ pub fn encode_path_for_filename(path: &Path) -> String {
         final_result = "root".to_string();
     }
 
-    fold_to_max_len(final_result, &path_str)
+    fold_to_max_len(final_result)
 }
 
 #[cfg(test)]
@@ -146,6 +172,59 @@ mod tests {
     /// The orchestrator nests one encoded path inside another, so a workspace
     /// under a long root ran past `NAME_MAX` and its terminal directory could
     /// not be created at all — `File name too long (os error 36)` (#1971).
+    /// The cap re-keys nothing a previous release could have written: a name
+    /// the filesystem accepts is returned exactly as before.
+    #[test]
+    fn a_name_the_filesystem_accepts_is_never_folded() {
+        // Grown a component at a time until the encoder has to fold. Length
+        // is not the signal here — a folded name is under the cap by
+        // construction, so waiting for one to exceed it never returns — the
+        // mark is.
+        let mut deep = String::from("/a");
+        while !encode_path_for_filename(Path::new(&deep)).contains(FOLD_MARK) {
+            deep.push_str("/bbbbbbbbbb");
+        }
+
+        // One component shorter is the longest name that fits, and it is
+        // untouched — a directory of exactly NAME_MAX bytes is live today.
+        let fits = deep.rsplit_once('/').unwrap().0.to_string();
+        let name = encode_path_for_filename(Path::new(&fits));
+        assert!(name.len() <= MAX_ENCODED_LEN);
+        assert!(
+            !name.contains(FOLD_MARK),
+            "a name that fits must be returned whole, or it stops finding what \
+             the last release wrote under it: {name}"
+        );
+    }
+
+    /// The encoder maps several spellings of one path onto one name, and a
+    /// folded name has to keep doing that — otherwise the head matches, the
+    /// digest does not, and one directory acquires two names.
+    ///
+    /// Only the equivalences the encoder actually provides are asserted. A
+    /// *trailing* separator is not among them: it becomes a trailing `_` that
+    /// nothing strips, so `/a/b/` and `/a/b` name different files here and
+    /// always have. That is worth knowing and is not folding's to fix.
+    #[test]
+    fn folding_keeps_the_encoders_normalisation() {
+        let long = format!("/{}", "a_directory_component/".repeat(30))
+            .trim_end_matches('/')
+            .to_string();
+        let canonical = encode_path_for_filename(Path::new(&long));
+        assert!(canonical.contains(FOLD_MARK), "long enough to fold");
+        for spelling in [
+            long.replace('/', "//"),             // doubled separators collapse
+            long.trim_start_matches('/').into(), // a leading one is stripped
+            long.replace('/', "\\"),             // either separator will do
+        ] {
+            assert_eq!(
+                encode_path_for_filename(Path::new(&spelling)),
+                canonical,
+                "spelling {spelling:?} names the same directory"
+            );
+        }
+    }
+
     #[test]
     fn an_over_long_path_folds_under_the_cap_and_stays_unique() {
         let deep = format!("/{}/workspace", "a_very_long_directory_name/".repeat(20));
--- a/crates/fresh-editor/plugins/lib/widgets.ts
+++ b/crates/fresh-editor/plugins/lib/widgets.ts
@@ -785,7 +785,12 @@ export function windowEmbed(options: {
     // that value is 0. Mapping to it keeps a placeholder rendering as
     // the blanks it is meant to render, instead of taking the panel
     // down with it.
-    windowId: Number.isFinite(options.windowId) && options.windowId > 0
+    // The range is closed at *both* ends: `u32` has a top as well as a
+    // bottom, and an id above it fails to deserialise exactly as a negative
+    // one does — same rejected spec, same frozen panel.
+    windowId: Number.isFinite(options.windowId) &&
+        options.windowId > 0 &&
+        options.windowId <= 0xffff_ffff
       ? Math.trunc(options.windowId)
       : 0,
     rows: options.rows,
--- a/crates/fresh-editor/plugins/lib/widgets.ts
+++ b/crates/fresh-editor/plugins/lib/widgets.ts
@@ -759,7 +759,9 @@ export function textInput(
  * reserved rectangle.
  *
  * `windowId` of 0 (or any unknown id) renders the placeholder
- * blanks without dispatching the per-window paint. */
+ * blanks without dispatching the per-window paint — which is also
+ * what a `windowId` that names no window at all is normalised to,
+ * see below. */
 export function windowEmbed(options: {
   windowId: number;
   rows: number;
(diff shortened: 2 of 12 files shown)


crates/fresh-editor-core/src/path_encode.rs
```diff
@@ -6,13 +6,85 @@
 
 use std::path::Path;
 
+/// The longest filename [`encode_path_for_filename`] will return.
+///
+/// `NAME_MAX` is 255 bytes on Linux and macOS, and **every caller appends to
+/// this** — `.json` for a workspace snapshot, `-3.txt` and `-3.log` for a
+/// terminal's scrollback — so the budget is not the whole of `NAME_MAX`. 200
+/// leaves each of them room.
+const MAX_ENCODED_LEN: usize = 200;
```
sinelaw: The cap re-keys names that worked before, with no migration.

`NAME_MAX` is 255, so every encoding in the **201..=255** band is created successfully today — the PR body's claim that "only names that overflow change, and those could not be created before" holds only above 255. The longest suffix any caller appends is 10 bytes (`.data.sock` / `.ctrl.sock` in `server/ipc/mod.rs`); the common ones are 5 (`.json`) or 0 (`terminal_dir_for`, `working_data_dir_for`, `project_state_dir`, `recovery::storage::with_scope` all join the encoding as a whole directory name).

For a working dir whose encoding is, say, 210 chars — very reachable, since the PR's own repro overflows only because the *doubly* encoded name passes 255, which means the singly encoded one sits in exactly this band — the first run after this change resolves to a different name everywhere the encoding is a key:

- `workspaces/<encoded>.json` / `<encoded>.<stable_id>.json` — `candidate_files_for_root` prefilters on the new encoded prefix, so the saved layout is not found and the editor starts with an empty session;
- `recovery/default/<encoded>/` — unsaved-buffer recovery data written before the upgrade becomes unreachable;
- `terminals/<encoded>/` — scrollback is orphaned;
- `workspaces/<encoded>/` (`project_state_dir`) — per-workspace config is orphaned;
- `<socket_dir>/<encoded>.ctrl.sock` — a daemon started by the old binary is no longer discoverable by `is_server_alive()`, so the client silently starts a second one against the same directory.

Either raise the cap to the real limit (`255 - <longest suffix>`, i.e. 245) so only genuinely impossible names change, or read the unfolded name as a fallback when the folded one is absent.

---

crates/fresh-editor-core/src/path_encode.rs
```diff
@@ -6,13 +6,85 @@
...
+    // can still land *inside* a `%XX` triple, which would leave a stray `%`
+    // or a half-escape that decodes as text, so back off to the escape's
+    // start.
+    let mut head = &encoded[..keep];
+    if head.ends_with('%') {
+        head = &head[..head.len() - 1];
+    } else if head.len() >= 2 && head.as_bytes()[head.len() - 2] == b'%' {
+        head = &head[..head.len() - 2];
+    }
+    format!("{head}{FOLD_MARK}{:016x}", fold_hash(path_str.as_bytes()))
```
sinelaw: Hashing `path_str` instead of `encoded` loses the encoder's normalisation.

`encode_path_for_filename` deliberately maps several spellings of one path onto one name: `\` and `/` both become `_`, runs of separators collapse, and leading separators are stripped. So today `/a/b`, `/a//b`, `/a/b/` and `a/b` all produce the same filename. With the digest taken over the raw path string that stops holding for anything long enough to fold: the heads are identical, the hashes are not, so one directory gets two different names.

`terminal_dir_for` and `working_data_dir_for` (`config_io.rs`) do not canonicalise their argument — `terminal_dir_for` is reached from `spawn_terminal_session_impl` with a caller-supplied `cwd` — so a `cwd` that carries a trailing separator (or a doubled one) now creates a second scrollback directory beside the existing one instead of landing on it.

`fold_hash(encoded.as_bytes())` keeps the same collision resistance (the full pre-truncation encoding still differs for two paths sharing a 183-char prefix) while leaving the result a pure function of the normalised form.

---

crates/fresh-editor/plugins/orchestrator.ts
```diff
@@ -5848,6 +5848,41 @@ interface LifecycleResult {
...
+// is already there — `git worktree move` refuses that — so an occupied name
+// takes the next free `-2`, `-3`, … instead of failing the archive.
+function archivedPathFor(repoRoot: string, s: AgentSession): string {
+  const graveyard = editor.pathJoin(
+    editor.getDataDir(),
+    "orchestrator",
+    slugify(repoRoot),
+    ".archived",
+  );
+  const name = editor.pathBasename(s.root) || s.hostLabel || "workspace";
```
sinelaw: The `s.hostLabel` fallback reintroduces exactly the bug this function fixes.

The whole point of the change is that a *display* string must not name a directory. `s.hostLabel` is one: it is written from a window snapshot's label (`hostLabel: s.label` at the reconcile sites, and `s.hostLabel = label` in the rename path), and `workspaceDisplayName` treats it as free text. It can contain `/`, spaces, em dashes — the same characters that produced `.archived/gamma-… · bash — root@vm: ~/…` and made `git worktree move` nest the graveyard.

It only fires when `editor.pathBasename(s.root)` is empty, which needs `s.root` to be a filesystem root (`Path::file_name` still returns `b` for `/a/b/`), so it is narrow — but if the branch is unreachable it should not silently fall back to an unvalidated string, and if it is reachable it produces the reported failure. Either drop straight to `"workspace"`, or run the fallback through the same single-component check.

---

crates/fresh-editor/plugins/lib/widgets.ts
```diff
@@ -759,15 +759,35 @@ export function textInput(
...
+    // the next keystroke that happened to produce a valid spec
+    // (sinelaw/fresh#1971: the orchestrator's synthetic ids for
+    // workspaces that have no window yet, which left the picker on
+    // "Archiving… / Waiting for git…" after the archive had finished).
+    //
+    // "No window" is a state this widget already has a value for, and
+    // that value is 0. Mapping to it keeps a placeholder rendering as
+    // the blanks it is meant to render, instead of taking the panel
+    // down with it.
+    windowId: Number.isFinite(options.windowId) && options.windowId > 0
```
sinelaw: The guard only closes the low end of the `u32` range.

The stated rule is "a window id the host cannot hold is *no window*", but `options.windowId > 0` lets anything above `u32::MAX` through: `windowEmbed({ windowId: 5_000_000_000, rows: 5 })` still fails `expected u32` on the host, serde still rejects the *whole* spec, and `updateFloatingWidget` still drops it — the same frozen panel the rest of this comment describes. `Math.trunc` handles the fractional case but not the magnitude one.

```suggestion
    windowId: Number.isFinite(options.windowId) && options.windowId > 0 &&
        options.windowId <= 0xffff_ffff
      ? Math.trunc(options.windowId)
      : 0,
```

---

crates/fresh-editor/src/workspace.rs
```diff
@@ -805,6 +805,14 @@ pub fn decode_filename_to_path(encoded: &str) -> Option<PathBuf> {
     if encoded == "root" {
         return Some(PathBuf::from("/"));
     }
+    // A name the encoder had to fold to fit `NAME_MAX` keeps only the head of
+    // the path plus a digest, so there is no path to give back. The encoder
+    // can never emit a literal `~` (a real one becomes `%7E`), which is what
+    // makes its presence a reliable answer rather than a guess. Callers
+    // already handle `None` by falling back to the name itself.
+    if encoded.contains('~') {
```
sinelaw, line 813: "Callers already handle `None` by falling back to the name itself" is only true for one of the two callers.

`main.rs` does fall back harmlessly (it just prints the raw session name). But `app/orchestrator_persistence.rs` does

```rust
let project_path = crate::workspace::decode_filename_to_path(&dir_name)
    .unwrap_or_else(|| PathBuf::from(dir_name.clone()));
...
if w.project_path.is_none() { w.project_path = Some(project_path.clone()); }
```

so for a folded directory — which the encoder can now produce, at `orchestrator_persistence.rs:203`/`:510`, for exactly the deep orchestrator paths this PR is about — the `windows.json` migration writes the *filename* (`…deep-directory-six~18289b17dcb27c00`) into every migrated window's `project_path` as a relative path that names nothing. That is persisted, not just displayed.

Skipping the entry (or leaving `project_path` unset) when the decode declines would be safer than storing a path that cannot resolve.

---

crates/fresh-editor/src/app/widget_runtime.rs
```diff
@@ -4022,4 +4064,121 @@ mod tests {
...
+        // The tree cannot carry it yet — the option row is not in the frame
+        // that is currently built — so the move is held rather than dropped.
+        assert_eq!(
+            editor.pending_panel_tree_focus,
+            Some((panel_key.clone(), "menu-pick:move:root".to_string())),
+            "a move with no element to land on is held for the next frame"
+        );
+
+        frame_the_shell(&mut editor);
+        editor.retry_pending_panel_tree_focus();
```
sinelaw: The regression test drives the replay by hand, so it cannot catch the fix being unwired.

`frame_the_shell` is documented as "the same two calls `Editor::render` makes around it", but `render` now makes three — it calls `retry_pending_panel_tree_focus()` right after `self.shell_ui = Some(ui)`. The helper does not, and this test compensates by calling it explicitly. Delete `render.rs:418` and #3137 returns in the real editor while this test still passes, which is the one regression the test exists to prevent.

Moving the call into `frame_the_shell` (keeping the helper honest to its own doc comment) and dropping the explicit call here would close that.

---

sinelaw (commented):
Reviewed the three fixes. The focus-replay mechanism (#3137) reads correctly to me: the registry is written before `focus_panel_widget_in_tree`, so the retry's `focus_key` re-check is a real staleness guard; `(holds, found)` is the right discriminator; and the replay sits after `ui.frame()` so the gain it raises is the last one `apply_settled_shell_messages` sees. The `windowEmbed` normalisation and the archived-directory rename are both the right shape.

Six comments, none blocking the mechanisms themselves — the substantive one is the 200-byte cap, which changes names that fit `NAME_MAX` today and so silently re-keys existing on-disk state rather than only fixing the overflow.

---

crates/fresh-editor-core/src/path_encode.rs
sinelaw: Right, and this was the serious one — thanks. Fixed in 766f171.

The claim in the PR body was wrong: folding is a rename, so every encoding in 201..=255 would have moved on first run and stopped finding what it named, exactly as listed.

I went further than 245 though, to `NAME_MAX` itself, because "255 − longest suffix" is wrong in both directions:

- **Too low.** Several callers append nothing and join the encoding as a whole directory name — `terminal_dir_for`, `working_data_dir_for`, `project_state_dir`, and the recovery scope. So a 250-byte encoding is live today as a directory, and 245 would still have re-keyed the 246..=255 band for exactly the terminal directories this PR is about.
- **Too high.** The longest suffix is not 10. `workspace_path_for` builds `<encoded>.<stable_id>.json`, and `generate_stable_id`'s `ws-<hex nanos>-<hex seq>` makes that around 28 bytes.

Since one encoding is shared by all of them, the only cap that renames nothing live is the filesystem's own limit. A name that overflows *with* a suffix but fits without one stays a caller-specific limit — unchanged from before this cap existed, and out of scope here.

Re-verified the original repro after the change: the terminal directory is created at exactly 255 bytes with its `.txt`/`.log` inside and no `os error 36`. New test `a_name_the_filesystem_accepts_is_never_folded` grows a path until the encoder folds, then asserts the one-component-shorter name comes back whole.

---

crates/fresh-editor-core/src/path_encode.rs
sinelaw: Correct on the substance — fixed in 766f171, now `fold_hash(encoded.as_bytes())`. The head was already a function of the normalised form and the digest was not, so a path long enough to fold could acquire two names for one directory. Collision resistance is unchanged, since the full pre-truncation encoding still differs for two paths sharing a 238-character prefix.

One correction to the example, though: **the trailing-separator case does not hold**, and never did. `trim_start_matches('_')` only strips a *leading* underscore, so a trailing separator survives as a trailing `_`:

```
'/a/b'   -> 'a_b'
'/a//b'  -> 'a_b'
'/a/b/'  -> 'a_b_'     <-- differs, before and after this PR
'a/b'    -> 'a_b'
'\a\b'   -> 'a_b'
```

So a `cwd` with a trailing separator already got its own scrollback directory on master; folding neither caused nor worsened it. The doubled-separator, leading-separator and backslash equivalences are real, and those are what the fix protects. `folding_keeps_the_encoders_normalisation` asserts exactly those three and documents the trailing case as a pre-existing property rather than pretending it holds — my first draft asserted it and failed, which is how I noticed.

---

crates/fresh-editor/plugins/orchestrator.ts
sinelaw: Agreed, and it was inconsistent of me to write the doc comment about display strings and then reach for another one as the fallback. Dropped in 766f171 — the fallback is now just `"workspace"`:

```ts
const name = editor.pathBasename(s.root) || "workspace";
```

You're right on both horns: unreachable means it should not silently accept an unvalidated string, and reachable means it produces the reported failure. `hostLabel` is written from the window snapshot's label at the reconcile sites and overwritten by the rename path, so it is free text with the same `/` and space problem. A root with no basename is a filesystem root, which owns no workspace, so the constant is the honest answer rather than a second guess.

The main path is unchanged and re-verified in tmux: archiving still lands at `.archived/good`, with the manifest keeping the display label.

---

crates/fresh-editor/plugins/lib/widgets.ts
sinelaw: Correct — I closed the end the bug happened to arrive from rather than the range the host actually accepts. `u32` has a top as well as a bottom, and an id above it is the same rejected spec and the same frozen panel. Suggestion applied in 766f171, with the comment saying why both ends are there.

---

crates/fresh-editor/src/workspace.rs
sinelaw: You're right, and the distinction that matters is the one you drew: `main.rs` only prints it, but this one **persists** it. My sentence was written for the harmless caller and applied to both.

Fixed in 766f171 — the migration now leaves `project_path` unset rather than storing the filename:

```rust
let project_path = crate::workspace::decode_filename_to_path(&dir_name);
...
if w.project_path.is_none() {
    w.project_path = project_path.clone();
}
```

I preferred this to skipping the entry: `continue` would drop every window in that directory from the migration, which is worse than a missing field. Unset is a state the store already handles — it is exactly where an entry that never carried a `project_path` sits — whereas `…deep-directory-six~18289b17dcb27c00` as a relative path is a value nothing can resolve and everything downstream would have to special-case.

And these are precisely the deep orchestrator roots that now fold (`orchestrator_persistence.rs:203`/`:510`), so this was reachable, not theoretical.

---

crates/fresh-editor/src/app/widget_runtime.rs
sinelaw: The best catch of the six — a regression test that cannot fail is worse than none, because it reports coverage it does not have. Fixed in 766f171, though moving the call into `frame_the_shell` alone turned out **not** to close it, and I checked rather than assuming:

```
# retry_pending_panel_tree_focus() deleted from render.rs, call moved into frame_the_shell
test a_focus_onto_a_widget_the_next_frame_builds_lands_on_that_frame ... ok
```

Still green, because the helper was still a *copy* of `render`'s sequence — it had simply acquired its own call. Two spellings of one step, which is the same fault one level up.

So the sequence is now a single function, `Editor::lay_out_shell_tree` (take the retained tree, `note_shell_layout`, `ui.frame`, put it back, replay the pending focus), called by both `render` and `frame_the_shell`. `render`'s inline block is gone, and `regions_of` / `PaneRects::read` read the tree back through `shell_ui.as_ref()` afterwards. Re-checked with the replay deleted from that one function:

```
test a_focus_onto_a_widget_the_next_frame_builds_lands_on_that_frame ... FAILED
test result: FAILED. 1 passed; 1 failed
```

Restored, both green, and the doc comment on the helper now says it calls the real thing rather than claiming to mirror it.

---

PR #3168, opened 2026-09-03
refactor(lsp): move Deno detection out of the editor and into a plugin

sinelaw:
Follow-up to #2981, on top of #3111. The runtime check it added is kept; what changes is where the convention lives, and what it leaves alone.

## Before / after

Setup for all of these: a project with `package.json`, `index.ts` and `git init`, workspace trusted, `typescript-language-server` on `PATH`. JS/TS are `auto_start: false`, so start the server with `Ctrl+P → Start/Restart LSP Server`.

### A `deno.json` in a project without Deno (#2981, kept from #3111)

1. Add `deno.jsonc` to the project. Do **not** install `deno`.
2. Open `index.ts`, start the LSP.

| before #3111 | after |
|---|---|
| `LSP (error)`, nothing spawned; both languages lost to a server that cannot start | `typescript-language-server --stdio` spawned, `LSP (on)` |

With `deno` installed the same steps spawn `deno lsp`, as before.

### TypeScript the user disabled

1. In config, set `lsp.typescript` to `enabled: false` (or pick "Disable LSP for typescript" in the LSP Servers popup once).
2. Add `deno.jsonc`, install `deno`, open `index.ts`, `Ctrl+P → Show LSP Status`.

| before (core version on master) | after |
|---|---|
| TypeScript replaced with `deno lsp`, **enabled** — the user's disable is undone | TypeScript untouched: still `typescript-language-server`, still disabled (popup offers "Enable LSP for typescript"); JavaScript, still enabled, switches to `deno` |

### Leaving a Deno project

1. Open a Deno project (`deno.jsonc`, `deno` installed) → `deno lsp`.
2. `Ctrl+P → Switch Project` to a plain npm project, open its `index.ts`, start the LSP.

| before | after |
|---|---|
| the per-window block re-evaluated at construction, so this worked — but a language disabled at runtime inside the Deno project came back enabled | `typescript-language-server` with `rootUri` = the new root; a language disabled inside the Deno project stays disabled |

## Why

`configure_lsp_servers` hard-coded one project convention — a root holding a `deno.json` or `deno.jsonc` is served by `deno lsp` — into window construction. That is a claim about *projects*, not about the editor: it cannot be turned off, inspected or replaced without a rebuild, every further convention costs another branch in a function whose job is wiring a manager, and #2981 had to fix it by adding the runtime check inside that same block.

## What this does

The block is gone, and so is `configure_lsp_servers`' `root` parameter — nothing else in it used the root. The convention is `plugins/deno_lsp.ts`, which:

* switches JavaScript and TypeScript to `deno lsp` when the active window's root has a `deno.json`/`deno.jsonc` **and** `deno` is on `PATH` — the #2981 condition, unchanged in meaning;
* restores what was configured before when a window without one becomes active, read from the live config before the first override so the user's own entry comes back;
* caches its decision per root.

Disable the plugin and the behaviour is gone; copy it and Fresh learns the same trick for another runtime.

## What it leaves alone

A language the user has turned off stays off, in both directions. The "Disable LSP for …" rows persist `enabled: false`; `registerLspServer` can only register an *enabled* server, so touching such a language would switch it back on. A user who configured `deno lsp` themselves is likewise left with it, runtime present or not.

## Two differences from the core version

* **Scope.** The choice follows the active window (`window_created`, `active_window_changed`, `buffer_activated`) rather than being fixed per window root at construction.
* **Timing.** It lands when the plugin loads, so an already-spawned server keeps its command until restarted. JS/TS are `auto_start: false` by default, so nothing has started that early.

## Open question: the switch can be persisted

`registerLspServer` writes into the runtime config as well as the window's manager (the core version touched only the manager), and `save_to_layer` saves the runtime config's delta against the lower layers. So any config save while a Deno project is active — the LSP popup's Disable/Enable rows, a settings toggle — writes `javascript`/`typescript = deno lsp` into the user's `config.json`, where it survives leaving the project and the next launch. I reproduced this: after "Disable LSP for typescript" in a Deno project, the user config held `deno` for both languages and a plain project auto-started `deno` for JavaScript.

I have not changed that here because it is the API's contract, and `rust-lsp.ts` relies on it (its full/reduced-memory choice is meant to persist). Two ways out, and which one is a design call:

1. a runtime-only override for plugin LSP registrations (`setLspServerOverride(language, config | null)` or similar): applied to every window's manager and to new windows, never written to the config. This plugin would then need no baseline at all — restore is "clear the override". It also fixes the same leak for language packs, whose servers currently land in the user's config on the next save.
2. accept the persistence and document it.

I'd take (1) if you agree; say the word and I'll do it in this PR or a follow-up.

## Tests

`tests/e2e/deno_lsp_plugin.rs` drives the real plugin through the harness with a stub `deno` on `PATH`: a Deno project with the runtime switches both languages, one without it keeps its configured server (#2981), a project with no marker file is untouched, a disabled language is neither switched nor re-enabled, and leaving a Deno project (`create_window_at` + `set_active_window`) restores the configured server. They take `pin_config_globals()` since `PATH` is a process global. I checked the positive case fails when the plugin's decision is neutered, so it tests the plugin rather than the default config.

Verified interactively as well (tmux, real keystrokes, logging stubs for both servers): the eight combinations of marker file × runtime installed × user-configured `deno lsp`, the disabled-language cases above, and Switch Project in both directions with and without the runtime.

Two harness findings, both handled in the plugin: `plugins_loaded` is not delivered in every embedding (the first evaluation happens at module load), and `editor.getEnv` answers `undefined` for an unset variable — a `!== null` check on `PATHEXT` reads as Windows on Linux and splits `PATH` on the wrong separator.

`cargo fmt --check`, `cargo check`/`clippy -p fresh-editor --all-targets` and the new tests are clean locally.

---

https://claude.ai/code/session_01SHsuKyMdSZ7soeFzDziRNw


sinelaw:
`test ubuntu-latest` is red on 2c009fd: **9545 of 9546 passed, one timed out** —

```
TIMEOUT [ 180.131s] (7399/9546) fresh-editor::all_tests remote_channel_timeout_tests::test_multiple_reconnections
```

I don't believe that one is this PR's, for three reasons:

1. **The diff doesn't touch that code.** It is a plugin file, the removal of the Deno block from `configure_lsp_servers`, three new e2e tests, a changelog line and a `tsconfig.json` entry. `remote_channel_timeout_tests` exercises `services::remote::channel` reconnection, which none of that reaches.
2. **It is nowhere near its budget.** Locally the whole module passes in 8.0s with that test at 6.9s — a 26× margin under nextest's 180s slow-timeout. A test that needs 6.9s and gets killed at 180s was starved, not broken.
3. **Master has the same failure mode.** Run [33623560543](https://github.com/sinelaw/fresh/actions/runs/33623560543) on master ends `9466 passed, 1 timed out` with `TIMEOUT [180.128s] … markdown_compose_scrollbar_markers::every_heading_is_marked_without_scrolling` — a different, equally unrelated test hitting the same 180s wall on a loaded runner. I also saw a third member of this same module (`test_reconnection_via_replace_transport`) time out in a local full-suite run of #3159, which does not contain this change.

So this reads as the runner-load flake this repo already has rather than something to fix here, and there is no existing fix to port. I have not touched the test: making it robust would mean changing timing constants in remote-channel code this PR has no business in, and skipping or quarantining it is not on the table.

What I *can* say against the alternative: this PR does add ~15s of test time to a ~1060s suite, so it is not literally load-neutral. If a re-run comes back red on the same test I will treat it as this PR's and dig into it properly rather than re-running again.

Re-running the failed job now (my one re-run — the first attempt was refused while the sibling macOS/Windows jobs were still in flight).

---

PR #3179, opened 2026-09-03
Status bar line number follows an added or removed cursor

### LINKED ISSUE #3167: Status bar shows a stale line number after Ctrl+Alt+Down (add cursor below)
## Steps to reproduce

Any file with at least three lines, e.g.:

```
abcd
abcd
abcd
```

1. Cursor at `Ln 1, Col 1`.
2. Press `Ctrl+Alt+Down` (add cursor below) — status bar shows `2 cursors`.
3. Press `Ctrl+Alt+Down` again — status bar shows `3 cursors`.
4. Press `Right`.

## Actual

| after | status bar |
|---|---|
| step 2 | `Ln 1, Col 1` |
| step 3 | `Ln 1, Col 1` |
| step 4 | `Ln 3, Col 2` |

The line number stays at `Ln 1` while cursors are being added, then jumps to `Ln 3` on the first arrow key — the single `Right` press appears to move the cursor two lines.

## Expected

The line number should follow the cursor the status bar is reporting, so it reads `Ln 2` after step 2 and `Ln 3` after step 3.

## Notes

- The column is reported correctly throughout; only the line number lags.
- Any real cursor movement refreshes it.

sinelaw:
Fixes #3167.

**Why:** the status bar's `Ln`/`Col` describes the cursor the next keystroke will act from, and that is the primary cursor — so when adding, removing or bulk-editing changes which cursor is primary (or where it is), the line must follow, exactly as the column already does.

**Symptom** (reproduced on master in tmux): `Ctrl+Alt+Down` twice shows `Ln 1, Col 1 · 3 cursors`; the first `Right` then jumps to `Ln 3, Col 2`. `Enter` with three cursors and `Esc` afterwards leave it stale the same way.

**Cause:** `primary_cursor_line_number` is a cache only `apply_move_cursor` refreshed, and only for the primary. `AddCursor`, `RemoveCursor` and the multi-cursor bulk path change the primary without a `MoveCursor`.

**Fix:** one `sync_primary_cursor_line_number` on `EditorState`, called from those three sites and from `apply_move_cursor` / `apply_bulk_edit`, which previously carried their own copies (with two different large-file fallbacks). A chokepoint on `Cursors` is not possible: it lives in `fresh-editor-core`, the cache on `EditorState` in `fresh-editor`.

**Tests:** `e2e::multicursor::test_status_bar_line_follows_added_and_removed_cursors` asserts the rendered status bar after each add, after a bulk `Enter`, and after `Esc`; fails on master at the first assertion. The `visual_comprehensive_a` snapshot's status bar changes from `Ln 6` to `Ln 19` for the same reason (its last `Ctrl+D` put the primary on line 19).


crates/fresh-editor/src/state.rs
```diff
@@ -1225,24 +1227,41 @@ impl EditorState {
...
+    /// relative line-number gutter) read instead of resolving the primary
+    /// cursor's line on every frame. [`Self::apply_move_cursor`] keeps it in
+    /// step for the cursor it moves, but only while that cursor is the
+    /// primary. Anything that changes *which* cursor is primary — adding one
+    /// (the new cursor becomes primary), removing one — or that writes cursor
+    /// positions without a `MoveCursor` (the multi-cursor bulk-edit path)
+    /// must call this too, or the status bar keeps reporting the line of a
+    /// cursor that is no longer the primary, while the column, read live,
+    /// is right (#3167).
+    pub fn sync_primary_cursor_line_number(&mut self, primary_position: usize) {
```
sinelaw: The description says this is now "one place that knows how to compute the value (including the large-file estimate) rather than four copies of the `match`", but a fifth copy in this same `impl` was left behind: `EditorState::apply_bulk_edit` (state.rs ~L1438-1442, the undo/redo path for every `BulkEdit`) still does its own

```rust
let primary_pos = cursors.primary().position;
self.primary_cursor_line_number = match self.buffer.offset_to_position(primary_pos) {
    Some(pos) => LineNumber::Absolute(pos.line),
    None => LineNumber::Absolute(0),
};
```

with a *different* large-file fallback (`Absolute(0)` instead of `primary_position / 80`). So after this PR a multi-cursor edit in a large file without line metadata estimates the line one way on the forward path (the new call in `event_apply.rs`) and another way when the same edit is undone or redone. It already has `cursors.primary().position` in hand, so it is a one-line switch to `self.sync_primary_cursor_line_number(primary_pos)`.

The same hand-rolled shape (with `unwrap_or(0)`) also lives in `app/navigation.rs:90`, `app/window/mod.rs:1928` and `app/workspace.rs:1590`. Not asking for all of those in this PR, but given CONTRIBUTING's "when a bug recurs, consider centralizing" and that the helper was introduced precisely so the computation is not copied, at least the copy in this file should go through it - and if the `app/` sites are converted later, `pub(crate)` would match `replay_bulk_marker_adjustments` right below rather than `pub`.

---

crates/fresh-editor/src/app/event_apply.rs
```diff
@@ -570,7 +570,18 @@ impl Editor {
...
+
+        // The cursors were written directly rather than through `MoveCursor`,
+        // so the primary's cached line number did not follow (#3167).
+        self.windows
+            .get_mut(&self.active_window)
+            .map(|w| &mut w.buffers)
+            .expect("active window present")
+            .get_mut(&active_buf)
+            .unwrap()
+            .sync_primary_cursor_line_number(primary_position);
```
sinelaw: This six-call lookup chain (`self.windows.get_mut(..).map(..).expect(..).get_mut(&active_buf).unwrap()`) is repeated verbatim a few lines below for `.highlighter.notify_edits(&edit_lengths)`. Nothing borrows `self` between the two, so they can share one binding:

```rust
let state = self
    .windows
    .get_mut(&self.active_window)
    .map(|w| &mut w.buffers)
    .get_mut(&active_buf)
    .unwrap();
state.sync_primary_cursor_line_number(primary_position);
state.highlighter.notify_edits(&edit_lengths);
```

(The `state` binding earlier in this function had to be dropped for the `split_view_states_mut()` borrow, which is presumably why the chain was re-spelled, but it only needs re-spelling once.)

---

sinelaw (commented):
Read-only review of bf8007a (no build/test run). The fix itself checks out; two inline comments, one on an overstated centralization claim with a real behavioural consequence, one on style.

What I verified by reading:

- All three sites reach the new helper. `AddCursor`/`RemoveCursor` go `apply_event_to_active_buffer` -> `Window::apply_event_to_keyed_buffer` -> `EditorState::apply` with no interception, and `Esc` emits per-cursor `RemoveCursor` events (`Cursors::remove_secondary` is unused in editor code). Multi-cursor `Enter` takes `apply_events_as_bulk_edit` via the `events.len() > 1` fallback in `apply_action_as_events`.
- Borrow shapes compile: `cursors` is a separate `&mut Cursors` parameter in `apply`/`apply_add_cursor`, and `primary_position` is read out of the `split_view_states_mut()` block before `self.windows` is borrowed again.
- `Cursors::remove` always reassigns `primary_id`, so `cursors.primary()` in the `RemoveCursor` arm cannot panic; on the last cursor `remove` is a no-op and the sync is harmless.
- Undo/redo of a cursor add inverts to `RemoveCursor`/`AddCursor`, both of which now sync.
- The test's expectations are right: after `End`, `Enter` on three cursors the primary (id 2, line 3) lands on line 6; `Esc` removes ids 1 and 2 in whatever HashMap order, and `remove`'s arbitrary `keys().next()` handoff always ends on id 0 (line 2). On master the cache stays at `Absolute(0)` after the first `Ctrl+Alt+Down`, so `Ln 2, Col 1` is indeed the first failing assertion. Assertions are on the rendered status bar, `Ctrl+Alt+Down` and `Esc` are bound in `keymaps/default.json`, and the `Ln {line}, Col {col}` format matches.

---

crates/fresh-editor/src/state.rs
sinelaw: Done in 67429b5: `apply_bulk_edit` now calls `sync_primary_cursor_line_number`, so forward and undo/redo of a bulk edit estimate the line the same way, and the helper is `pub(crate)`. The three `app/` copies are left for a follow-up as you suggest.

---

crates/fresh-editor/src/app/event_apply.rs
sinelaw: Done in 67429b5: one `state` binding serves both the sync and `notify_edits`.

---

sinelaw:
CI on 67429b5 failed `e2e::visual_regression::visual_comprehensive_a` (ubuntu and macOS; the test is ignored on Windows), and it was this PR's, in the intended direction: that snapshot ends with two `Ctrl+D` presses, so its status bar was recording the stale line — `Ln 6, Col 12` while the primary cursor sat on line 19. CI's diff was that single line, now `Ln 19, Col 12`; the snapshot is updated in 3d92085 and `origin/master` merged in.

---

PR #3180, opened 2026-09-03
Multi-cursor: a skip-over lands past its delimiter when an earlier cursor inserts

### LINKED ISSUE #3166: Multi-cursor: typing over an auto-closed `)` is dropped when another cursor inserts on an earlier line
Related to #3125 but a different arrangement, and not fixed by #3158 — that PR fixed the case where *every* cursor steps over an existing `)`. This one has one cursor inserting and one stepping over, and still produces the `a(;)` from #3125.

## Steps to reproduce

`mix.cpp`, two lines:

```
xx
a()
```

1. Put the cursor at the end of line 1 (after `xx`, `Ln 1, Col 3`).
2. `Ctrl+Alt+Down` — status bar shows `2 cursors`. The second cursor lands on line 2 at `Col 3`, i.e. between the `(` and the `)`.
3. Type `)`.
4. Type `;`.
5. Save.

## Actual

```
xx);
a(;)
```

The `;` on line 2 went **inside** the parens. After step 3 the status bar does not advance — the cursor on line 2 did not step over its `)`.

## Expected

```
xx);
a();
```

Cursor 1 has no `)` in front of it and inserts one; cursor 2 does and should step over it, leaving the `;` after the parens on both lines.

## Notes

- Swapping the two lines (`a()` first, `xx` second — so the stepping-over cursor is on the *earlier* line) works correctly and gives `a();` / `xx);`.
- Reproduces on master and on the #3158 branch.

sinelaw:
Fixes #3166.

## Symptom

Reproduced first on master, driven interactively in tmux with real keystrokes: `mix.cpp` = `xx` / `a()`, `End`, `Ctrl+Alt+Down` (status bar `2 cursors`), type `)`, type `;`, `Ctrl+S`.

| after | line 1 | line 2 | status bar |
|---|---|---|---|
| `)` | `xx)` | `a()` | `Ln 1, Col 3` — the second cursor did not step over |
| `;` (saved) | `xx);` | **`a(;)`** | |

The reverse ordering (`a()` above `xx`, with the cursor between the parens) gives `a();` / `xx);` on master, as the issue says.

## Cause

Exactly the "known adjacent defect" from #3158. A skip-over owns no edit: `handle_skip_over` emits one `MoveCursor`, `insert_position → insert_position + 1`, in the coordinates of the buffer *before* the keystroke. `apply_events_as_bulk_edit` shifts a cursor's `MoveCursor` for the other cursors' edits only when that cursor also carries an `Insert` at its own position (the auto-close shape); a move from any other cursor is taken as already post-edit, because that is what indent and the other bulk emitters produce. With the inserting cursor on the *earlier* line, its one byte is below the skipping cursor, the skip-over's move is applied unshifted, and the cursor lands one byte short — inside the parens.

"Shift any cursor that owns no edit" is the wrong discriminator (it breaks `test_multicursor_indent_with_selections`, as #3158 found), so this does not touch the applier.

## Fix

At the emitter. `insert_char_events` records the index of each `MoveCursor` the two skip-over paths push, and after the per-cursor loop `shift_skip_over_moves` adds to each one the net byte delta of the *other* cursors' `Insert`/`Delete` events at strictly lower positions. The skip-over moves reach the applier in post-edit coordinates, which is what its existing rule assumes for a cursor with no own insert. Auto-close cursors keep their pre-edit moves (the applier already shifts those). Only other cursors' edits count, so the dedent skip-over variant — which deletes and re-inserts its own line's indent and already moves relative to the result — and a cursor deleting its own selection are not double-counted. With one cursor there are no other cursors' edits and nothing changes.

**Contract this creates** (found in review): a multi-cursor list from `insert_char_events` is now meaningful only to `apply_events_as_bulk_edit` — its skip-over moves are post-edit while its auto-close moves are pre-edit, and applying it one event at a time through `EditorState::apply` would shift the skip-over moves twice. Every multi-cursor dispatcher already routes such lists to the bulk applier, so no production path does that; the doc comment on `insert_char_events` states it. The structural fix would be to carry the coordinate space on `Event::MoveCursor` itself so the applier never has to infer it. That is an event-enum change touching every emitter; I have not done it here and mention it as the long-term option.

## Verification

- `e2e::multicursor::test_type_over_auto_closed_paren_when_earlier_cursor_inserts`: real key events on a `.cpp` fixture with auto-close on, both orderings (with `a()` first, `End` then `Left` puts the cursor between the parens; the test asserts `Ln 1, Col 3` before adding the second cursor). Asserts the buffer after `)` (a setup guard — the buffer is the same on master) and after `;`, plus the rendered `a();` / `xx);`. For the `xx`-first ordering the `;` assertion is the exact bytes the tmux run above produced on master, so the test cannot pass there; the reverse ordering was not run in tmux and is covered by the test alone.
- Per the maintainer's instruction in this session, nothing was built or run locally after the master reproduction: the test's passing direction, `indent_dedent`, `multicursor`, `auto_close`, `bracket`, `smart_editing`, `paste`, `popup`, `undo_redo`, `virtual_space`, `indent`, clippy and the unit tests are all left to CI. `cargo fmt` is clean.
- An independent read-only review posted four inline findings; three are addressed in the follow-up commit, and the fourth (branch history, below) is left open for the maintainer.
- History note: the branch was first cut from a stale local `master`; `origin/master` was merged in rather than rewriting the pushed commit, and one follow-up commit restores the test file after a bad conflict resolution. The merge commit and the first commit do not build on their own. The diff against `master` is the three files above only; a squash-merge, or a rebase to one commit on request, gives a clean history.


crates/fresh-editor/tests/e2e/multicursor.rs
```diff
@@ -2257,3 +2257,50 @@ fn test_type_over_auto_closed_paren_with_multiple_cursors() {
...
+        config.editor.auto_close = true;
+        let mut harness =
+            EditorTestHarness::create(100, 24, HarnessOptions::new().with_config(config)).unwrap();
+        harness
+            .load_buffer_from_text_named("mix.cpp", fixture)
+            .unwrap();
+
+        // End of line 1, then a second cursor on line 2 at the same column,
+        // which is between `(` and `)` on the `a()` line.
+        harness.send_key(KeyCode::End, KeyModifiers::NONE).unwrap();
```
sinelaw: The second fixture cannot pass, with or without the fix. `End` is bound to `move_line_end` (`keymaps/default.json`), so on `a()\nxx\n` it puts the primary cursor at col 3, *after* the `)`, and `add_cursor_below` clamps col 3 to col 2 on `xx` (`cursor_position_on_line` uses `target_col.min(line_len)`). Neither cursor then has a `)` in front of it, so typing `)` is a plain insert at both: the buffer becomes `a())\nxx)\n`, and `assert_buffer_content(&after_paren)` expects `a()\nxx)\n`. The comment "which is between `(` and `)` on the `a()` line" is only true for the first fixture, where `End` on `xx` gives col 2. The issue's note that the swapped ordering already works presumes the cursor was placed between the parens, which these keystrokes do not do.

For the reverse ordering, press `Left` after `End` (or build both lines by typing `a(` with auto-close on, as `test_type_over_auto_closed_paren_with_multiple_cursors` does) so the stepping cursor is really inside the parens.

Two smaller points on what the test proves: the `after_paren` assertion does not discriminate the bug from the fix (on master the buffer after `)` is also `xx)\na()\n`; only the cursor differs), so it is a setup guard rather than evidence of the skip-over, which the comment above it implies. And the PR description's "the assertions are the exact bytes the tmux run above produced on master" covers the first ordering only; the tmux reproduction described there was never run with `a()` first.

---

crates/fresh-editor/src/input/actions.rs
```diff
@@ -865,6 +865,65 @@ fn handle_skip_over(events: &mut Vec<Event>, cursor_id: CursorId, insert_positio
...
+/// second one byte short of stepping over its `)` — the `;` typed next went
+/// inside the parens (#3166). Bringing the skip-over moves into post-edit
+/// coordinates here, at the emitter, keeps the applier's one rule true.
+///
+/// Only edits by *other* cursors at strictly lower positions count. A
+/// cursor's own edits are already folded into the move it emits: the
+/// dedent variant deletes and re-inserts its own line's indent and moves
+/// relative to the result, and a selection is deleted at the position the
+/// move starts from, not below it.
+fn shift_skip_over_moves(events: &mut [Event], skip_over_moves: &[usize]) {
```
sinelaw, line 890: After this change one event list from `insert_char_events` carries `MoveCursor`s in two coordinate spaces: an auto-close move is still pre-edit (the applier shifts it via `insert_at_cursor_pos`), a skip-over move is now post-edit. That is correct for `apply_events_as_bulk_edit`, and I traced every shape it can emit with two cursors in either order (auto-close, plain insert, selection delete then skip-over, surround, virtual-space gap, block pad, auto-dedent, dedent skip-over) through that applier and the rebased targets come out right. But the same list has a second consumer that applies it sequentially: `EditorState::apply`, where `apply_insert` calls `cursors.adjust_for_edit(position, 0, len)` and `apply_move_cursor` sets the position verbatim. That is how every multi-cursor unit test in this file consumes `action_to_events` (`for event in events { state.apply(&mut cursors, &event) }`, e.g. the `foo()\nfoo()` skip-over tests around line 5999). Under sequential application the pre-edit move was the correct one: the higher cursor's `MoveCursor` runs first, and the lower cursor's insert then shifts it. With the rebase, the `xx`/`a()` case applied sequentially lands cursor 2 at byte 8 (past the newline) instead of 7, because the +1 is applied twice.

No production path does this today: `handle_insert_char_editor`, `apply_action_as_events` and `handle_popup_type_char` all send `len() > 1` lists to the bulk applier, and the single-cursor path has no other cursors' edits to add. So this is not a user-visible bug, but the emitter's output is now only valid for one of its two consumers, and a unit test for #3166 written at the action layer (the layer #3158 said "covers" skip-over) would fail. At minimum the doc comment on `insert_char_events` should state that a multi-cursor list from it is only meaningful to `apply_events_as_bulk_edit`; that also strengthens the case for the structural fix you mention (a coordinate-space tag on `Event::MoveCursor`), since this is the second place the applier's inference had to be worked around.

---

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
...
+* **Typing over an auto-closed `)` lands past it even when another cursor inserts one** - with `xx` above `a()` and a cursor at the end of each line, typing `)` gave the first line its `)` but left the second cursor *inside* its parens, so the `;` typed next produced `a(;)` - the same symptom #3125 had from the other end, in the one arrangement that fix did not cover. A cursor stepping over an existing delimiter emits only a move, in the coordinates of the buffer before the keystroke, while the multi-cursor edit path takes a move from a cursor that inserted nothing as already accounting for the other cursors' insertions. The skip-over moves are now rebased onto the other cursors' edits before they are applied, so the stepping cursor lands one byte later when an earlier cursor inserted one (#3166).
```
sinelaw: "with `xx` above `a()` and a cursor at the end of each line" is not the reproduction: the second cursor is between the `(` and the `)` (col 3, as the issue states), not at the end of `a()`. A cursor at the end of that line has nothing to step over and the bug does not occur. Since this is the user-facing description of when the bug bites, it should say the second cursor sits inside the parens.

---

crates/fresh-editor/tests/e2e/multicursor.rs
sinelaw, line 1: Branch history, against CONTRIBUTING's "individual commits should pass `cargo check --all-targets`, `cargo fmt`, and `cargo clippy`": the four commits do not meet it.

- `7af80c220` (merge of `origin/master`): the conflict resolution in this file spliced the new test's loop into the body of `test_type_over_auto_closed_paren_with_multiple_cursors` and dropped the new test's closing brace (visible in `git show 7af80c220:crates/fresh-editor/tests/e2e/multicursor.rs`), so the test target does not compile at that commit.
- `d4ba9f427` (the fix): the filter reads `id != cursor_id && position < old_position` with `id: &CursorId` / `cursor_id: &mut CursorId` and `position: &usize` / `old_position: &mut usize`. `PartialEq<&mut B> for &A` exists, but there is no `PartialOrd<&mut B> for &A`, so `position < old_position` should not compile; the follow-up subject "compare the skip-over rebase's fields by value" reads as the fix for exactly that. Worth confirming, but if so the first commit also fails the per-commit check.
- `627c95c98` bundles that compile fix with the unrelated CHANGELOG entry, and `05f65d32` exists only to undo the merge's damage.

The PR description already flags this. Given the net diff is 3 files / 114 lines, a rebase onto `master` squashed to one commit (or fix + changelog) would satisfy the rule and leave a bisectable history; the merge commit and the restore commit carry no information worth keeping.

---

sinelaw (commented):
Reviewed by reading only (no build or test run), against `origin/master` (`147bd980`) and CONTRIBUTING.md.

**Fix logic.** The diagnosis holds: `apply_events_as_bulk_edit` applies a `MoveCursor` unshifted unless the same cursor has an `Insert` at its original position, and a skip-over has no edit, so `insert_position + 1` was applied in pre-edit coordinates. I traced `shift_skip_over_moves` through the applier for every shape `insert_char_events` emits with two cursors in either order: plain insert, auto-close (`Insert` of 2 bytes + pre-edit move, which the applier shifts itself), auto-dedent (`Delete` + `Insert` at line start), dedent skip-over (own `Delete`/`Insert` below the cursor, excluded by cursor id), selection delete followed by skip-over or insert, surround (two inserts), virtual-space gap and block pads. In each the rebased target is right, the skipping cursor's own edits are not double counted, and a cursor with no lower-position edits from others is unchanged, so the single-cursor path is unaffected. Undo/redo is unaffected (`BulkEdit` stores snapshots and old/new cursor lists). The event patterns and `CursorId` comparisons compile as written; clippy passed in CI.

**Findings** (inline):
1. The second fixture of the new test (`a()` first) cannot pass: `End` lands after the `)`, not between the parens, so there is no skip-over and the buffer becomes `a())`. The first fixture does fail on master and pass with the fix, as claimed.
2. The emitted list now mixes pre-edit (auto-close) and post-edit (skip-over) `MoveCursor`s, which is right for the bulk applier but wrong for the sequential `EditorState::apply` consumer the unit tests in `actions.rs` use. Not a production bug today; a contract note on `insert_char_events` at least.
3. CHANGELOG entry misdescribes the second cursor's position.
4. Commit hygiene: the merge commit does not compile the test target, the first commit most likely does not compile at all, and the branch needs a squash/rebase per CONTRIBUTING.

Not flagged: `assert_buffer_content` reads model state rather than the screen, but the surrounding tests do the same and the final assertions are on rendered output.

---

crates/fresh-editor/tests/e2e/multicursor.rs
sinelaw: Correct on all three counts — the second fixture was never inside the parens. Pushed in eb2cd3e: with `a()` first the test sends `Left` after `End`, asserts `Ln 1, Col 3` before adding the second cursor so the setup is checked rather than assumed, and the comment on the `)` assertion now says it is a setup guard (the buffer is identical on master) and that the `;` is the evidence. The description's "exact bytes from the tmux run" sentence applies to the first ordering only; I've left the description as is except for that, which I'll correct in the body.

---

crates/fresh-editor/src/input/actions.rs
sinelaw: Agreed that the list now has one valid consumer. Pushed in eb2cd3e: the doc comment on `insert_char_events` states that a multi-cursor list from it is meaningful only to `apply_events_as_bulk_edit` (skip-over moves post-edit, auto-close moves pre-edit), that sequential `EditorState::apply` would shift the skip-over moves twice, and that every multi-cursor dispatcher already routes such lists to the bulk applier. That is the second workaround for the applier's inference, as you say; the coordinate-space tag on `MoveCursor` remains the structural fix and this PR does not attempt it.

---

CHANGELOG.md
sinelaw: Fixed in eb2cd3e: "one cursor at the end of `xx` and one between the parens of `a()`".

---

crates/fresh-editor/tests/e2e/multicursor.rs
sinelaw: All accurate, including the `&usize < &mut usize` in the first commit (no `PartialOrd` impl for that pairing; the "compare by value" follow-up was exactly that). I have not rewritten the branch because the session's rules forbid rewriting pushed history; the net diff is the three files. Two ways to a clean history, either at the maintainer's word: squash-merge the PR, or tell me to rebase onto `master` as one commit (fix + test + changelog) and force-push this branch. Leaving this thread open for that decision.

---

PR #3181, opened 2026-09-03
Auto-indent: Enter on the blank last line of a text file keeps the cursor's column

### LINKED ISSUE #3165: Auto-indent: Enter on a blank line at the end of the file indents to the previous block
Split out from #1425. The shape reported there (a blank line with an unindented line below it) is fixed; this one — a blank line with nothing non-empty after it — still behaves the old way. #1425 is closed, so filing separately.

## Steps to reproduce

Plain text file, LF, spaces, `auto_indent` on. Five lines, where the fifth is empty and is the last line of the file:

```
    line1
    line2
        line3
        line4
```

1. Put the cursor on the empty fifth line, at `Ln 5, Col 1`.
2. Press `Enter`.
3. Type `x` and save.

## Actual

Cursor lands at `Ln 6, Col 9`. The saved file has 8 spaces before the `x`:

```
    line1
    line2
        line3
        line4

        x
```

## Expected

Cursor stays at column 1 and the file gets a plain unindented `x` — the same as when a non-empty line follows the blank line.

## Notes

- Only the plain-text path. The equivalent Python file (`def f():` / `    a = 1` / `    if a:` / `        b = 2` / blank last line) does **not** reproduce.
- Adding any non-empty line below the blank line makes it behave correctly.

sinelaw:
Fixes #3165.

**Symptom** (reproduced on master in tmux): plain-text file ending in an 8-space block and a blank last line; Enter on the blank line lands at `Ln 6, Col 9`, the saved file has `        x`.

**Cause:** `calculate_indent_pattern` copies the previous non-empty line's indent onto a blank line. The #1425 look-ahead overrides that only when a less-indented line follows; a blank last line has nothing after it.

**Fix:** a plain-text buffer (`state.language == "text"`, set from the path at open) keeps the cursor's column on a blank line — the #1425 leading-whitespace rule extended to blank lines, now one helper `leading_whitespace_column`. Column 1 of an empty line → 0; end of 8 inserted spaces → 8, so typing an indented block is unchanged.

**Decisions to confirm:**
- Plain text only. Code keeps indenting a blank line inside an unclosed block on every path: `calculate_indent` (tree-sitter + pattern fallback) and the new `calculate_indent_no_grammar` (a language with neither a bundled grammar nor regex rules — Rust in this build; CI caught it via `test_indent_after_empty_line_in_function_body`, twice: first when the rule sat on the shared fallback, then when the gate read the highlighter, which in that harness has no engine yet for `.rs`). Say the word to extend the rule to code.
- Consequence for text files: Enter at column 1 of an already blank line between two indented paragraphs gives 0, not the paragraph indent (VS Code does the same).

**Tests:** unit `test_enter_on_blank_last_line_keeps_cursor_column_no_language` (0 via `no_language`, 8 via `no_grammar`, plus end/mid-whitespace and tab cases); e2e `auto_indent::test_plain_text_enter_on_blank_last_line_does_not_indent` (status bar `Ln 5, Col 1` → `Ln 6, Col 1`, then the indented-block flow at `Ln 5, Col 9` / `Ln 6, Col 9`). Nothing run locally per maintainer instruction; CI verifies.


crates/fresh-editor/tests/e2e/auto_indent.rs
```diff
@@ -179,6 +179,54 @@ fn test_fallback_copies_previous_indent() {
...
+    harness.send_key(KeyCode::End, KeyModifiers::NONE).unwrap();
+    harness.assert_screen_contains("Ln 4, Col 14");
+    harness
+        .send_key(KeyCode::Enter, KeyModifiers::NONE)
+        .unwrap();
+    harness
+        .send_key(KeyCode::Enter, KeyModifiers::NONE)
+        .unwrap();
+    harness.type_text("y").unwrap();
+    harness.assert_buffer_content(
```
sinelaw: This assertion will panic even with the fix, so the test cannot pass in CI as written.

`harness_with_auto_indent()` calls `enable_shadow_validation()`, and `assert_buffer_content` compares the harness's shadow string against `expected` *before* it compares the real buffer. The shadow model's Enter arm (`update_shadow_for_key`, `KeyCode::Enter`) inserts a bare `'\n'` and knows nothing about auto-indent, and `send_key` never resyncs the shadow string after a key (only the cursor, and only after undo/redo). After the two Enters the shadow string is `"…line4\n\ny\n\nx"` while `expected` has the two 8-space lines the editor inserted, so this fails with "Shadow string mismatch (bug in test harness shadow tracking)" regardless of what the editor did.

That is exactly why every other test in this file only calls `assert_buffer_content` *before* an auto-indenting Enter and switches to `get_buffer_content()` + `contains` afterwards (`test_fallback_copies_previous_indent`, `test_rust_auto_indent_after_brace`, …). The first `assert_buffer_content` at line 209 only survives because the indent it exercises is empty.

Two ways out, and I'd take the first for the part of the flow it can cover, since CONTRIBUTING "E2E Tests Observe, Not Inspect" asks for rendered output:

- Observe the status bar, which is what the issue reporter saw: `Ln 5, Col 9` after the Enter at the end of `line4`, and `Ln 6, Col 9` after the second Enter — the latter is the very number the bug produced on the blank last line, so it doubles as a demonstration that the two blank-line cases are distinguished. `y` at `Col 10` on `Ln 6` then follows.
- The whitespace-only line 5 is invisible on screen, so if you want its 8 spaces pinned, do it the way the neighbours do: `assert_eq!(harness.get_buffer_content().unwrap(), "…")` with no shadow involvement — or drop `enable_shadow_validation()` for this test.

Either way the second half needs to stop going through the shadow check.

---

crates/fresh-editor-core/src/primitives/indent.rs
```diff
@@ -262,6 +262,12 @@ impl IndentCalculator {
             return indent;
         }
 
+        // A blank line keeps the cursor's own column (#3165). Not on the
+        // language path: see `cursor_column_on_blank_line`.
+        if let Some(indent) = Self::cursor_column_on_blank_line(buffer, position, tab_size) {
```
sinelaw: Placing this before `calculate_indent_pattern` makes the pattern function's whole "current line is empty" branch unreachable on the no-language path: every blank line now returns from `cursor_column_on_blank_line`, so `calculate_indent_pattern` is only ever entered here with a line that has content, where it degenerates to "current line indent, +tab_size if it ends in a trigger".

Two things follow that the PR does not say:

1. The #1425 forward look-ahead (`next_non_empty_line_indent_unless_closer`) is now dead for files with no language. Its regression test `test_enter_on_blank_line_above_unindented_uses_following_indent` still passes, but through the new rule, not through the code its comment describes; the look-ahead is left reachable only via `calculate_indent` for the few bundled C-family grammars, where no unit test drives it. Worth either a test on that path or a note in that test's comment, so the next person does not "fix" the look-ahead by reading a test that no longer exercises it.
2. The doc comments are now stale in three places and should be updated in this change: the `# Scope` section of `calculate_indent_pattern` (item 1 says it is "the heuristic used by `calculate_indent_no_language`" for blank lines), the tail of `indent_for_cursor_in_leading_ws` ("when the line has no content … the regular smart-indent logic takes over"), and the module doc ("uses the C-family pattern heuristic directly").

---

crates/fresh-editor-core/src/primitives/indent.rs
```diff
@@ -271,6 +277,72 @@ impl IndentCalculator {
...
+    /// file landed at column 9 under an 8-space block (#3165). In prose a
+    /// blank line ends what came before it; the previous block's indent is
+    /// not wanted, whether or not anything follows.
+    ///
+    /// Deliberately **not** applied on the language path
+    /// ([`calculate_indent`]): there the same backward scan is the fallback
+    /// for code tree-sitter cannot parse yet, where a blank line at the end
+    /// of an unclosed block *should* keep the block's indent
+    /// (`test_indent_after_empty_line_incomplete_syntax`).
+    fn cursor_column_on_blank_line(
```
sinelaw: This is a near-verbatim copy of `indent_for_cursor_in_leading_ws` directly below it: the line-start scan and the "everything before the cursor is whitespace, accumulate `col`" loop are identical, and only the final loop differs — that one returns `Some(col)` on seeing content, this one returns `None`. The two are exact complements of the same predicate, "is the cursor inside leading whitespace", split on whether content follows.

And in the only call sequence that exists, the second loop here (lines 334-342) can never return `None`: `indent_for_cursor_in_leading_ws` has already returned `None` for this position, which — given the first loop passed — means there is no content after the cursor. So the ~15 lines of after-cursor scanning are redundant at the one call site.

The simpler shape is a single helper returning the column when everything before the cursor on the line is whitespace (i.e. the shared first loop), with `indent_for_cursor_in_leading_ws` expressed as "that, if content follows" for the language path, and `calculate_indent_no_language` using the unconditional form. That states the actual rule this PR adopts for plain text in one sentence — "in leading whitespace, the new line gets the cursor's column, blank line or not" — and removes the copy-paste. CONTRIBUTING #12 ("when a bug recurs, consider centralizing") is about exactly this: #1425 and #3165 are the same rule reached from two sides.

Separately, the doc comment's "In prose a blank line ends what came before it" argument is fine, but the header still says this is *extending* the leading-whitespace rule while the code is a duplicate rather than an extension; whichever way you go, doc and code should match.

---

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
...
+* **Enter on the blank last line of a text file no longer indents to the block above** - in a plain-text file ending in an indented block and one empty line, Enter on that empty line put the cursor at `Ln 6, Col 9` and the next character eight spaces in. The fallback indenter for files without a language scanned backwards past the blank line and copied the previous block's indent; the look-ahead that stops it when a less-indented line follows (#1425) has nothing to look at when the blank line is the last one. On a blank line the new line now keeps the cursor's own column - none at column 1, the full indent when Enter is pressed at the end of the whitespace a previous Enter inserted, so typing an indented block is unchanged. Only files with no language are affected; code that tree-sitter cannot yet parse keeps indenting a blank line at the end of an unclosed block (#3165).
```
sinelaw: "Only files with no language are affected" is narrower than what the code does. `handle_insert_newline` reaches `calculate_indent_no_language` whenever the buffer's language has no *bundled* tree-sitter grammar **and** `rules_for_syntax_name` finds no regex family. That includes:

- `Language::HTML` (has an `indents.scm` but no bundled grammar — `ts_language()` is `None` — and no entry in `family_for_id`), and
- every syntect-only syntax the `Language` enum does not cover: YAML, TOML, Markdown, XML, SQL, Makefile, Dockerfile, INI, …

The PR description acknowledges this ("files whose language has neither a tree-sitter grammar nor regex indent rules … get the new behaviour too"); the user-facing entry should say the same rather than "no language". Something like "files with no language and languages without indent rules (HTML, YAML, Markdown, …)".

One consequence of the rule that neither the entry nor the description mentions, and that YAML users in particular will notice: on those files, Enter on a *pre-existing* blank line whose previous non-blank line ends in `:`/`{`/`[`/`(` used to indent by `tab_size` (the pattern indenter's backward scan added the trigger's delta); it now gives column 0. Typing `key:` then Enter is unchanged, since that line has content. If that is intended it is worth a clause here; if not, the new rule needs a carve-out.

---

sinelaw (commented):
Reviewed by reading (nothing built or run), against master and CONTRIBUTING.md.

**Verified as correct:** the diagnosis (on master, `calculate_indent_pattern`'s backward scan returns 8 for the reported fixture because the #1425 look-ahead is `None` at end-of-buffer); the unit test's first, third and fourth assertions fail on master and the second passes on master by design; the e2e's `Ln 4, Col 14` / `Ln 5, Col 1` / `Ln 6, Col 1` and both expected buffer strings match what the editor would produce; the language path is untouched and `test_indent_after_empty_line_incomplete_syntax` is unaffected; no existing unit or e2e expectation on the no-language path is contradicted; CRLF handling; single well-motivated commit with a CHANGELOG entry.

**Findings, in priority order:**

1. **The e2e test cannot pass in CI.** `harness_with_auto_indent()` turns on shadow validation, the shadow model inserts a bare `\n` on Enter, and `assert_buffer_content` compares the shadow string first — so the second `assert_buffer_content` (two editor-inserted 8-space lines) panics with "Shadow string mismatch" even with the fix. The neighbouring tests avoid `assert_buffer_content` after an auto-indenting Enter for this reason. Suggested screen-based assertions inline.
2. `cursor_column_on_blank_line` runs ahead of `calculate_indent_pattern`, which makes the pattern indenter's blank-line branch — including the #1425 look-ahead — unreachable on the no-language path; its regression test now passes through the new rule instead, and three doc comments are stale.
3. `cursor_column_on_blank_line` duplicates `indent_for_cursor_in_leading_ws` almost line for line and its after-cursor loop is redundant at the only call site; the two are complements of one predicate and should be one helper.
4. The CHANGELOG's "only files with no language" understates the reach (HTML and all syntect-only syntaxes — YAML, TOML, Markdown, … — take the same path), and the entry does not mention that Enter on a pre-existing blank line after a `:`/`{`-terminated line no longer adds a level in those files.

---

crates/fresh-editor/tests/e2e/auto_indent.rs
sinelaw: Right — that assertion would have failed on the shadow check regardless of the editor. Pushed: the test now observes the status bar (`Ln 5, Col 9` after the first Enter at the end of `line4`, `Ln 6, Col 9` after the second, `Ln 6, Col 10` after `y`) and compares `get_buffer_content()` directly for the whitespace-only line, with a comment saying why `assert_buffer_content` is avoided there.

---

crates/fresh-editor-core/src/primitives/indent.rs
sinelaw: Pushed: the three doc comments now say the no-language caller only reaches the pattern heuristic with content before the cursor, so its empty-line branch and the #1425 look-ahead are exercised only through `calculate_indent`; the `test_enter_on_blank_line_above_unindented_uses_following_indent` comment says which rule answers it now and where the look-ahead stays live. I did not add a language-path test for the look-ahead in this PR: it would need a fixture that tree-sitter fails on in a specific way, and per the maintainer's instruction nothing is run locally here, so I would be guessing at what the parser does.

---

crates/fresh-editor-core/src/primitives/indent.rs
sinelaw: Agreed, and done: one `leading_whitespace_column` returns the column plus whether content follows; `indent_for_cursor_in_leading_ws` is that filtered to content-follows (language path), and `calculate_indent_no_language` takes the column either way. The duplicate is gone and the rule is one sentence in the code comment.

---

CHANGELOG.md
sinelaw: Both points taken into the entry: it now names the languages that share the path (HTML, YAML, TOML, Markdown, XML, SQL, Makefile, …) and states the consequence — Enter at column 1 of an *already blank* line under a `:`/`{` line no longer adds a level, while Enter at the end of the `key:` line still does. That is intended (it is what VS Code does on those files too) and is part of the decision the PR asks the maintainer to confirm; a carve-out for trigger lines is a few lines if it is vetoed.

---

sinelaw:
CI on 664fbc0 failed `e2e::auto_indent::test_indent_after_empty_line_in_function_body` on all three platforms, and it was this PR's: a `.rs` file, Enter on an empty line inside an unclosed `fn` body, expected to indent to 4. Rust has no bundled tree-sitter grammar in this build, and under a plugin grammar the regex-rules tier is skipped as well (`rules_for_syntax_name` keys on the syntect syntax name, which a tree-sitter engine does not expose), so a Rust buffer reached `calculate_indent_no_language` — the blank-line rule was being applied to code.

Fixed in 3fbcf92 by gating on the buffer, not the function: `calculate_indent_no_grammar` is the old fallback and is used whenever the buffer has a language, even one Fresh has no grammar or rules for; `calculate_indent_no_language` (old fallback plus the blank-line rule) is used only when `highlighter.language()` is `None` — plain text and unknown extensions. Unit tests now pin both: the #3165 fixture gives 0 through the first and 8 through the second. The changelog no longer claims HTML/YAML/Markdown get the new behaviour; they do not, and neither does any code file. `origin/master` is merged in as well.

---

sinelaw:
`test_indent_after_empty_line_in_function_body` failed again on 521a061, still this PR's: the gate read `highlighter.language()`, and in that harness the `.rs` buffer has no highlighter engine when the keys arrive, so it was taken for plain text. cfc2291 gates on `state.language == "text"` instead — the id set from the path at open, which is what the auto-close code in the same file already keys on.

---

PR #3182, opened 2026-09-03
A plain mouse click collapses multiple cursors to one

### LINKED ISSUE #3125: [Bug] Inconsistent multi cursor input behavior

sinelaw:
Fixes #3125 — the mouse-click half, which #3158 left open on purpose. The type-over half was fixed there and its remaining arrangement is #3180.

## Symptom

Reproduced first on master, driven interactively in tmux with real keystrokes and an SGR mouse click: `foo bar` / `foo baz` / `foo qux`, `Ctrl+Alt+Down` (status bar `2 cursors`), click on line 3 after `foo`, type `Z`, `Ctrl+S`.

| after | status bar | file |
|---|---|---|
| click (master) | `Ln 3, Col 4` · **`2 cursors`** | |
| `Z` (master) | | `Zfoo bar` / `foo baz` / `fooZ qux` — two edits for one keystroke |
| `Z` (this branch) | `Ln 3, Col 4`, no cursor count | `foo bar` / `foo baz` / `fooZ qux` |

## What Fresh's clicks already meant, and what changes

`handle_editor_click` (`click_handlers.rs`) treats Shift-click and Ctrl-click as "extend the primary's selection to the click" (Ctrl because some terminals intercept Shift) and every other left click as "move the primary to the click". Drag starts from the click; double- and triple-click (word / line) have their own handlers and pass through this one on their first press. There is no click gesture that *adds* a cursor — Alt-click is not bound — so the only thing to preserve is the extend gesture.

**Decision to confirm:** a plain click now removes every other cursor and moves the survivor to the click; Shift/Ctrl-click still extends the selection and leaves the cursor set alone. This is the universal editor behaviour (VS Code, Sublime, Zed). The issue is labelled `discussion`, so: veto if you want it configurable, or if a modifier-click should keep the multi-cursor set rather than only the extend gesture. Alt-click, which means nothing to Fresh today, collapses like a plain click; it can be carved out in one line if a cursor-adding click is ever bound to it.

Two details, both following `Esc` (`RemoveSecondaryCursors`) so the two ways of leaving multi-cursor mode behave alike:

- **The survivor is the lowest-id cursor, not the primary.** The add-cursor commands allocate the next id from the cursor count, so a lone survivor with a higher id would be overwritten by the next `Ctrl+Alt+Down` (found in review). After the collapse it is the only cursor, so the status bar reports the click either way.
- **A collapsing click is an undo step.** The removals and the move go into one `Event::Batch`; removing a cursor is a write action, so `Ctrl+Z` after the click brings the cursors back, and the click truncates redo history — a plain `MoveCursor` did neither. Stated in the CHANGELOG. If you would rather the click not enter the undo stack, the removals can be applied unlogged, at the cost of making them unrecoverable.

## Verification

- `e2e::multicursor::test_plain_click_collapses_multiple_cursors`: real key events plus `mouse_shift_click` / `mouse_click`. Asserts `2 cursors` after the add, that a Shift-click keeps `2 cursors` and lands at `Ln 2, Col 4`, that a plain click leaves no cursor count on the status bar and lands at `Ln 3, Col 4`, that one typed character then changes exactly one line, and that a further `Ctrl+Alt+Down` yields `2 cursors` (the survivor is a normal cursor). The typed-character assertion is the file the tmux run above produced on master with the extra `Z`, so the test cannot pass there.
- Per the maintainer's instruction in this session, nothing was built or run locally after the master reproduction: the test's passing direction, the `multicursor` / `mouse` / `selection` suites, clippy and the unit tests are left to CI. `cargo fmt` is clean.
- An independent read-only review posted six inline findings; all are addressed in the second commit and answered on their threads.


crates/fresh-editor/src/app/click_handlers.rs
```diff
@@ -656,9 +656,50 @@ impl Editor {
...
+                .get(&self.active_window)
+                .and_then(|w| w.buffers.splits())
+                .map(|(_, vs)| vs)
+                .expect("active window must have a populated split layout")
+                .get(&split_id)
+                .map(|vs| vs.cursors.clone());
+            if let Some(cursors) = cursors {
+                for (cursor_id, cursor) in cursors.iter() {
+                    if cursor_id != primary_cursor_id {
+                        events.push(Event::RemoveCursor {
```
sinelaw: **Bug: keeping the primary (not the lowest id) makes the next `Ctrl+Alt+Down` / `Ctrl+Alt+Up` / `Ctrl+D` overwrite the surviving cursor.**

`add_cursor_below`, `add_cursor_above` and `add_cursor_at_next_match` in `app/clipboard.rs` all allocate the new id as `CursorId(self.active_cursors().count())` (lines 1560, 1611, 1642 on master), and `Cursors::insert_with_id` is a plain `HashMap::insert` with no collision check. `Action::RemoveSecondaryCursors` keeps the *lowest* id, so on master the survivor is always `CursorId(0)` and `count()` never collides. This PR keeps the *primary*, which after `Ctrl+Alt+Down` is the newly added cursor (`Cursors::add`/`insert_with_id` set `primary_id = id`).

Concrete sequence on this branch, from a fresh buffer: `Ctrl+Alt+Down` (cursors `{0, 1}`, primary 1) -> plain click (this code removes 0, keeps 1; `count() == 1`) -> `Ctrl+Alt+Down` again. The new cursor gets `CursorId(1)`, `insert_with_id(1, ..)` replaces the existing cursor 1, and the user ends up with a single cursor that has jumped one line down while the status message reports 2 cursors. The logged `AddCursor{1}` can't be undone either: its inverse `RemoveCursor{1}` hits the `len() <= 1` guard in `Cursors::remove` and is a no-op, so the original cursor position is gone.

Two ways out, either is fine; the second is the smaller change:
1. Keep the min-id cursor like `remove_secondary_cursors` does (emit `RemoveCursor` for every other id, and target the `MoveCursor` at the min id with *that* cursor's old position/anchor/sticky column). The status bar still reports the click, because after the removals the min-id cursor is the only one left and `Cursors::remove` re-picks the primary.
2. Fix the three allocators in `clipboard.rs` to `max(id) + 1`, which `add_cursors_to_line_ends` a few lines below already does with the comment "so we never collide with a cursor an undo could re-insert later".

Either way, a test with a second `Ctrl+Alt+Down` after the click would have caught this (see the comment on the test).

---

crates/fresh-editor/src/app/click_handlers.rs
```diff
@@ -656,9 +656,50 @@ impl Editor {
...
         };
 
+        // A plain click ends multi-cursor editing: every secondary cursor
+        // goes, and the primary is what moves to the click — as in VS Code,
+        // Sublime and Zed. Shift/Ctrl-click extends the primary's selection
+        // and leaves the set alone, so a selection built across several
+        // cursors is not lost to the click that extends it. Before this the
+        // click moved the primary and kept the rest, which may be out of
+        // the viewport, and the next keystroke edited every one of them
+        // (#3125). The removals and the move are one `Batch` so undo steps
```
sinelaw: **Undo/redo behaviour change the description does not state: a collapsing click is now a write action.**

`Event::is_write_action` is true for `RemoveCursor`, and therefore for this `Batch`. The lone `MoveCursor` a click used to log was read-only. Two consequences, both user-visible:

1. `EventLog::append` truncates redo history on a write action, and deliberately *skips logging* non-write events so that "navigation after undo does not destroy the redo chain" (its own comment, citing VS Code/Sublime). So: multi-cursor type `x`, `Ctrl+Z`, click elsewhere -> `Ctrl+Y` now does nothing. On master that click preserved redo.
2. `EventLog::undo` stops at the first write action, so `Ctrl+Z` after a collapsing click first restores the cursor set (and moves the primary back) instead of undoing the last edit. In the three editors the comment cites, a cursor-set change is never an undo step.

`Escape`/`RemoveSecondaryCursors` already has both properties today, so this is at least consistent, and maybe that is the answer. But the description only says "the event log sees the click as one step", and the CHANGELOG says nothing, so please either state it explicitly (description + CHANGELOG) or decide the click should not enter the undo stack.

Two smaller points on the "undo steps back over the click as a whole" claim: the inverse re-adds each secondary through `apply_add_cursor` -> `insert_with_id`, which makes the *last re-added* cursor the primary, so after undo the status bar reports a former secondary, not the cursor that was primary before the click; and since `cursors.iter()` is `HashMap` order, which secondary that is varies run to run (see the next comment).

---

crates/fresh-editor/src/app/click_handlers.rs
```diff
@@ -656,9 +656,50 @@ impl Editor {
...
+            let cursors = self
+                .windows
+                .get(&self.active_window)
+                .and_then(|w| w.buffers.splits())
+                .map(|(_, vs)| vs)
+                .expect("active window must have a populated split layout")
+                .get(&split_id)
+                .map(|vs| vs.cursors.clone());
+            if let Some(cursors) = cursors {
+                for (cursor_id, cursor) in cursors.iter() {
```
sinelaw: `Cursors::iter` walks a `HashMap`, so the order of the `RemoveCursor` events inside the batch is unspecified. That makes the logged/streamed event sequence for one click non-reproducible, and (via `inverse()`, which reverses the batch) makes which secondary ends up primary after an undo non-deterministic. Sorting by `cursor_id.0` before pushing costs nothing here and removes both. (`remove_secondary_cursors` has the same flaw; no need to fix it in this PR, but no need to copy it either.)

---

crates/fresh-editor/src/app/click_handlers.rs
```diff
@@ -656,9 +656,50 @@ impl Editor {
...
+        let mut events: Vec<Event> = Vec::new();
+        if !extend_selection {
+            let cursors = self
+                .windows
+                .get(&self.active_window)
+                .and_then(|w| w.buffers.splits())
+                .map(|(_, vs)| vs)
+                .expect("active window must have a populated split layout")
+                .get(&split_id)
+                .map(|vs| vs.cursors.clone());
```
sinelaw: Style: this is a fourth `windows.get(..).splits()..get(&split_id)` chain in the function and it clones the entire `Cursors` map only to iterate it once. The `cursor_snapshot` closure ~150 lines above already holds `vs.cursors` and reads the primary from it; collecting the secondary `(id, position, anchor)` triples there (or returning `vs.cursors.iter().filter(..).map(..).collect::<Vec<_>>()` from this closure instead of `vs.cursors.clone()`) avoids the clone and the repeated lookup. Likewise `move_event.clone()` on both branches below: push `move_event.clone()` only in the batch branch and move `move_event` itself into the other.

---

crates/fresh-editor/src/app/click_handlers.rs
sinelaw, line 1: The module doc (line 9-10) still says `handle_editor_click` dispatches to "cursor placement / multi-cursor add depending on modifiers". The PR's own premise, which I verified (no `ALT` handling anywhere in `mouse_input.rs`, `chrome/splits.rs` or this file), is that no click gesture adds a cursor. Since this PR redefines what a click does to the cursor set, that line should say so instead ("place the caret and collapse to one cursor; Shift/Ctrl extend the selection").

---

crates/fresh-editor/tests/e2e/multicursor.rs
```diff
@@ -2257,3 +2257,46 @@ fn test_type_over_auto_closed_paren_with_multiple_cursors() {
...
+    assert!(
+        !screen.contains("cursors"),
+        "expected a single cursor after a plain click, got:\n{screen}"
+    );
+    harness.assert_screen_contains("Ln 3, Col 4");
+
+    // Which is what makes the next keystroke land in exactly one place.
+    harness.type_text("Z").unwrap();
+    harness.assert_buffer_content("foo bar\nfoo baz\nfooZ qux\n");
+    harness.assert_screen_contains("fooZ qux");
```
sinelaw, line 2361: The test does prove the fix as far as it goes: I traced it on master (primary is the added cursor 1 on line 2, cursor 0 stays at byte 0, `Z` lands on lines 1 and 3), so the `!screen.contains("cursors")` assertion fails there, the earlier `"2 cursors"` check keeps it from passing vacuously on a narrow status bar, `find_text_on_screen` returns `(col, row)` in the order `mouse_shift_click`/`mouse_click` take, and `detect_multi_click` keys on position as well as time, so the second click (different row) is a single click, not a double.

What it does not cover is what happens to the surviving cursor afterwards, and that is where the id-collision bug in `click_handlers.rs` hides. Please extend it with one more step after this line:

```rust
// The survivor must still be a normal cursor: adding below it works.
harness
    .send_key(KeyCode::Down, KeyModifiers::CONTROL | KeyModifiers::ALT)
    .unwrap();
harness.assert_screen_contains("2 cursors");
```

On this branch that fails (the new cursor overwrites the survivor, count stays 1). A `Ctrl+Z` step asserting `"2 cursors"` comes back and `Ln 2, Col 4` is reported would also pin down the undo semantics the description claims.

---

sinelaw (commented):
Review by reading only (no build/test run), against CONTRIBUTING.md.

**Verified as correct:** the collapse only triggers on the caret-placement path (fold-toggle, `onClick`, widget/terminal/composite/non-scrollable panels all return earlier, unchanged); `Batch` is applied recursively by `EditorState::apply`, `collect_lsp_changes`, `adjust_other_split_cursors_for_event` (only Insert/Delete matter there) and `trigger_plugin_hooks_for_event` (so `cursor_moved` still fires); `track_cursor_movement` receives the same `MoveCursor` as before, so position history is unchanged; `modifies_buffer()` is false for the batch, so no preview promotion, no dirty flag, no LSP `didChange`; Shift/Ctrl-click, drag start, virtual-line clicks and the first press of a double/triple click all behave as the description says; the CHANGELOG's `Ctrl+D`/`Ctrl+Alt+Down` bindings and the reporter attribution are accurate; the product decision (plain click collapses, Shift/Ctrl-click extends and keeps the set, Alt-click collapses) is stated honestly. The regression test observes screen + buffer only, cannot pass vacuously, and fails on master for the reason claimed.

**Findings (inline):**
1. Bug: keeping the *primary* rather than the lowest id lets the next `Ctrl+Alt+Down`/`Up`/`Ctrl+D` allocate `CursorId(count())`, which now collides with the survivor and silently overwrites it (undo cannot recover it). Fix either by keeping the min id like `RemoveSecondaryCursors`, or by fixing the three allocators in `clipboard.rs` to `max+1`.
2. Undocumented behaviour change: the click batch is a write action, so it truncates redo history after an undo and becomes its own undo step. Consistent with `Escape` today, but the description/CHANGELOG should say so or the design should avoid it.
3. Nondeterministic `RemoveCursor` order (HashMap) in the logged batch.
4. Style: extra split lookup + full `Cursors` clone; stale module doc.
5. Test: add a second `Ctrl+Alt+Down` after the click (fails on this branch) and an undo step.

---

crates/fresh-editor/src/app/click_handlers.rs
sinelaw: Confirmed from the allocators in `clipboard.rs` and `insert_with_id`; a real bug. Pushed option 1: the survivor is the lowest id, as `RemoveSecondaryCursors` keeps it, and the `MoveCursor` targets that cursor with its own old position/anchor/sticky column. The test now adds a cursor after the click and asserts `2 cursors`, which the previous commit would have failed.

---

crates/fresh-editor/src/app/click_handlers.rs
sinelaw: Stated explicitly now, in the code comment on the batch, the CHANGELOG entry ("Like `Esc`, the collapse is an undo step: `Ctrl+Z` after the click brings the cursors back") and the PR description. I kept the click in the undo stack rather than applying the removals unlogged: consistency with `Esc` seemed worth more than preserving redo across a collapsing click, and an unlogged removal would make the cursors unrecoverable. That is part of the decision the PR asks the maintainer to confirm. Which cursor is primary after an undo is now deterministic (removals sorted by id, see the next thread).

---

crates/fresh-editor/src/app/click_handlers.rs
sinelaw: Done: the removals are sorted by id before they go into the batch.

---

crates/fresh-editor/src/app/click_handlers.rs
sinelaw: Done: the removed cursors are collected inside the existing `cursor_snapshot` closure (no clone, no fourth lookup), `extend_selection` is computed once above it, and `move_event` is moved rather than cloned, with position history taking the move back out of the batch.

---

crates/fresh-editor/src/app/click_handlers.rs
sinelaw: Updated: the module doc now says a plain click places the caret and collapses to one cursor, Shift/Ctrl extend the selection, and no click adds a cursor.

---

crates/fresh-editor/tests/e2e/multicursor.rs
sinelaw: Added the `Ctrl+Alt+Down` step with the `2 cursors` assertion. I did not add the `Ctrl+Z` step: the undo chain at that point also holds the typed `Z` and the added cursor, and with nothing run locally in this session I would be guessing at the exact sequence; the undo semantics are stated in the description and changelog instead.

---

PR #3183, opened 2026-09-03
Retained-mode UI: the design, focus and keys as the tree's, no panel on the buffer's scroll, and the web's panels from the display list

### LINKED ISSUE #3176: Retained mode: finish stage 3.2 — focus is described, not written (one authority per fact)
## The end goal

`docs/internal/fresh-editor-retained-mode-plan.md` §2 principle 4:

> **One authority per fact** — not two writers and a gate deciding which counts.

and `docs/internal/fresh-editor-plugin-widgets-end-state.md` §1:

> the end state is not "delete the runtime": it is **one behaviour authority and three projections**, with the *node* projection as the only one that answers the pointer, the keyboard and the geometry.
>
> What must die is not the collector but its **second life as a source of truth**.

This issue is about reaching that for **focus** specifically, because the migration is currently stopped in a state where focus has *two writers and no gate* — and that has already shipped one user-visible regression (#3137), whose fix (#3163) makes the two writers converge rather than removing one.

**Concretely, the end state is:** a panel's focused widget is a property of the description. A plugin says *"this node is focused"* in its spec; the tree settles focus onto it whenever that changes; `WidgetPanelState::focus_key` is a read-model, not a writer; and nothing anywhere calls "focus widget X" imperatively. When that holds, an entire class of bug becomes **unrepresentable** rather than fixed.

---

## Where it actually stands

Stage 3.2's own progress note is honest that it is partway:

> the plugin widgets are on the tree's ring, which leaves the registry's focused-key string a mirror written from one place rather than the authority on which control is live.

But `focus_key` is not written from one place. Per the end-state doc's own table it has **three** writers — the collector, `handle_widget_focus_advance`, and `UiFact::WidgetFocus` — plus `WidgetMutation::SetFocusKey` from plugin IPC. And the imperative counterpart, `Editor::focus_panel_widget_in_tree` (`app/widget_runtime.rs:1418`), writes the *other* direction.

So "which widget holds this panel's focus" currently has two authorities that can disagree, and **nothing decides which wins except frame ordering**.

### The gap the code already names

`on_the_ring` (`view/shell/widgets.rs:588`) marks `autofocus()` on the widget the panel says is focused — the declarative path — and then states the limit outright at line 617:

> A focus move made while the panel is already focused is the other moment, and **no description can express it** — `apply_autofocus` leaves focus alone once it is inside the scope — so that one is an imperative write, `Editor::focus_panel_widget_in_tree`.

That is exactly right, and `fresh-ui/src/focus/mod.rs:529` confirms it:

```rust
pub(crate) fn apply_autofocus(&mut self) {
    let alive = self.focus.is_some_and(|f| self.arena.get(f).is_some());
    if alive && self.focus.is_some_and(|f| self.in_active_scope(f)) {
        return;                        // focus already inside the scope: nothing happens
    }
```

Its doc lists three cases — *focus already inside the scope and nothing happens; a scope has just opened; the scope has closed*. **There is no fourth case for "the scope's wanted target changed while focus was inside it."** That missing case is the whole of the imperative escape hatch.

### What it cost (#3137)

Opening a dock dropdown is two plugin writes — a spec carrying the option rows, then `setFocusKey` onto one of them — both applied before the frame that describes either. So `focus_panel_widget_in_tree` asked the tree for an element that did not exist yet:

```
focus_panel_widget_in_tree: PROBE widget=menu-pick:move:root holds=true found=false
```

It silently did nothing, the tree kept the focus the dropdown was meant to take, and the *other* direction of the mirror wrote that disagreement back over the registry:

```
set_panel_focus_and_notify: firing `focus` widget_event  old=menu-pick:move:root new=sessions
dispatch_floating_widget_key: decision code=Down focus_key=Some("sessions") outcome=SmartKey("Down")
```

User-visible result: `↓` in "Move to Folder…" dismissed the popup, moved the dock selection and live-switched the workspace. Bisected to `8aba3b95a` — the commit that moved the authority to the tree and wired only one direction of the mirror. It is not in any release; it existed only on master.

This is also the migration hazard the plan itself warns about (§2):

> The question when migrating a surface is never "does it look the same" but "what did the old code *guarantee* that the description does not say".

The old code guaranteed a plugin could name a widget that was **about to** exist. The description cannot say that.

### What #3163 did, and why it is not this

#3163 holds the failed move in `Editor::pending_panel_tree_focus` (`app/mod.rs:1213`) and replays it after the frame that builds the element. It works and it is tested — but it keeps both writers and adds a **third** piece of focus state to make them converge. It is a convergence patch, deliberately sized not to pre-empt the architecture. It should be **deleted** by the work below, not built on.

---

## What "done" looks like

1. **A focus scope can express a changed target.** Add the missing `apply_autofocus` case: when the scope's `autofocus`-marked node is not the focused one *and the mark has changed since it settled*, re-settle onto it. (Generation counter on the scope, or comparing the marked key against the focused key — design call.) This is the one library change everything else depends on.
2. **`SetFocusKey` becomes a spec property, not a mutation.** The plugin's declared focus rides in the spec `on_the_ring` already reads; the IPC mutation stops writing panel state directly.
3. **Delete `focus_panel_widget_in_tree` and `pending_panel_tree_focus`.** With (1) and (2) there is no imperative focus write left to defer.
4. **`focus_key` becomes read-only.** It stays as the read-model `router::WidgetPanelView` and the description consult; the collector, `handle_widget_focus_advance` and `UiFact::WidgetFocus` stop writing it. This retires one of the six `WidgetPanelState` fields the end-state doc marks for death.
5. **The router stops string-matching one plugin's key conventions.** `input/router.rs:222` is explicit:

   > ACKNOWLEDGED RESIDUE … this dock branch hardcodes ONE plugin's widget-key conventions ("filter", "sessions", "project-pick:"/"menu-pick:" prefixes, `dock_menu_*` events) — orchestrator-specific panel semantics in host dispatch.

   In the end state the dropdown node owns its own keys via `on_key` and the host never asks "does a dropdown own `↓`". This is `end-state` §2.1's *"the node projection as the only one that answers … the keyboard"*.

## Exit condition

- No caller anywhere names a widget to focus imperatively.
- `WidgetPanelState::focus_key` has zero writers outside the tree's own settle.
- `router.rs`'s dock branch carries no plugin-specific key prefixes.
- The regression test added in #3163 (`a_focus_onto_a_widget_the_next_frame_builds_lands_on_that_frame`) still passes with `pending_panel_tree_focus` deleted — because the description now carries what the imperative write used to.

## Why it is worth doing rather than living with the patch

The dock has three inline dropdowns ("New Task… ▾", the project scope picker, Move-to-Folder) and they all funnel through the same `openDockMenu`, so they all depended on that one focus landing. Any future surface that grows a widget and focuses it in the same turn re-enters this. The patch makes the current instances work; only the declarative form makes the next one impossible.

---

Follow-up to #3163 (which fixes the symptom) and #3137 (the regression it caused). Related plan docs: `fresh-editor-retained-mode-plan.md` §3.2, `fresh-editor-plugin-widgets-end-state.md` §1.4/§2.1.

sinelaw:
## Objective

Make the UI a **retained-mode tree**: describe the screen once, and let the thing that laid it out answer every question about it. Closes #3176.

Today's bugs in this area are almost all one bug — two sources of truth for where something is, disagreeing by a frame. This PR removes the second source rather than synchronising it better.

## Before → after

| | Before | After |
|---|---|---|
| **Geometry** | painters drew *and* recorded (split areas, tab layouts, menu layouts, scrollbar thumbs, a window layout cache); everything else read those records back | the tree lays out; geometry is a query on it by key. No painter-side records |
| **Focus** | a `FocusManager` plus a ring walked over the spec, reconciled with the description after the fact | the description *is* the focus fact; the tree is its projection, settled on the frame that changes it |
| **Keys** | a pipeline with a tail that caught whatever nothing claimed | exactly one owner per key, named by the tree; a surface that declines hands the key on itself |
| **Pointer** | hit-test tables of byte ranges over a text projection of each panel | a press lands on the node it lands on |
| **The text pane** | drawn by a painter the rest of the UI then queried | a host leaf whose content pass settles the facts its leaf, the pointer, the popups and the web all read |
| **Painting** | two bands, several painters, passes that rewrote cells others had written | one fold, one caret; nothing runs after it but effects over finished cells |

Data flows one way: **model → description → layout → pane content pass → paint → fold to cells/DOM.**

## Scope

232 files, ≈ +25k / −44k. **The net deletion is the point** — this replaces mechanisms, it does not add a layer beside them.

Eight commits, grouped by concern, each building on its own:

1. the design document (and fifteen superseded plan/review docs deleted)
2. **S1+S2** focus and keys
3. **S3+S6+S5** the described surfaces — panels, menus, settings, file browser
4. **S7** the pane as a host leaf, one paint, and a provenance gate
5. **L15 + L12** two library rules
6. **L5** a text run's caret, selection and rows stated in bytes of its own string
7. master's #3194 `focusFollowsCursor`, ported onto the tree
8. the welcome page as a described pane — what a round of manual use found after the suite was green

## Duplicate mechanisms

**Retired** — no longer a source of truth anywhere: `FocusManager` and the spec-side focus ring · the layer stack · `WindowLayoutCache` and every reader · the painter's tab-strip, scrollbar and separator records · the panel hit-test tables · the key pipeline tail · `view/controls/` · `FileBrowserRenderer`/`Layout` · `MenuRenderer`/`MenuLayout` · `render_settings` · the second grid painter · the two-band render · the post-paint cell tints · `presents_blocking_overlay` · the split-level `view_mode`/`compose_width` in a saved session.

**Still present, deliberately, with the condition for removing each:**

| Leftover | Why it is still here |
|---|---|
| `painted` / `boxes` and the text projection | now only for two panel shapes: an *anchored* floating panel, and a spec containing a markdown document. Every other panel is described |
| the markdown document view | still an immediate-mode render per frame, over a shadow wrap of its own text — the last one on this path (S4) |
| `Panel::anchored_width`, `content_cols`, `Interior::avail_height` | the anchored panel's interior is built *at* a width instead of stating one; they go when it does |
| the gutter | still painter-side. Its shape is not what the plan assumed — its rows depend on view lines produced *after* layout, so it becomes its own leaf beside the content leaf, not runs |
| the scene's region views (web) | retire as each surface crosses (S9) |

**No new caching.** The facts a frame settles are re-settled by the next frame that draws them and cleared for what it does not — that is the whole invalidation story. The one memo (marks resolved to rows) replaces an older cache with a narrower key.

## Remaining gaps / next steps

- **S4** — the markdown document view becomes one wrapped run in a viewport. The library half is landed; what is owed is editor-side. This also frees the width parameter that blocks the anchored panel's `Auto`.
- **S3** — the kinds' keys from the node, and `painted`/`boxes` with them, once S4 removes the last exception.
- **S7** — the gutter as its own leaf.
- **S8** — theme and performance: retire `Paint::Lit`, resolve cache, benchmark.
- **S9** — the web fold for the remaining surfaces.

**Known cost to take up next.** Every routed key marks the description stale, so the next key lays the whole frame out (~5 ms at 80×24, debug). Fine for typing, and it wants the description's inputs versioned so a content-only change skips layout. It is not purely theoretical: on `markdown_compose_scrollbar_markers::heading_marks_survive_exploring_the_document`, which sends six page keys and settles after each, this branch measures **~18 % slower than master** (2.86 / 3.01 / 3.15 s against 2.55 / 2.49 / 2.54 s). Small in absolute terms, but that test waits on screen *stability* through an unbounded wait, and it has hit CI's 180 s cap on a loaded macOS runner before this branch existed — so the margin it was already short of is now shorter. Worth fixing at the cost, not by widening the test.

## Testing

- CI green across the matrix: three OS test suites, clippy, fmt, doc, schema, scene parity, the web playwright suite, musl, and the plugins-disabled build.
- A **provenance gate** runs on every frame the e2e corpus draws: it fails if any painter-written cell lands outside a host leaf's rectangle. Its first full run found five real defects, each fixed at its cause; it is now a standing invariant rather than a one-off audit.
- Unit: `fresh-editor` 2464, `fresh-editor-core` 1310, plus all of `fresh-ui`.
- Behavioural areas driven by hand as well as by test, because cell-identical output is necessary and not sufficient: the dock's keyboard, terminal mode and its PTY gate, tab drag, and the welcome page's navigation.
- **A round of manual use on a green branch found eight more**, all in the page class and none of them things the suite asks about: the compose column and its margins, the scrollbar's position and its theme, the pane's ground, motion keys the page claimed instead of inheriting, geometry slid rather than cut when it scrolled off the top, controls that did not light under the pointer, and a saved session handing one buffer's compose column to another. The page's cells are now identical to master's, colour for colour, and the last of these is the eighth commit.
- Two rounds found things tests would not have: the web suite caught the terminal slice three times, and an independent review of the branch produced six findings, all fixed.

## Merge history

Rebased on `master`. `master` merged #3194 mid-flight, which builds a feature on machinery this branch deletes; it is carried here re-expressed on the tree rather than dropped or reverted, and all 25 of its e2e tests pass. While the branch was un-mergeable against it, GitHub created no CI runs at all — worth knowing, since "no runs" looked like "no failures".

https://claude.ai/code/session_017128ATZ9CJCWqNPRT6WRfW


crates/fresh-editor/src/view/shell/buffer_host.rs
```diff
@@ -0,0 +1,930 @@
...
+            buckets: RefCell::new(None),
+        }
+    }
+
+    /// The marks on a `track`-long bar, bucketed from the facts' rows, plus
+    /// the hovered cell.
+    fn marks(&self, f: &BarFacts, track: u16) -> Rc<[Mark]> {
+        let ptr = Rc::as_ptr(&f.marks) as *const RowMark;
+        if let Some((p, t, h, m)) = self.buckets.borrow().as_ref() {
+            if *p == ptr && *t == track && *h == f.hover_cell {
```
sinelaw: **Bucket cache key is incomplete: `f.colors` is not part of it, and the `Rc` it keys on is not held.**

`marks()` reuses the cached `Rc<[Mark]>` whenever `(Rc::as_ptr(&f.marks), track, hover_cell)` matches, but the `Mark.theme` it baked in is `literal(color)` from `f.colors`, which `settle_pane_bars` re-resolves from the theme every frame. The marks `Rc` itself is served from `ScrollbarMarkerBuckets::cached_rows` keyed by `ProjectionKey` (marker/content/save versions + basis), which a theme switch does not touch.

Failure: open a buffer with theme-keyed scrollbar marks (unsaved-change marks, diagnostics), switch theme (`Settings → Theme`), and the bar's marks keep the old theme's literal colours until the marker set or buffer content changes.

Secondary: the cache stores the raw pointer without holding a clone of the `Rc`, so once the old allocation is freed a new `Rc<[RowMark]>` can land at the same address (two resolutions between two paints is enough) and the stale bucketing is served for a different mark set. Holding `Rc<[RowMark]>` in `Bucketed` and comparing with `Rc::ptr_eq`, plus including `f.colors` (or its pointer) in the key, closes both.

---

crates/fresh-editor/src/app/render.rs
```diff
@@ -8172,24 +7220,157 @@ impl Editor {
...
+                .collect()
+        };
+        let win = self.active_window_mut();
+        win.retain_pane_handles(|pane| live.contains(&pane));
+        visible
+            .into_iter()
+            .map(|pane| {
+                let h = win.pane_handle_for(pane).clone();
+                if interiors.contains_key(&pane) || groups.contains_key(&pane) {
+                    h.clear_rows();
```
sinelaw: **A pane with no text pass this frame keeps a stale `caret`/`caret_shown`; `Window::pane_caret()` can then report a caret cell for a pane that shows none.**

`clear_rows()` clears only `rows`. `settle_views` iterates `prepared.contents`, i.e. only the panes the text pass drew, so a pane that becomes a described panel (a plugin panel mounted into it) or a hidden sibling of a maximized pane is never re-settled and keeps its last `caret: Some(..), caret_shown: true`.

`Window::pane_caret()` is `self.panes.values().find_map(..)` over a `HashMap`, so whichever pane happens to iterate first with `caret_shown` wins.

Failure: pane A is active with the caret shown; a plugin mounts a panel into A (`git_log`, package list) or A is hidden by maximizing B; the next frames' `pane_caret()` can return A's old cell and `maybe_start_cursor_jump_animation` animates a trail to a cell with no caret in it.

Suggest settling `caret: None, caret_shown: false` alongside `clear_rows()` here (or having `clear_rows` reset them), and having `pane_caret()` read only the active pane's handle rather than the first one that says it shows a caret.

---

crates/fresh-editor/src/view/shell/buffer_host.rs
```diff
@@ -0,0 +1,930 @@
...
+    /// start (a press does) or to nothing (a hover does); `fallback` is the
+    /// position a pane with no rows answers.
+    pub(crate) fn resolve(
+        &self,
+        col: u16,
+        row: u16,
+        allow_gutter: bool,
+        fallback: usize,
+    ) -> Option<ClickTarget> {
+        let rows = Some(self.rows.clone());
```
sinelaw: **Efficiency: the whole row map is deep-cloned on every `text_byte_at`, which fresh-ui computes per pointer sample.**

`fresh-ui/src/hit.rs` calls `text_byte_at(target, pos)` in `propagate` (every press, move, release and wheel routed to the leaf) and in `fire_at` (enter/leave), so `PaneView::resolve` runs on every mouse motion over a pane. `Some(self.rows.clone())` copies every `ViewLineMapping` — two `Vec`s per visual row — just to satisfy `screen_to_buffer_position_with_overshoot`'s `&Option<Vec<ViewLineMapping>>` parameter.

The PR's stated goal is that bare motion over text costs nothing; this makes it cost an O(rows × columns) allocation per sample on a tall pane. `click_handlers.rs` (lines 75 and 320) clones `v.rows` again for the same press.

Changing the helper to take `Option<&[ViewLineMapping]>` (or `&[ViewLineMapping]`) and passing `&self.rows` removes all three clones.

---

crates/fresh-editor/src/view/shell/tabs.rs
```diff
@@ -0,0 +1,912 @@
...
+        let w = str_width(&txt);
+        if skip >= w {
+            skip -= w;
+            continue;
+        }
+        let room = available.saturating_sub(rendered);
+        if room == 0 {
+            break;
+        }
+        let shown: String = txt.chars().skip(skip).take(room).collect();
```
sinelaw: **`skip` and `room` are column counts but the cut is by `chars()`; with a wide-character tab name the node is sized wider than the room it has.**

`skip`, `available`, `rendered` and `room` are all `str_width` columns, but the piece is cut with `txt.chars().skip(skip).take(room)`, and then the node is given `w(Sizing::Cells(shown_w))` where `shown_w = str_width(&shown)`.

Failure: a tab named `日本語ファイル.txt` is the last one visible with `room == 3`: `take(3)` yields three CJK characters, `shown_w == 6`, the node is 6 cells wide in a strip row fixed at `width`, so the children overflow the row and the trailing `>` arrow / pinned `+` are pushed out of it. On the scrolled side, `skip(skip)` skips too few columns and the window starts left of where `offset` says.

The legacy painter had the same char-based cut but only painted cells into a clipped buffer; here the explicit width makes the overflow visible. Cutting on grapheme/width (accumulating `str_width` per grapheme, as `prompt_line::window_start` does) fixes both edges.

---

crates/fresh-ui/src/focus/mod.rs
```diff
@@ -599,6 +726,197 @@ impl<M: 'static> Ui<M> {
...
+        let prev = self.settled_marks.insert(scope, id.clone());
+        match prev {
+            Some(p) if p == id && !released => return,
+            Some(_) => {}
+            // A confinement seen for the first time — the panel just took
+            // the keyboard around a focus that was already inside it. Its
+            // mark is where the description says focus is, and that is the
+            // entry landing this scope never got; a confinement that marks
+            // nothing leaves focus where it found it.
+            None if mark.is_none() && !released => return,
```
sinelaw, line 783: **Doc and code disagree on the first-settle case.**

The function doc (line 737) says: "A scope the tree has never settled records its mark and moves nothing: a mark that was already there when the scope came into view is not a decision made since." The match does the opposite: `None` with `mark.is_some()` falls through to `focus_element(mark)`. The inline comment on the `None if mark.is_none()` arm explains why the code is right ("the entry landing this scope never got"), so the fn-level doc is the stale half.

Cost: this is the library rule the whole S2/S7 focus design leans on; a reader of the doc would expect a first-seen confinement with a mark to leave focus alone and would debug the wrong thing. Please align the doc with the arm's comment.

---

crates/fresh-editor/src/app/input_dispatch.rs
```diff
@@ -414,225 +414,189 @@ impl Editor {
...
+            );
+        }
+        self.rerender_widget_panel(key);
+    }
+
+    /// Take the active window's prompt down, and its toolbar panel with it.
+    ///
+    /// The one way a prompt closes: the toolbar's registry entry lives
+    /// exactly as long as the prompt that shows it.
+    pub(crate) fn drop_prompt(&mut self) -> Option<crate::view::prompt::Prompt> {
```
sinelaw, line 575: **"The one way a prompt closes" is not the only way: prompts replaced by assignment bypass this and leak the toolbar panel.**

`plugin_commands.rs:2450` and `:2486` (`StartPrompt` / with initial text), plus the `prompt = Some(..)` sites in `search_ops.rs:1044`, `lsp_actions.rs:109/327`, `settings_prompts.rs` and `file_explorer.rs`, overwrite an existing `prompt` in place. The old `Prompt` drops with its `toolbar` key, but the registry entry `PanelKey(plugin, PROMPT_TOOLBAR_PANEL_ID)` stays mounted.

Failure: live grep is open with its toolbar, the user has Tabbed onto a toggle; a binding reachable from inside the prompt starts another plugin prompt. The toolbar panel leaks, and the next `setPromptToolbar` from live_grep finds it mounted and takes `mount_prompt_toolbar`'s `update` branch, carrying the stale `focus_key` (the toggle) and instance states into the fresh prompt — so the new toolbar opens with focus on a control the user never focused this session, and the description marks it as the focus holder.

Calling `self.drop_prompt();` before each assignment (as `file_open_input.rs` now does) keeps the invariant this doc states.

---

sinelaw (commented):
Single-pass code review at high effort (no Agent tool available in this session, so this is one reviewer's pass over the diff, not the multi-agent fan-out + verify). Six inline findings, most severe first: the scrollbar mark bucket cache ignoring theme colours, a stale caret on panes with no text pass, prompt replacement leaking the toolbar panel, a per-mouse-sample row-map clone, a wide-character tab cut, and a doc/code mismatch in the focus library's first-settle rule.

Note: the uncommitted working tree in the session (pipeline tail and `Flow::Observe` deleted, `PaneKey`/`SidebarKey`) matches what the PR body describes but is not in this head; nothing in it changed these findings.

---

crates/fresh-editor/src/view/shell/buffer_host.rs
sinelaw: Fixed in 1298bb4 by deleting the memo rather than widening its key: the bar buckets its marks on every paint. The costly half (marks to rows against the buffer) is already done once per change on the editor's side; what the leaf does is arithmetic over those rows, cheaper than the paint that asks for it, and one fewer key to get wrong.

---

crates/fresh-editor/src/app/render.rs
sinelaw: Fixed in 1298bb4: after the text pass settles the panes it drew, `Window::clear_carets_except` clears the caret of every other pane's handle, so a pane that mounted a panel or is hidden behind a maximized sibling has no caret at all (`PaneHandle::clear_caret`; rows are kept). `pane_caret()` then only ever finds the caret this frame placed.

---

crates/fresh-editor/src/view/shell/buffer_host.rs
sinelaw: Fixed in 1298bb4: `screen_to_buffer_position*` take `Option<&[ViewLineMapping]>`; `text_byte_at` borrows the rows, and the drag path borrows the view for the resolve and drops it before the mutations. The per-press sites keep a clone because their `Ref` would cross a `&mut self`; they hand a borrow on.

---

crates/fresh-editor/src/view/shell/tabs.rs
sinelaw: Fixed in 1298bb4: the piece is cut by `columns(txt, skip, room)`, which walks graphemes accumulating `str_width`, so a piece is never wider than its room and the scrolled window starts where the offset says (`a_piece_is_cut_by_columns_not_chars`).

---

crates/fresh-ui/src/focus/mod.rs
sinelaw: Fixed in 1298bb4: the doc now says what the arm does — a scope seen for the first time takes its mark as the entry landing it never got, and one that marks nothing leaves focus where it found it.

---

crates/fresh-editor/src/app/input_dispatch.rs
sinelaw: Fixed in 1298bb4: `Editor::set_prompt` is the one way a prompt opens; it calls `drop_prompt` first, so the previous prompt's toolbar panel is unmounted. All 27 `prompt = Some(..)` sites go through it (`a_prompt_opened_over_another_unmounts_its_toolbar`).

---

PR #3184, opened 2026-09-03
Read the whole of a long search match, and bind the keys you can name

### LINKED ISSUE #1128: Key binding issue when using *

### LINKED ISSUE #3186: Search &amp; Replace: results can't be scrolled horizontally
Split out of #1580, which collected several unrelated Search & Replace reports from an external document and is down to this one item. You noted in April it would be easier to act on split.

## The report

From the original list:

> Can't scroll horizontally in the search panel, if a line is long there's no way to see the rest of it (should work with left/right arrow keys and shift+mouse wheel).

## Where it stands

**Half is fixed.** #3154 made a result row window onto its own match, so the matched text is on screen even when it sits far along a long line. That is the part with one right answer.

**Panning is not.** The rest of the line is still unreachable, and the two gestures the report names both do nothing:

- `Left` / `Right` are the Tree's collapse/expand (`crates/fresh-editor-core/src/widgets/kinds/tree.rs`), so binding them as asked would cost tree navigation.
- `Shift`+wheel arrives as a *vertical* notch with the modifier set — terminals rarely send `ScrollLeft`/`ScrollRight` — and nothing reads the modifier, so it scrolls the list down.

Three things stood in the way, all recorded in #3154's description:

- No widget kind has a horizontal offset; `PaintedWindow::offset` counts rows.
- The plugin cut each row to the panel's width before the host saw it, so a host-side offset would have scrolled onto padding — real panning needs the plugin to send wider rows and the host to slice, which changes the row-building contract.
- The chord has to come from somewhere in an already-dense panel mode.

## Reproduction

```sh
mkdir -p /tmp/srdemo && cd /tmp/srdemo && git init -q
python3 -c "open('long.txt','w').write('HEADMARK ' + 'x'*250 + 'needle' + 'y'*100 + 'TAILMARK' + 'z'*30 + '\n')"
fresh --no-restore
```

`Alt+A`, search `needle`, `Tab` to the results tree, `Down` onto the match row. On `64b75a6` neither `HEADMARK` nor `TAILMARK` can be brought on screen by any key or wheel gesture.

#3184 implements this.

sinelaw:
## What this is for

Two things a reader could not do.

**Read a match that is not near the start of its line.** A Search & Replace row showed one window onto its line and nothing could move it — so on a minified file, a long prose line, or anything behind wide characters, the match sat off the edge with no way to bring it into view. Issue #1580; #3154 fixed where the window rests and stopped there.

**Bind a key by its name.** Most keypad names, the media keys and the modifier keys parsed as nothing, and an entry naming one was dropped at load with a warning badge and no explanation. Issue #1128.

---

### Reading a long match

The fixture is a 400-column line with markers at each end and the match in the middle, plus a short line and one behind 60 CJK characters.

**Before.** The long row shows the head of the line and the match is nowhere on screen; the wide-character row is drawn past the panel edge and cut by the terminal. `Shift`+arrows and `Shift`+wheel do nothing at all:

```
[v] cjk.txt:1   - 漢                                                                              needleyyyyyyyyyyyyyy
[v] long.txt:1  - HEADMARK xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx...
[v] plain.txt:1 - alpha needle beta
```

**After.** Every row rests on its own match and fits the panel:

```
[v] cjk.txt:1   - 漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢漢needleyyyyyyyyyyyyy…
[v] long.txt:1  - …xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxneedleyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy…
[v] plain.txt:1 - alpha needle beta
```

`Shift+Right` — the tail of the line, and the marker at the far end:

```
[v] long.txt:1  - …yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyTAILMARKzzzzzzzzzzzzzzzzzzzzzzzzzzzz…
[v] plain.txt:1 - alpha needle beta
```

`Shift+Left` — the head, which nothing could reach before:

```
[v] long.txt:1  - HEADMARK xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx…
[v] plain.txt:1 - alpha needle beta
```

`Shift+Home` returns every row to its own match. `Shift+End` takes the row you are on to the end of its line. `path:line` never moves — it is which match the row *is* — and `plain.txt`, which already fits, never moves either.

### `Ctrl`+arrow in the panel

The panel focuses its results tree once a search settles, so this is where a reader lands after typing.

| | |
|---|---|
| **before** | `Ctrl+Left` changes the status bar — `Ln 1, Col 17` → `Col 11` → `Col 10` — while nothing on screen moves. The key reaches the buffer behind the panel and runs a word-wise motion on a cursor nobody can see. |
| **after** | The status bar does not move. In a search field the same chord moves by word, which is what it means. `Ctrl+PageUp`/`PageDown` still switch buffers. |

### `Shift`+wheel

| | |
|---|---|
| **before** | Scrolls vertically — the same as not holding Shift. Terminals report the gesture as an ordinary vertical notch with the modifier set, and nothing read the modifier. |
| **after** | Scrolls sideways, over the results and over an ordinary buffer, one pan step a notch. |

### Binding a keypad key

With `{ "key": "kp_enter", "modifiers": ["ctrl"], "action": "duplicate_line" }` in the config:

| | |
|---|---|
| **before** | `[⚠ 1]` in the status bar — the entry did not parse and was dropped |
| **after** | No warning, and `Ctrl+Enter` duplicates the line |

The same for `kp_0`…`kp_9`, `kp_equal`, the keypad navigation keys, the media keys and the modifier keys. `docs/configuration/keyboard.md` now lists every name that parses, including the honest part: a terminal reports keypad keys as their main-keyboard equivalents, so `kp_multiply` is an alias for `*` rather than a key of its own.

---

## How, briefly

**Panning.** The plugin used to cut each row to the panel's width before the host saw it, so there was nothing to the left of the window and nothing to the right. It now sends the row whole and says which part matters; the host does the fitting. That is the division of labour the vertical axis already has, and it is why a pan costs no plugin work and no round trip — one number moves and the visible rows repaint.

The pan is one number shared by every row, applied as an offset from where each rests and clamped per row against what that row can show. Rows of different lengths slide together rather than drifting apart, and "pan to the end" is answered for the row under the selection, since rows of unequal length cannot all be at their end at once. The fit is measured in display columns rather than codepoints, which is what fixes the wide-character case.

**Key names.** "What does this key mean" was written out once per terminal encoding the input parser decodes, and again — partially — as a list of names the config accepts, so extending one never extended the others. There is now one table per family, in the parser, and both sides resolve through it. Coverage stops being something to maintain, the keyboard reference is generated from those tables, and a test fails when the two drift.

## Try it

<details>
<summary><b>Setup</b> — copy-pasteable, uses a scratch <code>HOME</code> so your own config is untouched</summary>

```sh
cargo build
FRESH=$PWD/target/debug/fresh

mkdir -p /tmp/srdemo && cd /tmp/srdemo && git init -q
printf 'alpha needle beta\n' > plain.txt
python3 - <<'EOF'
open("long.txt", "w").write("HEADMARK " + "x"*250 + "needle" + "y"*100 + "TAILMARK" + "z"*30 + "\n")
open("cjk.txt",  "w").write("漢"*60  + "needle" + "y"*200 + "ENDMARK\n")
EOF

export DEMO_HOME=/tmp/srdemo-home
mkdir -p "$DEMO_HOME/.config/fresh"
cat > "$DEMO_HOME/.config/fresh/config.json" <<'JSON'
{ "keybindings": [ { "key": "kp_enter", "modifiers": ["ctrl"], "action": "duplicate_line" } ] }
JSON

HOME="$DEMO_HOME" "$FRESH" --no-restore
```

If you compare against `master`, use a **separate `HOME` per binary**: the embedded-plugin cache is shared, and reusing one silently runs the other build's plugin.
</details>

`Alt+A`, type `needle`, `Down` onto a match row. Then `Shift+Left` / `Shift+Right`, `Shift+Home` / `Shift+End`, `Shift`+wheel, and `Ctrl`+arrows. With focus in the **Search** field instead of the tree, `Shift+Left` still extends the selection and the rows do not move.

## Notes

- **`Shift`+wheel changes ordinary buffers too**, not just this panel. Flagged because it is user-visible beyond the issue; the alternative is two rules for one gesture.
- A match row's payload grows, since a row has to carry more than the visible slice for there to be anything to pan onto. Bounded, and disclosed — the per-frame cost that amplifies it is pre-existing and filed as #3195.
- The general codepoint-versus-display-column confusion elsewhere in this layer is #3185. This change is column-correct on the path it adds, which is what fixes the symptom; changing what the existing width hints *mean* is a contract change with every widget as a consumer.
- The originally reported #1128 symptom — `Ctrl+*` stripped by a legacy-encoding terminal before Fresh sees the key — is not fixable here and is untouched.

Closes #1128, closes #3186. Refs #1580, #3185, #3195.

https://claude.ai/code/session_0187KwTwpiLC2xQrfFiwJtum

### DIFF
--- a/crates/fresh-editor-core/src/widgets/kinds/tree.rs
+++ b/crates/fresh-editor-core/src/widgets/kinds/tree.rs
@@ -16,6 +16,8 @@ use crate::widgets::render::{
 
 pub struct Tree;
 
+use crate::widgets::render::PAN_COLUMNS;
+
 impl WidgetImpl for Tree {
     fn on_wheel(
         &self,
@@ -208,6 +210,49 @@ impl WidgetImpl for Tree {
             }
             "Left" | "Right" => {
                 lateral(spec, widget_key, panel, key == "Right", fx);
+            }
+            // Panning. `Left`/`Right` are collapse/expand — the tree meaning
+            // every OS tree widget and the ARIA tree pattern give them — so
+            // sideways takes the one arrow chord whose conventional meaning a
+            // read-only, single-select tree does not have: extend selection.
+            // (`Alt`+arrows are back/forward, `Ctrl`+arrows word-wise; both
+            // are bound.) Issue #1580.
+            "S-Left" | "S-Right" | "S-Home" | "S-End" => {
+                let bounds = crate::widgets::render::pan_bounds(spec, viewport.cols, None);
+                // `S-End` is "the end of the row I am on", not "the end of
+                // the longest row": the pan is shared, rows of unequal length
+                // cannot all sit at their tail at once, and answering with
+                // the longest one leaves the selected row clamped with
+                // keystrokes still to spend before it moves.
+                let selected = resolve(spec, widget_key, &panel.instance_states).selected;
+                let end = usize::try_from(selected)
+                    .ok()
+                    .map(|r| crate::widgets::render::pan_bounds(spec, viewport.cols, Some(r)).1)
+                    // A row with nowhere to go — a file header among match
+                    // rows — would make the key a no-op while the rows around
+                    // it still had a tail to show. Fall back to the longest.
+                    .filter(|&r| r > 0)
+                    .unwrap_or(bounds.1);
+                let delta = match key {
+                    "S-Left" => Some(-PAN_COLUMNS),
+                    "S-Right" => Some(PAN_COLUMNS),
+                    // Home is where each row's content says it should rest —
+                    // its own match — not the head of the line. The head is a
+                    // few more `S-Left`s away, and a reader who wants the
+                    // match back should not have to pan to find it.
+                    "S-Home" => None,
+                    // Far enough that the per-row clamp lands every row on
+                    // its own tail — and no further, so `S-Left` walks back
+                    // from the end of the longest row rather than from a
+                    // number no content justifies.
+                    _ => Some(end),
+                };
+                if !panel.pan_h(widget_key, delta, bounds) {
+                    // Already home, or already at the value asked for: say so,
+                    // so the key can mean something else further out rather
+                    // than being swallowed by a tree that did nothing.
+                    return super::KeyDisposition::Pass;
+                }
             }
             "Enter" => {
                 if let Some(ev) = activate_event(spec, widget_key, panel) {
@@ -667,6 +712,10 @@ fn render_widget_tree(
     // The offset is the *last paint's*, not the tree's: the scroll fold
     // reads back its own previous value and republishes it below.
     let prev_scroll = ctx.painted(tree_key).map(|w| w.offset).unwrap_or(0);
+    // Sideways is the reader's, not the paint's — one delta for the whole
+    // tree, clamped per row against that row's own length so rows of
+    // different lengths slide together rather than drifting apart.
+    let h_pan = ctx.h_pan(tree_key);
 
     // Compute the visible (un-collapsed) flat slice of the
     // full `nodes` list. A node at depth d is visible iff
@@ -810,6 +859,7 @@ fn render_widget_tree(
                 rows: visible_rows,
                 items: visible_rows / per_node.max(1),
                 offset: scroll,
+                cols: panel_width,
             },
         );
     }
@@ -851,6 +901,7 @@ fn render_widget_tree(
             card_borders,
             panel_width,
             indent_cols,
+            h_pan,
         );
         let mut entry = rendered.entry;
         let is_selected = abs_idx as i32 == effective_sel_abs;
--- a/crates/fresh-editor-core/src/widgets/render.rs
+++ b/crates/fresh-editor-core/src/widgets/render.rs
@@ -561,6 +561,12 @@ pub struct RenderContext<'a> {
     /// Settings dialog, a toolbar, most tests) starts every window at
     /// the top, exactly as an absent instance-state offset used to.
     pub prev_painted: Option<&'a HashMap<String, PaintedWindow>>,
+    /// How far each keyed rows-widget is panned sideways, in display columns.
+    /// The reader's own fold, carried here for the same reason
+    /// `prev_painted` is: the painter is where a window is resolved, and
+    /// sideways is a window like any other. `None` (a stateless render)
+    /// pans nothing, exactly as an unset entry does.
+    pub h_pan: Option<&'a HashMap<String, i32>>,
 }
 
 impl RenderContext<'_> {
@@ -571,6 +577,14 @@ impl RenderContext<'_> {
     pub fn painted(&self, key: Option<&str>) -> Option<PaintedWindow> {
         let k = key.filter(|k| !k.is_empty())?;
         self.prev_painted?.get(k).copied()
+    }
+
+    /// Sideways pan for `key`, in display columns. Zero for a keyless widget
+    /// — a pan is addressed to a widget and an unkeyed one cannot be named.
+    pub fn h_pan(&self, key: Option<&str>) -> i32 {
+        key.filter(|k| !k.is_empty())
+            .and_then(|k| self.h_pan?.get(k).copied())
+            .unwrap_or(0)
     }
 
     /// Whether `key` names the focused widget. Empty keys never match.
@@ -628,6 +642,10 @@ pub struct RenderOptions<'a> {
     /// previous render published. A host that keeps a panel mounted
     /// MUST thread it, or every repaint starts its lists at the top.
     pub prev_painted: Option<&'a HashMap<String, PaintedWindow>>,
+    /// See [`RenderContext::h_pan`] — how far each rows-widget is panned
+    /// sideways. A host that keeps a panel mounted MUST thread it, or every
+    /// repaint slides the reader's rows back to their resting window.
+    pub h_pan: Option<&'a HashMap<String, i32>>,
 }
 
 /// Render a spec to a [`RenderOutput`] under explicit [`RenderOptions`].
@@ -668,6 +686,7 @@ pub fn render_spec_with_options(
         marker_gutter: opts.marker_gutter,
         avail_height: opts.avail_height,
         prev_painted: opts.prev_painted,
+        h_pan: opts.h_pan,
     };
     let mut next_state = HashMap::new();
     let collected = render_collected(spec, prev, &mut next_state, ctx, panel_width);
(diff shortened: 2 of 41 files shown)


crates/fresh-editor-core/src/widgets/kinds/tree.rs
```diff
@@ -209,6 +218,32 @@ impl WidgetImpl for Tree {
...
+                    "S-Left" => Some(-PAN_COLUMNS),
+                    "S-Right" => Some(PAN_COLUMNS),
+                    // Home is where each row's content says it should rest —
+                    // its own match — not the head of the line. The head is a
+                    // few more `S-Left`s away, and a reader who wants the
+                    // match back should not have to pan to find it.
+                    "S-Home" => None,
+                    // Far enough that the per-row clamp lands every row on
+                    // its own tail; `pan_h` bounds the stored value.
+                    _ => Some(i32::MAX / 4),
```
sinelaw: **`S-End` leaves `S-Left` permanently dead.**

`S-End` stores `i32::MAX / 4` = 536,870,911 in `h_pan`, and that value is only clamped *per row at paint time* — the stored number keeps it. `S-Left` walks it back 8 columns a press, so after one `Shift+End` the reader needs roughly 67 million `Shift+Left` presses before any row moves again. `Shift+Home` is the only escape.

Concretely: focus the results tree, press `Shift+End` (every row lands on its tail), press `Shift+Left` → `pan_h` stores `536870903`, which still clamps past `max_start` for every row, so the screen does not change.

The same debt accrues on a smaller scale in the ordinary direction: every `S-Right`/Shift+wheel press past the longest row's tail still adds +8, so the reader has to press `S-Left` exactly as many times before the rows start moving.

`pan_h` also returns `true` for all of these, so each dead press is `Consumed` and triggers a full panel re-render.

A content-shaped bound would fix both — e.g. clamp the stored value against the widest row's `max_start` at paint time and write it back, or make `S-End` a saturating "go to the end" rather than a huge delta.

---

crates/fresh-editor/plugins/lib/fresh.d.ts
```diff
@@ -1440,6 +1440,20 @@ type TreeNode = {
...
+	* the window can only start at the head of the line — which is exactly
+	* where a search result's match usually is not (issue #1580). Naming the
+	* span lets the host rest the window on it instead, and the reader pans
+	* away from there.
+	*
+	* Chars rather than columns because that is the unit a plugin can count:
+	* it has the string, not the terminal's width table. The host converts.
+	* Out-of-range values are harmless — they resolve to the end of the text.
+	*/
+	windowAnchor?: TextWindowAnchor | null;
```
sinelaw, line 1481: **`TextWindowAnchor` is referenced here but never declared.**

```
$ grep -rn TextWindowAnchor crates --include=*.ts
crates/fresh-editor/plugins/lib/fresh.d.ts:1456:	windowAnchor?: TextWindowAnchor | null;
```

The new struct was not registered in `crates/fresh-plugin-runtime/src/ts_export.rs` — it needs an arm in `decl_for` (next to `"TreeNode" => Some(fresh_core::api::TreeNode::decl(&cfg))`, ~line 182) and an entry in the emitted-type-name list (next to `"TreeNode", // Used by WidgetSpec::Tree.nodes`, ~line 404).

The repo's own `plugins/check-types.sh` doesn't catch it because it runs with `--skipLibCheck`, which suppresses errors *inside* `.d.ts` files. The consequence today is that `TreeNode.windowAnchor` resolves to an error type, so `treeNode({ windowAnchor: … })` gets no checking at all — and any plugin author whose tsconfig checks lib files sees `Cannot find name 'TextWindowAnchor'`.

---

crates/fresh-editor/plugins/search_replace.ts
```diff
@@ -830,27 +764,38 @@ function flatItemKey(item: FlatItem): string {
...
-function renderFlatItemEntry(item: FlatItem, W: number): TextPropertyEntry {
-  if (!panel) return { text: "" };
+/** One rendered row: the entry, plus the char span the row exists to show.
+ *
+ *  The span is the host's to act on — it rests the row's window on it when
+ *  the row is wider than the panel, and the reader pans away from there. The
+ *  plugin no longer cuts the string itself: it sends the capped context whole
+ *  and says where the interesting part is, which is what makes panning left
+ *  able to reach the head of the line at all. See `TreeNode.windowAnchor`. */
+type RenderedRow = { entry: TextPropertyEntry; anchor?: { start: number; len: number } };
```
sinelaw: **`RenderedRow.anchor` omits `pinned`, so this file no longer type-checks.**

`renderFlatItemEntry` returns `anchor: { pinned, start, len }`, but the declared shape is `{ start: number; len: number }`. Excess-property checking on the contextually-typed literal rejects it. Reproduced with the repo's own checker:

```
$ plugins/check-types.sh search_replace.ts
search_replace.ts(904,7): error TS2353: Object literal may only specify known
  properties, and 'pinned' does not exist in type '{ start: number; len: number; }'.
```

(The four `lib/widgets.ts` errors in the same run are pre-existing.) It works at runtime only because JS keeps the extra property.

```suggestion
type RenderedRow = { entry: TextPropertyEntry; anchor?: { pinned: number; start: number; len: number } };
```

---

crates/fresh-editor/src/input/keybindings.rs
```diff
@@ -1689,6 +1689,276 @@ impl BindingSource {
...
+    },
+    KeyName {
+        names: &["pageup"],
+        code: KeyCode::PageUp,
+    },
+    KeyName {
+        names: &["pagedown"],
+        code: KeyCode::PageDown,
+    },
+    // Lock and system keys. A terminal speaking the kitty keyboard protocol
```
sinelaw, line 1777: **The same gap is still open for the media and standalone-modifier keys.**

This block closes the `{:?}`-spelling hole for `CapsLock`…`Menu` (kitty 57358–57363), but `kitty_functional_key` also decodes 57428–57440 into `KeyCode::Media(_)` and 57441–57454 into `KeyCode::Modifier(_)`, and neither has a name here.

Concretely: on a terminal in kitty keyboard mode, record a binding on the volume-mute key in the keybinding editor. `keybinding_editor/editor.rs:1122` calls `key_code_to_config_name(KeyCode::Media(MuteVolume))`, which falls through to `_ => format!("{:?}")` and writes `"key": "Media(MuteVolume)"`. At load, `parse_key` lowercases it, finds no table entry, sees it is not one character and does not match `f<n>` → `None`, `warn_invalid_key` fires, and the binding the user just created silently does nothing. That is exactly the #1128 failure mode, unchanged.

`config_names_round_trip` can't see it because it only iterates the codes already present in the tables.

---

crates/fresh-editor-core/src/widgets/render.rs
```diff
@@ -2770,6 +2789,171 @@ pub struct RenderedTreeRow {
...
+    // `total - budget` leaves the last column of the line permanently
+    // unreachable — panning right ran out one character short of the end.
+    let max_start = total.saturating_sub(budget.saturating_sub(1));
+    let rest = resting_start_col(content, budget, anchor).min(max_start);
+    let start_col = (rest as i64 + pan as i64).clamp(0, max_start as i64) as usize;
+
+    let slice_start = byte_offset_at_visual_column(body, start_col);
+    let lead = start_col > 0;
+    let lead_cols = usize::from(lead);
+    let content_budget = budget - lead_cols;
```
sinelaw, line 2895: **A panned window can render `budget + 1` columns when `budget == 1`.**

With `budget == 1` and any `start_col > 0`, `content_budget` is 0: the slice is empty, but both markers are still pushed, so the result is `"……"` — two columns for a one-column budget.

```rust
window_row_body("abcdefghij", 1, None, 3)
// max_start = 10, start_col = 3, lead = true, content_budget = 0
// full_end == slice_start, trail = slice_start < content.len() → true
// text == "……"  (2 cols, budget 1)
```

Reachable on a real row whenever `panel_width - prefix_cols - pinned_cols == 1` (a narrow sidebar plus a long `path:line` pinned head) and the tree is panned; the row then overruns the panel edge. `window_row_body_never_splits_a_wide_cell` can't catch it because it only exercises `pan == 0`, where `lead` is always false.

Suppressing the trailing marker when `content_budget == 0` (there is nothing between the two markers to elide) fixes it.

---

crates/fresh-editor/plugins/search_replace.ts
```diff
@@ -904,46 +849,63 @@ function renderFlatItemEntry(item: FlatItem, W: number): TextPropertyEntry {
...
   if (panel.searchPattern) {
-    highlightMatches(displayCtx, panel.searchPattern, panel.useRegex, panel.caseSensitive, ctxOverlays);
+    highlightMatches(context, panel.searchPattern, panel.useRegex, panel.caseSensitive, ctxOverlays);
   }
 
   const segments: StyledSegment[] = [
     { text: location, style: { fg: C.lineNum } },
     { text: " - " },
-    { text: displayCtx, overlays: ctxOverlays },
+    { text: context, overlays: ctxOverlays },
```
sinelaw, line 884: **The spec's text payload grows ~4–5x per row.**

Every row now ships the whole `CONTEXT_HARD_CAP`-capped context (up to 512 chars) where `windowAroundMatch` used to ship a panel-width slice (~80–120). With `MAX_RESULTS = 10000` that is roughly 5 MB of context text in the `WidgetSpec` against ~1 MB before, plus one `InlineOverlay` per pattern hit anywhere in the full context rather than only in the visible slice.

That spec crosses the QuickJS→Rust boundary on every streaming flush, and `Editor::panel_interior` / `pane_panel_interior` (`app/render.rs:8253`, `8359`) deep-`clone()` it **on every frame**, so the per-frame clone scales with it. Searching a common token in a repo with minified assets is the case that bites.

Panning genuinely needs more than the visible slice, so this is a real cost of the feature rather than an outright bug — but 512 for *every* row is a flat cap that ignores the panel width. Sizing it to a few multiples of `innerWidth` would keep the pan usable at a fraction of the payload.

---

sinelaw (commented):
Careful diff pass over the panning + key-name work. `cargo check -p fresh-editor -p fresh-editor-core -p fresh-input-parser` is clean; the plugin type-check is not (see the `search_replace.ts` comment). Six findings inline, roughly in severity order:

1. `S-End` stores a pan of `i32::MAX / 4` that `S-Left` can never walk back — `Shift+Home` becomes the only escape.
2. `fresh.d.ts` references `TextWindowAnchor`, which is never declared: the type was not registered in `ts_export.rs`.
3. `RenderedRow.anchor` omits `pinned`, so `check-types.sh` fails on `search_replace.ts` with TS2353.
4. Media and standalone-modifier keycodes still serialise to unparseable `{:?}` spellings — the #1128 failure mode, unclosed.
5. `window_row_body` renders two markers (2 columns) for a `budget == 1` panned window.
6. Every row now carries up to 512 chars of context, ~4–5x the old payload, in a spec that is deep-cloned per frame.

The pinned-head / overlay-splitting arithmetic in `render_tree_row` and the keypad table unification both read correctly to me; the column-vs-codepoint fix and the wide-cell tests are good.

---

crates/fresh-editor-core/src/widgets/kinds/tree.rs
sinelaw: Fixed in f65bbc0, and thank you — this was the worst of the six.

The stored pan is now bounded by what the content justifies. Two numbers rather than one, because a pan is a delta from where each row *rests*: a match 259 columns along a 403-column line has a great deal to its left and almost nothing to its right, and a single symmetric bound is wrong on whichever side is shorter by the difference. `render::pan_bounds` answers that from the spec — `(max(2·anchor.start, capped by the row), max(row_len − anchor.start))` — and it is O(1) per row, so a pan keystroke over a ten-thousand-row tree stays a length walk rather than a width walk.

Both conversions are upper bounds in the safe direction: a row's byte length is at least its column count (a two-column glyph costs three bytes), and an anchor's char offset is at most its column offset doubled. So nothing that was reachable stops being reachable — `pan_left_reaches_the_line_head` and `pan_right_reaches_the_line_tail` are the guards for that, and both still pass.

Your two other observations came along with it:

- **Dead presses no longer repaint.** Past the bound `pan_h` returns `false`, so the key is `Pass` rather than `Consumed`.
- **`pan_bounds` returns `(0, 0)` for every kind whose paint does not thread the pan**, which is every kind but `Tree`. So a Shift+wheel notch over a `List` now falls through instead of being swallowed into a pan nothing renders. The wheel path (`pan_widget_by_key`) resolves the widget from the spec to ask.

I did not take the write-back-from-paint option. It is the only way to get the *exact* clamp — that needs the panel width, which only the paint knows — but it means a new paint→state channel, and the residual after this fix is bounded by roughly a panel width rather than by `i32`.

New test `test_search_replace_pan_end_is_walkable_back`: `Shift+End`, then 60 × `Shift+Left` (8 columns each, more than the fixture's whole 403-character line) must show `HEADMARK`. Against the old code it fails with the row still parked mid-line.

---

crates/fresh-editor/plugins/lib/fresh.d.ts
sinelaw: Correct, and fixed in f65bbc0. `TextWindowAnchor` is registered in `ts_export.rs` — the `decl_for` arm and the emitted-name list, both where you said — and `fresh.d.ts` regenerated, so the type is declared right after `TreeNode`.

The `--skipLibCheck` point is the part worth keeping: the field was completely unchecked for plugin authors and nothing said so. Regenerating surfaced a second, real disagreement — `#[serde(default)] pinned: u32` produces `pinned: number` (required) in the generated type, while `widgets.ts` declared the option as `pinned?: number`. `treeNode` now fills the default in rather than passing the field through, which is the right place for it: the wire and the type then agree, and a caller with no pinned head does not have to say `pinned: 0`.

---

crates/fresh-editor/plugins/search_replace.ts
sinelaw: Confirmed and fixed in f65bbc0. I reproduced it exactly as you did — `search_replace.ts(904,7): error TS2353` — which also means the PR body's claim that this file "adds none to the four pre-existing `lib/widgets.ts` errors" was wrong. It added one. Body corrected.

I took your suggestion's shape but pointed it at the generated type rather than restating the fields:

```ts
type RenderedRow = { entry: TextPropertyEntry; anchor?: TextWindowAnchor };
```

which is only possible because the sibling finding got `TextWindowAnchor` declared. Spelling the three fields out here would have left two places to change and no check that they agree — the same drift that made the whole keypad half of this PR necessary. `check-types.sh search_replace.ts` is now back to exactly the four pre-existing `lib/widgets.ts` errors, and a full `check-types.sh` run matches the branch without this commit file-for-file and error-for-error.

---

crates/fresh-editor-core/src/widgets/render.rs
sinelaw: Fixed in f65bbc0 — `trail` is now `content_budget > 0 && full_end < content.len()`, so a one-column window is the lead marker and nothing else.

Your diagnosis of why the test missed it is the more useful half, so I widened `window_row_body_never_splits_a_wide_cell` rather than adding a separate case: it now sweeps pans (`-40, -1, 0, 1, 5, 40`) as well as budgets, over an ASCII body as well as a CJK one. Against the unfixed code it fails with

```
budget 1, pan 1: "……" is 2 columns
```

which is the assertion I ran before writing the fix.

---

crates/fresh-editor/src/input/keybindings.rs
sinelaw: You are right, and the last line is the finding. Fixed in f65bbc0.

The keys are one table now, `fresh_input_parser::media_modifier`, sitting beside `keypad` and resolved by the decoder and the config parser alike — so `kitty_functional_key` delegates the whole 57428–57454 range instead of carrying twenty-seven arms, and `key_code_to_config_name` writes `volume_mute` and `left_hyper` instead of `Media(MuteVolume)`. Unlike the keypad these are not aliases (nothing on the main keyboard means "mute"), and the generated `docs/configuration/keyboard.md` now says so in its own section.

**`config_names_round_trip` was circular, which is the real defect.** It enumerated the name tables and asked whether the names name themselves; it could never have reported a code that was in no table. It now sweeps every Private Use Area codepoint the parser decodes:

```rust
codes.extend((0xe000..=0xf8ff).filter_map(fresh_input_parser::kitty_functional_key));
```

That required making `kitty_functional_key` public, which I think is right on its own terms: it is the honest answer to "which keys can arrive here", and a test that wants to check every one of them against something else has no other way to enumerate them without restating the list and testing the restatement.

Removing the new `Media(_) | Modifier(_)` arm makes it fail with

```
Media(Play) serialises to "Media(Play)", which does not parse back
```

The sweep also picked up F13–F35, which round-trip already; the docs said `f1` through `f24` and now say `f35`.

---

crates/fresh-editor/plugins/search_replace.ts
sinelaw: This is the one I have not changed, and I want to be explicit about why rather than quietly leaving it.

**The payload growth is the feature's price.** You cannot pan onto text you never sent, and the plugin does not know the pan — that is the whole point of the split this PR argues for. So a row has to carry more than the visible slice. The multiplier is bounded (~2.7× the regex work, ≤525 bytes a row against ≈200) and the `CONTEXT_HARD_CAP` of 512 is unchanged, so this is a larger constant rather than a new asymptote.

**Sizing to a few multiples of `innerWidth` does not buy much where it hurts.** At the panel widths this case actually appears in — a full-width results panel, 100–130 columns — `4 × W` is 400–520, which is 512 to within noise. It only saves on narrow panels, which are also the ones with few enough columns that the payload was never the problem. And it makes the pan range depend on the panel width, so the same search pans further after you widen the window, which I would rather not have to explain.

**The per-frame deep clone is the part that deserves fixing, and it is not mine.** `panel_interior` cloning the whole spec every frame turns a linear payload into a linear payload *per frame*; that was already true at ~1 MB and this PR made it more visible. Filed as #3195, with the `Rc<WidgetSpec>` option, the cached-`Interior` option, and a note to measure before assuming the clone is where the time goes.

So: real, disclosed in the PR body's cost paragraph, and the amplifier has an issue. Leaving this thread open — if you would rather the cap be width-derived anyway, say so and I will do it here.

---

PR #3188, opened 2026-09-04
A panel that opens hidden is told the width it was given

sinelaw:
Opening the FILES sidebar (`F`) in Review Diff showed every filename elided to a few characters in a panel with columns to spare, and it stayed that way until the panel divider was dragged.

## What was happening

The sidebar elides its own rows, so it needs the panel's width, and the only way a plugin learns that is `viewport_changed`. A side panel that starts hidden is laid out for the first time on the frame it is **shown**, and after that it never changes size on its own — so that first layout is the only announcement its width ever gets.

The host skipped exactly that one: a split with no entry in `previous_viewports` was treated as "not established yet", and by the end of the same frame the snapshot recorded the new geometry, so the change was gone. Traced on a real review, the FILES leaf appears once with `previous=None` and is never reported again — the plugin had a width for the diff and the sticky header, and none at all for the sidebar.

With no width to work from, the plugin fell back to a ratio of `state.viewportWidth`, which is seeded from `getViewport()` — whichever split holds focus when the review opens. Open a review from a narrow pane (a file explorer, or the sidebar of the review being replaced) and that seed is a few dozen columns, so the ratio hits the plugin's 12-column floor and every row in a 40-column sidebar is cut to five characters. Dragging the divider produced a width *change*, which the host did report — hence "it corrects itself after I resize the panel border".

## The change

- `app/render.rs`: a split reports the geometry it is born with. `previous=None` now counts as a change instead of being skipped.
- `plugins/audit_mode.ts`: the stopgap guess (used only until the host's first report lands) takes its share of the **review group's** width — the toolbar spans it — rather than of an unrelated split's viewport. Panel geometry is also dropped when a review session ends: the next session's panels are different buffers in a differently sized group.

## Tests

- `review_diff_ux_bugs::test_files_panel_rows_fit_a_narrow_panel` — the symptom end to end. In a 10-column sidebar, rows built to the 12-column guess are clipped by the host, taking the `…` and the status letter with them (`   mod_00…` before, `   mod_… M` after).
- `buffer_group_panel_geometry` (+ a small probe plugin) — pins the contract the plugin side now relies on: show a hidden panel and its width arrives, with no resize to prompt it.

Verified by reverting the fix and re-running: the narrow-panel test fails on the old code and passes on the new.

https://claude.ai/code/session_016xHWjwJ8GS9ejGs2YeWSLd

---
_Generated by [Claude Code](https://claude.ai/code/session_016xHWjwJ8GS9ejGs2YeWSLd)_


crates/fresh-editor-core/src/widgets/registry.rs
```diff
@@ -571,20 +571,48 @@ impl WidgetPanelState {
...
+            let new_state = WidgetInstanceState::Tree {
+                expanded_keys: current.expanded,
                 // Re-pinning the *same* index (which the orchestrator
                 // dock's `refreshOpenDialog` does on every probe-poll
                 // repaint) must preserve a user scroll — otherwise the
                 // refresh would snap the view back to the selection a
                 // beat after a mouse scroll. Only an actual selection
                 // change re-arms scroll-follows-selection. Mirrors the
                 // List branch below.
+                user_scrolled: current.user_scrolled && index == current.selected,
```
sinelaw, line 600: **The spec's `selectedIndex` becomes permanently inert for this tree, and the stored index is absolute into a `nodes` list that gets rebuilt.**

`tree::resolve_seeded` prefers a stored `Tree` entry over the spec seed. Before this change the FILES sidebar had *no* `Tree` entry (a selection write recorded a `List` entry, which `resolve` ignores), so every repaint re-derived the highlight from `buildFilesPanelSpec`'s `selected = filesTree.keys.indexOf(trackedKey)` — a *key* lookup, correct across rebuilds. After this change, the very first `pointSidebarAtCurrentFile()` (which runs on the first `renderFilesPanel`, since `filesSelectedNodeKey` starts `""`) installs a `Tree` entry, and from then on the host's stored **absolute index** wins forever.

Concrete failure: review open, highlight on `src/mod_004.rs` at node index 7. The watch poll picks up a newly-modified `src/aaa.rs`, which sorts first, so every node after it shifts by one. `renderFilesPanel` rebuilds the tree and `pointSidebarAtCurrentFile` early-returns (`nodeKey === filesSelectedNodeKey`, the current file did not change), so nothing re-pins the index — the sidebar now highlights `src/mod_003.rs`. Same for anything else that reshapes the node list without changing the current file (staging a file, a filter change, a file disappearing).

Note `buildFilesPanelSpec` still documents the old contract at audit_mode.ts:2124 ("The spec's `selectedIndex` is authoritative on every render"), which this makes false.

---

crates/fresh-editor-core/src/widgets/registry.rs
```diff
@@ -571,20 +571,48 @@ impl WidgetPanelState {
...
+        let tree_spec = crate::widgets::find_widget_by_key(&self.spec, widget_key)
+            .filter(|spec| matches!(spec, WidgetSpec::Tree { .. }));
+        if let Some(spec) = tree_spec {
+            // Through the resolver, so a first write inherits the spec's
+            // `expanded_keys` seed instead of recording an empty set and
+            // collapsing every directory in the tree.
+            let current =
+                crate::widgets::kinds::tree::resolve(spec, widget_key, &self.instance_states);
+            let new_state = WidgetInstanceState::Tree {
+                expanded_keys: current.expanded,
```
sinelaw, line 592: **The same write also freezes `expanded_keys`, so the spec's `expandedKeys` seed stops taking effect after the first selection.**

Copying `current.expanded` is right for preserving what the user opened, but the entry it lands in is what makes `tree::resolve` stop consulting the spec. `buildFilesTree` recomputes `out.groupKeys` on every rebuild and passes it as `expandedKeys: filesTree.groupKeys` — i.e. "all groups open" is a *seed*, re-asserted on every repaint, and it silently becomes dead the moment `pointSidebarAtCurrentFile` first calls `setSelectedIndex` (first paint of the sidebar).

Concrete failure: open a worktree review (one category, group keys `dir:unstaged src`, `dir:unstaged tests`) — frozen. Press `s` to stage a file. `fileGroups()` now yields two categories, so the tree gains `cat:staged` / `cat:unstaged` and `dir:staged …` keys, none of which are in the frozen set — those rows render collapsed and their files vanish from the sidebar until the user expands each one by hand. Filtering (`/`) hits the same thing, because `compressDirChains` changes the compressed directory keys with the file set.

---

crates/fresh-editor-core/src/widgets/registry.rs
```diff
@@ -571,20 +571,48 @@ impl WidgetPanelState {
...
+        // **Which state this records is the widget's kind, not whatever
+        // happens to be in the map.** An untouched `Tree` has no entry at
+        // all — the render walk stopped writing one back — and recording a
+        // `List` entry for it made the write invisible: `tree::resolve`
+        // honours a `Tree` entry or falls back to the spec's seed, so the
+        // selection snapped back to whatever the plugin's last repaint had
+        // seeded. Clicking a row in the review's FILES sidebar left the
+        // highlight where it was for exactly that reason, and only looked
+        // right when a repaint happened to re-seed it at the clicked row.
+        let tree_spec = crate::widgets::find_widget_by_key(&self.spec, widget_key)
```
sinelaw, line 584: **Third consequence of the same "an entry now always exists" flip: `Tree::on_focus_change`'s blur clear becomes effective, and the FILES sidebar can't recover from it.**

`Tree::on_focus_change(gained: false)` calls `panel.set_selected_index(key, -1)` (tree.rs:139). Until now that was a no-op for the sidebar — it wrote a `List` entry that `resolve` ignored, so the highlight survived on the spec seed. With this change it really clears the tree's selection.

Concrete failure: click a file row in the FILES sidebar (this focuses the tree via `deliver_widget_hit`'s click-to-focus), then click the panel's filter `text` field. `set_panel_focus_and_notify` → `notify_widget_focus_change` blurs the tree → selection is `-1` → the "you are here" band disappears. The plugin cannot put it back: `pointSidebarAtCurrentFile` early-returns on `nodeKey === filesSelectedNodeKey`, and its mirror never changed. The highlight stays gone until the reader happens to move to a *different* file. (Keyboard `/` is unaffected — the plugin's own `SetFocusKey` path doesn't run `on_focus_change`.)

---

sinelaw (commented):
Reviewed the two commits (`render.rs` viewport birth report + `audit_mode.ts` width basis, and the `set_selected_index` tree fix). The width half looks sound to me: the toolbar spans the group, its `viewport_changed` now fires on the group's first frame, and the listener is registered synchronously after the single `await` in `openReviewPanels`, so the ordering works out. `previous_viewports` is cleared and repopulated per frame, so a hidden→shown panel is correctly "new" each time. `cargo check --tests -p fresh-editor` is clean.

The three comments below are all one root cause in `set_selected_index`: the tree now *always* gets an instance-state entry, and `tree::resolve` prefers that entry over the spec for **both** `selected_index` (an absolute node index) and `expanded_keys`. That is exactly the fix for the click case, but it also silently retires the plugin's per-repaint seeds, which `audit_mode.ts` still relies on for everything the click path doesn't cover.

A targeted way out, if you want the fix without the fallout, is to keep the write but not let it outrank a *changed* spec — e.g. only honour a stored `Tree` entry when the spec's `expanded_keys`/node identity hasn't moved since it was written, or have `renderFilesPanel` push `setSelectedIndex`/expansion unconditionally after every rebuild instead of guarding on `filesSelectedNodeKey`.

---

crates/fresh-editor/src/app/render.rs
```diff
@@ -1044,14 +1044,21 @@ impl Editor {
...
+                // the first time reports too: its birth geometry is a change
+                // from "no geometry at all", and it is the only chance a
+                // plugin gets to hear it. Skipping it (the old behaviour)
+                // swallowed exactly the frame a panel appears on — a buffer
+                // group's side panel is laid out for the first time when it
+                // is *shown*, and after that its size never changes on its
+                // own, so a plugin that lays its own rows out to the panel
+                // width was left guessing until the user happened to drag
+                // the divider. The review diff's FILES sidebar elided every
+                // filename to a guessed width for exactly that reason.
```
sinelaw: **The code change is right; the mechanism this comment (and the commit message, the CHANGELOG and the new test's module doc) describes is not.** I traced it with `RUST_LOG=fresh::app::render=trace` on `test_files_panel_rows_fit_a_narrow_panel`, on master:

```
split=LeafId(SplitId(2)) current=(0, 10, 13) previous=None changed=false   <- group created, FILES *visible*, 10 cols
split=LeafId(SplitId(4)) current=(0, 38, 11) previous=Some((0, 16, 11))    <- next frame: FILES hidden, diff grows
...
split=LeafId(SplitId(2)) current=(0, 10, 13) previous=Some((0, 10, 13)) changed=false   <- forever after, incl. the show
```

`create_buffer_group` builds the group with **every** panel visible, and `openReviewPanels` only hides the side panels afterwards — so the host renders one frame with FILES laid out at 10 columns before `applyPanelVisibility` lands. That frame is the `previous=None` one that was dropped. When the user later presses `F`, the panel comes back at *exactly the same* geometry (leaf id and `SplitViewState` survive hiding — `set_buffer_group_panel_visible` says so), so `changed` is false and there is nothing left to report. That is why the width was never announced, and it is why firing on first sight fixes it.

"A side panel that starts hidden is laid out for the first time on the frame it is *shown*" is the opposite of what happens here, and the difference is load-bearing: a panel that really *is* hidden before its first layout keeps the terminal-sized viewport `SplitViewState::with_buffer(tw, th)` gave it, so showing it *is* a change and master already reported it. (Verified below on the new probe test.) Worth rewriting this comment, the commit message and the CHANGELOG bullet — "A panel that opens hidden was never told the width it was given" is not what the trace shows, and the accurate version ("the geometry a panel is born with is the only announcement it ever gets, and it was being dropped") is a better justification for the change than the one written.

---

crates/fresh-editor/tests/e2e/plugins/buffer_group_panel_geometry.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    run_command(&mut harness, "TestVP: Create");
+    harness
+        .wait_until(|h| h.screen_to_string().contains("MAIN-PANEL-MARKER"))
+        .unwrap();
+
+    // Hidden: never laid out, so there is nothing for the host to report.
+    assert_eq!(
+        reported_side_width(&mut harness),
+        -1,
```
sinelaw: **This test passes on master, and it isn't a near-miss — it exercises a different code path from the bug.** The PR admits it passes both ways; here is why, so the decision to keep or strengthen it can be made on facts.

`tvp_create` hides `side` and `rail` inside the `.then()`, before the group is ever laid out. Traced on master:

```
split=LeafId(SplitId(2)) current=(0, 120, 40) previous=None      <- side, never laid out
split=LeafId(SplitId(4)) current=(0, 120, 40) previous=None      <- rail, ditto
```

Those are the `SplitViewState::with_buffer(terminal_w, terminal_h)` values the constructor hands every panel, and they never move while the panel is hidden. So `TestVP: Show` takes `side` from `(120, 40)` to `(29, 37)` — a real change, which **master already reports**. The test's premise comment ("Hidden: never laid out, so there is nothing for the host to report") is true here and is exactly what makes the test blind: the review's FILES panel *was* laid out once, at its final geometry, before being hidden, which is the case the fix is for.

Two ways to make it earn its place, either of which I'd take over dropping it:
1. Have the probe plugin hide `side` only after the group has been laid out once (a `setTimeout(0)`/`flush` before `setBufferGroupPanelVisible`), i.e. the review's actual sequence. Then `previous` on the show frame is the real 29 columns and the report only exists with `None => (true, None)`. (The `-1`-while-hidden assertion has to go, since with the fix the birth frame reports the width before the hide.)
2. Keep this shape but assert the *birth* report instead: after `TestVP: Create` the probe should already know `side`'s width, with no `Show` at all. That is the contract the plugin side now depends on, and it is the one this change actually adds.

As written it also cannot catch the ordering hazard in my other comment, because the probe subscribes to `viewport_changed` at module load rather than after creating the group.

---

crates/fresh-editor/src/app/render.rs
```diff
@@ -1044,14 +1044,21 @@ impl Editor {
...
                 let (changed, previous) =
                     match self.active_window().previous_viewports.get(split_id) {
                         Some(previous) => (*previous != current, Some(*previous)),
-                        None => (false, None), // Skip new splits until they're established
+                        None => (true, None),
```
sinelaw, line 1067: **The panel's width is now announced exactly once, on a frame the plugin is not guaranteed to be subscribed for, and there is no second chance.** This is the part of the design I'd push back on, not the one-line change itself.

With the fix, the trace for the review shows every group leaf firing on the group's birth frame and `split=2` (FILES) never firing again for the rest of the session — including on the `F` that shows it. `openReviewPanels` registers `editor.on("viewport_changed", on_review_viewport_changed)` at its *end*, after `applyPanelVisibility()` and `updateMagitDisplay()`. That works today only because of how the runtime happens to interleave the plugin task with renders. Demonstrated: with this PR applied, inserting a single line before the registration —

```ts
await editor.flush();
editor.on("viewport_changed", on_review_viewport_changed);
```

— makes `test_files_panel_rows_fit_a_narrow_panel` fail again with the *identical* symptom (`"   mod_00…"`). So the fix's effectiveness rests on an ordering nothing enforces, nothing documents, and no test covers — and there is no fallback: `listSplits()` walks `mgr.visible_leaves()`, which excludes buffer-group panels, so a plugin has no way to *ask* for a panel's geometry.

The cheap hardening is on this side of the boundary: when `set_buffer_group_panel_visible` makes a panel visible, remove that leaf from `previous_viewports` so the next render reports it as new. Then "the panel is on screen" always produces an announcement, whoever is listening by then, and the one-shot race stops mattering. (It also makes the probe test's premise true rather than accidental.)

---

crates/fresh-editor/plugins/audit_mode.ts
```diff
@@ -1749,12 +1749,21 @@ let filesPanel: WidgetPanel | null = null;
...
+ *  narrow pane — a file explorer, or the FILES sidebar of the review being
+ *  replaced — therefore seeded it with a few dozen columns, and every row
+ *  in a 40-column sidebar came out elided to the 12-column floor below. */
 function panelWidthOf(panel: 'files' | 'comments'): number {
     const known = state.panelWidths[panel];
     if (known && known > 0) return known;
-    return Math.max(12, Math.floor(state.viewportWidth * (panel === 'files' ? FILES_PANEL_RATIO : 0.15)));
+    const group = state.panelWidths["toolbar"];
+    const basis = group && group > 0 ? group : state.viewportWidth;
+    return Math.max(12, Math.floor(basis * (panel === 'files' ? FILES_PANEL_RATIO : 0.15)));
```
sinelaw: **The new basis is a better guess, but it fixes nothing the PR's own test covers, and the `12` floor still overshoots the case that produced the bug report.** Checked: with the host change reverted and *only* this plugin change in place, `test_files_panel_rows_fit_a_narrow_panel` fails identically (`"   mod_00…"`). The reason is right here — in that test the group is 39 columns, so `floor(39 * 0.22) = 8`, and `Math.max(12, 8)` still hands back `12` for a panel the host made 10 columns wide. The rows are still built too wide and still clipped.

That is fine as far as it goes (the host change is what fixes it), but it means the floor is the remaining hazard in this function: for any review whose FILES panel is under 12 columns, the guess is guaranteed to overshoot, and the guess is used for every paint that happens before the host's single report is processed. Since the host now reports at group creation, `renderFilesPanel` skips hidden panels, and the elide is clamped by `Math.max(4, W - …)` downstream, I don't have a reproduction on this branch — but the floor is why "too wide" is the failure direction rather than "too narrow", and clamping to the *known* group width (`Math.min(basis, …)`) would cost nothing and remove it.

Worth stating in the comment above that this path is now near-dead: the host reports a panel's width on the frame the group is created, so the guess only covers paints issued before that report is delivered.

---

sinelaw (commented):
Second pass, focused on checking the PR's own story against the code rather than on the diff.

**Both fixes are in the right place; one of the two diagnoses is described wrongly, and I verified each claim by running the code.**

What holds up:
- **The tree fix is exactly right.** The shell renderer that paints a panel mounted in a split (`view/shell/widgets.rs::live_selection`) reads the selection from a `Tree` instance state and otherwise falls back to the spec seed — a `List` entry hits the `_ => seed` arm, so the click's write really was invisible. `test_clicking_a_file_moves_the_sidebar_highlight` fails with `registry.rs` reverted to master and passes with it, so it pins the behaviour.
- **`None => (true, None)` fixes the reported symptom.** `test_files_panel_rows_fit_a_narrow_panel` fails with that one line reverted and passes with it.
- **Blast radius of firing on first sight looks contained.** Every `viewport_changed` listener in the tree (`dashboard`, `welcome_screen`, `markdown_compose`, `markdown_toc`, `code-tour`, `audit_mode`) guards on a dimension/layout-key change or on its own buffer id, so the extra birth event is at worst one redundant repaint; the hook is fired under a `try_read` with no mutable borrow of `self`, so there is no re-entrancy. A hidden panel does *not* get a bogus report — its leaf is absent from the visible grouped layout, so the `buffer_id` lookup drops the hook (confirmed by the probe test's `-1` assertion still holding). I ran the whole `e2e::plugins::` suite locally: 543 passed, 0 failed.
- **The plugin-side toolbar basis and the `stop_review_diff` clearing are harmless**; no reader iterates `panelWidths`' keys, and every reader guards on `> 0`.

What I'm raising, all verified rather than suspected:
1. The diagnosis in the render.rs comment / commit message / CHANGELOG is wrong about *why* the report went missing — the FILES panel is laid out once at group creation, at its final geometry, before the plugin hides it; it is that birth frame that was dropped, not a first layout "on the frame it is shown".
2. `buffer_group_panel_geometry` doesn't pin the contract it says it pins, for a specific reason, with two concrete ways to make it do so.
3. The panel's width is now announced exactly once, on a frame the plugin may not yet be subscribed for — adding one `await editor.flush()` before `openReviewPanels`' listener registration reproduces the original bug with this PR applied. Suggested one-line hardening in the host.
4. `panelWidthOf`'s new basis does not change the outcome in the case the new test covers; the `12` floor is the remaining sharp edge.

I also went hunting for a fifth issue and can report it as **not** a problem, because the earlier review comment on this PR claims it is: routing tree writes through `tree::resolve` does **not** freeze `expandedKeys` for the FILES sidebar — see my reply on that thread.

---

crates/fresh-editor-core/src/widgets/registry.rs
sinelaw: **Checked this one against the running editor and it does not reproduce — expansion is not frozen for this panel.** Posting the evidence rather than leaving the finding standing, because it would otherwise read as a blocker.

The reasoning above is correct about `registry.rs` and `tree::resolve`: after this change the first `set_selected_index` really does install a `Tree` entry whose `expanded_keys` is the spec seed of that moment, and `resolve_seeded` really does prefer it forever after. I instrumented both ends and saw exactly that:

```
SSI-TREE key=files-tree index=3 expanded=1
RENDERTREE files-tree prev_expanded=["dir:unstaged src/"] spec=["cat:staged", "dir:staged src/", "cat:unstaged", "dir:unstaged src/"] nodes=7
RENDERTREE-VIS files-tree visible=2
```

So `kinds::tree`'s collector does compute a collapsed tree from the frozen set. But that is not the code that paints a widget panel mounted into a split. That path is `crates/fresh-editor/src/view/shell/widgets.rs`, whose `WidgetSpec::Tree` arms take the selection from instance state (`live_selection`) and the expansion **from the spec**:

```rust
let sel_abs = live_selection(cx, key, *selected_index);
let expanded: HashSet<String> = expanded_keys.iter().cloned().collect();
let visible = crate::widgets::collect_visible_tree_indices(nodes, item_keys, &expanded);
```

and it says so in its own doc comment at line 483: *"A `Tree`'s `expanded_keys` is still read from the spec here, and agrees today because the plugins that drive it re-send the spec."* `audit_mode` re-sends `expandedKeys: filesTree.groupKeys` on every rebuild, so it is in exactly the group that keeps agreeing.

End-to-end check of the scenario in the comment — open a worktree review with 3 modified files (one category), click a file row, click the `src` directory row (so `pointSidebarAtCurrentFile` stops re-pinning), press `S` to stage:

```
▼ STAGED (1)
 ▼ src
    mod_002.rs M
▼ UNSTAGED (2)
 ▼ src
    mod_000.rs M
    mod_001.rs M
```

Everything open, nothing hidden. The filter case is the same code path.

The two sibling findings on this PR (the stale absolute index, and blur clearing the highlight) I *did* reproduce, both via `live_selection` — see my replies there.

---

crates/fresh-editor-core/src/widgets/registry.rs
sinelaw: **Reproduced end to end on this branch.** 5 modified files (`src/mod_000..004.rs`), review open, FILES sidebar shown, keys on the diff; click `mod_004.rs`, then write to `src/main.rs` (tracked, modified, sorts before every `mod_*`) and let the 2 s watch poll pick it up:

```
AFTER CLICK  selected=Some("   mod_004.rs M           ")
AFTER RESCAN selected=Some("   mod_003.rs M           ")
```

The highlight slides one row up onto the wrong file and stays there — `pointSidebarAtCurrentFile` early-returns because `state.filesCurrentKey` did not change, so nothing re-pins the stored absolute index. The mechanism is `view/shell/widgets.rs::live_selection`, which prefers the stored `Tree` entry's `selected_index` over the spec's key-derived seed.

Worth noting the cost of fixing it is small and local: `pointSidebarAtCurrentFile` could drop its `nodeKey === filesSelectedNodeKey` early-return (or re-pin whenever `state.dataRevision` moves), so a rebuild always re-asserts the index the key resolves to. That keeps the host change and restores the key-based semantics the sidebar was relying on.

---

crates/fresh-editor-core/src/widgets/registry.rs
sinelaw: **Reproduced, and it is a behaviour change against master.** Same setup (5 files, review open, sidebar shown, keys on the diff): click `mod_004.rs`, then click the panel's `[Filter files: ]` field.

On this branch:
```
AFTER FILE CLICK   selected=Some("   mod_004.rs M           ")
AFTER FILTER CLICK selected=None
```

With `registry.rs` alone reverted to master:
```
AFTER FILE CLICK   selected=Some("   mod_000.rs M           ")
AFTER FILTER CLICK selected=Some("   mod_000.rs M           ")
```

So master keeps a highlight (on the wrong row — that is the bug this PR fixes), and the branch has none at all. `selected_sidebar_row` reads the `popup_selection_bg` band inside the sidebar columns, so this is the band actually disappearing, not a test artefact.

Whether that is a regression depends on what the band means in this panel. `Tree::on_focus_change` clears a blurred tree's selection deliberately, to avoid a second focus ring — but in the FILES sidebar the band is also "the file you are reading", which is true whether or not the tree has the keyboard, and the plugin has no way to put it back. If the intent is to keep it, the narrow fix is plugin-side: have `pointSidebarAtCurrentFile` re-pin unconditionally (see the sibling thread — the same change also fixes the stale-index case).

---

sinelaw:
Acted on the review in 48a9573. Findings and what changed:

**Stale absolute index** and **blur clearing the highlight** — both real, both reproduced here. One cause: `pointSidebarAtCurrentFile` early-returned whenever the current file had not changed, so nothing re-asserted the index after a rebuild, and nothing put the band back after `Tree::on_focus_change` cleared it. It now re-pins from the key on every repaint. Two new guards, each verified to fail without the re-pin: `test_sidebar_highlight_survives_a_rebuild_above_it` (lands on `mod_003.rs`) and `test_sidebar_highlight_survives_focusing_the_filter` (band gone entirely).

**The mechanism in the comment was wrong** — accepted. The group is created with every panel visible and laid out, and the plugin hides the side panels afterwards; the dropped frame was the panel's birth, not a first layout on show. The `render.rs` comment, the CHANGELOG bullet and the test docs are rewritten to say that.

**One-shot announcement** — accepted, and the suggested hardening is in: `set_buffer_group_panel_visible` drops the leaf's snapshot when a panel becomes visible, so the next frame reports it and a plugin that missed the birth frame still gets a width.

**The geometry probe test** — deleted, along with its plugin. Both suggested strengthenings were tried and neither discriminates: with `side` visible at creation and `rail` hidden after, and again with nothing hidden at creation, the test still passes with either host rule reverted, because hiding a panel re-lays the group out and manufactures a change that is reported regardless. It pinned nothing, so it is gone rather than kept as decoration. `test_files_panel_rows_fit_a_narrow_panel` remains the guard and does fail without the host change.

**The `12` floor** — the guess is now clamped so it cannot claim more columns than the review has, and the comment says the path only covers paints issued before the host's first report.

**Frozen `expanded_keys`** — left as is, on your own evidence: the shell renderer reads `expanded_keys` from the spec, and `audit_mode` re-sends it every rebuild. That thread stays open rather than resolved, since nothing changed for it.

---

PR #3194, opened 2026-09-05
Welcome screen: keep the caret and the focus ring on the same thing

sinelaw:
The welcome page is a document with a real caret in it *and* a widget panel with a focus ring, and the two were independent. That gave the reader two "you are here" markers that could point at different things, and the disagreement was not cosmetic: Tab advanced focus down the page while the caret stayed on the top row, and an arrow key or a click moved the caret while `Enter` still fired whatever the last Tab had left focused — off screen, unasked for, in one case a persisted setting.

Focus and the caret are two answers to one question, so they are made one answer.

## What changed

A new `WidgetPanelOptions.focusFollowsCursor`, which the welcome page mounts with. The host then maintains both directions for a panel mounted into a buffer the reader moves a caret through:

* a focus move (Tab, Shift+Tab, a plugin's `setFocusKey`) seats the caret on the focused widget's own first cell — which reveals it, so the page no longer needs a reveal of its own;
* a caret move focuses the widget it landed on, or clears focus when it landed on prose, so `Enter` on a paragraph does nothing.

A focus region is a **control, not a row**: the hit areas already carry a byte span per row, and that span is the region. The three door cards share every one of their rows and the three verbs sit on one line, so a caret row cannot say which of them the reader is on — only a column can. It resolves *nearest* by the same distance function the click path already uses for two side-by-side lists, because a caret reading down a page keeps whatever column it was in, and that column is very often a framed card's border or the gap between two verbs.

Third commit: a blank row between each level banner's rule and the sentence under it. Set tight they read as one two-line heading — the sentence looked like a subtitle of the rule rather than the first line of the level it introduces.

## One handler per route, and there are two routes

The editor writes a caret through exactly two, and both now end in `sync_widget_focus_to_cursor`:

1. **The event pipeline** (`Editor::apply_event`'s `Event::MoveCursor`) — the same fork `cursor_moved` fires from, so every user-visible move travels it: arrows, page keys, a click on the text, Goto Line, navigate back/forward, a search hit, a diagnostic jump.
2. **The direct seating path** (`Editor::seat_buffer_cursor`) — the host and plugins putting a caret somewhere, which is how `setBufferCursor` and `scrollToWidget` work, and so how the page's own `1`/`2`/`3` level jumps do.

That second path used to be two copies of the same split-walk; `handle_set_buffer_cursor` and `ScrollAlign::Minimal`'s un-latch now share one function with it, so a third copy cannot drift. `column_distance` was lifted out of `row_select_hit` for the same reason — the click path and the caret path ask the same question about the same coordinates.

The gate in front of it (`WidgetRegistry::has_focus_follower`) allocates nothing and never touches the buffer: this runs on every cursor move in the editor, almost all of them in ordinary text buffers no panel is mounted into.

## Three things the host has to get right

All learned from this page.

**A focus region is a control, not a row** — see above. Resolving by row also made Tab between two controls of one row impossible: the move to the second seats the caret on the row they share, and the row hands focus straight back to the first.

**A card anchors at its top-left cell**, and seating asks whether the caret already *resolves* to the widget before moving it — not whether it is inside it. Arrowing *up* into a card's last row focuses the card, and seating on the card's top row would throw the reader back over everything they just walked past; and a caret arriving at a card heading resolves to the fold arrow at the left margin, so an "is it inside it" test would yank the caret leftwards at every card on the way down.

**Tab from nothing focused starts beside the caret.** "Nothing focused" is not this panel at rest — it is the caret on prose, which on a page that is mostly prose is most rows. Falling back to the ring's first entry sent every Tab from below the fold to the top of the document: read down to Level 3, press Tab, and you were back on the startup switch.

## Removed from the page

Three workarounds go with this: the page's own reveal on `focus`, its read-back of card-header rows, and `/` jumping to Level 1 to bring the finder field it had just focused back on screen — which would now move the caret off that field, taking the focus with it.

## Tests

Not run locally, by request — CI takes them.

* `tab_brings_the_caret_to_the_control_it_focused` — the second Tab must leave the caret on the first door card's own row.
* `tab_between_two_cards_on_one_row_carries_the_caret_across_it` — Tab from door 1 to door 2 moves the caret right along the row they share, and `Enter` then opens Level 2 rather than Level 1.
* `tab_after_reading_down_the_page_does_not_jump_back_to_the_top` — Tab from Level 3 stays in Level 3.
* `moving_the_caret_onto_prose_disarms_the_focused_control` — Tab onto the startup switch, click the tagline, press Space: the persisted setting must not change.
* `the_caret_arriving_at_the_finder_field_types_into_it` — walk down to the finder's input with the arrow key (counted from what is on screen, not from the card's structure) and type; the characters must land in the field.
* `a_level_banner_has_air_before_its_description` — the row under the Level 1 rule is blank and the description is two rows down.
* Four `WidgetRegistry` unit tests: two controls on one row are told apart by column and the nearest wins in the gap; a clickable non-stop row is not a focus region; `anchor_of_widget` is the first cell in paint order; the Tab ring seeds from the caret in both directions and wraps; a panel that did not opt in is invisible to all of it.

https://claude.ai/code/session_01T19PZp7YRwntSbMQdTMdHQ


crates/fresh-editor-core/src/widgets/registry.rs
```diff
@@ -1000,6 +1024,183 @@ impl WidgetRegistry {
...
+    /// An empty list is an answer, not the absence of one: a caret
+    /// parked on a paragraph means *nothing* is focused, and a caller
+    /// that read it as "leave focus alone" would keep the last Tab's
+    /// target armed under an Enter aimed at prose. `None` is reserved
+    /// for "no panel here asked for this".
+    ///
+    /// Focusable, not merely hit-testable: `tabbable` is the panel's own
+    /// Tab ring, so a row of a list whose entries are clickable but not
+    /// stops (the finder's results) carries no candidate, exactly as Tab
+    /// would never land there.
```
sinelaw: Two doc comments are concatenated here, and the first one documents a function that does not exist.

Lines 1027-1050 describe a *list*-returning API ("every widget of it a caret on `row` could be said to be on — left to right", "The list is normally empty or one long", "An empty list is an answer"). The item they are attached to is `has_focus_follower`, which returns `bool`. Its real doc starts at line 1051 ("Whether any panel mounted into `buffer_id`…"). The list version was evidently an earlier draft of `focus_target_at`, which now lives at 1096 with its own (row+column) doc.

Rustdoc will render the whole 24-line stale block as `has_focus_follower`'s documentation. Delete 1027-1050.

---

crates/fresh-editor-core/src/widgets/registry.rs
```diff
@@ -1000,6 +1024,183 @@ impl WidgetRegistry {
...
+    ) -> Option<(PanelKey, String)> {
+        let panel_key = self
+            .panels_for_buffer(buffer_id)
+            .into_iter()
+            .find(|pk| self.get(pk).is_some_and(|p| p.focus_follows_cursor))?;
+        let panel = self.get(&panel_key)?;
+        let mut best: Option<(&HitArea, usize)> = None;
+        for hit in &panel.hits {
+            if hit.buffer_row != row || !panel.tabbable.contains(&hit.event.widget_key) {
+                continue;
```
sinelaw, line 1122: The doc above claims "Focusable, not merely hit-testable… a row of a list whose entries are clickable but not stops (the finder's results) resolves to nothing". That is not what this filter tests. `panel.tabbable` holds **widget keys**, so the predicate is "does *some* hit with this key belong to the ring", not "is *this* hit focusable".

The welcome screen depends on the difference. `doorRow()` emits one `button` per row of a door card, all seven sharing `key: jump:${d.n}`, with `focusable: verb` — true on exactly one row. Because the key is in `tabbable`, all seven rows resolve here as focus regions. That is what makes "caret anywhere in the card focuses the card" work, and it works by accident of key reuse, not by anything this code or the doc says. The finder's results only fall out because they happen to be keyed `hit:${i}` individually.

Consequence: give the door card's focusable row its own key (a perfectly ordinary refactor) and six-sevenths of every card silently stops being a focus region, with no test naming the invariant. Either filter on the hit's own focusability (record it on `HitArea` at collect time) or state plainly in the doc that focusability is per key and that a non-focusable widget sharing a focusable sibling's key is deliberately a focus region.

---

crates/fresh-editor-core/src/widgets/registry.rs
```diff
@@ -1000,6 +1024,183 @@ impl WidgetRegistry {
...
+            }
+            let d = column_distance(hit, col_byte);
+            if best.is_none_or(|(b, bd)| d < bd || (d == bd && hit.byte_start < b.byte_start)) {
+                best = Some((hit, d));
+            }
+        }
+        let key = best
```
sinelaw: "Nearest" here is unbounded, which makes the headline claim of the doc — "**A focus region is the widget's own painted span, not its row**" — false for the common case.

There is no maximum distance. On any row that carries exactly one tabbable hit, `best` is always `Some`, so the focus region *is* the whole row, however far away the caret is. A caret at column 0 of a row whose only control is painted at columns 60-70 focuses that control and arms it under Enter. The doc justifies unboundedness with "a caret in the gutter of a framed card or in the gap between two verbs" — a gap of one or two columns — but nothing in the code limits it to that.

This also weakens the other stated guarantee, that "an empty string is an answer": for a single-control row it can never be the answer, so a reader whose caret is 50 columns from the only button on the line still has it armed.

Give `column_distance` a cap (a few columns, or the widget's own height/width neighbourhood) and return `""` past it, and add a test for "caret far from the only control on the row focuses nothing".

---

crates/fresh-editor-core/src/widgets/registry.rs
```diff
@@ -1000,6 +1024,183 @@ impl WidgetRegistry {
...
+    /// anchors at its own first column rather than at the row's.
+    pub fn anchor_of_widget(&self, buffer_id: BufferId, key: &str) -> Option<(u32, usize)> {
+        self.panels_for_buffer(buffer_id)
+            .into_iter()
+            .filter_map(|pk| self.get(&pk))
+            .flat_map(|p| p.hits.iter())
+            .filter(|h| h.event.widget_key == key)
+            .map(|h| (h.buffer_row, h.byte_start))
+            .min()
+    }
```
sinelaw: Three new lookups, three different and inconsistent notions of "which panel are we talking about":

- `focus_target_at` (1101): `panels_for_buffer(...).find(|pk| focus_follows_cursor)`
- `tabbable_from_caret` (1163): the same `find`
- `anchor_of_widget` (here): **no** filter at all — it flat-maps hits across *every* panel mounted into the buffer and takes the global min.

Two problems. (1) `panels_for_buffer` collects from a `HashMap`, so its order is arbitrary; with two focus-follows panels on one buffer, `find` picks a nondeterministic one, and it can pick a different one on two consecutive calls in the same frame. (2) `anchor_of_widget` will happily return a hit from a panel that never asked for `focusFollowsCursor`, or from another panel that reuses the key — so `seat_cursor_on_focused_widget` can seat the caret using coordinates from a panel other than the one whose focus just moved.

None of the three takes the `PanelKey` the caller already has. `seat_cursor_on_focused_widget` and `sync_widget_focus_to_cursor` both know the panel; pass it in and drop the re-derivation, or make `panels_for_buffer` deterministic and assert there is at most one focus-follower per buffer.

---

crates/fresh-editor/src/app/widget_runtime.rs
```diff
@@ -1557,6 +1579,218 @@ impl Editor {
...
+        let position = view_states.values().find_map(|vs| {
+            vs.buffer_state(buffer_id)
+                .map(|bs| bs.cursors.primary().position)
+        })?;
+        let (line, col) = self
+            .active_window()
+            .buffers
+            .get(&buffer_id)
+            .map(|st| st.buffer.position_to_line_col(position))?;
+        Some((line as u32, col))
```
sinelaw, line 1868: `view_states` is a `HashMap<LeafId, SplitViewState>` (`Buffers::split_view_states`), so `values().find_map(...)` returns an arbitrary split's caret. The doc says "from whichever split is carrying that buffer's cursor" as though there were one; each split showing a buffer has its own `cursors`, and they can differ.

Two concrete consequences with the page open in two splits:

- `seat_cursor_on_focused_widget`'s guard (1610-1618) may consult split B's caret while the reader is Tabbing in split A, so it either skips a seat it should make or makes one it should not — depending on hash iteration order, i.e. it can differ between runs.
- `handle_widget_focus_advance`'s new seed (1248-1258) can seed split A's Tab from split B's caret, so Tab in the split you are looking at jumps to a widget beside the *other* pane's caret.

And `seat_buffer_cursor` then writes the result to *every* split showing the buffer (1710-1712), so a Tab in one pane yanks the other pane's caret too.

This needs the leaf id, not a `find_map`: take the effective active split (as `apply_event_to_active_buffer` does) and read that split's cursor, and consider scoping the write the same way.

---

crates/fresh-editor/src/app/widget_runtime.rs
```diff
@@ -1557,6 +1579,218 @@ impl Editor {
...
+
+    /// Put `buffer_id`'s caret at `position` in every split showing it.
+    ///
+    /// The editor writes a caret through exactly two routes — the event
+    /// pipeline (`Editor::apply_event`, which every user-visible move
+    /// travels: arrows, page keys, a click, Goto Line, back/forward,
+    /// a search hit) and this one, for the host and plugins seating it
+    /// directly. Both end in `sync_widget_focus_to_cursor`, so a
+    /// `focusFollowsCursor` panel cannot be reached by a caret move that
+    /// leaves its focus behind.
```
sinelaw: "The editor writes a caret through exactly two routes … Both end in `sync_widget_focus_to_cursor`, so a `focusFollowsCursor` panel cannot be reached by a caret move that leaves its focus behind."

There are at least four, and two of them bypass the sync:

- `plugin_dispatch.rs:3705` — `CreateVirtualBufferWithContent`'s `initial_cursor_line` calls `set_buffer_cursor_in_splits` directly (and with `splits_for_buffer` only, no grouped subtrees).
- `plugin_dispatch.rs:4118` — the display-virtual-buffer path, also direct.

Both write a caret into a buffer that can perfectly well have a focus-follows panel mounted (the welcome screen *is* a plugin virtual buffer), and neither calls `sync_widget_focus_to_cursor`, so the invariant this doc asserts as a guarantee does not hold. Either route them through `seat_buffer_cursor` — which is exactly what `handle_set_buffer_cursor` was changed to do — or drop the "exactly two routes" claim and say the sync is best-effort.

An unverifiable "this is the only route" comment is worse than none: the next caret-writing call site will be added by someone who read it and believed it.

---

crates/fresh-editor/src/app/widget_runtime.rs
```diff
@@ -1557,6 +1579,218 @@ impl Editor {
...
+                        .get(&inner_leaf)
+                    {
+                        if vs.active_buffer == buffer_id && !splits.contains(&inner_leaf) {
+                            splits.push(inner_leaf);
+                        }
+                    }
+                }
+            }
+        }
+        splits
```
sinelaw, line 1748: This is a verbatim copy of `plugin_dispatch.rs:4095-4116`, which the PR leaves in place. The PR removed one copy (from `handle_set_buffer_cursor`) and added this one, so the count of "the split walk for a buffer, including grouped subtrees" went from 2 to 2, not to 1 — and `plugin_commands.rs:1229` now advertises the helper as "shared with the widget runtime's own caret moves so the two cannot drift" while a third copy sits untouched in `plugin_dispatch.rs`.

Delete the `plugin_dispatch.rs` copy and call `self.splits_showing_buffer(buffer_id)` there too. (`plugin_dispatch.rs:3696-3702` is a fourth, grouped-subtree-less variant of the same walk that should probably become this call as well.)

Also, inside the loop this re-resolves `self.windows.get(...).buffers.splits()` and re-`expect`s it once per inner leaf, having already resolved it above — hoist it.

---

crates/fresh-editor/src/app/widget_runtime.rs
```diff
@@ -1244,13 +1244,31 @@ impl Editor {
...
             }
-            None if delta >= 0 => ring[0].clone(),
-            None => ring[(n - 1) as usize].clone(),
+            None => unfocused_seed(self).unwrap_or_else(|| {
+                if delta >= 0 {
+                    ring[0].clone()
+                } else {
+                    ring[(n - 1) as usize].clone()
+                }
+            }),
```
sinelaw: `unwrap_or_else` here collapses four different `None`s into one silent "jump to the top of the document":

1. this panel did not ask for `focusFollowsCursor` (the correct, expected case — every other panel in the editor);
2. `panel_buffer_caret` found no split carrying the buffer;
3. `tabbable_from_caret` found a focus-follows panel, but it was a *different* one on the same buffer (it re-derives the panel and ignores `panel_key` entirely — see the registry comment);
4. no hit in `panel.hits` matches any key in `ring`, i.e. the ring and the last paint disagree.

Cases 2-4 are bugs, and the fallback makes them look exactly like case 1 — a reader below the fold pressing Tab and landing on the startup switch, which is precisely the symptom the new seeding was written to eliminate. The author would never find out.

Split the "did not opt in" answer from the failure answers: have `tabbable_from_caret` take the `PanelKey` and return `Option<Option<String>>`, or check `panel.focus_follows_cursor` at this call site and `tracing::warn!` on the remaining `None`s.

---

crates/fresh-editor/src/app/widget_runtime.rs
```diff
@@ -1543,6 +1561,10 @@ impl Editor {
         // The tree is the ring; the registry is its mirror. Written in step,
         // before the plugin is told, so nothing reads one against the other.
         self.focus_panel_widget_in_tree(panel_key, &new_key);
+        // On a `focusFollowsCursor` panel the caret comes along, before
+        // the plugin hears about the move — so a `focus` handler that
+        // reveals the widget is revealing a row the caret is already on.
+        self.seat_cursor_on_focused_widget(panel_key, &new_key);
```
sinelaw: This makes `set_panel_focus_and_notify` mutually recursive with itself — `set_panel_focus_and_notify` → `seat_cursor_on_focused_widget` → `seat_buffer_cursor` → `sync_widget_focus_to_cursor` → `set_panel_focus_and_notify` — and nothing in the code or the comments says so.

It does terminate (the inner call's guard at 1610-1618 compares against the caret the outer call just wrote), but the re-entrancy is placed *between* the registry write and the `focus` event, so when the inner call fires the outer one is mid-flight. Two observable consequences whenever `focus_target_at(anchor_of(X)) != X`:

- Plugins receive the `focus` events in reverse order: the inner (final, correct) key first, then the outer (already superseded) key. Any plugin mirroring focus from these events — `welcome_screen.ts` does exactly this at line 2333 — ends up holding the stale key while the registry holds the new one.
- `notify_widget_focus_change(old_key, new_key)` at 1573 then runs with a `new_key` that is no longer the focus key.

`anchor_of_widget` and `focus_target_at` are not inverses (the first ignores `tabbable` and searches all panels, the second breaks column ties leftmost), so the disagreeing case is not hypothetical.

At minimum, seat the caret *after* the notify/event pair, and add a re-entrancy guard flag with a `debug_assert!` so a second-level entry is loud rather than silently reordering events.

---

crates/fresh-editor/src/app/event_apply.rs
```diff
@@ -212,6 +212,30 @@ impl Editor {
...
+                    .rev()
+                    .find(|e| matches!(e, Event::MoveCursor { .. }))
+                {
+                    let buffer_id = self.active_buffer();
+                    let position = *new_position;
+                    self.sync_widget_focus_to_cursor(buffer_id, position);
+                }
+            }
+            _ => {}
+        }
```
sinelaw: The comment says "every way the caret moves … lands here". This match sees two of them.

- Nested batches are missed: the `find` is over `events.iter()` one level deep, so a `Batch` containing a `Batch` whose last event is a `MoveCursor` syncs nothing.
- Edits are missed: `Insert`, `Delete`, `BulkEdit` all move the caret and none of them reach this. The welcome screen's buffer is read-only so it does not bite today, but `focusFollowsCursor` is a public `WidgetPanelOptions` field — the first panel mounted into an editable buffer gets focus that silently stops tracking the caret as soon as the user types.
- It syncs to the event's `*new_position` rather than to the cursor position that actually resulted. `apply_event_to_keyed_buffer` has already run above, so the real caret is available and may differ (clamping, virtual space).

The robust shape is to read the resulting primary cursor position once, after step 1, and sync from that for any event where it changed — which also collapses the two arms into one and removes the nested-batch hole.

Separately: this block is labelled "2b" but sits after step 3 (search-overlay re-evaluation) and immediately before the block labelled "3". Renumber.

---

crates/fresh-editor/src/app/window/mod.rs
```diff
@@ -1911,7 +1911,16 @@ impl Window {
...
+                    let cursor = view_state.cursors.primary_mut();
+                    cursor.move_to(position, false);
+                    // An absolute placement is a jump, not a vertical move, so
+                    // it clears the goal column — the same thing every other
+                    // jump does (`new_sticky_column: None` on a diagnostic
+                    // jump, a search hit, a click). Without this the *next*
+                    // Up/Down snapped the caret back to a column the reader
+                    // left long ago: Tab onto the third door card seated the
+                    // caret on it and one Down threw it back into the first.
+                    cursor.sticky_column = None;
```
sinelaw, line 1938: This changes behaviour for callers that did not ask, and the loop it sits in is the reason.

`set_buffer_cursor_in_splits` applies to **every** split showing the buffer, not just the one being seated. So a plugin `SetBufferCursor` (or the new `seat_buffer_cursor`) now wipes the goal column of a split the user is actively arrowing through in another pane: they press Down, Down, Down through a ragged file in pane B, a plugin seats the caret in pane A, and their next Down lands at a different column. Before this commit that could not happen.

The justification given is entirely about the welcome screen's own Tab-then-Down. That is a real bug, but the fix belongs where the jump is made (`seat_buffer_cursor`, for the split it is seating), not in the shared multi-split writer.

If it must stay here, the argument to make is why "an absolute placement is a jump" is true for *all four* current callers including the two `plugin_dispatch` ones — and that argument is not in the comment.

---

crates/fresh-editor/src/app/sidebar.rs
```diff
@@ -518,18 +522,40 @@ impl super::Editor {
...
+    /// dock or centred panel into the column all say so out loud. A
+    /// quiet mount waits in a hidden column and is there the moment the
+    /// reader opens it.
+    pub(crate) fn reveal_sidebar(&mut self) {
+        if self.active_window().file_explorer_visible {
+            return;
+        }
+        self.active_window_mut().file_explorer_visible = true;
+        if self.file_explorer().is_none() {
+            self.init_file_explorer();
```
sinelaw, line 562: A global behaviour change for every plugin, argued from one page's layout bug, and the argument does not hold up.

The stated cause is "The pane it stole those columns from was mid-layout: the welcome screen had already composed against the full width, so its cards fell out of their row and its rows wrapped." But the welcome screen already listens to `viewport_changed` and repaints when its `layoutKey` moves (`welcome_screen.ts:1896-1932`), and the section is mounted from `buffer_activated` — i.e. while the page is the active buffer of its split, which is exactly when `viewport_changed` is delivered to it. If the column opening really did leave the page composed at the old width, the bug is a missing or mis-ordered repaint on that path, and it will bite again the next time the pane changes width for any other reason (opening the dock, a vertical split).

What this commit does instead is make "mount a section into a hidden column" silently produce nothing visible, for every plugin, forever — the classic shape of a fallback that hides a bug. `markdown_toc` mounts `startBlurred: true`, so its section now goes nowhere on a closed sidebar; that may well be desirable, but it is a separate product decision landing as a side effect.

Two asks: (a) confirm the layout corruption is not simply a missed repaint before making this global; (b) add a test for the claim in the last paragraph — "waits in a hidden column and is there the moment the reader opens it" — because nothing in this PR covers mount-while-hidden-then-reveal.

---

crates/fresh-editor/plugins/welcome_screen.ts
```diff
@@ -1926,38 +2274,72 @@ function openFinderHit(index: number): void {
...
-    // label you can no longer read — which also stopped Tab and
-    // Shift+Tab being inverses, because a jump forgets where it came
-    // from.
-    //
-    // A minimal reveal has none of those cases. There is nothing left
-    // for the page to know about its own layout.
-    if (bufferId !== null && k.length > 0) {
-      editor.scrollToWidget(bufferId, k, "minimal");
-    }
+    lastFocusedWidget = k;
```
sinelaw: The bug the new comment describes is still here, one line above it.

Line 2324 (`if (k.length > 0) lastFocusedWidget = k;`) is unchanged and still runs for **every** event type before the `focus` branch. Line 2333 only fixes the `focus` case. So there are now two writers of the same mirror, and the unguarded one fires on `activate`, `select`, `change` and `cancel` too.

That matters because the page has widgets the host never focuses. `doorRow` sets `focusable: false` on six of a card's seven rows, and `find_widget_by_key` (host, `actions.rs:18`) returns the *first* match for a key — the first door row — so the click-to-focus check in `deliver_widget_hit` finds a non-focusable widget and does **not** move focus. Clicking a door card therefore fires `activate` with `widget_key = "jump:1"`, line 2324 sets `lastFocusedWidget = "jump:1"`, and the host's focus key is unchanged. Same for a click on a finder result (`focusable: false`, keyed `hit:${i}`).

Result: the mirror names a widget nothing is focused on — the exact state the comment says was fixed. If host focus was `finderField` at the time, `finderFocused()` now returns false while the field still has focus and still takes typed characters, so Up/Down silently stop walking the results.

Delete line 2324 and let the `focus` branch be the sole writer.

---

crates/fresh-editor/plugins/welcome_screen.ts
```diff
@@ -1926,38 +2274,72 @@ function openFinderHit(index: number): void {
...
+    return true;
+  }
+  return args.event_type === "focus" || args.event_type === "blur";
+}
+
 editor.on("widget_event", (args) => {
+  if (args.panel_id === CONTENTS_PANEL_ID) {
+    handleContentsEvent(args);
+    return;
+  }
```
sinelaw, line 2346: `handleContentsEvent` is documented as "Returns true when it handled one, so the page's handler can stay about the page", and its last line computes `args.event_type === "focus" || args.event_type === "blur"` for that purpose — but the caller discards the value and returns unconditionally. The boolean is dead, and the doc describes a contract nothing uses.

Either drop the return type and the final expression (the routing is already decided by `panel_id`), or actually use it:

```ts
if (args.panel_id === CONTENTS_PANEL_ID) {
  if (!handleContentsEvent(args)) { /* log the unhandled type */ }
  return;
}
```

As written, an event type the Contents section starts emitting that nobody handles is swallowed with no trace.

---

crates/fresh-editor/plugins/welcome_screen.ts
```diff
@@ -1342,7 +1493,142 @@ function buildSpec(): WidgetSpec {
  *  whole document up the screen. */
 function render(): void {
   if (!panel) return;
+  outline = [{ label: "Top of the page", key: "startupToggle", depth: 0 }];
   panel.set(buildSpec());
```
sinelaw: "A table is a second copy of the outline… letting it say what it is building costs one push per section and cannot disagree with the page." Except for this line, which is the table: the first outline entry is hand-written here, and its `key: "startupToggle"` is a string that must match a widget key defined in `hero()`. Rename that widget and the Contents section's first row points at nothing — and `scrollToWidget` on an unknown key does nothing visible, so the failure is silent. That is precisely the failure mode the paragraph above claims to have designed out.

Two more consequences of building the outline as a side effect of `buildSpec()`:

- `outlineSelected` is an **index** into a list that is discarded and rebuilt on every `render()` — and `render()` runs on every keystroke in the finder. It happens to be stable only because every `card()`/`banner()` call is unconditional today. Add one card behind `if (repoFiles !== null)` and the mark silently points at a different section. Store the widget key and derive the index at paint time.
- `card()` pushes before it calls `body()`, so a nested `card()` inside a body would be recorded out of document order. Nothing does that yet; nothing stops it either.

Have `hero()` push its own entry, the way `card()` and `banner()` do.

---

crates/fresh-editor/plugins/welcome_screen.ts
```diff
@@ -1610,8 +1915,18 @@ let lastKey = "";
...
+  // a source file would be a table of contents for something that is not
+  // on screen — and it would sit where the Markdown outline goes, which
+  // is the same column and the same idea.
+  if (bufferId !== null && e.buffer_id !== bufferId) {
+    unmountContents();
+    return;
+  }
   if (bufferId === null || e.buffer_id !== bufferId) return;
   if (editor.getActiveBufferId() !== bufferId) return;
+  mountContents(false);
```
sinelaw: Three things in seven lines.

1. Line 1927 (`if (bufferId === null || e.buffer_id !== bufferId) return;`) now has a dead disjunct: the new guard above already returned for every case where `bufferId !== null && e.buffer_id !== bufferId`. Only the `bufferId === null` half can still fire. Collapse it.

2. Unmount/remount churn. Every switch away from the page destroys the section and every switch back re-`mountSidebarSection`s it. `markdown_toc` deliberately avoids exactly this (`markdown_toc.ts:406-420`: update in place, remount only when the title changes) because a remount is a re-mount — it re-derives rows, re-inserts into the sidebar order, and resets state. With this PR a remount also re-runs the `startBlurred` decision. Switching between the welcome page and a file repeatedly will shuffle sidebar section order against `markdown_toc`'s section.

3. The mount is silent on failure. `mountContents(false)` mounts blurred, and `reveal_sidebar` is now deliberately not called for a blurred mount — so if the column is closed the reader gets no outline and no indication one exists other than the `Contents` button further up a page they may have scrolled past. Worth at least a comment saying so, given the button is the only recovery.

---

crates/fresh-editor/tests/e2e/plugins/welcome_screen.rs
```diff
@@ -439,3 +549,447 @@ fn opening_a_file_leaves_a_welcome_page_the_reader_was_looking_at() {
...
+        .send_key(KeyCode::Char('z'), KeyModifiers::NONE)
+        .unwrap();
+    harness.wait_for_async_quiescence(4).unwrap();
+
+    let screen = harness.screen_to_string();
+    let field = screen
+        .lines()
+        .find(|l| l.contains("find ["))
+        .unwrap_or_default()
+        .to_string();
```
sinelaw, line 889: This test passes vacuously if the caret never reaches the finder at all.

`put_the_caret_in_the_finder` asserts only that the caret starts above the field; it never verifies arrival. The `wait_until(contains("README.md"))` here does not verify it either — `finderCard()` renders its hit rows unconditionally, with no query and regardless of focus (`welcome_screen.ts:881-945`), so the repo's files are listed whether or not the field has focus. And the final assertion is a negative (`!field.contains('z')`), which is trivially satisfied when nothing ever put the caret in the field.

So the exact regression this test is named for — "the caret arrives at the field, Down cycles the list forever" — would still be caught, but every *other* way of failing to get there (an off-by-one in the row count, the field scrolling out of view, focus never moving) turns the test green.

Fix: assert the caret is in the field before the two Downs — type a character, assert it appears between the brackets, then Backspace it out — so the test's premise is established rather than assumed.

---

crates/fresh-editor/tests/e2e/plugins/welcome_screen.rs
```diff
@@ -64,6 +91,99 @@ fn open_welcome(harness: &mut EditorTestHarness) {
...
+        .expect("the finder field is on screen");
+    let caret_row = caret_position(harness).1;
+    assert!(
+        caret_row < field_row,
+        "expected the caret above the finder field, got {caret_row} and {field_row}"
+    );
+    for _ in 0..(field_row - caret_row) {
+        harness.send_key(KeyCode::Down, KeyModifiers::NONE).unwrap();
+    }
+    harness.wait_for_async_quiescence(4).unwrap();
```
sinelaw: `field_row - caret_row` is a count of **screen** rows, used as a count of Down presses, which move by **buffer** lines. The two are only equal while nothing between the caret and the field wraps and the pane does not scroll. Both are realistic on this page: it is composed to `measure()` and the pane scrolls as soon as the caret nears the bottom, at which point `field_row` is stale for the remaining presses.

The helper then returns without checking anything, so a wrong count is invisible here and shows up as a confusing failure (or a vacuous pass) in each of the four tests that use it.

Loop instead of counting: press Down until the caret's row equals the field's row (re-reading both each iteration), with a bound and a panic naming where it stopped — the same shape `scroll_until` already uses.

---

crates/fresh-editor/tests/e2e/plugins/welcome_screen.rs
```diff
@@ -439,3 +549,447 @@ fn opening_a_file_leaves_a_welcome_page_the_reader_was_looking_at() {
...
+            .unwrap();
+    }
+    harness.wait_for_async_quiescence(4).unwrap();
+
+    assert_eq!(
+        caret_readout(&harness),
+        before,
+        "a printable key still moves the caret. Screen:\n{}",
+        harness.screen_to_string()
+    );
```
sinelaw, line 1026: The observable ("the caret did not move") does not cover the claim ("the page binds no printable key") for two of the six keys under test.

Space is the clear one. Re-add `["Space", "welcome_space"]` and this test still passes: the caret is on prose here so nothing is focused, `activateFocused()` returns false, `dispatch(widgetKey("Space"))` reaches a panel with no focused widget, and the caret does not move. The key would be bound again — stealing the character from the finder, which is the entire point of removing it — and the test would be green.

`/` is similar in the reverse direction: it passes today because `welcome_focus_find` used to call `jumpTo("1")`; a regression that re-binds `/` to focus the field *without* jumping also stays green.

Test the property directly rather than a side effect of it: put the caret in the finder field, type all six characters, and assert every one of them lands between the brackets. That is the thing that actually breaks when a printable key is bound, and it holds for Space and `/` as well as the digits.

---

crates/fresh-editor/tests/e2e/plugins/welcome_screen.rs
```diff
@@ -64,6 +91,99 @@ fn open_welcome(harness: &mut EditorTestHarness) {
...
+        if let Some(pos) = harness.render_observing_cursor().unwrap() {
+            return pos;
+        }
+        harness.wait_for_async_quiescence(1).unwrap();
+    }
+    panic!(
+        "no frame drew the caret — the page is scrolled away from it. Screen:\n{}",
+        harness.screen_to_string()
+    );
+}
```
sinelaw, line 128: Two problems with this helper, both of which make failures point at the wrong thing.

1. It returns the **first** frame that draws a caret, not a settled one. The docstring's premise is that "a frame drawn while the page is repainting can have the caret off-view and hidden" — but the converse is equally true: a frame drawn before an async repaint lands can show the caret at its *old* position. Every caller uses the returned coordinates as the assertion subject (`tab_brings_the_caret_to_the_control_it_focused`, `tab_between_two_cards_on_one_row_carries_the_caret_across_it`, `a_caret_seated_by_tab_keeps_its_column_on_the_next_arrow_key`), so this can pass or fail on a stale frame. Callers do `wait_for_async_quiescence(4)` first, which mostly hides it — meaning the retry loop is doing nothing in the happy path and papering over a race in the unhappy one.

2. The panic asserts a cause it has not established: "no frame drew the caret — the page is scrolled away from it". A hidden caret can equally mean focus went to the sidebar or the dock, the pane lost the buffer, or the render never happened. Say what was observed, not why.

If a bounded wait is wanted, wait for the caret to be *stable* across two consecutive frames rather than merely present.

---

crates/fresh-core/src/api.rs
```diff
@@ -1860,6 +1860,29 @@ pub struct WidgetPanelOptions {
...
+    /// key, a click on the text) focuses the widget on the row it landed
+    /// on — or clears focus, when the row carries none. "Nothing
+    /// focused" is a state this option produces constantly, so a panel
+    /// declaring it almost certainly wants `autoFocusFirst: false` too.
+    ///
+    /// `None` reads as `false`: every panel written before this field
+    /// keeps focus and cursor independent.
+    #[serde(default)]
+    #[ts(optional)]
+    pub focus_follows_cursor: Option<bool>,
```
sinelaw, line 1890: "a panel declaring it almost certainly wants `autoFocusFirst: false` too" understates this: the combination `focusFollowsCursor: true` + `autoFocusFirst: true` (the *default*, since `auto_focus_first()` is `unwrap_or(true)`) is actively broken, and nothing rejects it.

`autoFocusFirst` re-seeds focus onto the first tabbable widget **on every repaint** — that is the behaviour `welcome_screen.ts` documents at its `WidgetPanel` construction, and the reason it passes `false`. Combine that with this option and every repaint now also drags the caret: repaint → focus := ring[0] → `set_panel_focus_and_notify` → `seat_cursor_on_focused_widget` → the reader's caret teleports to the top of the document. A plugin repainting on a timer or on `git status` landing would move the user's cursor out from under them, and the plugin author's only warning is one clause of a prose paragraph.

Make it enforced rather than advised: have `focus_follows_cursor()` returning true imply `auto_focus_first() == false`, or reject the combination at mount with a clear error.

---

crates/fresh-editor/src/app/widget_runtime.rs
```diff
@@ -1557,6 +1579,218 @@ impl Editor {
...
+        byte_in_row: usize,
+    ) -> Option<usize> {
+        let state = self.active_window().buffers.get(&buffer_id)?;
+        let start = state.buffer.line_start_offset(row as usize)?;
+        let end = match state.buffer.line_start_offset(row as usize + 1) {
+            // Less the newline that begins the next row.
+            Some(next) => next.saturating_sub(1),
+            None => state.buffer.len(),
+        };
+        Some((start + byte_in_row).min(end))
```
sinelaw, line 1719: This clamp is silent, and when it fires it does not do what the caller asked: the caret is seated at end-of-row instead of on the widget.

That is not a harmless approximation here, because the seated position is immediately fed back through `sync_widget_focus_to_cursor`, which resolves focus by *nearest column*. On a row carrying several controls, end-of-row is nearest the rightmost — so if `byte_in_row` for the third door card exceeds the stored row length, Tab focuses card 3, the caret is clamped to end-of-row, the sync re-resolves, and focus lands wherever "rightmost on the row" happens to be. Focus and the caret both end up somewhere the user did not select, with nothing logged.

Whether it fires today depends on whether the panel's `raw` rows are stored padded to the pane width — which is a property of the renderer, not of anything asserted here. That is exactly the "happens to work because of something elsewhere" shape.

Add a `tracing::warn!` (or `debug_assert!`) when `start + byte_in_row > end`: either the hit geometry and the buffer text have drifted, or a row really is unpadded, and both are things the author should hear about rather than absorb.

---

sinelaw (commented):
Reviewed the whole diff against the surrounding code. The feature works, and the prose is persuasive, but several of the load-bearing claims in the comments do not survive checking against the code. Themes:

**Comments that assert invariants the code does not hold.** `seat_buffer_cursor`'s "exactly two routes … both end in `sync_widget_focus_to_cursor`" is false — `plugin_dispatch.rs:3705` and `:4118` write carets directly. `focus_target_at`'s "focusable, not merely hit-testable" is a *key*-level test, not a hit-level one, and the door cards work only because of that. "A focus region is the widget's own painted span, not its row" is not true for any row with a single control, because the nearest-match has no distance bound. And `registry.rs` carries a 24-line doc block for a list-returning function that no longer exists, glued to the front of `has_focus_follower`.

**Undeclared re-entrancy.** `set_panel_focus_and_notify` is now mutually recursive with itself via `seat_cursor_on_focused_widget` → `seat_buffer_cursor` → `sync_widget_focus_to_cursor`. It terminates, but the recursion sits *between* the registry write and the `focus` event, so when `anchor_of_widget` and `focus_target_at` disagree — and they are not inverses — plugins get the focus events in reverse order and the plugin-side mirror ends up stale.

**Nondeterminism.** `panel_buffer_caret` picks an arbitrary split's caret out of a `HashMap`; `panels_for_buffer(...).find(...)` picks an arbitrary focus-follower panel; `anchor_of_widget` searches every panel on the buffer with no filter at all. Three new lookups, three different notions of "which panel", none of them taking the `PanelKey` the caller already has.

**Fallbacks that erase the difference between "not applicable" and "broken".** The `unwrap_or_else` in the Tab seed collapses four distinct `None`s into the very jump-to-top behaviour the seeding was written to prevent. `buffer_offset_in_row`'s clamp silently relocates a caret whose landing position then re-drives focus resolution.

**Scope creep, argued from one page.** `reveal_sidebar` changes behaviour for every plugin, justified by a welcome-screen layout symptom that the page's own `viewport_changed` repaint should already cover, and `sticky_column = None` lands in the shared multi-split cursor writer rather than at the jump site.

**Tests.** `walking_down_the_page_does_not_get_stuck_in_the_finder` passes vacuously if the caret never reaches the field (the finder renders its hits regardless of focus). `the_page_binds_no_printable_key` does not cover Space, the key most worth covering. `put_the_caret_in_the_finder` converts screen rows into keypress counts without verifying arrival, and four tests depend on it.

Individual comments have specifics and suggested fixes.

---

crates/fresh-editor-core/src/widgets/registry.rs
sinelaw: Fixed — the stale 24-line block is deleted. It was an earlier draft of `focus_target_at`'s doc, left concatenated onto `has_focus_follower` when that function was split out.

---

crates/fresh-editor-core/src/widgets/registry.rs
sinelaw: Right, and I've stated it rather than filtered it. Focusability is per **key**, and that is load-bearing: a card several rows tall becomes one focus region precisely because its rows share one key with `focusable` set on the row that names the action. Filtering per hit would break that, and there is no other mechanism for a multi-row region.

So the doc says so plainly now instead of claiming "focusable, not merely hit-testable", and `focus_regions_are_per_key_not_per_hit` pins the invariant you point out had no test — split those rows into separate keys and it fails.

---

crates/fresh-editor-core/src/widgets/registry.rs
sinelaw: Half accepted: the doc's headline claim was false and is gone. The cap is not, and here is why.

A cap has to separate two cases on this page: a caret at column 0 of a framed card's row, where the field's span starts at column ~17 (must resolve to the field, or arrowing down through the card focuses nothing in it and typing goes nowhere), and a caret at column 0 of the top row, where the switch sits at ~73. Any constant between those is tuned to this page's gutter widths, and would silently change which of them works when a card's inset changes. That is the magic-constant shape the rest of this review is against.

So the rule is now stated as what it is — **the row is the region; the column decides which control on it** — including the consequence you name: on a row with one control, that control is the answer for every column of the row. `a_lone_control_owns_its_whole_row` asserts exactly that, so it is a decision with a test rather than an unbounded surprise.

---

crates/fresh-editor-core/src/widgets/registry.rs
sinelaw: Fixed, all three. Every lookup now takes the caller's `PanelKey` — `focus_target_at`, `anchor_of_widget` and `tabbable_from_caret` — and none of them re-derives the panel or reaches across panels.

For the callers that genuinely start from a buffer there is one deterministic resolver, `focus_follower_of`, which returns `None` and logs when two panels on one buffer both ask to track the caret. Two panels tracking one caret has no right answer, so picking one out of a `HashMap` was picking arbitrarily between two wrong ones.

---

crates/fresh-editor/src/app/widget_runtime.rs
sinelaw: Fixed — `panel_buffer_caret` reads the **effective active split**, which is the split whose caret the reader is moving and the one `apply_event` writes through. The `values().find_map` is gone, and with it both order-dependent outcomes you describe.

I left `seat_buffer_cursor` writing every split showing the buffer: that is the pre-existing contract of `set_buffer_cursor_in_splits` (a plugin's `setBufferCursor` moves the buffer's caret, not one pane's), and narrowing it is a separate change with its own callers to weigh.

---

crates/fresh-editor/src/app/widget_runtime.rs
sinelaw: Fixed by making the claim true rather than softening it. `plugin_dispatch.rs:3705` and `:4118` both call `seat_buffer_cursor` now, so they go through the sync — they write into plugin virtual buffers, which is exactly the kind of buffer that carries a focus-following panel.

The doc also says why it is an invariant rather than an observation, and names those two initialisation-shaped paths, so the next caret write has somewhere to read that it has to come through here.

---

crates/fresh-editor/src/app/widget_runtime.rs
sinelaw: Fixed — and it was worse than 2 copies. Routing the two `plugin_dispatch` sites through `seat_buffer_cursor` removed both walks *and* the two sync bypasses you flagged in the comment above. There is one `splits_showing_buffer` now.

The repeated re-resolve inside the loop is hoisted: the manager and the view-state map are destructured once.

---

crates/fresh-editor/src/app/widget_runtime.rs
sinelaw: Fixed. "Did not opt in" is its own match arm now — `None if !follows_cursor => end_of_ring` — and it is the only silent one. The remaining `None`s (no caret to read, or a ring naming widgets the last paint did not place) `tracing::warn!` with the panel key before falling back, because both are disagreements between a panel's spec and its paint rather than a resting state.

---

crates/fresh-editor/src/app/widget_runtime.rs
sinelaw: Fixed, both halves.

The seat is now the **last** statement of `set_panel_focus_and_notify`, after `notify_widget_focus_change` and the `focus` event — so a disagreeing resolve can no longer deliver the inner (final) key before the outer (superseded) one, and `welcome_screen.ts`'s mirror cannot be left holding the stale one.

The recursion is declared: the call chain is drawn in the comment, and `Editor::seat_focus_depth` counts it so "at most one re-entry" is a `debug_assert!`. You're right that `anchor_of_widget` and `focus_target_at` are not inverses — which is exactly why an assertion is worth more here than the prose that was standing in for one.

---

crates/fresh-editor/src/app/event_apply.rs
sinelaw: Fixed — all three holes, by taking the shape you suggest. The block reads the caret the buffer **ended up with** after step 1 and syncs from that, unconditionally, so there is no event match left at all: nested batches, every edit, and clamped positions are covered by construction rather than by enumeration.

Cost for every other buffer in the editor is one bool (`has_focus_follower`) before anything is touched. And it's renumbered to 3b, in place.

---

crates/fresh-editor/src/app/window/mod.rs
sinelaw: Kept, with the argument you asked for put into the code.

The loop is the reason it belongs here rather than the reason it doesn't: this writes every split showing the buffer, so each of those carets **has just been teleported to a position it did not walk to**. A goal column is a memory of the column the reader was aiming for while moving vertically; keeping it across a jump they did not make is keeping a memory of somewhere they no longer are. Your pane-B reader has their caret moved by the plugin either way — the goal column surviving that move is the bug, not its preservation.

And it does hold for all four callers, which the comment now says: `setBufferCursor`, `scrollToWidget`, a virtual buffer's `initialCursorLine`, and the focus↔caret seat are all absolute placements, none of them a vertical motion.

---

crates/fresh-editor/src/app/sidebar.rs
sinelaw: (a) You're right that the layout corruption is recoverable — the page does repaint on `viewport_changed` — and I've taken that argument out of the comment. It was the symptom that found this, not the reason for it.

The reason is what it does to a reader: `markdown_toc` mounts a section for every Markdown buffer, so opening a `.md` file yanked a thirty-five-column sidebar — and a newly created file explorer — open on someone who had deliberately closed it. Every plugin doing this passes `startBlurred: true`, which is a plugin saying "take nothing from the reader for this"; the host was overriding that. Reversing it makes the flag mean what it says.

(b) Added: `an_outline_mounted_into_a_closed_sidebar_appears_when_it_is_opened` — the section is absent with the sidebar closed, and present after `Ctrl+B`, with nothing of the page's own pressed in between.

---

crates/fresh-editor/plugins/welcome_screen.ts
sinelaw: Fixed — that line is deleted and the `focus` branch is the sole writer.

Your trace is exactly right: `find_widget_by_key` returns the first match, which for a door card is a `focusable: false` row, so a click never moves host focus while the old line moved the mirror. The comment now records that path, including the consequence you name — host focus still in the finder field while `finderFocused()` reads false, so Up/Down silently stop walking the results while the field is still taking what is typed.

---

crates/fresh-editor/plugins/welcome_screen.ts
sinelaw: Fixed by dropping it. Routing is decided by `panel_id` at the call site, so the boolean had no job: `handleContentsEvent` returns `void` and the doc no longer describes a contract nothing uses.

---

crates/fresh-editor/plugins/welcome_screen.ts
sinelaw: Fixed, both parts.

`startupRow()` pushes its own outline entry now, the way `card()` and `banner()` do — no hand-written table, and no string that has to match a widget key defined elsewhere.

And the mark is the widget **key**, not an index: `outlineSelected` is a `string`, and the one place an index is computed (`outlineIndex()`) does it at the moment it is used. Your conditional-card scenario no longer moves the mark.

The nested-`card()` ordering I've left: nothing does it, and `card()` pushing before `body()` is what gives the outline document order for every case that exists.

---

crates/fresh-editor/plugins/welcome_screen.ts
sinelaw: 1. Fixed — collapsed to `if (bufferId === null) return;` followed by the buffer-mismatch branch.

2. Kept, and I'd argue it's the point rather than churn to be avoided: one outline per buffer, so leaving this page has to take its section down, or it sits under the next file in the slot that file's own outline goes in. `markdown_toc` unmounts on leaving a Markdown buffer for the same reason; what it avoids is remounting to change a *title*, which this never does.

3. Added — the comment now says a closed sidebar stays closed and that the `Contents` button is the recovery.

---

crates/fresh-editor/tests/e2e/plugins/welcome_screen.rs
sinelaw: Fixed, by establishing the premise where it belongs. `put_the_caret_in_the_finder` ends in `assert_the_finder_takes_typing`: it types a character, asserts it appears between the brackets, and backspaces it out. Four tests get the premise proved rather than assumed, and this one can no longer pass by never arriving.

It also clicks the field rather than counting Down presses — see the reply on the helper itself.

---

crates/fresh-editor/tests/e2e/plugins/welcome_screen.rs
sinelaw: Fixed. The shared helper clicks the field, so the screen-rows-as-buffer-lines conversion is gone from the four tests that only need the field focused, and it verifies arrival by typing.

The walk survives in `the_caret_arriving_at_the_finder_field_types_into_it`, where it is the subject rather than the setup — and there it's the loop you describe: re-read the caret row and the field row each iteration, bounded, with a panic naming where it stopped. That test also waits for the page's `git` probes to land before scrolling, because a repaint under the caret is what made reading its row unreliable in the first place — which showed up as a real flake once I ran the suite under parallel load.

---

crates/fresh-editor/tests/e2e/plugins/welcome_screen.rs
sinelaw: Fixed — the test asserts the property. It puts the caret in the finder, types `/0123` and a space, and asserts `find [/0123 ` appears: every one of the six characters, in order, in the field. Re-binding Space fails it now, which was your point.

The caret-did-not-move check is kept as a second half, with the caret away from the field, so "the key does nothing when you are not typing" is still covered.

---

crates/fresh-editor/tests/e2e/plugins/welcome_screen.rs
sinelaw: Fixed, both.

It waits for the caret to be *settled* — two consecutive frames drawing it in the same place — so a frame drawn before a pending repaint lands can no longer be the assertion subject.

And the panic reports what was observed ("no two consecutive frames drew the caret in the same place", with the last frame's value) rather than asserting a cause it hasn't established. Worth adding that your first point was not theoretical: under parallel load the caret really can stay hidden for the whole bound, which is how I found that one test's setup was reading a screen row through an async repaint.

---

crates/fresh-core/src/api.rs
sinelaw: Fixed by making the combination unrepresentable rather than advised against: `auto_focus_first()` is `!self.focus_follows_cursor() && …` now, documented at both fields.

Your framing is the reason. `autoFocusFirst` defaults to `true`, so the broken pair was what a plugin got by *not* thinking about it — and one clause of a prose paragraph is no defence against a default.

---

crates/fresh-editor/src/app/widget_runtime.rs
sinelaw: Fixed — it `tracing::warn!`s with the buffer, row, byte and row end when a span runs past its row.

The comment records your point about why it isn't a harmless approximation: end-of-row is nearest the rightmost control, so the sync re-resolves focus to whatever sits there, and both focus and caret end up somewhere the reader did not choose. Kept as a clamp rather than a bail, so the caret still lands on the right row while the log names the drift.

---

PR #3221, opened 2026-09-07
Quitting accounts for every workspace, and never deletes work it was not shown

### LINKED ISSUE #3189: Orchestrator: quitting from a clean workspace loses unsaved changes in other workspaces
When using multiple workspaces through the Orchestrator dock, quitting Fresh appears to check for unsaved changes only in the currently active workspace.

If another workspace has a dirty buffer but the currently active workspace is clean, Fresh exits without showing the unsaved-changes prompt. When Fresh is reopened, the file in the other workspace is restored from disk and the unsaved changes are lost.

### Steps to reproduce

1. Start Fresh with the Orchestrator dock and multiple workspaces.
2. Open **Workspace A**.
3. Open an existing file in Workspace A and modify it without saving.
4. Wait longer than the configured autorecovery interval. In my case, autorecovery is configured to run every 2 seconds.
5. Switch to **Workspace B**, which has no dirty buffers.
6. Exit Fresh while Workspace B is active.
7. Fresh exits without asking what to do with the unsaved changes in Workspace A.
8. Start Fresh again.
9. Return to Workspace A.
10. The file is still open, but the unsaved modifications are gone and its contents have reverted to the saved version on disk.

### Expected behavior

When quitting Fresh from an Orchestrator session, Fresh should check for dirty buffers across **all open workspaces**, not only the currently active workspace.

Unsaved buffers in other workspaces should either:

- cause the normal save/discard/exit-recoverable prompt to appear, or
- be preserved by Hot Exit/recovery so they can be restored after restarting Fresh.

### Control case

The same scenario works correctly without multiple Orchestrator workspaces.

If I modify a file and quit Fresh while that workspace/editor is active, Fresh detects the unsaved changes and asks what to do.

If I select **Exit (recoverable)** and start Fresh again, the unsaved modifications are correctly restored.

This suggests that autorecovery/hot-exit itself is working, but dirty buffers belonging to other Orchestrator workspaces are not being considered when Fresh is exited from a clean workspace.

### Impact

This can cause **silent data loss**.

The particularly problematic part is that there is no warning when exiting. From the user's perspective, the unsaved buffer is still open in another Orchestrator workspace and autorecovery is enabled, so there is no indication that those changes will be discarded on exit.

sinelaw:
Fixes #3189.

## Problem

Quitting only ever looked at the workspace on screen. `Ctrl+Q` from a clean one exited with no prompt while another held unsaved work — and the exit then deleted that workspace's recovery files. Gone, no warning at any point.

## Objective

Unsaved work is never destroyed without the user asking.

## Decisions & trade-offs

- **Recovery data leaves the store only when the user resolves it** — saves, or explicitly discards. Anything this session never held in memory is kept indefinitely, not aged out: any expiry is a guess at when the user stopped caring, and being wrong once costs an edit. Costs some disk.
- **Stamp entries with the owning workspace** and recover per workspace, on activation. New metadata field, backward compatible via a path-prefix fallback.
- **Retarget the active-window pointer** for shutdown work instead of duplicating ~10 per-window helpers. Only sound off the render path; confined to shutdown and documented as such.
- **Confirmation on workspace close stays with the Orchestrator UI.** This layer only guarantees the content is recoverable — the Archive/Delete dialogs still don't mention unsaved buffers, which is follow-up work.

## Solution

Every quit-time and exit-time operation now spans all workspaces rather than the active one: the prompt (and it names them), save, discard, the Save-As chain, the recovery flush, and the periodic sweep. Crash recovery returns each buffer to its own workspace when that workspace is activated. Exit deletes an entry only if a live buffer backed it and it wasn't preserved.

## Before / after

| Step | Before | After |
|---|---|---|
| Dirty buffer in `wsA`, switch to clean `wsB`, `Ctrl+Q` | exits silently | prompts |
| …the prompt | — | `3 buffers have unsaved changes (wsA: 2, wsB-1).` |
| …press `s` | saves `wsB` only, deletes the rest | saves every workspace |
| …press `q` (recoverable) | `wsA`'s recovery files deleted | preserved, restored next start |
| Archive/Delete a dirty background workspace | its work is dropped | flushed to recovery first |
| Crash with 3 dirty workspaces | edits since you left a workspace are lost | all present |
| …restart | everything piles into the foreground workspace | each workspace gets its own, on activation |
| Quit having visited only some workspaces | the unvisited ones' data is deleted | untouched |

## Verification

Every row above was driven in tmux against both this branch and a `fbddf641` baseline build. Also: four workspaces, crash, four restarts each visiting a different subset — the last workspace's edit came back intact having been skipped three times, then all four saved to their own files. Regression tests cover the prompt, cross-workspace save/discard, per-workspace crash recovery, and the keep-what-you-can't-account-for rule.

Full evidence in the comments below.

https://claude.ai/code/session_01U14RqE6EeBisx85NZ1LnB9


sinelaw:
Pushed `5fa3343`, folding in two more paths with the same active-window blind spot that the original two fixes don't reach. Auditing every deletion site turned these up.

**A crash still lost a background workspace's edits.** The *periodic* auto-recovery save ran against the active window only, so a buffer edited and then left behind had recovery data no newer than the last tick it was on screen for. The first commit fixed the exit flush — but an exit is exactly what a crash doesn't get, which is the case recovery exists for. The sweep is cross-window now, and costs almost nothing: a background window's buffers can't become `recovery_pending` again without edits, and edits only happen while a window is active, so each is written once after the user switches away and then skipped. The rate limit stays one editor-wide tick off the active window's clock, with every window's stamp advanced together so a later switch doesn't re-tick on a stale timer.

**Closing a workspace dropped its buffers without preserving them.** `close_window` tears the `Window` down with everything in it and asks nothing about unsaved changes. `flush_window_recovery` writes them out first; afterwards no live buffer backs the entry, so `end_session_accounting` preserves rather than cleans it and the content survives to the next start.

That leaves the *confirmation* on close unfixed, deliberately. `close_window` is what the Orchestrator calls after the user has already picked Delete / Archive / Kill, and it returns a bool into a plugin dispatch loop with nowhere to host a prompt — asking belongs to the UI offering those actions. This layer's job is making the answer recoverable either way. Worth a separate issue if you want the prompt.

Two tests added alongside the existing ones: a background workspace's dirty buffer reaching recovery from the periodic tick alone (no quit, no shutdown), and closing a workspace flushing its unsaved content and that content surviving the exit after it.

### For completeness, the deletion sites that remain

| Path | Deletes | Can it lose unshown work? |
|---|---|---|
| `end_session_accounting` (only exit path) | entries a live buffer backed and that weren't preserved — saved or discarded; plus unaccounted entries past `max_recovery_age_secs` | No |
| `delete_buffer_recovery` (on save, on buffer close) | that one buffer's entry | No — resolves the id through the active window, and `BufferId` is globally unique (shared atomic allocator), so a background window's id is simply absent and the call no-ops. It can never delete the wrong workspace's entry. |
| `accept_recovery` in `recover_all_buffers` | each entry *after* loading its content into a buffer | No |
| `end_session` / `end_session_preserving` | — | No production callers left; tests and the module doc example only |
| `discard_all_recovery`, `cleanup_old`, `cleanup_orphans` | — | No callers anywhere |

Two known behaviours this PR does not change, for the record:

- `recover_all_buffers` and `try_restore_hot_exit_buffers` open every entry into whichever workspace is active, regardless of origin. Not data loss, but files can land in the wrong workspace — and it's slightly more visible now that more entries survive.
- `cleanup_old` / `cleanup_orphans` have no callers, so the only GC is the age check in `end_session_accounting`. Orphans can sit for `max_recovery_age_secs` (7 days). That's the deliberate cost of not deleting what we can't account for.

Verification status: the first commit's behaviour is confirmed end-to-end in tmux (prompt, preserved chunk, edit restored on the next start). These two follow-ups are covered by the new tests and compile clean with no new warnings, but per the request the suite wasn't run locally — CI is their first behavioural run.

---

sinelaw:
## tmux scenario matrix, A/B against the pre-fix binary

Built `fbddf641` as a baseline binary alongside the branch build and drove both through the same tmux sequences (120x30 pane, scratch `HOME`). Files `a1.txt`/`a2.txt` in `wsA`, `b1.txt` in `wsB`, `c1.txt` in `wsC`; workspaces created through the real Orchestrator dialog, switched through the dock, Archive/Delete through the dock's right-click context menu.

| # | Scenario | Baseline `fbddf641` | This branch |
|---|---|---|---|
| 1 | 3 dirty buffers across 2 workspaces, `Ctrl+Q` from the workspace whose own buffer is dirty but which isn't holding the others | prompt says **"1 buffer has unsaved changes"** | **"3 buffers have unsaved changes"** |
| 2 | …then `s` (save and quit) | only `b1` written. `a1`/`a2` **not** written **and** their recovery deleted — recovery store 0 entries, both edits gone for good | all three written to disk; store cleaned to 0 (all resolved) |
| 3 | …then `d` (discard) | — | nothing written, store 0, restart shows no `*` anywhere in either workspace (no resurrection) |
| 4 | …then `q` (recoverable) | (1-buffer form of this is the issue's own repro: entries deleted) | 3 entries preserved; restart brings `a1*`/`a2*` back in `wsA` and, on diving in, `b1*` in `wsB` — **each in its own workspace** |
| 5 | Cancel (Esc) at the prompt | — | stays running, both workspaces still dirty and intact |
| 6 | Live periodic sweep: 4 dirty across 3 workspaces, `wsC` active | — | all 4 entries current on disk while running |
| 7 | **Edit, switch away inside the auto-save interval, keep working elsewhere, SIGKILL** | the backgrounded edit has **no recovery entry at all**; restart brings `a1.txt` back *clean* — edit lost | swept by the cross-window tick; restart recovers `LATE-EDIT a1 original` |
| 8 | SIGKILL with 4 dirty across 3 workspaces | — | all 4 recovered with contents intact |
| 9 | **Archive** a dirty background workspace, then quit cleanly | entry survives the archive but the following clean exit **deletes it** (store 0) — edit gone, no prompt at any point | preserved through both (1 entry, `EDIT-A1 a1 original`) |
| 10 | **Delete** a dirty background workspace, then quit cleanly | — | preserved |
| 11 | `Ctrl+Q` with nothing dirty, dock + terminals + panels present | — | exits immediately, no spurious prompt (the composite/hidden/virtual filter holding) |

Row 4 also closes the gap I flagged earlier: the full workspace-restore path — not just `--no-restore` — puts each recovered buffer back in **its own** workspace.

### Three things the runs turned up

1. **Neither confirm dialog mentions unsaved editor buffers.** Archive's says it will "SIGKILL all workspace processes / close the editor workspace / move the worktree to `.archived/`"; Delete's says "Uncommitted changes will be lost", which reads as a git statement. Nothing warns that a modified buffer is about to be dropped. The content is now safe either way, but this is the confirmation gap in the plugin — worth its own issue.

2. **Crash recovery funnels everything into the foreground workspace.** After the row-8 SIGKILL, `b1` and `c1` came back as tabs in `wsA` rather than in their own workspaces. Not data loss, and pre-existing, but more visible now that background workspaces reliably have recovery data.

3. **`--no-restore` on the *first* run writes no workspace snapshot**, so a later plain restart has nothing to reattach to — the dock loses the extra workspaces and the recovered buffers have nowhere to go. Recovery data itself survived (verified). Pre-existing `--no-restore` behaviour, unrelated to this PR, but it's why the issue's own repro steps don't show the restore half.

---

sinelaw:
Pushed `102081b`.

### 1. The prompt says where the work is

A bare count is close to useless when the unsaved buffers are in a workspace you can't see — the exact case this prompt now exists to catch. It names the workspaces whenever any of the unsaved work is outside the active one:

```
3 buffers have unsaved changes (wsA: 2, wsB-1). (s)ave and quit, (d)iscard and quit, (q)uit (recoverable), (C)ancel?
```

A workspace holding more than one buffer is qualified with its count; one holding a single buffer is named bare. When everything unsaved **is** in the workspace on screen it keeps the old plain form — that's what a single-workspace user sees, and naming their one workspace would be noise. The location rides in as a parenthetical, so the four new locale strings are each locale's existing translation with one placeholder spliced in, not new prose to translate (60 added lines across 15 files, no reformatting).

### 2. Per-workspace recovery

The recovery store is shared by the whole session, and entries recorded *what* was unsaved but not *whose* it was — so after a crash, buffers from three projects came back as tabs in one, and the foreground workspace consumed entries belonging to workspaces the user hadn't even opened.

- Entries are stamped with the owning workspace's `stable_id`. Keyed on that rather than the project root, because two Orchestrator workspaces may deliberately share a worktree, which makes the root ambiguous and the stable id not.
- Recovery is per workspace, on activation: the foreground one claims its own at startup, every other one when the user first dives in.
- Legacy entries (no stamp) fall back to the longest workspace root prefixing the original path, and the single startup pass also sweeps up anything no open workspace claims — so nothing regresses into being silently dropped. That widening is startup-only; on later activations it stays strict, or each workspace in turn would adopt its own copy of the same unattributable entry.
- Entries are **read, not consumed**. Each becomes a live modified buffer that keeps writing to the same entry, leaving the exit accounting to decide. That is what makes quitting after visiting only some workspaces safe — the unvisited ones' work is still on disk.

`RecoveryStorage::save_recovery` / `RecoveryService::save_buffer` keep their signatures; the owner rides on new `*_owned` variants, so the ~20 existing call sites are untouched.

### tmux verification

| Check | Result |
|---|---|
| 2 dirty in background `wsA` + 1 in active `wsB-1`, `Ctrl+Q` | `3 buffers have unsaved changes (wsA: 2, wsB-1).` |
| all dirty buffers in the active workspace | plain form, no workspace named |
| stamps on disk | `wsA/a1.txt -> ws-…-0`, `wsA/a2.txt -> ws-…-0`, `wsB/b1.txt -> ws-…-1` |
| crash, restart, foreground `wsA` | `a1.txt*` and `a2.txt*` only — `b1` no longer leaks in |
| then dive into `wsB` | `b1.txt*` with its edit, adopted there |
| crash across 3 workspaces, restart, **visit only `wsA`**, quit | all 3 entries still on disk |
| next start, dive `wsB` then `wsC` | each comes back with its own edit |

### Still open

The Archive / Delete confirm dialogs still say nothing about unsaved editor buffers. That one is a different surface: `WindowInfo` carries no unsaved count, so it needs a new host API field plus its TS binding, then dialog text and keys across the 15 locales in `orchestrator.i18n.json` (which the plugin-i18n completeness test enforces). Happy to do it — just flagging that it's plugin/UI work rather than more of the data-loss core, so it may belong in its own PR.

Tests added: prompt names the workspaces; prompt stays plain when it shouldn't; crash recovery restores each buffer into its own workspace and only on activation; quitting without visiting a workspace keeps its recovery data. Per the standing request the suite is still not run locally — CI is the first run.

---

sinelaw:
## Multiple restarts with partial workspace visits

Four workspaces (`wsA` launch + `wsB`/`wsC`/`wsD` via the Orchestrator dialog), one dirty buffer each, SIGKILL, then four restarts each visiting a different subset. No code changes — this exercises `102081b` as pushed.

Entries were stamped per workspace before the crash:

```
wsA/a1.txt -> ws-18d33dbafe61d0fb-0
wsB/b1.txt -> ws-18d33dc1a5855ccc-1
wsC/c1.txt -> ws-18d33dcbeacab6cf-3
wsD/d1.txt -> ws-18d33dd63248da81-5
```

| Restart | Visited | Prompt on quit | Entries after quit |
|---|---|---|---|
| 1 | `wsA` only | `1 buffer has unsaved changes.` (plain — the only dirty buffer is in view) | 4 |
| 2 | `wsA`, `wsB` | `2 buffers have unsaved changes (wsA, wsB-1).` | 4 |
| 3 | through `wsC` | `3 buffers have unsaved changes (wsA, wsB-1, wsC-2).` | 4 |
| 4 | through `wsD` | `4 buffers have unsaved changes (wsA, wsB-1, wsC-2, wsD-3).` | — |

`wsD`'s `EDIT-D1 d1 original` came back intact on restart 4 after three restarts of never being opened. Each workspace showed only its own buffer on every restart, contents byte-identical each time — no accumulation, no duplication, no drift. Stable ids held across all four restarts, so attribution never degraded to the path-prefix fallback.

Closing the loop, `s` from restart 4 wrote all four to their own files and cleaned the store to 0:

```
wsA/a1.txt   EDIT-A1 a1 original
wsB/b1.txt   EDIT-B1 b1 original
wsC/c1.txt   EDIT-C1 c1 original
wsD/d1.txt   EDIT-D1 d1 original
entries: 0
```

### The same shape on the baseline

Three workspaces, crash, one restart into `wsA` with no dives. `fbddf641` produced a genuinely tangled state:

- `c1.txt*` and `b1.txt*` were pulled into `wsA`.
- Diving to `wsC` afterwards showed it **empty** — its buffer now lives in `wsA` instead.
- Diving to `wsB` showed `b1.txt*` there too, so the same file was open **and separately dirty in two workspaces at once** (`wsA` still had its copy). Two independent modified buffers over one path is worse than either losing it or moving it.
- The quit prompt back in `wsA` said `2 buffers`, with a third dirty `b1` sitting in `wsB`.

One correction on that baseline run: I launched it without the `a1.txt` argument, so my `EDIT-A1` keystrokes landed in a scratch buffer. That is why two `[No Name]*` tabs appear in it — my scripting slip, not a defect. It does not affect the three findings above, which are all about `b1`/`c1`.

---

crates/fresh-editor-core/src/recovery_types.rs
sinelaw, line 163: shouldn't this accept the new parameter?

crates/fresh-editor-core/src/recovery_types.rs
```diff
@@ -144,6 +144,23 @@ pub struct RecoveryMetadata {
     /// Original file size (0 for new buffers, needed for reconstruction)
     #[serde(default)]
     pub original_file_size: usize,
+
+    /// Durable id (`Window::stable_id`) of the workspace whose buffer this
```
sinelaw: avoid overly verbose and wordy comments. only include info that can't be determined from the code (motivation/goal/objective/tradeoff/decision). same for all other comments

crates/fresh-editor-core/src/recovery_types.rs
sinelaw: Yes — fixed in `45e604a`. `new` now takes `workspace_id` instead of hardcoding `None` and leaving the caller to patch it in; a constructor that silently drops one of its struct's fields invites exactly that.

The re-stamp in `save_recovery_owned` stays, though, and isn't redundant: it's what carries an entry across a buffer moving between workspaces (Extract Tab to New Workspace), where the metadata already exists and is being updated rather than constructed.

---

crates/fresh-editor-core/src/recovery_types.rs
sinelaw: Done in `45e604a`, applied across everything this branch added — net **−118 lines** of comment.

The rule I used: drop anything restating the code, keep motivation, trade-offs and decisions. So the `workspace_id` doc went from ~14 lines to 6, keeping only *why* it's keyed on `stable_id` rather than the project root (two workspaces may share a worktree) and what `None` means for readers. Same treatment for `with_window_retargeted`, `end_session_accounting`, `adopt_recovery_for_active_window`, `flush_window_recovery`, the auto-save sweep, and the inline comments in `close_window`, `set_active_window` and the quit paths.

Where I kept length it's carrying something not inferable from the code — e.g. `with_window_retargeted`'s "only safe for synchronous, non-rendering work", which is a constraint on callers rather than a description of the body.

---

PR #3223, opened 2026-09-07
Editing a huge single-line file costs the screen, not the line

sinelaw:
Follows #3162, which made *opening* a huge single-line file bounded. Everything that happens afterwards was not.

## The problem

On a 19 MB one-line JSON, measured by attributing every buffer read in a frame to its call site:

| | before |
| --- | --- |
| steady-state frame (wrap off) | **12.6 MB read** |
| one keystroke | **13.7 MB** |
| one `Down` press | **12.7 MB** |
| width measured per frame | 801 KB |

The file, several times over, per keypress. On the reported 50 MB file that is unusable, and `Down` stopped at the bottom of the first screen and never went further.

Two causes, the same shape at different layers.

**The layout unit was a constant, not the pane.** With soft wrap off a row was chopped at `MAX_SAFE_LINE_WIDTH` — 10,000 columns, ~83× the pane — so a 40-row frame built ~400,000 characters to draw ~4,800. It also left the viewport counting rows one way while the renderer drew them another, which is why the cursor could walk the rows on screen and then stop.

**Structural questions were answered by reading text.** "Where does this line start", "where does it end", "how many lines are below this one" — asked several times per keystroke by the scroll clamp, the visibility check, the viewport-end calculation, the fold and indentation-guide probes, and the vertical motions — each answered by reading the line, in 100 KB pieces, on a file whose line is the file.

## Results

| | before | after |
| --- | --- | --- |
| frame, wrap off | 12.6 MB | **150 KB** |
| keystroke, wrap off | 13.7 MB | **178 KB** |
| `Down`, wrap off | 12.7 MB | **300 KB** |
| frame, wrap on | 8.1 MB | **222 KB** |
| keystroke, wrap on | 9.2 MB | **268 KB** |
| width measured per frame | 801 KB | **238 bytes** |
| edit at a far column of a long line | 38.5 MB | **2.3 MB** |

Verified interactively on the reported 50 MB file, in a debug build, not only in tests:

| | before | after |
| --- | --- | --- |
| cursor step at byte 52,428,800 | — | 51–63 ms |
| `End` from byte 0 | no-op | 1137 ms (indexes the line) |
| `Home` back, `Ctrl+End`, `Ctrl+Home` after that | — | 50–57 ms |
| insert one char at byte 52,428,804 | 2287 ms | **30–34 ms** |
| backspace there | 3602 ms | **30–41 ms** |

Swept there for regressions across click, drag-select, shift-wheel, menu, go-to-line, find, and type/undo at a far column, plus the same sweep with wheel and scrollbar drag on an ordinary multi-line file.

## The commits

**`core: answer where a line begins and ends without reading the line`** — three primitives that walk the piece tree's own storage instead of copying text out of it, all with a mandatory cap, where `None` means "not within reach" rather than "not there". They also *fill in* the tree's per-piece line-feed count as they cross a piece, so the second search of a region is answered from metadata; that count already survives edits, so typing does not discard it. `LineIterator` gains a bounded backward search whose result is snapped to a character boundary, and `counters::work` gives every test a clock-free budget in bytes read. Also fixes the wrap machine measuring every token's width twice, and adds a property test holding the counts these searches write against the bytes.

**`The row a pane draws is the pane's width, and structure is walked not read`** — one logical line is one visual row with wrap off on a large file; every line-boundary search in the layout and movement paths goes through the primitives above, and the ones that cannot answer say so rather than guessing. Four places had read a budget that ran out as an answer: the scroll clamp (which dragged the viewport to byte 0 on a pane of ten-kilobyte lines), a cut row that stepped over the line below it, the window-end walk, and the plugin line offer. Bracket matching and the `lines_changed` offer are sized by the frame rather than by the file, and the ASCII fast path — dead code, gated on a condition no production caller met — is reachable again. Includes the `Home`/`End` fixes: both name a position, so neither may answer with a budget.

**`Viewport edge shading: not over the caret, and not on a guess`** — an edge no longer shades the band the caret sits in, nor shades on a fact the editor does not have (a pane with no scrollbar, or a file too large to count lines on, where the thumb is placed by a byte fraction).

**`A row a pane draws is a window into its line, not the line up to it`** — the row was built as a *prefix*, every column from the line's start out to the right edge, so a frame's cost and its reach both scaled with how far right the view had scrolled. It now reads only the window it shows. See below.

**`Paging, selecting and the horizontal clamp count lines, not read pieces`** — the last callers still asking the reader for "the next line" and getting back the next 100 KB piece of the line the cursor is already on. Page up/down walked pieces, so a page on a one-line file moved thirty-nine pieces along it; select-to-line-start/end selected to a piece boundary; scrolling by rows with wrap off moved the top by the backward search's 64 KB bound. All four now walk logical line starts. The visible one was the horizontal clamp: it needs the line's length to keep the view out of the empty space past a line's end, and measuring by reading returned a piece — so the code took the larger of that and the cursor's own column, which on a longer line makes the length *equal* the cursor's column and the clamp then states that the cursor is the last visible column. Every press of Left dragged the window one column with it and pinned the caret to the right-hand edge.

**`A keystroke costs the same at either end of a line`** — cursor movement at a far column was fixed by the commits above, but *typing* there still took 2.3 s and backspace 3.6 s, while the same keystroke at byte 0 of the same file took 48 ms. Attributed rather than guessed, by tagging every buffer read with its caller: one insert read the line **twice**, from two unrelated places. `line_start_and_blank_prefix` walked backwards in 4 KB blocks to find the line start (18 MB in 4,417 reads) — it now asks `prev_line_start_within` instead, and settles its blank-prefix question at the first non-blank byte to the left of the caret, which is exact and costs only the indentation. And `collect_lsp_changes` converted the edit's byte offsets into LSP's UTF-16 positions — which needs the line counted from its start — on every buffer-modifying event, after which the send path discarded the result because the buffer has no server. A log on that path confirmed it fired twice for two keystrokes. The side-effect-free head of that same guard is now asked *before* the changes are built.

## Two things worth a reviewer's attention

**The `End` bug was found by running the editor, not by the tests.** The test covering it passed while the editor failed: it used a 2.6 MB file with the large-file threshold moved to 1 MB, which is the one window where neither of the two bounds bites. The build stops a line at `max_lines × MAX_LINE_BYTES` (~4.4 MB, because it counts 100 KB read *pieces* rather than lines), and below the 10 MB threshold the row is chopped at `MAX_SAFE_LINE_WIDTH` instead — 4.4 MB is under 10 MB, so with default settings no file size could draw a scrolled-away column. The test now runs on the module's 19 MB fixture, past both bounds, and lands together with its fix in the fourth commit.

**Guarding the LSP conversion had to wrap the send step whole, not just the collection.** Skipping only the conversion leaves an empty change set, and the empty-set branch exists to snapshot the *entire document* for the BulkEdit case — so guarding the cheap half alone would have replaced an 18 MB read with a 50 MB one.

**Reaching the end of a line indexes the file.** The walk records what it crosses, so afterwards the status bar switches from `Byte N` to `Ln/Col`. That is an honest report of what the editor now knows, but it is a visible change.

## Behaviour changes worth a decision

**With soft wrap off, a large file with no line breaks draws one row and `Down` has nowhere to go.** That is what the setting means on such a file, and soft wrap — which walks it row by row and is unaffected — is what reading one vertically is for. The alternative was to keep inventing rows by chopping the line at a fixed column, which is a worse version of what soft wrap already does and is what left the viewport and the renderer disagreeing.

**On a file too big to index, plugin decorations that consume source lines** (blame, live diff, compose reflow) see only the visible text. A row injected with no source byte of its own cannot be placed by a byte walk; teaching the walk the whole transform stage stays open and nothing here blocks it.

**Known gap:** a long-line file *below* the large-file threshold keeps the old wrap-off row cost and the 10,000-column chop. The bounded searches and the bracket and plugin bounds still apply to it; only the row-width change is scoped out.

**Known gap:** the bracket overlay still reads 2 MB per keystroke — the bulk of what remains after the last commit. Its match search is already viewport-bounded, but on a single-line file `viewport_start .. viewport_end` spans the whole line, so the margin degenerates and the 1 MB cap is what binds. Bounding it by what is actually *drawn* needs the pane's cell window plumbed into the overlay, and tightening the rainbow nesting depth — already approximate at 1 MB — changes visible colours. That is a separate change with its own decision to make, not a bound swap.

**Scope notes:** the windowed row is skipped for binary buffers, whose rows are built byte-wise by a different reader. The byte/column conflation it relies on predates this branch — exact for ASCII, approximate for wide characters, as before.

## Tests

Every cost claim is pinned on **bytes read from the buffer**, not on the clock, so they mean the same on a loaded CI runner as on an idle laptop (CONTRIBUTING §3), and each bound is stated against a byte-identical multi-line control — same bytes, different shape, so a gap between the two is the shape costing something. The row model, `Down`, the dropped lines, the arrow keys on long-line files, the reported `End`/`Home` sequence, stepping left from the end of a long line, and the shading are asserted on rendered output (§2).

The edit-cost test states its invariant as a *ratio* — what a keystroke costs does not depend on the caret's column — rather than as an absolute figure, so it keeps its meaning if the fixture changes. Without its fix it reads 39,671,146 bytes against 128,375 for the same keystroke at byte 0.

Where a fix lives on a path the rendered view does not take — the byte-based vertical motions, the scroll clamp — the failing test is a unit test on that function rather than an end-to-end one that would pass either way; the e2e tests over those areas are labelled as the coverage they are.

Twelve e2e tests in `large_file_edit_bounded.rs`, the shading tests in `semantic/scroll_edges.rs`, and unit tests for the new primitives, the leaf-count property, the repaired fast path (including that an escape sequence split across two tokens stays invisible), and the single width measurement.

https://claude.ai/code/session_01MEmagq6rum6Fs9RyqJXgGn

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,9 +36,12 @@ For live updates on Fresh, [follow me on X](https://x.com/TheNoamLewis).
     * A multi-line paste now arrives whole instead of only its last line (#3215)
     * A killed editor leaves it usable instead of exiting straight out of raw mode and the alternate screen (#3197)
     * No longer loses recent scrollback (#3151)
-* **Opening large files**
-    * Huge single-line files (tens of MB) no longer loop or stall for cursor movement, scrolling and paging (#1806)
+* **Large single-line files (minified JSON and friends)**
+    * Opening one no longer loops or stalls for cursor movement, scrolling and paging (#1806)
     * Very large binary/zip files open instantly instead of freezing or risking an out-of-memory crash (#3142, reported by @mommysgoodpuppy)
+    * **Editing one is no longer slow.** A frame, a keystroke and an arrow press now cost the size of the screen rather than the size of the line: on a 19 MB one-line JSON a frame reads ~150 KB instead of ~12 MB, and a keystroke ~180 KB instead of ~14 MB. The scans behind that — where a line starts, where it ends, how many lines are below this one — are bounded, and a stretch already found to hold no line break is remembered rather than re-walked on every frame
+    * With **soft wrap off**, a logical line is now one visual row, so a file with no line breaks draws one row and `Down` has nowhere to go — turn soft wrap on to read such a file vertically. Rows used to be chopped at a fixed 10,000 columns, which put ~400,000 characters through layout to draw ~4,800 and left the viewport and the renderer counting rows differently
+    * Plugin decorations that consume source lines (blame, live diff, compose reflow) are bounded to the visible text on such a file, so a plugin can no longer make a frame read the line
 * **Wave screensaver** now ends when you return to the window (or the GUI window), not just on the next keypress (#3204)
 * **`deno.json`** no longer disables TypeScript on a project without Deno (#2981, reported by @atk)
 * **Keymap chords** - a single-key binding on a chord prefix now fires, and deleting a keybinding frees the key (#3171)
--- a/crates/fresh-editor-core/src/config.rs
+++ b/crates/fresh-editor-core/src/config.rs
@@ -1163,9 +1163,15 @@ pub struct EditorConfig {
     /// from its background, the one inside it two thirds, and the third
     /// row in normally. Constant, not animated — the same rows are
     /// shaded whether the view is moving or still, so scrolling is a
-    /// plain shift of the text through a fixed gradient. The top edge
-    /// only shades when there is something above it, so the first lines
-    /// of a file are not dimmed for no reason.
+    /// plain shift of the text through a fixed gradient.
+    ///
+    /// An edge shades only where there is something beyond it to trail off
+    /// into, so the first and last lines of a file are not dimmed for no
+    /// reason, and neither is a file that fits its pane. An edge whose facts
+    /// are missing — a pane with no scrollbar, or a file too large to count
+    /// lines on — is left alone rather than shaded on a guess. So is the edge
+    /// the cursor is sitting in: the shading helps reading, and the line being
+    /// edited is the one line the reader is certainly looking at.
     #[serde(default = "default_true")]
     #[schemars(extend("x-section" = "Display"))]
     pub viewport_edge_fade: bool,
--- a/crates/fresh-editor-core/src/counters.rs
+++ b/crates/fresh-editor-core/src/counters.rs
@@ -1,8 +1,12 @@
-//! Global performance counters for observability and testing.
+//! Performance counters for observability and testing.
 //!
-//! Atomic counters that can be incremented from anywhere and read/reset in tests.
-//! All operations use relaxed ordering — these are best-effort metrics, not
-//! synchronization primitives.
+//! Two kinds, for two different questions:
+//!
+//! * [`Counters`] — process-global atomics for I/O that happens on whatever
+//!   thread does it (disk reads, recovery chunks). Relaxed ordering: these are
+//!   best-effort metrics, not synchronization primitives.
+//! * [`work`] — per-thread counters for the layout path, where a test needs to
+//!   measure exactly the work *it* caused.
 
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::LazyLock;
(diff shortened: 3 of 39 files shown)


sinelaw:
`test ubuntu-latest` is red on 8cd8371, and I don't believe it's this PR's.

**What failed.** One test out of 9,815, and it did not assert — it ran out of time:

```
TIMEOUT [180.012s] (5657/9815) fresh-editor::all_tests
  e2e::plugins::review_diff_line_staging::test_review_visual_stage_modified_line_pair
Summary [1679.162s] 9815 tests run: 9814 passed, 1 timed out, 163 skipped
```

**Why it isn't mine.**

- `test macos-latest` ran the same test on the same commit and passed.
- It passes locally 10 times out of 10, in 1.09–1.26 s against the 180 s cap — a ~140× margin, which is not a test that got slower.
- It never reaches the code this commit changed. It drives the review panel with `n`/`v`/`j`/`s`, so the insert path holding `line_start_and_blank_prefix` isn't entered at all. And `lsp_change_could_be_sent` is a strict prefix of the guard chain already at the top of `send_lsp_changes_inner` — metadata, `lsp_enabled`, URI, buffer state — all of which return `false` with no side effect. So whenever the new check skips the work, the old code reached the same early return having done nothing; the two are equivalent, and neither is on this test's path.
- The test's shape is `wait_for_screen_contains` / `wait_until` polls over git subprocesses and an async panel repaint, which under runner contention fails exactly this way: a wait that never settles, ending at the harness timeout rather than at an assertion.
- `master`'s own CI is intermittently red — 5 of its last 15 runs failed.

**No fix ported, because there is nothing to port:** no change on master or elsewhere addresses this, and the test is sound everywhere I can reach it. Making it robust would mean bounding those waits, which is unrelated to this PR and shouldn't be smuggled into it.

**Re-run pending.** `rerun-failed-jobs` is refused with `403 This workflow is already running` while `test windows-latest` is still going. I'll spend the one re-run as soon as the run completes, and keep watching until the PR is green.

---

Issue #3214, opened 2026-09-07
some line input prompts get cut-off in standard terminal width 80


Korkman:
For example this line input exceeds small terminal widths (80 and lower) with 98 characters:

`1 buffer has unsaved changes. (s)ave and quit, (d)iscard and quit, (q)uit (recoverable), (C)ancel?`

Quite predictable, it's horribly long in German locale with 133 characters:

`1 Buffer hat ungespeicherte Änderungen. (s)peichern und beenden, (v)erwerfen und beenden, (q)eenden (wiederherstellbar), (A)bbrechen?`

I suggest to implement word-wrapping for line inputs. At the same time maybe reconsider modal dialogs in place of them for major prompts like this.

sinelaw:
Reproduced on 0.4.10 (debug build of `main`), Linux, tmux **80x24**, default config.

Steps:
1. `printf 'hello world\nsecond line\n' > t.txt`
2. `tmux new-session -x 80 -y 24 'fresh --no-restore t.txt'`
3. Type any character (buffer becomes dirty).
4. Press `Ctrl+Q`.

The prompt row is hard-cut at column 80 — no wrap, no ellipsis:

```
1 buffer has unsaved changes. (s)ave and quit, (d)iscard and quit, (q)uit (recov
```

The full string is 99 columns (`prompt.quit_modified_hot_one` in `locales/en.json`), so `erable), (C)ancel?` is lost and the Cancel key is not discoverable at all. `de.json` is 179 columns, so it loses everything from `(v)erwerfen` onwards.

---
_Reproduced by an automated agent driving the editor interactively in tmux._

---
_Generated by [Claude Code](https://claude.ai/code)_

sinelaw:
**Fixed on master** (`06dafb9dd`, 0.5.1 build) — by the modal-confirmation change in 0.5.2 (#3320), which replaced the bottom-row prompt entirely.

Same repro as before: 80x24 tmux, dirty buffer, `Ctrl+Q`.

English:
```
┌──────────────────────────────────────────────────────────────────────────┐
│ Unsaved Changes                                                          │
│──────────────────────────────────────────────────────────────────────────│
│ 1 buffer has unsaved changes.                                            │
│──────────────────────────────────────────────────────────────────────────│
│ [ Save and Quit ]   Discard and Quit     Quit (recoverable)     Cancel   │
└──────────────────────────────────────────────────────────────────────────┘
```

[...]

PR #3224, opened 2026-09-08
fix(prompt): wrap long confirmation prompts on narrow terminals (#3214)

56steve:
## What this fixes

Issue #3214. On an 80-column terminal, the quit-with-unsaved-changes prompt and the other single-key confirmation prompts were hard-cut at the last column with no wrapping and no ellipsis. The English message is 98 columns wide:

`1 buffer has unsaved changes. (s)ave and quit, (d)iscard and quit, (q)uit (recoverable), (C)ancel?`

so `(C)ancel?` fell off the end and the Cancel key was not discoverable at all. German is worse at 133 columns and loses even more of the options.

## The change

When a confirmation prompt (one that takes no typed input) is wider than the row, its message now wraps across rows instead of clipping the tail. Two coordinated changes in the described shell tree:

- `view/shell/prompt_line.rs`: `build()` returns a wrapping text node when the message overflows and there is no input, and the prompt row is now sized `Auto` instead of a fixed one cell.
- `view/shell/frame.rs`: the prompt region is `Auto` when shown (and zero when hidden), so the row takes the rows it needs and the body above gives them up. This is the same way popups and modal dialogs already size themselves.

Input prompts (Search, Open, Go to line) are left alone. Their label is short, and it is the query rather than the label that can overflow, so they keep their single scrolling row with its existing horizontal-scroll behavior. The trigger is purely structural (message wider than the row), so there is no per-prompt-type list to keep in sync.

## Testing

- Added a test that a long confirmation message wraps across rows, keeps `(C)ancel` visible, and never lets a visual row exceed the width.
- Added a test that a short confirmation message stays on one row, so the wrap only engages on real overflow.
- Existing prompt row tests (caret scrolling, click-to-byte, file-open truncation, selection) still pass unchanged.
- Full `fresh-editor` lib suite: 2452 passed, 0 failed.
- `cargo fmt --check` clean, `cargo clippy -p fresh-editor` clean on the touched files.


crates/fresh-editor/src/view/shell/frame.rs
```diff
@@ -537,7 +537,15 @@ pub fn frame_tree(f: Frame) -> Node<UiMsg> {
                     None => row(),
                 },
             )
-            .h(cells(f.prompt_line)),
+            // `Auto` when shown, so a confirmation prompt that wraps on a
+            // narrow terminal takes the rows it needs (issue #3214); zero
+            // when hidden, keeping the reserved-but-empty rectangle callers
+            // anchor to.
+            .h(if f.prompt_line {
+                Sizing::Auto
```
sinelaw: Height is unbounded. This and the `Sizing::Auto` in `prompt_line` both grow with no cap and no fallback to ellipsis.

The German message is 133 columns. On a 30-column terminal it wants 5+ rows, taken from the body. On a short terminal that is most of the frame.

This also overlaps #3059, which clamps the suggestions box to the rows above the prompt. The two need to agree on who owns those rows.

crates/fresh-editor/src/view/shell/prompt_line.rs
```diff
@@ -66,15 +66,32 @@ fn ellipsis_ink() -> String {
...
 
 /// The row at `width` cells: the label, then the input's window.
 fn build(p: &PromptRow, width: u16) -> Node<UiMsg> {
+    // A confirmation prompt carries its whole message, including its hotkey
+    // options, in `message` and takes no typed input. When that message is
+    // wider than the row it used to be hard-cut at the last column, hiding
+    // trailing options like "(C)ancel" (issue #3214). Wrap it across rows
+    // instead. Input prompts keep their single scrolling row: their label is
+    // short and it is the query, not the label, that can overflow.
+    if p.input.is_empty() && p.dir.is_none() && str_width(&p.message) > width as usize {
```
sinelaw: `p.input.is_empty()` is also true for an input prompt before the user types anything.

On a narrow terminal a long label takes this branch — e.g. "Replace with (regex): " at 22 columns on a 20-column terminal. The branch returns early and emits no cursor, so the caret disappears, then the layout snaps back to the scrolling row on the first keystroke.

Test whether the prompt accepts typed input, not whether input happens to be empty right now.

sinelaw (commented):

The wrap is correct and the fixture is the real 98-column string. Two issues inline.

sinelaw (commented):

Follow-up. Reproduced #3214 against the installed 0.5.1 binary.

80x24 terminal, edited a buffer, Ctrl+Q. The prompt row comes out exactly 80 chars, cut mid-word:

```
1 buffer has unsaved changes. (s)ave and quit, (d)iscard and quit, (q)uit (recov
```

`(recoverable), (C)ancel?` is gone, so Cancel is not discoverable. Confirmed.

Measured the locale strings after placeholder expansion: en is 99 cols, de is 134. At 30 columns de needs 5 rows, which is the unbounded-height case raised in the other comment.

crates/fresh-editor/src/view/shell/frame.rs
56steve: Good catch, fixed in 68300ae1a.

Every normal prompt is back to a single row. Only a confirmation prompt grows, and it stops at three rows. If the message still doesn't fit, the third row ends in "…" with whatever the user typed after it, so the German quit prompt at 30 columns takes three rows instead of five and the buffer keeps the rest of the screen.

On the overlap with #3059: suggestions only show for input prompts, and those stay one row, so a wrapped confirmation and the suggestions list don't show up together. The list is also anchored to the top edge of the prompt row, so even if they did, it would sit above the wrapped rows rather than on them. `the_list_stays_above_a_prompt_row_that_wraps` covers that.

This also turned out to be why CI was red. Sizing the row to its content for every prompt let ordinary prompts shrink to zero rows, so "Save as:" and the command palette query never drew, and the code tour tests timed out. Those pass locally now.

crates/fresh-editor/src/view/shell/prompt_line.rs
56steve: Agreed, and it was a bit worse than that. The confirmation prompts take their answer as typed input too (press `d`, then Enter), so the wrap switched off as soon as the user pressed a key and the prompt went back to one clipped row.

Fixed in 68300ae1a. Wrapping is now decided by the prompt type: `PromptType::is_confirmation()` lists the prompts whose message spells out their answers, and only those wrap. Input prompts always keep their single scrolling row, so "Replace with (regex): " on a 20-column terminal stays on one row with its caret, typed or not. In the wrapped layout the typed answer and the caret follow the message.

There are unit tests for the typed-answer case and the long-label case, and an e2e test that drives the real quit prompt at 80 columns, types `d`, and checks "(C)ancel" is still on screen.

PR #3245, opened 2026-09-10
fix(terminal): route split resize shortcuts to the editor

kimprap:
## Motivation
With the integrated terminal focused and keyboard capture off, configured split-resize shortcuts are forwarded to the PTY instead of resizing the dock.

## Changes
- Admit `IncreaseSplitSize` and `DecreaseSplitSize` as terminal UI actions. The production change is exactly two eligibility arms.
- Preserve keyboard capture, explicit/context/global binding precedence, ordinary PTY input, active terminal, and the existing parent first-child ratio semantics.
- Add actual Kitty parser/resolver/handler coverage and a keyboard-driven, rendered-output E2E using a real shell. No default keymap or resize-policy change.

## Verification
- `cargo test -p fresh-editor --lib app::terminal_input::tests:: -- --nocapture`: 7 passed.
- `cargo test -p fresh-editor --lib input::keybindings::tests:: -- --nocapture`: 68 passed.
- `cargo test -p fresh-editor --test all_tests terminal_resize:: -- --nocapture`: 4 passed.
- Terminal UI command-registry regression: 1 passed.
- Isolated native PTY smoke with configured Kitty Ctrl+Shift+Super+L/K: dock height 10 → 8 → 10, rendered top row 30 → 32 → 30; shell markers arrived intact, editor text/tabs stayed unchanged, and the shell/application exited cleanly.
- Exact-base production with only the reviewed regressions retained compiled and produced two behavioral unit failures. Its rendered dock flow showed the resize key corrupting the shell marker; the unsatisfied semantic wait was externally stopped and all owned processes were reaped.
- `cargo check --all-targets`, `cargo fmt --check`, and `cargo clippy`: passed.

## Known upstream baseline failure
`crates/fresh-editor/plugins/check-types.sh` still fails 19 of 53 checks. All 97 plugin input files are byte-identical to base `dad823c6fd5bd72a3d39a8922b01813e53417d99`; independently executed baseline and candidate outputs match byte-for-byte, with zero new diagnostics. This PR does not change plugin API/config type definitions or repair those unrelated failures. The plugin check is **not** reported as passing.

This branch contains only terminal resize routing. The independent duplicate-copy fix is #3244.


sinelaw (commented):

No findings.

The production change is two match arms in the terminal UI fallthrough allowlist. Tests cover capture on and off, binding precedence across the global/normal/terminal/file_explorer contexts, and real Kitty byte sequences through the parser.

sinelaw (commented):

Follow-up. Reproduced the reported behavior against the installed 0.5.1 binary.

Config bound `ctrl+alt+l` to `increase_split_size` (when: normal). Opened the dock terminal with Alt+`, keyboard capture off.

- Terminal focused, 3x Ctrl+Alt+L: divider stayed at row 17. No resize.
- Ctrl+] to focus the editor, same 3 presses: divider moved 17 -> 18.

Same binding, same keys, resizes only when the editor has focus. Confirms the shortcut is swallowed in terminal focus.

sinelaw:

My earlier rebase of this branch onto master broke the build. master added a `chord: String` field to `config::Keybinding`, and the three struct literals here list every field, so they failed with E0063 on clippy, `check (no plugins)` and all three test jobs. The rebase did not conflict because the literals and the struct are in different files.

Pushed 6357c2c7f setting `chord: String::new()` in the three literals. These are key+modifiers bindings, so the field is empty and no behaviour changes. `cargo check -p fresh-editor --tests` is clean locally.

PR #3251, opened 2026-09-10
Orchestrator dock & dialogs redesign: widget-library kinds, SSH host picker, machines, five-state attention, notifications, CLI verbs

sinelaw:
Implements `docs/internal/orchestrator-ux-redesign.md`. The Orchestrator's dock and its New Workspace / Run Agent dialogs were built on the `Raw` escape hatch — the plugin painting strings the host had no model for. This moves them onto real widget kinds, adds the kinds the library was missing, and lands the semantic work the design note left open behind them: a five-state `agentState`, an attention line, notifications, an SSH host picker, a machine registry, remote worktrees, and a CLI.

63 files, ~11.5k added. The bulk is the plugin (`orchestrator.ts`, +3.9k/−1.1k) and its strings; the rest is the library additions the plugin needed, the host plumbing under them, the CLI verbs, and the design notes.

## The goal

A workspace switcher that tells you which of your agents needs you, and a dialog that can start one anywhere — locally, on a machine over ssh, or in a cluster — without the user learning three different forms. Everything below follows from those two.

## Design decisions, and what they cost

**The dialog is one field set, not three.** `Machine` is always the host control (Local first, then saved machines, then `~/.ssh/config` aliases, then `Other host…` / `Kubernetes…` / `Add machine…`), and one `Project Path` roots the workspace in every mode. The cost is that `Project Path` means a local path in one mode and a remote one in another; it is paid with a per-mode hint under the field rather than with separate forms, because the alternative was three field sets that drift.

**Disclosure is value-driven, so there is no `Advanced…` fold.** A field appears when the value above it makes it meaningful. The cost is that the form's shape changes as you type — which is why:

**Every section is reserved at its tallest shape.** Switching `Launch in`, `Machine`, the agent or the worktree toggle changes what is *in* a section, never where the sections are. The cost is blank rows in the short shapes and a real risk of getting a reservation wrong (see below); the benefit is that a dialog which re-centres under the cursor is unusable. A terminal too short for the fixed layout gets a content-sized one instead — the one place the guarantee is traded away rather than the modal overflowing.

**Reservations are measured from the form, not from live state.** Both reservation bugs found here were the same mistake: a `rowsOf` probe inherits whatever the live form says about a field the builder branches on, and only the fields the probe explicitly overrides are pinned. `tailRowsMax` inherited `target`, where the current-workspace shape has no tail at all, so the two shapes reserved different heights. The class is noted in the register; a probe is only as good as its pin list.

**Tab stops are derived from the renderer, not listed beside it.** The focus cycle used to enumerate the local worktree group's keys, so when the remote group arrived it was unreachable by keyboard. Both tails now branch on the same conditions the renderer does.

## Remote worktrees

A New Workspace on an ssh machine offers the same worktree a local one does. The repository is simply on the other side of a connection; branching a remote checkout used to mean sshing in by hand.

**One round trip per settled path.** `BatchMode`, a connect timeout, and a remote script that answers in one go whether the path is a repository, what its root is and what it has checked out. Debounced at 500 ms against the local probe's 200, because each of these is a connection rather than a `stat`.

**"Could not ask" is not "the answer is no."** These are different problems with different fixes, and calling the first one "not a git repository" sends the user to inspect a path that may be fine. A host that answered and has no repository disables the toggle; a host that never answered *keeps* it, because one asleep while the user typed is reachable by the time they press Create. The cost is a create that can still fail at the remote — accepted, because the row then reports the remote's own error, which is better information than a refusal this side guessed at.

Telling the two apart was the interesting part. `ssh` passes the remote command's exit status through, which makes "no repository" and "could not connect" the same number; the first implementation read them apart with a regex over ssh's prose. The remote script now swallows its own failures and exits 0, so a non-zero status means the transport and nothing else.

**The worktree is made before the attach, and the remote resolves its own paths.** The session has to root at a directory that exists, so `git worktree add` runs first and the transport is re-rooted at the absolute path the remote reports. This side knows neither the remote's `$HOME` nor its repository layout, and a guess at either is a path that works for one user and silently not for another — so both are resolved on the far side, under the same `~/.fresh/worktrees/<repo>/<name>` convention the local flow uses. It is idempotent for the same reason the local one is: `git worktree add` onto an existing directory fails, which would error the row on a retry of an interrupted attempt.

**Two deliberate asymmetries with the local form.** With the toggle off, the remote form drops `Checkout branch` instead of showing it — nothing performs a remote in-place checkout, and offering the field promised one. And on the unreachable path no refs are collected at all, so the create falls through to bare `git worktree add`, letting git name the branch after the worktree directory rather than having this side invent a ref for a repository it could not reach.

## Widget library (`fresh-editor-core`, `fresh-ui`, `fresh-core`)

- **`Radio`** — an inline single-select group, host-owned selection, ←/→ and Home/End, and a description arm that splits the row into one press target per option at the formatter's own byte ranges. Too narrow for its panel, it sheds the gaps between options and then the label column rather than running off the edge.
- **`Label`** — one row of static text, styled, indentable into a form's field column, optionally wrapping or eliding. This is what every `Raw` in the dialogs was standing in for.
- **Right-aligned label gutter** — `labelAlign: "right"` honoured by `Text`, `Toggle`, `Number`, `Dropdown` and `Radio` through the existing `label_width`.
- **Click outside closes things.** A dropdown's pop-over closes on a press anywhere on its own row *except* the `[value ▼]` button: the anchored-node exemption is geometric — the pressable piece inside the anchor's rectangle, not the whole row. A keyed `Overlay` reports a `dismiss` on an outside press, and the press still reaches its target.
- **`getHomeDir`** — a plugin reading a dotfile asks the editor for the home it resolved rather than reading `$HOME`, so the form's options stop depending on whoever ran the tests.

Both new kinds are description-only: the base branch removed `WidgetImpl::collect` and `render_spec`, so they call the row formatters in `render.rs` instead of carrying collectors.

Per `retained-mode-ui.md`'s working rule — *a library change needs a caller in the same PR and a test that fails without it* — `WidgetMutation::SetRadio`, a keyed `Label`, and the `RADIO_GAP` export were dropped before merge. All three were added for symmetry and called by nothing.

## Dock

Five-state `agentState`: `working` / `blocked` (quiet, and the last screen lines read as a question) / `done` (quiet after a ≥1.5 s burst that happened while the window was not active) / `idle` / `unknown`. Rows show `*` `●` `✓` `·` `?`, folders roll up `●n ✓n`, and the header counts only while a count is non-zero. Redraws after a resize or a dock toggle do not count as work.

**The detection heuristics are data, not code.** They are guesses about other people's programs and will be wrong in ways this repo cannot predict, so the built-in set yields to a file in the data dir and to a published URL when its version is newer. A file whose rules are all unusable falls back to the built-ins rather than leaving nothing detecting, and a rule with a catastrophic regex is refused rather than run. `Explain State` shows the decision chain, so a wrong badge is diagnosable rather than mysterious.

## CLI (`fresh --cmd …`)

`workspace list`, `agent list/get/explain`, `agent wait` (exit 3 on timeout), `agent start` (exit 1 when it exited first or stayed silent). It reuses `script run`'s channel and capability token deliberately: that token is the channel's whole security posture, and a second unauthenticated front door would need its own audit. A flag written with nothing after it is a usage error rather than a silent default, and the exit codes read a sentinel the script emits rather than a substring of its own JSON.

## Not built, and why

- **Persistence by default** (a bare `fresh` attaching to a daemon): a launch-model change for every user, not a dock change. `fresh -a` and in-app detach already exist; flipping the default and adding a session switcher is its own decision with its own test surface.
- **A layout tier inside a workspace** (saveable pane arrangements): needs a host concept between window and buffer. Nothing in the plugin can stand in for it.

## What this leaves open

`docs/internal/orchestrator-layout-asks.md` registers every place this work still does layout by hand — the plugin measuring the screen, pre-fitting a string, or building a description only to count its rows — each with the description primitive that would retire it, what it costs and what it blocks. Nothing in it is a behaviour bug. The largest remaining item is the dock menu's wall-clock guard, which waits on keyed widget nodes; an attempt at it is in this branch's history, reverted, with what it ran into.

## Verification

e2e coverage follows the new layout and the cases review found untested. The worktree tests wait for the *disabled* row before asserting it clears — `!contains("non-git")` holds before the probe has run at all, and the "New branch name" row renders whenever `projectPathIsGit !== false`, so both obvious assertions passed against the bug they were written for. The IPv6 destination split is caught through the pending row's `ssh:<destination>` label, which is where the parsed destination first becomes observable.

Beyond the suite, every stage was driven by hand in tmux with colour captures reviewed before pushing. The remote worktree was exercised end to end against a real OpenSSH server on a user port with a planted `~/.ssh/config` host and a real repository: probe, branch read off the remote, worktree created on a new branch, transport re-rooted, and a live remote shell inside it on the expected branch — including the non-git, unreachable, and unreachable-then-woken paths.

https://claude.ai/code/session_01XDjTpSe3pvwLxaLVAJZVes


sinelaw:
CI note on head 1963602: `test ubuntu-latest` is red on a single timeout, `remote_channel_timeout_tests::test_reconnection_via_replace_transport` (9866 of 9867 passed). That test exercises the remote channel's transport replacement, which this PR does not touch — its only change under `src/server/` is a test-fixture field for the new `label_align` mount option — and the same test passed on the previous head's Linux run and passes locally in 2.4 s. Nothing to port; I will re-run the failed job once the macOS and Windows jobs on this head finish, and keep watching.

---

sinelaw:
Same head, `test macos-latest`: 9811 of 9812 passed, including all 13 New Workspace completion-popup tests the 75%-width change was for; the one timeout is `e2e::markdown_compose_scrollbar_markers::heading_marks_survive_exploring_the_document`, which this PR does not touch (no markdown or scrollbar-marker change in the diff) and which passes locally in 3.5 s. `test windows-latest` is green. Re-running the two failed jobs once, now that the run has finished.

---

PR #3255, opened 2026-09-11
Delete the widget text projection: plugin panels render only through the UI tree

sinelaw:
## Why

Plugin panels (`WidgetSpec` trees) were being rendered **twice**, by two separate systems:

1. **The text projection** (`render_collected` and friends in `fresh-editor-core`) — walked the spec and produced strings, plus geometry on the side: click rectangles, a layout-box arena, and the scroll window each list was painted into.
2. **The retained-mode description** (`view/shell/widgets.rs`) — turns the same spec into `fresh-ui` nodes that the tree lays out, hit-tests and scrolls.

Both ran on every panel. Each owned part of the truth, and they disagreed: two answers for how wide a widget is, two focus rings, two ideas of where a click landed. That is the defect class this migration exists to end, and it had already shipped several times.

This PR finishes the migration and deletes the projection. The tree is the only thing that lays a panel out.

Three things blocked that, and each is cleared here:

- **The markdown document view** re-wrapped its own text, so it needed a renderer with a width. It is now a wrapped text run inside a viewport, and the tree does the wrapping.
- **An anchored popup couldn't size itself to its content**, because a divider inside it was a string of `─` of a computed length — so measuring the box asked a question the box was in the middle of answering. A divider is now a background the backend tiles, which contributes nothing to the measurement.
- **A pane-mounted panel's buffer** — which the status bar's `Ln`/`Col`, code-tour's `getBufferText` and the page reading row all read — was produced by the projection. It is now derived from the tree's own rows.

## Decisions and trade-offs

- **The plugin-facing API is unchanged.** `WidgetSpec` is the same; kinds still own their state, their keys and what a press means. Only the *rendering* half was removed.
- **A pane panel still has a text buffer**, because things outside the panel legitimately read it. It is now derived from the tree instead of rendered separately, so there is still one authority. Cost: the panel's subtree is painted once per layout to produce it (written only when the rows actually change).
- **One behaviour change:** that buffer now fills on the *next* frame rather than synchronously at mount.
- **`fullWidth` on a button changed meaning** — from "pad the label with spaces out to the panel width" to an actual width. Same appearance, but a menu inside a popup can now ask for full-width rows (before, that stretched the popup to the whole panel), so the orchestrator plugin stopped padding its own menu labels to align them.
- **Some dead state was flagged rather than removed** (`user_scrolled` on three instance-state variants now has no writer) to keep the diff bounded. It is recorded in the doc.

## What changed

**Added**
- `fresh-ui`: `Draw::Rule` (a glyph tiled across a rect, so a divider is a background not text), `Ui::text_rows_in` (the rows a wrapped run was laid out into), `Ui::paint_subtree` (one subtree's display list), and a byte-under-pointer fallback so a captured drag still reports where it is.
- `app/pane_mirror.rs` — derives a pane panel's buffer from the tree's rows.

**Removed** (~9,600 lines from `fresh-editor-core` alone)
- The text projection: `render_collected`, `WidgetImpl::collect` and every kind's collector, the container assemblers, `CollectedOutput` / `RenderOutput` / `RenderContext` / `RenderOptions`, `render_spec*`, `render_button`.
- Its geometry: `HitArea`, `LayoutBox` and `layout_box.rs`, `PaintedWindow`, `WidgetPanelState::{painted, boxes}`, and `List`/`Tree::on_wheel`, which moved a scroll window nothing drives any more.
- The floating panel's painted fallback: `FloatingWidgetState::entries`, `Spot`'s row/column counts, `Panel::{height, anchored_width}`, `render_floating_spec`.
- `widget_text_drag` and `PointerGrab`, replaced by the run's own pointer capture.

**Changed**
- The markdown document view is `viewport(text_runs(...).wrap(Hanging))`; `Up`/`Down`/`Home`/`End`, clicks and drag-selection are resolved from the rows the tree produced.
- Every mount, update and host re-render *resolves* the spec (carry state, clamp focus) instead of rendering it.
- A panel's box measures itself on both axes; `Panel.interior` is no longer optional.
- `docs/internal/retained-mode-ui.md` rewritten as one as-built record with an open-work register; `shell-stylesheet-design.md` folded into it.

## Review

Six findings from a review of this PR are fixed in c225ff8 — three were real bugs introduced here (a pane-mounted panel's widget could not be focused; a drag off the *top* of a document selected to the bottom; a prose drag crossing a terminal pane forwarded its motion into that PTY). The two behavioural ones have regression tests, each confirmed to fail against the code before the fix.

## Verification

- `fresh-editor-core` lib: 1214 passed · `fresh-editor` lib: 2503 passed · `fresh-ui` suite green.
- e2e subset (page, search, git_log, widget, welcome, panel, dock, orchestrator, settings, markdown, code_tour, sidebar, popup, dropdown): 1069 passed.
- Both feature configurations build with `--all-targets`.
- Driven by hand in a terminal: the dock's context menu hugs its content while every row's highlight spans it, the welcome page renders (rules, cards, prose, code samples), and the status bar's `Ln`/`Col` tracks the tree-derived buffer while paging.

https://claude.ai/code/session_01LgoWy44P4YkRntEqARMad4


crates/fresh-editor/src/app/shell_host.rs
```diff
@@ -2180,6 +2258,49 @@ impl Editor {
...
+                slot,
+                widget,
+                byte,
+                mods,
+            } => {
+                let Some(pk) = self.panel_key_of_slot(&slot) else {
+                    return;
+                };
+                if self.widget_registry.focus_key(&pk) != Some(widget.as_str()) {
+                    self.apply_ui_fact(
```
sinelaw, line 2278: **A prose press can't focus a pane-mounted panel's widget.**

`panel_key_of_slot` was extended with the `Slot::Pane` case, but the focus half is delegated to `UiFact::WidgetFocus`, whose own slot match is still

```rust
let key = match slot {
    Slot::Dock => ..., Slot::Floating => ..., Slot::Sidebar(i) => ...,
    _ => None,
}
.or_else(|| match slot { Slot::PromptToolbar => ..., _ => None });
let Some(key) = key else { return; };
```

so `Slot::Pane(..)` falls out and the arm returns without focusing anything.

`code-tour.ts` mounts each tour as a virtual buffer in the shared Utility Dock — a pane-mounted widget panel — and its prose carries `key: "prose"`. Clicking that prose therefore: resolves `pk` fine, silently fails to focus, and `move_prose_caret` moves the `TextEdit`. But the description only emits `selection_bytes` under `cx.is_focused(Some(k))`, so no caret and no selection is painted, and `prose_vertical_key` bails because `focus_key != "prose"` — so `Up`/`Down`/`Home`/`End` do nothing either. The press is effectively inert.

The old path went through `deliver_widget_hit` / `UiFact::WidgetHit`, which *does* have a `Slot::Pane` arm.

---

crates/fresh-editor/src/view/shell/sidebar.rs
```diff
@@ -334,6 +334,7 @@ fn panel_body(i: usize, p: &super::panel::Interior) -> Node<UiMsg> {
                 marker_gutter: interior.marker_gutter,
                 hovered_item_key: interior.hovered_item_key.clone(),
                 hovered_popup_row: interior.hovered_popup_row.clone(),
+                reveal: fresh_ui::behavior::anchor::Anchor::new(),
```
sinelaw: **This throws away the host's reveal anchor.**

Every other surface passes the interior's anchor down — `dock::column`, `splits::panel_content` and `overlay_prompt::toolbar_band` all say `reveal: i.reveal.clone()` — and `Editor::panel_interior` puts the host-owned handle (`prose_reveal_for(&key)`) on `Interior::reveal` for `PanelSlot::Sidebar(_)` too. Here a brand-new `Anchor` is constructed *inside the `layout_reader` closure*, so it is also a different anchor on every layout pass.

`Interior::reveal`'s own doc states the consequence: "an anchor binds to its element on mount and a fresh one each frame would bind to nothing." `Editor::move_prose_caret` calls `self.prose_reveal_for(panel_key).reveal_byte(..)` on the host's anchor, which is bound to nothing here — so in a plugin's sidebar section a markdown document's viewport never scrolls to follow the caret: pressing `Down` past the bottom of the window (or clicking below it) moves the caret out of view and the rows stay put.

```suggestion
                reveal: interior.reveal.clone(),
```

---

crates/fresh-editor/src/view/shell/widgets.rs
```diff
@@ -2057,218 +2098,126 @@ fn node_body(spec: &WidgetSpec, width: u16, cx: &Ctx<'_>, site: Site) -> Node<Ui
...
+                                    slot,
+                                    widget: kp.clone(),
+                                    byte,
+                                    mods: e.mods,
+                                }))
+                            }),
+                        )
+                        .on(
+                            GestureKind::Move,
+                            std::rc::Rc::new(move |e: &Event| {
```
sinelaw, line 2189: **The `Move` handler never claims the event** — the `Press` and `Release` handlers beside it both call `e.stop()`, this one does not.

`hit.rs::propagate` only reports `claimed` when `ctl.flow == Flow::Stop`; returning a message is not enough. So during a prose drag `Dispatched::claimed` is `false` (outside a modal layer, where `pointer_owned()` would cover it), and `handle_mouse_event` falls past the tree into `pane_content_takes_pointer` → `try_forward_mouse_to_terminal`.

That path used to be gated on the grab this PR deletes:

```rust
-        let chrome_drag_active = super::chrome::pointer_grab(self).is_some();
-        if !chrome_drag_active && !context_menu_open {
+        if !context_menu_open {
```

and `pointer_grab` returned `Some(PointerGrab::WidgetText)` for exactly this drag. Concretely: with a code-tour prose panel in the Utility Dock beside a terminal pane, press on the prose and drag the pointer over the terminal — each motion report is now forwarded to the shell running there.

Also worth noting this handler fires on plain hover motion too (no button held), emitting a `WidgetProseDrag` fact per motion event that the host discards because `prose_drag` is `None`.

---

crates/fresh-editor/src/app/shell_host.rs
```diff
@@ -2180,6 +2258,49 @@ impl Editor {
...
+            UiFact::WidgetProseDrag { widget, byte, .. } => {
+                let Some((pk, w)) = self.prose_drag.clone() else {
+                    return;
+                };
+                if w != widget {
+                    return;
+                }
+                // Past the text is the end of it: a drag that leaves the last
+                // row selects to the document's end, as it did in the buffer.
+                let target = byte.unwrap_or_else(|| self.prose_len(&pk, &w));
```
sinelaw: **"Past the text" is both directions, not just past the end.**

`Event::text_byte` is `None` whenever the pointer is outside the run's rectangle — above the first row exactly as much as below the last (`TextRun::text_byte_at` returns `None` for `local.y < 0`, and `text_byte_under` skips any child whose rect does not contain the point). Collapsing that to `prose_len()` means an upward drag off the top of the document extends the selection *forward to the end* of it.

Repro: press in the middle of a markdown document, drag the pointer up above the viewport — the selection jumps to cover everything from the anchor to the end of the document instead of growing toward the start.

The replaced `handle_widget_text_selection_drag` clamped into the region's row band in both directions ("dragging past either edge selects to the visible edge"). Recovering that needs the drag's direction: e.g. compare `e.pos.y` against the run's rect in the `Move` handler and send `byte: Some(0)` above it, `None` (→ end) below it.

---

crates/fresh-editor/src/app/mod.rs
```diff
@@ -1504,12 +1516,6 @@ pub struct Editor {
 }
 
 /// See [`Editor::widget_text_drag`].
```
sinelaw: Orphaned doc comment: `WidgetTextDrag` went, its `/// See [...]` stayed, and it now documents `FLOATING_PANEL_BUFFER_ID` while linking to a field that no longer exists (`widget_text_drag` → `prose_drag`). That is a `broken_intra_doc_links` warning plus an unrelated opening sentence in the rendered docs.

```suggestion
```

Same pattern at `crates/fresh-editor/src/app/mouse_input.rs:927`, where `/// Handle mouse drag event` is now attached to `clear_active_window_drag_state` after `handle_mouse_drag` was deleted.

---

crates/fresh-editor/src/app/render.rs
```diff
@@ -6404,6 +6408,20 @@ impl Editor {
     /// is the same question without the clone, and the surfaces that route a
     /// press by it (`view::shell::dock::column`,
     /// `render_floating_widget_panel`) must keep asking the same one.
+    /// The reveal anchor a panel's markdown document is scrolled through —
+    /// the same one every frame, because an anchor binds to its element on
+    /// mount and a new one each frame would bind to nothing.
+    pub(crate) fn prose_reveal_for(
```
sinelaw, line 6414: `prose_reveal` only ever grows: this is the sole writer and it is `entry(..).or_insert_with(..)`. Nothing removes an entry when a panel unmounts, while the sibling map added in this PR does (`self.pane_mirrors.remove(panel_key)` on the unmount path in `plugin_dispatch`).

A `PanelKey` is `(plugin, id)`, so opening and closing many code-tour tabs leaves one live `Rc<Anchor>` per tour for the life of the process. Worth dropping it alongside `pane_mirrors` on unmount.

---

sinelaw (commented):
Reviewed the diff against `master` (d042390) for correctness. The library-side pieces (`Draw::Rule`, `paint_subtree`, `text_rows_in`, the `text_byte_under` fallback) and the `pane_mirror` derivation look sound; the findings cluster around the new prose (markdown document) input path and one dropped anchor.

6 findings, most severe first — 4 substantive, 2 nits.

---

crates/fresh-editor/src/app/shell_host.rs
sinelaw: Confirmed and fixed in c225ff8. The match was `panel_key_of_slot` with the `Pane` leg missing, so the arm now calls that resolver — the same one the press beside it already used — and the duplicate is gone.

Regression test: `a_focus_fact_from_a_pane_reaches_its_panel` mounts a prose panel on a pane's buffer, checks the tree calls it `Slot::Pane(..)`, and applies the fact. Against the code before the fix it fails with `focus_key == Some("")`, exactly as described.

---

crates/fresh-editor/src/view/shell/sidebar.rs
sinelaw: Right — taken as suggested in c225ff8: `reveal: interior.reveal.clone()`, matching `dock::column`, `splits::panel_content` and `overlay_prompt::toolbar_band`.

---

crates/fresh-editor/src/view/shell/widgets.rs
sinelaw: Both halves fixed in c225ff8, using the idiom the repo already has for this (`splits.rs`, `tabs.rs`):

```rust
if !e.captured {
    return None;
}
e.stop();
```

`e.captured` answers the hover half — a bare motion across the run is not this run's drag, so it travels on and emits no fact — and `stop()` keeps the real drag from reaching `pane_content_takes_pointer`, which is what `PointerGrab::WidgetText` was doing. `Event::captured`'s own doc asks for exactly this ("a listener that treats a move as a drag reads this").

---

crates/fresh-editor/src/app/shell_host.rs
sinelaw: Confirmed — fixed in c225ff8, close to your suggestion but with the direction carried as a fact rather than pre-resolved to a byte: `WidgetProseDrag` gains `above: bool`, set in the `Move` handler from `e.local.y < 0`, and the applier maps `None + above → 0`, `None → prose_len`. The reason for the extra field rather than `byte: Some(0)` is that the run has the rectangle but not the document, and the applier has the document but not where the pointer went — this way neither guesses the other's half.

The existing end-to-end test now drags upward past the first row; against the old code it fails with `Some((26, 129))` (anchor → end) where it wants `Some((0, 26))`.

---

crates/fresh-editor/src/app/mod.rs
sinelaw: Both deleted in c225ff8 — this one and the `/// Handle mouse drag event` left on `clear_active_window_drag_state`.

---

crates/fresh-editor/src/app/render.rs
sinelaw: Fixed in c225ff8 — dropped on both unmount paths, right beside the `pane_mirrors.remove(..)` the anchor should have followed in the first place (`plugin_dispatch` and `widget_runtime`).

---

Issue #3258, opened 2026-09-11
Persistent LSP diagnostic errors possibly related to vim keybinds


thedadams:
I have been experiencing stale LSP diagnostics. A previous fix definitely made it better, but I am still seeing issue. I have found two reproductions. The steps are a little hard to explain so I created a video. One thing I want to point out (I say it in the video): in the second reproduction I was seeing the error reported in both files consistently when I first discovered that reproduction. I tried it 4 or 5 times and the "redeclared in this block" error was shown in both files. It only showed in one file in the video.

https://caps.thedadams.com/s/ngy361amth0zbh2

sinelaw:
Reproduced, deterministically, in an e2e test.

**Symptom**

Deleting a line with vi mode's `dd` leaves the language server holding the pre-delete text. Diagnostics for the deleted line stay on screen, hover on the file stops working, and saving does not clear either. Typing whitespace, or deleting the line character by character with `x`, does clear it.

**Steps to reproduce**

1. Open a file in a language with a running LSP server.
2. Enable Vi mode.
3. Put the caret on a line that produces a diagnostic (an unused variable, a duplicate declaration).
4. Press `dd`.
5. Save.

Expected: the diagnostic goes away; hover keeps working.
Actual: the diagnostic stays, hover returns "No hover information available", and the server's copy of the document still contains the deleted line — no `didChange` is sent for the `dd` at all.

The same applies to any plugin-driven edit, not only `dd`. Vi mode's `x`/`dw` go through a different path and do notify the server, which is why character-by-character deletion behaves differently.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3261, opened 2026-09-13
fix(lsp): synchronize plugin text edits with language servers

thedadams:
Vi mode implements dd through editor.deleteRange. Direct plugin text-edit commands mutated buffers without emitting didChange, leaving the language server on stale text and diagnostics stuck on deleted lines.

Build changes against the target buffer before applying plugin inserts and deletes, then notify each attached server. Cover the failure with a regression test that verifies the server converges and clears its diagnostic.

Addresses https://github.com/sinelaw/fresh/issues/3258


sinelaw:
Heads up: I arrived at the same fix independently before seeing this PR, and have now rebuilt that work on top of yours — your commit is cherry-picked unchanged as the base of #3265, with the differences as separate commits so they are reviewable on their own.

Three adjustments to this commit:

- `collect_lsp_changes_for_buffer` is derived without checking `lsp_change_could_be_sent` first. That converts byte offsets to UTF-16 by reading the line up to the edit, which `apply_event_to_active_buffer` gates for cost reasons (megabytes per edit on a one-long-line file). It is paid here on every plugin edit, including buffers with no server attached.
- `handle_insert_at_cursor` sends unconditionally, without the `edited` guard its two siblings have. If the buffer or split does not resolve, the edit is a no-op but the `didChange` still goes out.
- `collect_lsp_changes_for_buffer` returning an empty change set for an unknown buffer is silent.

One CI item: `handle_plugin_command` is compiled only under the `plugins` feature, so `plugin_delete_range_notifies_lsp` fails `check (no plugins)` with E0599. This PR has no CI runs on it (fork workflows need approval), so it has not surfaced here — it did on mine.

Separately, the second symptom in #3258 (the error showing in two files) has another cause that this PR does not touch: `notify_lsp_save_buffer` takes the URI from the buffer it is given but the text and language from whichever buffer is focused, so Save All sends each server the focused file's contents under another file's URI. Fixed in #3265 with a test.

---

sinelaw:
Thanks! Cherry-picking this work into #3265 which includes a few more fixes on top

Issue #2447, opened 2026-06-22
vi mode: Vim parity gaps and unified plugin refactor (tracking)


sinelaw:
(Automated summary)

vi mode is implemented in `crates/fresh-editor/plugins/vi_mode.ts` (~4200 lines). Six open issues report Vim parity gaps. All reproduced side-by-side against Vim 9.1 (identical keystrokes, buffer/status as ground truth):

- #2437 — `cw` consumes trailing whitespace; should act like `ce`.
- #2438 — `>>` / `<<` and visual `>` / `<` are no-ops.
- #2439 — `i"` / `a"` (and `di"`/`ci"`) do not search forward on the line; only work when cursor is already inside the quotes.
- #2441 — `f`/`t`/`F`/`T` not recognized in operator-pending (`df`/`dt`/`cf`/`ct` hang); `;`/`,` repeat is a no-op.
- #2442 — `j`/`k` onto a shorter line park the cursor one past the last char; `x` then deletes the newline and joins lines.
- #2443 — dot-repeat (`.`) of `o`/`O`/`a`/`A` does not replay the command and corrupts the buffer.

## Root cause

The plugin has no shared model for motions, operators, cursor position, and inclusive/exclusive ranges; handlers call host editor actions directly. Each issue is a symptom of that gap. The `vi-operator-pending` mode has no `f`/`t`/`F`/`T`/`;`/`,` and no `>`/`<` bindings. (Ref: PR #2398 discussion.)

## Proposed approach

Introduce a core engine the keybindings dispatch into:

1. Cursor model with a normal-mode clamp (cursor ≤ last char) + remembered desired column — #2442.
2. Motion→Range abstraction `{ start, end, inclusive|exclusive, linewise|charwise }`: makes `f`/`t` usable in operator-pending and stored for `;`/`,` (#2441); `cw`→`ce` special case (#2437); text-object forward-search rule (#2439).
3. Operator-consumes-Range; add `>`/`<` indent operators routing to the host indent/dedent actions — #2438.
4. Unified change-record + single insert-session for dot-repeat: replay the command, then the captured typed text — #2443.

Also covered by the same model (raised in PR #2398 review, not separately filed): visual `vW`/`vB`/`vE` selection range; `*`/`#` forward scan when cursor is off-word; `}` at EOF; stale `Ln`/`Col` readout after `setBufferCursor` motions.

[...]

PR #3269, opened 2026-09-14
vi mode: give the plugin a state model and a motion/range core

sinelaw:
Part of #2447.

## The objective

The six parity bugs tracked under #2447 are closed. This is the half of that issue left over — its root-cause section:

> The plugin has no shared model for motions, operators, cursor position, and inclusive/exclusive ranges; handlers call host editor actions directly. Each issue is a symptom of that gap.

So: build the model, then find out what its absence was hiding.

## The problems

**The state didn't distinguish possible from impossible.** `ViState` was a mode string with its associated data as sibling nullable fields. An operator with a text-object modifier but no operator, a visual range with no anchor, a `LastChange` missing the fields its own replay branch reads — all were representable, and only convention kept them from happening.

**Modal state outlived its buffer.** A pending operator, a visual anchor and an insert session are byte offsets into one buffer, and nothing reset them when the active buffer changed.

**Every command was written once per mode.** `w`, `b` and `e` each existed three times — normal, operator-pending, visual — and the two word families twice more on top of that. The uppercase family computed byte offsets and knew Vim's awkward rules (`dw` on a line's last word stops at the line end; `cw` is `ce`); the lowercase family composed editor actions and did not.

**Four implementations of "apply an operator to a range"**, disagreeing about the register's shape, where the caret lands, and whether `c` flushes before entering insert. Each disagreement was a bug.

**And the plugin had never been measured against Vim.** Until it was, the gaps were whatever someone had noticed.

## Decisions and trade-offs

**Verification drove the work.** A harness runs the same keystrokes through headless `vim -s` and through the editor in tmux and diffs the saved file. It found parity bugs no amount of reading would have surfaced, because they live on the async boundary between plugin and editor — and it was the safety net that made restructuring the command layer worth attempting at all.

**The reference is Vim, including where Vim is strange.** `v$d` really does join the next line up, unlike `d$`; `di(` really does scan forward past the end of the line when the caret is outside every pair. Both are implemented as Vim has them, with the comment that stops the next reader from "fixing" it.

**Tests that encode current behaviour aren't a safety net.** Several tests here asserted the plugin's pre-Vim behaviour, having been written by observing what it did. Their expectations are corrected, with the reasoning recorded at each site.

**The existing computed path was generalised, not replaced.** The uppercase word motions already carried Vim's end-of-line rules; the lowercase ones were folded into that rather than the other way round, so the rules were inherited instead of reimplemented.

**An unprimed review was commissioned against the finished branch** — no steer, no summary of what to look for. It found ten real bugs, one of them a regression introduced earlier in the same branch, and one commit message that claimed an invariant the code did not hold. All are fixed; the claim was made true rather than reworded.

## The design

**State** is three kinds of thing, separated by lifetime: `state`, what the current mode owns, cleared by `switchMode`; `memory`, what outlives any mode (`.`, `;`, `n`, the paste shape); and `LastChange`, a discriminated union with one variant per command shape, each carrying exactly the fields its replay reads.

**Motions** resolve to a target and a *kind* — exclusive, inclusive or linewise. The kind is the whole of the difference between `dw` and `de`, and between `dw` and `dj`; it was open-coded at six sites that could disagree, and several did.

**Ranges** carry the one property that outlives resolution — charwise or linewise — and `applyOperator` is the only code that acts on one. Operator + motion, text objects, find-char, the doubled line operators and the visual modes all resolve a range and hand it over.

**Modes are adapters, not implementations.** Normal mode moves the caret to the target, operator-pending turns it into a range, visual mode makes it the head. One resolver each.

**Buffer switches** reset the modal state and leave `memory` alone, since Vim's `.` and `;` are recorded as commands rather than offsets.

## Result

**128 keystroke sequences produce byte-identical files to Vim 9.1** — motions, operators, text objects, find-char, both visual modes, indent, dot-repeat and edge cases — with 16 new e2e regression tests alongside.

Not yet folded in: find-char and the paragraph motions still resolve their own ranges, and the `select_*` composition survives for motions with no pure resolver. One core with two stragglers.

https://claude.ai/code/session_01DhmH75y5k21URZeKqkJ4LH


sinelaw:
**`supply-chain audit` is failing, and it is not this branch's.**

[RUSTSEC-2026-0285](https://rustsec.org/advisories/RUSTSEC-2026-0285) — rustls accepts TLS 1.3 handshake messages across encryption-level boundaries — was published against `rustls 0.23.40`, which reaches us through `ureq` → `rustls-platform-verifier`.

Why it isn't this PR's:

- the branch touches no dependency file at all (`git diff --name-only origin/master...HEAD` matches nothing under `Cargo.*`);
- `master` pins the identical `rustls 0.23.40`, so it is equally affected — its last CI run simply predates the advisory, and `cargo-deny` fetches the database fresh on every run;
- the failing check is `advisories`; `licenses` and `sources` both pass.

The remedy the advisory names is a one-line lockfile bump:

```
cargo update -p rustls   # to >= 0.23.45
```

I have deliberately not applied it here. A TLS dependency bump inside a vi-mode plugin change is not something a reviewer should have to take on trust alongside 2,000 lines of unrelated diff, and it fixes `master` rather than this branch — it belongs in its own commit where it can be seen. Re-running the job would fail identically, since the advisory is not going to un-publish.

Everything else on this PR is green.

---

Issue #988, opened 2026-02-09
support dap


asukaminato0721:
https://microsoft.github.io/debug-adapter-protocol//

PR #3277, opened 2026-09-14
Add LSP code lens support

sinelaw:
Adds LSP code lens support. Lenses render as a line of virtual text above the code they describe, and clicking one runs it. On by default (`editor.enable_code_lens`).

## How it works

- Fresh asks the server for `textDocument/codeLens` when a file opens or changes, and resolves any lens that needs `codeLens/resolve`.
- Each lens becomes a virtual line, reusing the machinery inlay hints and git blame already use — so there is no new drawing code.
- Clicking a lens sends its command back to the server as `workspace/executeCommand`. For most servers that is the whole story and nothing else is needed.
- Some lens commands are meant for the *client* to run, not the server. LSP does not define what those mean, so the core interprets none of them: a plugin claims the command name (`registerLspClientCommands`) and handles it on the `lsp_execute_command` hook. `rust-lsp.ts` uses that for rust-analyzer's Run lenses, turning them into a terminal running cargo.

## Gaps

- **Client-side commands need a plugin.** Lenses the server executes itself work out of the box. Lenses that hand the client something to run do not — the client has to know what the payload means, and that knowledge lives in a plugin. rust-analyzer's Run lenses are the only ones wired up so far; any other server doing the same thing needs its own plugin to claim its commands.
- **rust-analyzer's Run lenses need the bundled `rust-lsp` plugin.** A build without plugin support gets no Run lens at all — not even an inert one, since declaring the command is also what makes rust-analyzer emit the lens. Server-executed lenses are unaffected.
- **No Debug.** Only Run works. The Debug lens is deliberately not shown, because fresh has no debugger yet (#988), and not declaring the command keeps a dead lens off the screen.
- **Not visible out of the box.** Language servers are blocked until the workspace is trusted, and rust's LSP `auto_start` defaults to false — so nothing appears until both are dealt with.

## Tested

Unit and e2e tests cover rendering, the click path, and the command routing. Also checked by hand against a real rust-analyzer: lenses line up with the symbol they describe, and clicking `▶︎ Run Test` opens a terminal that runs the test.

https://claude.ai/code/session_01S6GqV2iS6vCc4Q56iDTt6H


crates/fresh-editor/src/app/plugin_dispatch.rs
```diff
@@ -6288,6 +6293,37 @@ impl Editor {
...
+        let running = self
+            .lsp()
+            .map(|lsp| lsp.running_servers())
+            .unwrap_or_default();
+        for language in running {
+            tracing::info!(
+                "Restarting LSP for '{}' so it re-handshakes with the newly claimed client commands",
+                language
+            );
+            self.handle_restart_lsp_for_language(language);
```
sinelaw: **Re-handshake restart re-enables servers the user opted out of.**

`handle_restart_lsp_for_language` goes through `LspManager::manual_restart`, which is the *user-initiated* restart path and carries its side effects (manager.rs:1645-1653):

```rust
self.clear_cooldown(language);
self.disabled_languages.remove(language);
self.allowed_languages.insert(language.to_string());
```

`allowed_languages` is sticky for the session (only removed on an explicit stop/disable), and `force_spawn` treats it as "user explicitly started this language", spawning **every** configured server for it "even if individually disabled or marked not-auto-start" (manager.rs:1287-1303).

Scenario: a user configures two rust servers — `rust-analyzer` (`auto_start: true`) and a second entry with `enabled: false` they keep around but never want running. They open a Rust file, rust-analyzer autostarts, then `rust-lsp.ts` finishes loading on the plugin thread and calls `registerLspClientCommands`. This loop marks `rust` as manually allowed, and the disabled second server is spawned. Nothing the user did asked for that.

A targeted re-handshake (shut down + respawn without touching `allowed_languages` / `disabled_languages` / cooldown) would avoid inheriting the user-intent semantics of `manual_restart`.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
```diff
@@ -6288,6 +6293,37 @@ impl Editor {
...
+    fn handle_register_lsp_client_commands(&mut self, commands: Vec<String>) {
+        if !crate::services::lsp::client_commands::register_all(commands) {
+            return; // nothing new — every claim was already known
+        }
+
+        let running = self
+            .lsp()
+            .map(|lsp| lsp.running_servers())
+            .unwrap_or_default();
+        for language in running {
```
sinelaw: **Every restarted language gets the *active buffer's* path as its root hint.**

`handle_restart_lsp_for_language` derives `file_path` from `self.active_buffer()` (line 6329-6333) and hands it to `manual_restart` → `force_spawn` → `resolve_root_uri(language, file_path)`, which runs `detect_workspace_root(path, markers)` with *that language's* root markers (manager.rs:915-941).

That is fine when the caller picked the language (the `restartLspForLanguage` plugin command), but here the loop restarts **all** running languages while the path stays the active buffer's.

Scenario: a workspace with a Rust crate and a `web/` TS app, both servers running, a Rust file focused. `rust-lsp.ts` claims `rust-analyzer.runSingle`; the TS server is torn down and respawned with `detect_workspace_root("<crate>/src/main.rs", ["package.json", "tsconfig.json"])` — no TS marker above the Rust file, so the TS server comes back rooted inside the Rust crate and loses its project.

Each language's root should be resolved from a buffer of that language (or the root the handle already had) rather than from whatever happens to be focused.

---

crates/fresh-editor/src/services/lsp/client_commands.rs
```diff
@@ -0,0 +1,62 @@
...
+use std::collections::BTreeSet;
+use std::sync::{LazyLock, RwLock};
+
+static CLIENT_COMMANDS: LazyLock<RwLock<BTreeSet<String>>> =
+    LazyLock::new(|| RwLock::new(BTreeSet::new()));
+
+/// Claim `commands` for plugin handling. Idempotent; returns whether any
+/// of them were not already claimed, which is what tells the caller a
+/// running server needs to re-handshake to hear about them.
+pub fn register_all(commands: Vec<String>) -> bool {
```
sinelaw, line 44: **Claims are write-only: nothing ever unregisters them, and the registry outlives the plugin *and* the `Editor`.**

There is no `unregister`/`clear`, and `CLIENT_COMMANDS` is a process-global `static`. Two consequences:

1. **Unload/disable leaves a dead lens.** A user disables `rust-lsp` (or a plugin fails to re-register after a reload). `is_registered("rust-analyzer.runSingle")` still returns true, so `execute_code_lens_command` keeps taking the early-return branch and `run_hook` goes to zero handlers. Clicking ▶︎ Run now does *nothing at all* — no terminal, no status message, no fallback to the server. Compare `lsp_uri_schemes`, the analogous registry, which at least resets with the `Editor`.
2. **Test-order coupling.** `test_rust_analyzer_runnable_is_translated_to_an_argv` loads the embedded `rust-lsp.ts`, which claims `rust-analyzer.runSingle` for the rest of the test binary's process. Any other test in that binary asserting a `rust-analyzer.runSingle` lens reaches the server via `workspace/executeCommand` would pass or fail depending on ordering, and `test_claimed_client_command_goes_to_the_plugin_not_the_server` / `..._still_goes_to_the_server` are already sharing the registry across harnesses.

A per-plugin claim map with removal on unload (and an editor-scoped registry, or at least a reset hook for tests) would close both.

---

crates/fresh-editor/src/app/lsp_requests.rs
```diff
@@ -3894,21 +3799,33 @@ impl Editor {
...
-                .unwrap_or_else(|| Err("no CodeLens-capable LSP server".to_string()))
+
+        // A `Command` the server did not advertise in `executeCommandProvider`
+        // is the client's to interpret, and LSP says nothing about what any
+        // such command means — its `arguments` are that server's own
+        // extension. So the core interprets none of them: a plugin claims the
+        // name (`registerLspClientCommands`) and receives it here. Sending a
+        // claimed name to the server instead would just be rejected, since the
+        // server never offered to run it.
+        if crate::services::lsp::client_commands::is_registered(&command.command) {
```
sinelaw, line 3810: **A claimed name with no live handler silently swallows the click.**

`dispatch_client_lsp_command` only returns `Err` when `serde_json::to_string` on the arguments fails (which never happens for a value that came out of `serde_json`), so the `Err` branch below is effectively dead. Everything else — zero registered handlers, a handler that threw, a plugin that was unloaded — is indistinguishable from success, and the function returns without a status message.

Scenario: the `rust-lsp` plugin errors during load after its `registerLspClientCommands` command was already queued and drained (or the user disables it later; see the `client_commands.rs` comment). The user clicks ▶︎ Run Test and gets absolutely nothing: no terminal, no error, and the core deliberately doesn't fall back to `workspace/executeCommand`.

`click_handlers.rs` already guards this kind of dispatch with `has_hook_handlers(...)`. Doing the same here — and falling back to the server path, or at least `t!("lsp.code_lens_failed")`, when nobody is listening — would make the failure visible.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
```diff
@@ -6288,6 +6293,37 @@ impl Editor {
...
+    /// Plugins load on a background thread, so a server autostarted while a
+    /// file opens really can win that race. In the common case nothing is
+    /// running yet and this is a no-op.
+    fn handle_register_lsp_client_commands(&mut self, commands: Vec<String>) {
+        if !crate::services::lsp::client_commands::register_all(commands) {
+            return; // nothing new — every claim was already known
+        }
+
+        let running = self
+            .lsp()
```
sinelaw: **Only the active window re-handshakes.**

`Editor::lsp()` is `Some(&self.active_window().lsp)` (editor_accessors.rs:1219), and each window owns its own `LspManager`. The doc comment above says the restart "is not optional bookkeeping", but it is skipped for every non-active window.

Scenario: the user has two windows open, each with a rust server running, and the plugin thread delivers `registerLspClientCommands` while window B is focused. Window A's rust-analyzer keeps the capabilities it handshook with, so files opened in window A never get runnable CodeLens entries — and no later event re-triggers the restart, because `register_all` returns `false` for an already-known claim. The user sees ▶︎ Run in one window and not the other, permanently.

Iterating `self.windows` (collecting `(window_id, language)` pairs first) would cover this.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
```diff
@@ -6288,6 +6293,37 @@ impl Editor {
...
+            .map(|lsp| lsp.running_servers())
```
sinelaw: **Universal-scope servers can never be restarted by this path.**

`running_servers()` maps each handle to `scope().label()`, and `LanguageScope::label()` returns `"universal"` for an empty scope (manager.rs:62-64). Feeding that back into `manual_restart("universal", ..)` is a no-op twice over:

- its shutdown loop explicitly skips universal handles (`if !self.handles[i].handle.scope().is_universal() && ...`, manager.rs:1659);
- `force_spawn("universal")` then short-circuits on `scope().accepts("universal")`, which is `true` for any universal scope, and returns the *existing* handle (manager.rs:1176-1187).

So the call reports `(true, "LSP server for universal started")`, `reopen_buffers_for_language("universal")` runs, `"universal"` is permanently added to `allowed_languages` — and the universal server keeps the `experimental.commands.commands: []` it handshook with. A user running a universal server alongside `rust-analyzer` gets no runnable lenses from it, with a success message in the status bar.

---

crates/fresh-editor/plugins/rust-lsp.ts
```diff
@@ -267,4 +267,146 @@ editor.on("action_popup_result", (data) => {
...
+    if (kind !== "cargo" && kind !== "shell") {
+      throw new Error(`unsupported runnable kind '${String(kind)}'`);
+    }
+    runnable = args[0] as Runnable;
+  } catch (e) {
+    editor.setStatus(`Rust LSP: could not read runnable for '${data.title}': ${String(e)}`);
+    return;
+  }
+
+  const task = runnableToTask(runnable);
```
sinelaw: **`runnableToTask` is the one unprotected call, and it only validates `kind`.**

The `try` above checks `args[0].kind` is `"cargo"` or `"shell"` and nothing else, then casts. `runnableToTask` immediately does `[...cargo, ...a.cargoArgs]`, `a.executableArgs.length`, `[a.program, ...a.args]` — each a `TypeError` if the field is missing — and it sits between the two `try` blocks, so nothing catches it.

Scenario: a server (or a future rust-analyzer) sends `{"label":..,"kind":"shell","args":{"program":"./x","cwd":"/w"}}` with no `args` array. The spread throws, the hook handler rejects, and the user gets no terminal and no message at all. The Rust version this replaces reported `"invalid rust-analyzer runnable: {error}"` via serde, because every field was required at deserialize time.

Simplest fix: move `const task = runnableToTask(runnable);` inside the first `try` (its `catch` already produces a good "could not read runnable for '<title>'" message), and/or check the array fields alongside `kind`.

---

crates/fresh-editor/tests/e2e/plugins/lsp_client_commands.rs
```diff
@@ -0,0 +1,418 @@
...
+    let script_path = create_runnable_lsp_script(dir.path(), &runnable);
+
+    let mut harness = EditorTestHarness::create(
+        80,
+        24,
+        HarnessOptions::new()
+            .with_project_root()
+            // The real embedded `rust-lsp.ts` is what claims the command and
+            // does the translation, so it has to actually load.
+            .with_forced_embedded_plugins()
```
sinelaw, line 388: **Nothing waits for the embedded plugin's claim before the lens is clicked.**

Unlike the two tests above — which use `load_plugin` and pump the loop until the plugin is in — this one relies on background embedded-plugin loading and goes straight to `open_with_lens` + `click_lens`. The only waits in between are for LSP readiness and for `"Run Lens"` on screen, neither of which observes `registerLspClientCommands` having been drained.

If the claim lands after the click, `is_registered("rust-analyzer.runSingle")` is `false`, the command goes to the fake server (which answers `result: null`), the recorder never runs, and `wait_until(|_| argv_file.exists() && ..)` hangs to its timeout. If it lands *between* server start and the click, `handle_register_lsp_client_commands` tears the fake server down and respawns it mid-test, which is its own source of flakiness.

Compare `plugins_dir_in_working_dir.rs`, which waits on an observable effect of the embedded plugin before proceeding. Something equivalent here (e.g. wait until a `rust-lsp`-contributed command/status is visible) would make the ordering deterministic.

---

sinelaw (commented):
Code review of the plugin-API extraction (4 commits on top of `claude/pr2614-base`). The routing fork itself reads correctly — `is_registered` gates the hook vs. `workspace/executeCommand` cleanly, the capability list is now a pure function of the registry, and moving the hit test onto `PaneView::rows` is the right fix now that `settle_views` takes `layout.view_line_mappings`.

8 findings, clustered in two places:

**`handle_register_lsp_client_commands`** — reusing `manual_restart` for the re-handshake pulls in the whole user-initiated-restart semantics: it makes every running language "manually allowed" (spawning `enabled: false` / `auto_start: false` servers), re-roots each of them from the *active buffer's* path regardless of language, silently no-ops on universal-scope servers, and only touches the active window.

**Dead-claim handling** — the registry is a write-only process-global with no removal, and `execute_code_lens_command` swallows a claimed command without checking anyone is listening, so an unloaded/failed plugin turns its lenses into no-ops with zero feedback.

Plus one unguarded throw in `rust-lsp.ts` and one race in the new e2e test.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
sinelaw: Confirmed and fixed in 44b5345. `manual_restart` is a user action and I should not have reused it — `allowed_languages.insert` makes `force_spawn` take the "user explicitly started this" branch (manager.rs:1288), which spawns every configured server for the language, `enabled: false` ones included.

Replaced with `LspManager::respawn_for_client_capability_change`, which only puts back what was already running: it touches neither `allowed_languages` nor `disabled_languages` nor the cooldown, returns early when no server for the language is up, and logs an error if it cannot restore a server it just stopped.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
sinelaw: Confirmed and fixed in 44b5345. The caller now resolves each root from a buffer of *that* language rather than from whatever is focused:

```rust
let file_path = self.windows.get(&window_id).and_then(|w| {
    w.buffers.iter().find_map(|(buffer_id, state)| {
        (state.language == language)
            .then(|| w.buffer_metadata.get(buffer_id))
            .flatten()
            .and_then(|meta| meta.file_path().cloned())
    })
});
```

So the TS server in your example is re-rooted from a TS file, not from the focused Rust one.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
sinelaw: Confirmed and fixed in 44b5345. The point about it being unrecoverable is the sharp part — `register_all` returns `false` on any later call, so a window skipped once stays stale for the session.

The caller now collects window ids up front and walks every window's manager, resolving the root per (window, language).

---

crates/fresh-editor/src/app/plugin_dispatch.rs
sinelaw: Confirmed and fixed in 44b5345 — passing the literal `"universal"` as a language was plainly wrong, and the "succeeds while doing nothing" part made it invisible.

Two changes: `languages_with_running_servers()` replaces `running_servers()` here and only ever yields real languages (it flattens `scope().languages()` and skips universal scopes), and `has_universal_server()` lets the caller log a warning naming the case it cannot reach, instead of reporting success for it.

A universal server still keeps the capabilities it started with — that is inherent, since it cannot be addressed by language — but it is now stated rather than silently assumed handled.

---

crates/fresh-editor/src/app/lsp_requests.rs
sinelaw: Confirmed and fixed in 44b5345. `execute_code_lens_command` now checks for a live handler before claiming the command, and reports `lsp.code_lens_failed` when there is none.

One deviation from the suggestion: it uses `has_subscribers`, not `has_hook_handlers`. Both answer from the same `event_handlers` map, but `has_hook_handlers` blocks on a round-trip to the plugin thread (`PluginRequest::HasHookHandlers`), which may itself be mid-hook — and this runs on every lens click.

I did not add the fallback to `workspace/executeCommand`. A claimed name is by definition one the server did not advertise, so sending it produces a server-side error rather than a useful outcome; reporting locally says more.

---

crates/fresh-editor/plugins/rust-lsp.ts
sinelaw: Confirmed and fixed in 44b5345, taking the suggested fix: `runnableToTask` is inside a `try` with its own `catch`, so a payload with a valid `kind` but missing `cargoArgs`/`args` reports "could not read runnable for '&lt;title&gt;'" instead of rejecting the handler silently.

I kept the validation at `kind` rather than checking each array field — the flattening is the thing that actually needs them, so letting it throw and catching there keeps one place responsible for the shape.

---

crates/fresh-editor/src/services/lsp/client_commands.rs
sinelaw: Partly addressed; the rest I'm deliberately leaving, so splitting the two halves.

**(1) Dead lens after unload — symptom fixed, cause not.** 44b5345 makes `execute_code_lens_command` check for a live handler before taking the claimed branch and report `lsp.code_lens_failed` when there is none, so the click is no longer "absolutely nothing". The claim itself still is not withdrawn.

Withdrawing it needs more than a `remove`: `RegisterLspClientCommands` carries no plugin identity (commands that need one take an explicit `plugin_name` field, as `SetGlobalState` does), so it needs that plumbed through, an unload hook to act on, and then a *second* re-handshake pass to drop the capability from servers that already advertised it — otherwise rust-analyzer keeps sending lenses nothing will run. That is a bigger change than this PR should carry, and it is now noisy rather than silent, so I would rather do it separately.

Worth noting `lsp_uri_schemes`, the registry this is modelled on, is also insert-only — but you are right that mine is worse, not equivalent: it is a process-global `static`, so it outlives the `Editor` too. That part is forced. The other reader is `create_client_capabilities`, which runs on the LSP task with no `Editor` in reach, which is why the registry cannot be editor-scoped.

**(2) Test-order coupling — fixed, and it had already bitten.** The argv test's wait for the claim (added in 44b5345) read this same global, so a claim left by an earlier test in the binary satisfied it while that harness's `rust-lsp.ts` had not run — reintroducing the race it was meant to close, now landing in the new no-handler path. 88e0b60 waits on per-harness rendered output instead: the palette showing a command only `rust-lsp.ts` contributes, following `plugins_dir_in_working_dir.rs`. No test depends on the global registry now.

---

crates/fresh-editor/tests/e2e/plugins/lsp_client_commands.rs
sinelaw: Confirmed and fixed in 0585c38 — and the first attempt at fixing it was wrong in a way worth recording.

That attempt waited for `is_registered("rust-analyzer.runSingle")`. But that registry is process-global (the other finding), so a claim left by an earlier test in the same binary satisfies the wait immediately while *this* harness's `rust-lsp.ts` has not run and has no `lsp_execute_command` handler — the same race, now landing in the new "claimed but nobody is listening" path.

So it waits on per-harness rendered output instead, as suggested: the palette showing a command only `rust-lsp.ts` contributes, which it registers before the claim.

Two things that cost a hang before it worked, both because `wait_until` has no timeout — an unmeetable condition hangs the suite rather than failing it:

- the palette ellipsises its columns, so a 58-character *description* never renders in full at 80 columns. The harness is 120 wide now and matches a command name;
- the matched text has to be disjoint from the typed filter, or the prompt line satisfies the wait before any result renders. Filter `"Rust LSP"`, match `"Configure Mode"`.

The test now passes in ~1s where it previously ran until the runner killed it.

---

sinelaw:
`test macos-latest` is red on 0585c38, and it is not this PR's failure — but it is a real bug, not a flake, so here is the root cause rather than a re-run.

**Failing:** `fresh-editor-core keys::tests::a_sequence_round_trips_as_its_presses`. `test ubuntu-latest` passed the same test on the same commit, and macOS was green on the previous head; this PR touches no file under `crates/fresh-editor-core/src/keys/` (`git diff 03ba5f5..HEAD -- crates/fresh-editor-core/src/keys/` is empty).

It is a `proptest`, so at the default 256 cases it only fails when the generator happens to produce the offending input. At `PROPTEST_CASES=20000` it reproduces reliably, and shrinks to two counterexamples:

```
keys = [Key { code: Char('𝘈'), mods: KeyModifiers(CONTROL) }]   # U+1D608
key  =  Key { code: Char('🄰'), mods: KeyModifiers(CONTROL) }    # U+1F130
```

`parsing_then_printing_preserves_meaning` fails on the same input.

**Cause.** The printer lowercases a char "because uppercase would parse back as Shift+it" (keys/mod.rs:379), and the parser infers SHIFT from `c.is_uppercase()` (keys/mod.rs:356). Those two are not inverses. `𝘈` and `🄰` are uppercase *by Unicode property* but lower to themselves, so the printer writes `C-𝘈` unchanged and the parser reads it back as CONTROL|**SHIFT** — a modifier nothing wrote. The generator's filter (`c.to_lowercase().eq(once(c))`) selects for exactly that class, which is why the property test finds it and everyday use does not.

**Proposed fix** — infer SHIFT only when the char has a *distinct* lowercase:

```rust
let lowered = c.to_lowercase().next().unwrap_or(c);
if c.is_uppercase() && lowered != c {
    mods |= KeyModifiers::SHIFT;
}
KeyCode::Char(lowered)
```

Verified locally: 28/28 `keys::` tests pass at `PROPTEST_CASES=40000` (156× the default), including both failures. `A` → `Shift+a` still holds.

I have **not** included it here — it is unrelated to code lens and would widen this PR. Happy to send it as its own PR; it is a six-line change and worth pinning the two seeds in `proptest-regressions/` alongside it so the case stops being a lottery.

---

sinelaw:
## Performance review of the whole feature — open items for a maintainer

Reviewed the full feature diff (`03ba5f5~4..HEAD`, i.e. the four original commits plus the work on top), looking for new per-frame, per-keystroke and per-cursor-move work. Nothing here blocks correctness; the tests pass and it behaves correctly by hand. Marking the PR draft so these get a decision before it merges.

Roughly in order of how much I'd care.

### Per frame

- `ViewLineMapping` gained `virtual_text_namespace: Option<String>` (`app/types/layout.rs:37`, `view/ui/view_pipeline.rs:76`), so every rendered row carries a heap string slot — the struct is built fresh for every visual row, every frame.
- Filling it allocates: `ns.as_str().to_string()` per virtual text (`transforms.rs:598` and `:630`), then `namespace.clone()` per wrapped row of it, then another `.clone()` when the mapping is built — about 2–3 `String` allocations per virtual-line row per frame.
- All of that exists to answer one yes/no question on click (`is this row a code lens?`, compared as `!= Some("lsp-code-lens")`); an interned id, an enum, or a `&'static str` would carry the same information with no allocation.
- The render-path virtual-text query scans *every* virtual text in the buffer and filters afterwards (`virtual_text.rs:914`, called from `transforms.rs:519`), so it is O(all virtual texts) per frame, not O(visible) — pre-existing, but code lens makes it bigger by adding an entry per function/impl/mod.

### Per edit / per keystroke

- Code lens has no debounce at all, while inlay hints wait 500 ms (`window/mod.rs:4333`) — worth asking why the two differ.
- A server refresh re-requests lenses for **every open buffer of that language** (`async_messages.rs:1995`), not just the visible one, so 20 open Rust files means 20 requests per refresh.
- Every response clears the whole `lsp-code-lens` namespace and rebuilds all entries from scratch (`lsp_requests.rs:1484`, `:1646`) rather than diffing, so a file with many lenses churns all of its virtual texts and markers on each refresh.
- Each lens entry owns a marker in the interval tree, so lens count adds to per-edit marker adjustment — the tree keeps this cheap per marker, but it scales with file size.
- The rebuild allocates a fresh `String` per lens title plus an overlay vec each time, so garbage is proportional to lens count per refresh.
- Responses are version-guarded (good — no stale application), but the flip side is that while you type continuously every response is discarded, so the request and round-trip are pure waste until you pause.

### Smaller

- The code-actions popup is rebuilt once per server response, and each rebuild re-clones every lens command for the line — rust-analyzer's `Runnable` payload is a sizeable JSON blob, so that is a real copy.
- `code_lens_commands_for_cursor` scans the buffer's whole lens list per popup build instead of indexing by line — linear, fine at today's sizes.

### Checked and clear

- The click hit-test resolves against the pane's rows as a slice, no per-click clone of the row list.
- The `code_lens` menu-context value is a config bool read once per frame.
- The client-command registry lock is taken only when a lens is actually run, never per frame.

### If only one thing gets done

The debounce. It is the pattern inlay hints already use, and it collapses the request storm, the wasted round-trips and most of the rebuild churn at once. The per-row `Option<String>` is the other one worth doing, and it is a contained change — the namespace only ever needs to be compared, never owned.

---

Issue #2100, opened 2026-05-25
Laggy on a serial terminal


jetpax:
Wow, the console editor I have been looking for for SO long, thank you!

While Fresh works perfectly on a serial terminal connected to my Raspberry Pi 2W @115200, it is very laggy

mcedit, nano, micro for example, on the same platform,  are not.

Do I need to set something in settings?

sinelaw:
Do you know if it's a CPU issue (is CPU near 100% when fresh is processing your input keys) or just a serial port communication speed issue?

sinelaw:
Thanks for the report! Fresh is sending several full repaints per keystroke where most changed nothing. The extra repaints come from background plugin activity (e.g. the live-diff debounce timer) waking the screen even when nothing visible changed.

Fix in #2104 stops those no-op repaints — about 20–30% fewer bytes per keystroke, and cursor moves no longer emit redundant frames. That should make typing noticeably snappier over serial. More serial-specific tuning may follow.

---
_Generated by [Claude Code](https://claude.ai/code)_

sinelaw:
Turning off current-line highlighting and animations should also improve it (they are configurable - use the Open Settings command to bring up the settings UI)

sinelaw:
Merged some improvements that should make this better

sinelaw:
Triage: labeling as `deferred` for now. While partial improvements have already been merged (#2104), the core symptom here — perceived input latency over a physical serial link (Raspberry Pi 2W @ 115200 baud) — is bandwidth/hardware-dependent and not something we can reliably reproduce in our automated e2e/tmux test harness. Further serial-specific tuning is likely to be iterative rather than a single isolated fix. Keeping it open for ongoing work.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3278, opened 2026-09-14
Add a symbol breadcrumb trail under the tab bar

### LINKED ISSUE #2913: breadcrumb trail/navigation for buffer

sinelaw:
Fixes #2913. Supersedes #2917 — same feature, by @asukaminato0721, whose commit is the first one here. That branch is on a fork I cannot push to, so the follow-up work landed here instead. Close whichever you prefer.

## What it does

A row under the tab bar shows the symbol path at the cursor, from the LSP's document symbols, hanging off a root:

```
> InventoryStore > restock
```

It follows the cursor as you move, and clicking a crumb jumps to that declaration. The crumb under the pointer lights up. When the row is too narrow the outer crumbs drop off behind a `… > `, keeping the innermost — the function you are in matters more than the module it lives in.

The row stays put whatever the caret is inside. A caret between two functions has no trail to name, and the row shows its root alone — `>` — rather than disappearing and shifting the buffer a line every time the caret crosses a gap.

Off with `editor.show_breadcrumbs` (default on). A buffer whose symbols were never fetched — no server, or a server that reports none — has no row at all, so a file without an LSP looks exactly as it did.

## How it is built

The plugin (`lsp_navigation.ts`) asks the server for document symbols, works out which ones enclose the cursor, and hands the trail to the host with `setBreadcrumbs`. The host describes each crumb as its own node in the shell tree, so a click is answered by the crumb under the pointer rather than by hit-testing against remembered rectangles.

A crumb carries the **LSP position** its server reported — a line and a UTF-16 character — not a byte offset. Resolving a byte is work only the editor can do cheaply, since it has the line index; asking the plugin for one meant reading the source line of every symbol in the document, on activate and after every edit. The click resolves the position instead, and refines it onto the symbol's name for servers that report flat `SymbolInformation`.

Trails are held **per pane**, tagged with the buffer they describe. Two panes can show one buffer and each has its own caret, while a trail names the scopes around one caret.

Whether a pane draws a row is a separate question from what its trail says, and a **trail source** answers it: once the plugin has published for a buffer the row stays, empty or not. That needs two statements rather than one, so `setBreadcrumbs(id, [])` means "the caret is between symbols, keep the row" and the new `clearBreadcrumbs(id)` means "nothing to say here, take it away". Gating on a running server instead looked equivalent and is not: it put a permanent row on buffers whose server never reports a symbol.

## Three things worth a reviewer's eye

**A new plugin event.** `handle_lsp_initialized` already re-requests semantic tokens, folding ranges and inlay hints for buffers that opened before the handshake finished. Plugins were the one waiter with no signal, so this adds `lsp_ready { language, server_name }` beside them. Breadcrumbs use it to fill in trails whose first fetch failed because the server was not up. A diagnostics push stays as a fallback for the other failure — server up, request still did not return — which no event covers.

**Two new plugin calls, `getLineStartForPosition` / `getLineEndForPosition`.** `getLineStartPosition`/`getLineEndPosition` answer from the piece tree's line index, so they answer nothing at all on a buffer that has none — every large file in byte-offset mode. The new pair takes the byte a caller already holds and scans out from it, bounded, needing no index; vi-mode's `$`, `D` and `^` moved onto them, having done nothing on such a file. The handler checks the range it scanned was actually loaded rather than trust `slice_bytes`, which returns short data for an unloaded region instead of failing — no answer is recoverable, a confident wrong byte is not. The line-number pair is unchanged and still the right call when a line number is genuinely what you have.

**The kind filter is per language, and the plugin's.** A trail leaves out what is not a scope — variables, fields, constants, literals — because servers that report locals were otherwise turning a comprehension into `> InventoryStore > total_value > s > n`. It names what it drops rather than what it keeps, since the kind a server picks for a scope varies: rust-analyzer reports an `impl` block as `Object`, so an allow-list of the obvious scopes silently cost every Rust method the type it belongs to. And a kind only means what a language's server uses it for — `Module` carries a `namespace` in TypeScript and an `import` in Python — so the drop set is per language. Which kinds those are is a fact about language servers, so it lives with the plugin rather than in the editor's config:

```json
{ "plugins": { "lsp_navigation": { "settings": {
    "breadcrumbAllSymbols": false,
    "breadcrumbNonScopeKinds": { "python": [2, 13] }
} } } }
```

Also here: four `ui.breadcrumb_*` theme keys, named by each builtin theme — the strip's ground, line-number grey for the ancestors, and the inactive-tab foreground for the symbol the cursor is in. They started as one hard-coded set of dark greys, which drew a near-black band across `light` and defeated `high-contrast`.

## Known limits

- **Mouse only.** There is no keyboard route to a crumb, deliberately — no pane chrome is in the focus ring today, tabs included, so making breadcrumbs the exception would be a focus-model decision rather than a breadcrumb one. `Ctrl+P → Go to LSP Symbol` covers keyboard navigation and already preselects the enclosing symbol.
- **No toggle command.** `editor.show_breadcrumbs` is config-only and global; there is no palette entry or keybinding the way `show_status_bar` has one.
- **A pane earns its trail by having focus**, since that is the caret the plugin read. A pane that has never held focus draws its root but no crumbs until it does.

## Tests

End-to-end against a fake LSP server: the trail renders and updates, a click lands on the right symbol, a click refreshes the trail for where it landed, recovery when the first symbol request fails, a local staying out of the trail while a scope reported under an unfamiliar kind stays in, two panes on one buffer not sharing a trail, and the same reported symbols reading differently as TypeScript and as Python. Unit tests on the shell tree: each crumb placed where the click resolver expects it, a crumb elided off the left having no node to click at all, hover lighting only the crumb under the pointer — painted through the fold, not inspected — and an empty trail still drawing a row. Unit tests on the position resolver: a selection range taken as given, a flat declaration start refined onto its name, an absent label left alone, and UTF-16 columns resolving to the right byte.

The per-pane test was checked against the old buffer-keyed lookup before landing: it fails there with both panes drawing one trail, which is the bug it exists for.

A row that appears where it should not moves every line below it, so it breaks tests that have nothing to do with breadcrumbs — hover popups, fold gutters, a diagnostic marker, a whole-UI snapshot. Two rounds of CI failures here were exactly that, and they are the reason the gate is a trail source rather than a running server, and the reason a buffer with no symbols is cleared rather than published empty. One of those tests counted screen rows from the top of the pane to find its marker; it now finds the line, because what sits above the buffer is chrome that changes.

## Driven by hand

Against six real servers — **rust-analyzer**, **gopls**, **tsserver**, **pylsp**, **pyright** and **bash-language-server** — each one's `documentSymbol` output probed directly for the kinds it reports, then the trail checked live in the editor:

| Server | Trail | What the filter dropped |
| --- | --- | --- |
| pylsp | `> Warehouse > restock` | `total`, `names`, `s`, `n` (`Variable`), `LIMIT` (`Field`) |
| pyright | `> Warehouse > restock` | `LIMIT` (`Constant`), parameters (`Variable`) |
| gopls | `> (*Warehouse).Restock` | `Name` (`Field`) |
| tsserver | `> Inventory > Warehouse > restock` | `settings` (`Constant`) and its `Property` children |
| bash-ls | `> deploy_service` | `REGION`, `name`, `count` (`Variable`) |
| rust-analyzer | `> outer > inner > impl Warehouse > restock` | `name` (`Field`) |

Only rust-analyzer reports a scope as `Object`; TypeScript reports object literals as `Constant`/`Property`, so naming the filter by what it drops adds the `impl` block and no TypeScript noise. Also walked a caret through a file with real gaps at module level, checking the buffer's first line never moved: `> Warehouse > restock` inside a method, `>` on the blank line after it, `> helper` in the next function, `>` again. And by hand: a crumb click lands on the identifier rather than the range start, the trail refreshes after that click, the row elides to `> … > restock` at a narrow pane, hover lights one crumb, a file with no server has no row, and the row's colours follow the theme across `dark`, `light`, `dracula`, `nord`, `solarized-dark`, `high-contrast` and `nostalgia`.

https://claude.ai/code/session_01SRGrvpxQgYZPg9ZqPkWxDT


sinelaw:
`test ubuntu-latest` failed on one timeout: `e2e::vi_mode_bugs::test_vi_visual_line_indent_stops_at_selection`, 180s, with the other 9833 passing. It is not this PR's — the same test on the same commit passed on macOS and Windows, and it passes locally on this branch in 0.39s, so a 180s timeout is a hang under load rather than a logic break. This PR touches only the breadcrumb trail and LSP navigation, which share no code path with vi mode. I have re-run the job.

Worth noting separately: `wait_until` and `wait_for_buffer_content` have no deadline, so a test whose condition never becomes true hangs until nextest's 180s slow-timeout instead of failing with an assertion. That is why this surfaces as a timeout with no diagnostic, and it will keep making load-sensitive flakes expensive to diagnose.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
```diff
@@ -2534,9 +2536,18 @@ impl Editor {
             .expect("active window present")
             .get_mut(&actual_buffer_id)
             .and_then(|state| {
-                let len = state.buffer.len();
-                let content = state.get_text_range(0, len);
-                buffer_line_byte_offset(&content, len, line as usize, want_end)
+                let line = line as usize;
+                let start = state.buffer.line_start_offset(line)?;
```
sinelaw, line 2540: **`getLineStartPosition`/`getLineEndPosition` now return `null` on any buffer without a line index.**

`Buffer::line_start_offset` → `PieceTree::line_range` starts with `let line_count = self.line_count()?;`, and `line_count()` is `None` whenever a leaf's `line_feed_cnt` is unknown — i.e. a large file opened in byte-offset mode before a line scan has run (`active_buffer_has_line_index`, `viewport.rs:400`, `render_buffer.rs:239` all branch on exactly this). The deleted `buffer_line_byte_offset` scanned the materialized text, so it answered correctly in that state; the new code returns `None` for every line, in range or not.

Concrete failures:
- `plugins/lsp_navigation.ts` `attachLineText`: `start`/`end` come back `null`, so the line is skipped, `sym.lineStartByte` stays `-1`, and `navigateToSymbol` early-returns — picking a symbol in `Go to LSP Symbol` silently does nothing on a large file (it used to jump).
- `plugins/vi_mode.ts` uses these for `dd`, `G`, visual-block columns, `>`/`<` ranges (lines 632, 1079, 1159, 1749, 1792, 2283-2284, 2522, 3067); all of those become no-ops on a large file.

The perf win is worth keeping, but it needs a fallback (or at least `has_line_feed_scan`-style handling) for the no-line-index case rather than silently reporting "out of range".

---

crates/fresh-editor/plugins/lsp_navigation.ts
```diff
@@ -402,14 +438,175 @@ function parseSymbols(result: unknown): SymbolItem[] {
...
+  }
+}
+
+// The server was not up when we asked; now it is.
+editor.on("lsp_ready", (data) => retryMissingTrails(data.language));
+
+// And the other way a fetch fails: the server was up but the request did not
+// come back. There is no event for that, so take the next sign of life —
+// diagnostics — as the cue to try again.
+editor.on("diagnostics_updated", () => retryMissingTrails());
```
sinelaw, line 603: **The retry is not bounded, and it spawns language servers as a side effect.**

The doc comment above `retryMissingTrails` says a buffer whose symbols loaded is never asked again — true — but a buffer whose symbols *can never* load is asked forever, because `refreshBreadcrumbs` deliberately leaves no cache entry on a rejected fetch.

`BufferInfo.language` is a non-empty `string` for every buffer (`"text"`, `"markdown"`, `"json"`, …), so the `if (!language || !filePath)` guard does not catch buffers that simply have no server. Those reach `fetchSymbols`, `handle_send_lsp_request` rejects the callback ("LSP server for 'text' is not running" / "is unavailable"), and the buffer stays permanently "missing". Every `diagnostics_updated` — which rust-analyzer emits continuously — then re-schedules a fetch for *every* such buffer, at ~1 request per 250 ms per buffer for as long as diagnostics keep arriving.

Worse, `handle_send_lsp_request` calls `lsp.try_spawn(&language, None)` before it fails. `retryMissingTrails()` walks `listBuffers()`, not just the active one, so a diagnostics push on a Rust file will start `pylsp`/`tsserver` for a Python or TS buffer sitting in a background tab that the user never opened.

Suggest recording a negative result (e.g. distinguish "server said no" from "server not up yet", or cache `[]` after N failures / when the rejection is not a transient one), and scoping the diagnostics retry to the buffer whose URI the diagnostics are for.

---

crates/fresh-editor/plugins/lsp_navigation.ts
```diff
@@ -402,14 +438,175 @@ function parseSymbols(result: unknown): SymbolItem[] {
...
+editor.on("buffer_activated", (data) => {
+  const cached = breadcrumbSymbols.get(data.buffer_id);
+  if (cached) publishBreadcrumbs(data.buffer_id, editor.getCursorLine());
+  else editor.setBreadcrumbs(data.buffer_id, []);
+  scheduleBreadcrumbRefresh(data.buffer_id);
+});
+
+editor.on("cursor_moved", (data) => {
+  // `line` is 1-indexed here; LSP symbol ranges are 0-indexed.
+  if (data.cursor_id === 0) publishBreadcrumbs(data.buffer_id, data.line - 1);
```
sinelaw, line 571: **`cursor_id === 0` is not "the primary cursor" — the trail freezes in multi-cursor mode.**

`Cursors::add` sets `self.primary_id = id` for each newly added cursor (`cursor.rs:260-263`), and `insert_with_id` does the same on undo/redo. So as soon as the user adds a second cursor, the primary's `CursorId` is ≥ 1 and this guard never matches: the breadcrumb row keeps showing whatever scope the caret was in *before* the extra cursor existed, and stays wrong until the cursors collapse back (`remove_secondary` restores the min id).

The same filter also defeats the new `Editor::notify_cursor_moved`, which emits with `active_cursors().primary_id()` — so a breadcrumb click while multiple cursors are live jumps the caret but does not refresh the trail, which is exactly the case commit "refresh the trail after a crumb click" was added for.

Filtering on "is this the primary cursor" needs a signal the hook actually carries; comparing against `0` only happens to work in the single-cursor case.

---

crates/fresh-editor/src/view/ui/mod.rs
```diff
@@ -24,6 +24,7 @@ pub use crate::primitives::text_edit;
 
 // Runtime-only modules (depend on state, services, input, etc.)
 #[cfg(feature = "runtime")]
+#[cfg(feature = "runtime")]
```
sinelaw: Stray duplicate attribute — `#[cfg(feature = "runtime")]` is now applied twice to `pub mod expanded_menus_cache;`. Harmless to the compiler, but it is unrelated to breadcrumbs and looks like an accidental edit; worth dropping from the diff.

```suggestion
#[cfg(feature = "runtime")]
```

---

sinelaw (commented):
Read the whole diff. The shell-tree half (crumb nodes carrying their own LSP position, elision from the left, the `PaneChrome`/`reference_split_layout` parity sweep) looks solid, and `cargo check -p fresh-editor --all-features` is clean here. Four findings, one of which is a regression outside the breadcrumb feature itself.

1. **`handle_get_line_position` (plugin_dispatch.rs)** — the piece-tree rewrite returns `null` for every line on a buffer with no line index (large files in byte-offset mode), where the deleted scan answered correctly. Breaks `attachLineText` in this same PR, and a dozen `vi_mode.ts` motions.
2. **`retryMissingTrails` (lsp_navigation.ts)** — unbounded retry for buffers whose language has no server, and it spawns servers for background buffers via `try_spawn` on every `diagnostics_updated`.
3. **`cursor_id === 0` (lsp_navigation.ts)** — not equivalent to "primary cursor"; the trail stops following the caret in multi-cursor mode, including after a crumb click.
4. **`view/ui/mod.rs`** — duplicated `#[cfg(feature = "runtime")]`, an unrelated stray edit.

---

crates/fresh-editor/src/view/ui/mod.rs
sinelaw: Right, and accidental — removing `pub mod breadcrumbs;` took the module but not its attribute. Fixed in c8b3946.

---

crates/fresh-editor/plugins/lsp_navigation.ts
sinelaw: Right on both counts, including the second one — the click path was the case the guard defeated. Fixed in cb4a056: `cursor_moved` now carries `is_primary`, filled in `apply_event_with_hooks` where the cursor set is in scope, and the plugin reads that instead of comparing against `0`.

Added a unit test that builds a two-cursor set and asserts both directions: the cursor added last reports primary, and cursor 0 does not once another exists. That is the case that was silently wrong.

---

crates/fresh-editor/plugins/lsp_navigation.ts
sinelaw: Both halves confirmed, and the second was made worse by my own change — the handler this replaced only retried the active buffer; walking `listBuffers()` turned a bounded annoyance into an unbounded one. Fixed in cb4a056 along the two lines you suggested.

Retries now carry a per-buffer budget (`MAX_TRAIL_RETRIES`), which is what separates "the server was not up yet" from "this language has no server" — the two are indistinguishable from the plugin, since both arrive as a rejected promise and neither leaves a cache entry. A trail that can be fetched needs a retry or two; one that never can stops asking. Edits, reverts and tab switches are not rationed, only the event-driven retries.

The diagnostics retry is scoped to the buffer whose path matches `data.uri`, so a push from one language's server no longer reaches another language's background tabs — which is what was calling `try_spawn` for servers the user never asked for.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
sinelaw: The mechanism is right — `line_range` opens with `line_count()?`, which is `None` while any leaf's `line_feed_cnt` is — but I dug into the two consequences and one of them does not hold.

**`attachLineText` / Go to LSP Symbol is not affected.** `send_lsp_did_open_for_buffer` does `let Some(text) = state.buffer.to_string() else { return }`, and `to_string` is documented "Returns None if any buffers are unloaded (lazy loading)". So a lazily-loaded large file never gets a `didOpen`, the server never sees the document, and there are no symbols to pick. It did not jump before and does not now.

**And the deleted scan was not a working fallback.** It called `state.get_text_range(0, len)` → `get_text_range_mut`, which "handles lazy loading automatically" — it force-loads every unloaded chunk. It answered correctly by materialising the entire file, on *every* call. That is the thing byte-offset mode exists to avoid, so restoring a fallback would put it back; I do not think that is the fix.

What is left is real but narrower, and it is the contract rather than the behaviour: `getLineStartPosition` documented `null` as "line number out of range", and it quietly came to mean "or this buffer has no line index". A caller reading the old doc has no way to know to ask for `Action::ScanLineIndex`. Both API docs now say so, and `handle_get_line_position` records why `None` there is deliberate (cb4a056).

That leaves the vi-mode call sites you listed genuinely degraded on an unscanned large file. I have not touched them — it is a dozen call sites in a feature this PR does not otherwise go near, and whether they should trigger the scan, degrade quietly, or tell the user is your call rather than mine. Happy to take it on as a follow-up if you want it.

---

crates/fresh-editor/src/app/chrome/splits.rs
```diff
@@ -25,6 +25,38 @@ use fresh_i18n::t;
...
+        };
+        let Some(position) = self.active_window().buffers.get(&buffer_id).map(|state| {
+            breadcrumb_target(&state.buffer, line as usize, character as usize, label)
+        }) else {
+            return;
+        };
+        self.focus_split(pane, buffer_id);
+        let old_position = self.active_cursors().primary().position;
+        self.active_window_mut()
+            .set_buffer_cursor_in_splits(buffer_id, position, &[pane]);
```
sinelaw, line 65: **A breadcrumb click leaves every secondary cursor behind, re-opening #3125 on a new path.**

`set_buffer_cursor_in_splits` only touches `view_state.cursors.primary_mut()` (`app/window/mod.rs:1945`), so the other cursors stay where they were. Every other caret-placing click goes through `handle_editor_click`, which deliberately drops the whole multi-cursor set first — see the `kept_id` block and its "A plain click ends multi-cursor editing … the next keystroke edited every one of them (#3125)" comment at `app/click_handlers.rs:403-425`. Concretely: `Ctrl+Alt+Down` twice to get three cursors, click a crumb, type one character — it is inserted at the class name *and* at the two off-screen cursors.

Fix: collapse to a single cursor (the lowest id, as the click path does) before placing it, rather than calling `set_buffer_cursor_in_splits`.

---

crates/fresh-editor/src/app/window/mod.rs
```diff
@@ -2855,6 +2866,27 @@ impl Window {
...
+    /// height and never seen.
+    pub(crate) fn pane_breadcrumbs(
+        &self,
+        chrome: &HashMap<LeafId, crate::view::shell::splits::PaneChrome>,
+    ) -> HashMap<LeafId, Vec<fresh_core::api::BreadcrumbItem>> {
+        self.panes_with_buffers()
+            .into_iter()
+            .filter(|(leaf, _)| chrome.get(leaf).is_some_and(|c| c.breadcrumbs))
+            .filter_map(|(leaf, buffer)| {
+                let items = self.breadcrumbs.get(&buffer)?;
```
sinelaw: **Two panes showing the same buffer share one trail, so one of them is always describing the other pane's caret.**

The trail is stored per `BufferId` (`Window::breadcrumbs`) but carets are per pane: `split_pane_impl` gives the new split its own `SplitViewState::with_buffer(…, current_buffer_id)` with its own `cursors` (`app/split_actions.rs:110-114`), and `set_buffer_cursor_in_splits` writes each split's primary independently. So after `Ctrl+\` on `foo.rs` and moving the right pane's caret into `fn bar`, *both* rows read `… > bar` while the left caret is still in `fn foo`. Clicking the left pane's `bar` crumb then jumps the left caret to a symbol it never pointed at.

The plugin side has the same shape — `breadcrumbSymbols`/`publishBreadcrumbs` are keyed by buffer id and only ever publish for `getActiveBufferId()` — so fixing this needs a pane (or cursor) dimension on both sides, not just in `pane_breadcrumbs`.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
```diff
@@ -2308,6 +2283,32 @@ impl Editor {
...
+        let id = fresh_core::BufferId(buffer_id as usize);
+        for window in self.windows.values_mut() {
+            if window.buffers.contains_key(&id) {
+                if items.is_empty() {
+                    if window.breadcrumb_owners.get(&id) == Some(&plugin_name) {
+                        window.breadcrumbs.remove(&id);
+                        window.breadcrumb_owners.remove(&id);
+                    }
+                } else {
+                    if window.breadcrumbs.get(&id) != Some(&items) {
```
sinelaw: **Every caret move now forces an extra full repaint, in every buffer, even when the trail is byte-identical or empty.**

This equality check correctly skips the state write, but `SetBreadcrumbs` falls into the `_ => true` arm of `has_visual_commands` in `async_messages.rs:1671`, and `process_plugin_commands`'s return value is OR'd straight into `needs_render` (`async_dispatch.rs:659-661`). The plugin publishes unconditionally on every `cursor_moved` (`lsp_navigation.ts:587` → `editor.setBreadcrumbs` at :535), including for plain-text/terminal buffers that have no symbols at all, so holding an arrow key doubles the frame rate with nothing changing on screen. That is exactly the case the `SetStatusBarValue` arm above it was added for (#2100: "invisible on a fast terminal, but real lag over serial").

Fix: give `SetBreadcrumbs` its own arm that compares against the stored trail, the way `SetStatusBarValue` does.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
```diff
@@ -2534,9 +2542,18 @@ impl Editor {
             .expect("active window present")
             .get_mut(&actual_buffer_id)
             .and_then(|state| {
-                let len = state.buffer.len();
-                let content = state.get_text_range(0, len);
-                buffer_line_byte_offset(&content, len, line as usize, want_end)
+                let line = line as usize;
+                let start = state.buffer.line_start_offset(line)?;
```
sinelaw, line 2613: **`getLineStartPosition`/`getLineEndPosition` now return null for *every* line on any file ≥10 MB, silently breaking unchanged callers.**

`line_start_offset` → `piece_tree.line_range` bails on `line_count()?`, which is `None` whenever any leaf has an unknown line-feed count (`piece_tree.rs:1594`) — i.e. for every lazily loaded file, and the line scan that fixes that is only ever started by the user (`Action::ScanLineIndex` / Go to Line; nothing auto-starts `start_incremental_line_scan`). The deleted `buffer_line_byte_offset` answered these correctly, and even short-circuited `line == 0` to `Some(0)`. Concretely: open a 50 MB log with vi_mode on and press `$` or `D` — `lineEndTarget` (`plugins/vi_mode.ts:632`) gets `null` and the motion silently does nothing; `diff_nav.ts:50` and `audit_mode.ts:8585` no-op the same way.

Materializing the file per call was clearly the wrong trade, but "no answer" is not equivalent — consider falling back to the incremental scan, or at least distinguishing "out of range" from "no index" so callers can react.

---

crates/fresh-editor/plugins/lsp_navigation.ts
```diff
@@ -402,14 +438,209 @@ function parseSymbols(result: unknown): SymbolItem[] {
...
+  20, // key
+  21, // null
+  22, // enum member
+  24, // event
+  25, // operator
+  26, // type parameter
+]);
+
+function showsEveryKind(): boolean {
+  const cfg = editor.getConfig() as
```
sinelaw: **The whole merged config is re-serialized into a fresh JS object graph on every caret move.**

`showsEveryKind()` is called from `breadcrumbTrail`, which runs on every `cursor_moved` and `buffer_activated` — including in buffers with no symbols, since `publishBreadcrumbs` does not check the cache first. `getConfig` clones the cached `Arc<serde_json::Value>` and then walks all of it through `rquickjs_serde::to_value` (`quickjs_backend.rs:2862-2870`); the editor side goes to the trouble of an `Arc::ptr_eq` cache to avoid re-serializing per frame (`plugin_dispatch.rs:316-329`), and this undoes the JSON→JS half of that saving ~30×/s while an arrow key is held.

Fix: read the flag once in `refreshBreadcrumbs` (or cache it and refresh on config change) rather than per trail computation.

---

crates/fresh-editor/src/app/chrome/splits.rs
```diff
@@ -1460,3 +1492,91 @@ impl Editor {
...
+    fn an_absent_label_keeps_the_reported_position() {
+        let buffer = Buffer::from_str_test(SRC);
+        let reported = breadcrumb_target(&buffer, 1, 4, "NotOnThisLine");
+        assert_eq!(reported, buffer.lsp_position_to_byte(1, 4));
+    }
+
+    /// Columns are UTF-16 offsets, so a line with wide characters before the
+    /// name still resolves to the name's byte.
+    #[test]
+    fn utf16_columns_resolve_to_the_right_byte() {
```
sinelaw, line 1593: **This test asserts nothing about UTF-16: the wide characters are on a different line from the position being resolved.**

The fixture is `"# 日本語\nclass Store:\n"` and the assertion resolves `(line 1, character 6)`. `lsp_position_to_byte` converts the UTF-16 column only within `line`'s own bytes (`buffer/mod.rs:2996-3008`), and line 1 is pure ASCII, so byte 6 and UTF-16 offset 6 are the same number — the test passes identically against an implementation that treats `character` as a raw byte offset. Line 0 is never consulted.

To actually pin the conversion, put the wide characters *before* the name on the same line, e.g. `"class 日本語Store:"` asserted at `character: 6+3`.

---

sinelaw (commented):
Independent review of the full diff (34 files). Six findings, none of them style: one multi-cursor correctness bug on the crumb-click path, one wrong-output bug when the same buffer is open in two panes, a forced-repaint-per-caret-move regression of the kind #2100 fixed, a silent capability loss in `getLineStartPosition`/`getLineEndPosition` for ≥10 MB files that hits three shipped plugins, a per-caret-move config re-serialization in the plugin, and a test whose fixture cannot exercise the UTF-16 conversion it claims to pin.

I also traced and cleared: the `want_end` rewrite in `handle_get_line_position` (the `line_start_offset(line+1) - 1` form matches the deleted scanner for LF, CRLF, empty last line and missing trailing newline, since `line_count = lf + 1`); `breadcrumb_target`'s UTF-16 handling and its search-from-reported-column fallback (`from` is always a char boundary because `lsp_position_to_byte` walks chars); the `is_primary` plumbing on both hook paths, against `Cursors::add` setting `primary_id` (`model/cursor.rs:262`); the elision maths in `first_visible`/`lay_out` (widths reconcile, CJK is measured in columns, `total_w <= 1` returns an empty row); the `PaneChrome` parity test's new `continue` guard (it only skips starved panes where `reference_split_layout` is already known to be the worse of the two, and that function is test-only); the retry budget and `lsp_ready`/`diagnostics_updated` wake-ups (bounded, and `sendLspRequest` spawns no server the didOpen path would not already have spawned); and plugin-unload / buffer-close cleanup (consistent with `status_bar_values`).

---

crates/fresh-editor/src/app/chrome/splits.rs
sinelaw: Confirmed and fixed in 456ce03: the crumb click now collapses to the lowest id before placing the caret, as `handle_editor_click` does.

---

crates/fresh-editor/src/app/window/mod.rs
sinelaw: Fixed in 41551be: trails are keyed by pane now, tagged with the buffer they describe, and a publish lands on the focused pane whose caret the plugin read. The new e2e test fails against the old buffer-keyed lookup with both panes drawing one trail, which is the case you described.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
sinelaw: Fixed in 456ce03: `SetBreadcrumbs` has its own arm now, which asks whether the publish would change what is drawn — including which pane it would write to, so the check stayed honest when trails moved to pane keys.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
sinelaw: Fixed in 456ce03, by giving callers the right-shaped call rather than a fallback: `getLineStartForPosition`/`getLineEndForPosition` take the byte a caller already holds and scan out from it, bounded and needing no index, and vi-mode's `$`, `D` and `^` moved onto them. The handler verifies the range it scanned was actually loaded, since `slice_bytes` returns short data for an unloaded region rather than failing — a confident wrong byte would be worse than none. The line-number pair is unchanged; `diff_nav` and `audit_mode` genuinely have a line number rather than a byte, so they still need the index.

---

crates/fresh-editor/plugins/lsp_navigation.ts
sinelaw: Fixed in 456ce03: the flag is read once and dropped on `config_changed`, which also redraws so toggling the setting still shows immediately.

---

crates/fresh-editor/src/app/chrome/splits.rs
sinelaw: Right, and it was worse than it looked: with the label present the name search finds `Store` whatever the column arithmetic does, so moving the wide characters onto line 1 alone would still not have pinned it. Fixed in 456ce03 by doing both — wide characters before the name on the line being resolved, and an empty label so `breadcrumb_target` returns the converted position verbatim.

---

sinelaw:
`test macos-latest` is red on two nextest 180s timeouts — `e2e::markdown_compose_scrollbar_markers::heading_marks_survive_exploring_the_document` and `e2e::lsp_order::test_did_open_sent_before_hover` — and neither is this PR's: both run with this PR's plugin not loaded (one with plugins disabled outright, the other with only `markdown_compose` copied in), both pass on ubuntu and windows on this same commit, and across the ~4900 tests common to both logs that runner was about 25% slower than master's, with the LSP and plugin tests the *least* affected rather than the most.

The underlying cause is that both settle on `wait_until`, which has no deadline, so they are capped by the slow-timeout rather than failing an assertion; the per-page `wait_until_stable` loop in the markdown test can therefore run unbounded under load. I have spent the one re-run and am not widening this PR to re-time an unrelated test, but I can push a bounded settle if you would rather have it here.

---

sinelaw:
**Do not merge as is** — holding this for a performance pass over the diff. Nothing below is a correctness bug; the feature behaves as described. The issue is that several of its costs are paid by people who are not using it, on the cursor-move and render paths.

- Every arrow key in a plain text file sends a pointless "clear breadcrumbs" message to the editor, forever — `publishBreadcrumbs` (`lsp_navigation.ts:602`) falls straight into `clearBreadcrumbs` whenever a buffer has no cached symbols, and every caret move reaches it.
- Handling that message allocates a throwaway `String` per pane, every time: `plugin_name.to_string()` sits inside the filter closure in `handle_clear_breadcrumbs` (`plugin_dispatch.rs:2319`). The `set_breadcrumbs` path next to it borrows correctly.
- Turning breadcrumbs off in the config only hides the row. `show_breadcrumbs` is read once, in `pane_chrome` (`window/mod.rs:2783`); the plugin never consults it, so it still issues `documentSymbol` on every buffer activation and every debounced edit, and still keeps the cache.
- The trail is recomputed by scanning *every* symbol in the file on every cursor move — `breadcrumbTrail` (`lsp_navigation.ts:578`) is an O(n) filter plus a sort plus two array allocations, running in QuickJS, to produce one to four crumbs.
- That scan is bigger than it needs to be: the non-scope kinds (fields, variables, constants, enum members — usually the majority of what a server returns) are kept in the cache and re-rejected on every caret move, rather than dropped once at fetch time.
- The plugin caches the whole flattened symbol tree for every file opened, released only on `buffer_closed` (`lsp_navigation.ts:468`).
- The trail is deep-copied twice per pane per frame — `items.clone()` in `pane_breadcrumbs` (`window/mod.rs:2895`), then again by `Rc::from(items.to_vec())` in `breadcrumbs::surface` (`breadcrumbs.rs:65`), where the `Rc` was presumably meant to avoid the copy.
- Each keystroke spawns a 250ms timer that usually does nothing (`lsp_navigation.ts:656`). The debounce itself is correct; this is garbage, not a hazard.

Worth recording the other direction too: replacing the full-buffer materialisation in `handle_get_line_position` with the piece tree's line index takes that call from O(buffer) with an allocation to an O(log n) descent, and `attachLineText` now costs three round trips per *unique line* instead of per symbol. For anyone using the symbol finder on a large file that is a substantial net win.

The first three are small and self-contained. The kind pre-filter and the `Rc` are slightly larger. Replacing the linear scan with a binary search over the sorted start lines is the only one worth leaving for a follow-up.

---

PR #3298, opened 2026-09-15
clips: the orchestrator dock, filmed against the real thing — cleanup in three cuts, badges in one

sinelaw:
Five specs and two fixtures, filming the orchestrator dock. Four cuts of it being **tidied** — 25 workspaces with auto-generated names nobody can read, filed into seven folders and renamed in place — and one of it **reporting**: eleven agents working, six finishing, one asking, then the toolbar taken apart.

**Nothing about the dock is mocked.** The rows are real workspaces on real git worktrees with real branches and real uncommitted work; the folders, the moves and the renames are real orchestrator API calls, and every state badge is `sessionState` reacting to what a terminal actually printed. Only the agents are scripted, so the takes are repeatable.

| Spec | Shape |
|---|---|
| `fresh-dock-cleanup.json` | the long one — 51.7s, the whole story with the agent pane |
| `fresh-dock-cleanup-short.json` | 14s, square, for a feed |
| `fresh-dock-cleanup-short-vertical.json` | the same at 9:16 |
| `fresh-dock-cleanup-focus.json` | 6.55s — the dock alone, arranging and renaming, nothing else |
| `fresh-dock-ready.json` | 7.5s — the badges, then the toolbar's controls lifted out and labelled |

The four cleanup cuts come off one capture block; the short and focus generators copy it from the long spec rather than restating it, so they cannot drift apart when the fixture changes.

## Requires sinelaw/tui-clips#10 and #11

#10 adds `swipe`, `"transition": "wipe"`, `tag`, `render.crt` and the `mark`/`at` capture steps. #11 adds `"transition": "shatter"` and `explode` on a solo beat. Rendering against `tui-clips@main` fails at load naming the unknown key — loudly, not silently — so land those first.

### `✓` could never appear for a workspace created this session

`sessionState` reads `workedSinceSeen`, which the `terminal_output` handler arms only once the current output burst has lasted `WORK_MIN_MS`. The burst clock starts on `lastOutputAt === null`:

```js
if (s.lastOutputAt === null || now - s.lastOutputAt >= IDLE_AFTER_MS) {
  s.burstStartAt = now;
}
```

Two of the six `orchestratorSessions.set` sites — the one that turns a finished build into a live row (`runLocalCreate`) and `createLiveSession` — left the field **absent** rather than null. `undefined` matches neither arm: not `=== null`, and `now - undefined` is `NaN`, which is not `>=` anything. So `burstStartAt` was never set, `now - (s.burstStartAt ?? now)` was `0` on every output event for the row's whole life, and the flag could never arm.

Those rows reached `working`, `blocked` and `idle` fine, but a row that had plainly been working and then went quiet fell through to the dim `·` instead of the green `✓`, and the folder roll-ups never counted it. Only rows that had been through a `terminal_exit` (which does assign `null`) could ever get there.

Both literals also carried `state: "running"`, which is not an `AgentState`; `plugins/check-types.sh` was already flagging those two lines and flags neither now.

Diagnosed by loading the plugin from an isolated config dir with a probe in the handler — `sinceBurst: 0ms` on all 20 events of a 2.5s burst, `burstStartAt: undefined` throughout, same object identity each time. `e2e::orchestrator_done_badge` drives the flow it hid in (`[ + New ]` → Create Workspace → give the terminal work → leave before it starts → wait out the idle window) and asserts on the glyph in the dock row, cut at the dock's wall so the editor beside it cannot supply a `✓`. It fails on the parent commit and passes on the fix.

The fixture is not allowed to paper over the product. If a badge will not appear, that is the news.

### The dock's attention line is withdrawn

`● 2 need you · ✓ 1 done` sat between the action row and the list, drawn only while one of the counts was non-zero. That conditional is the part that did not survive contact: a row that is absent and then present is a row that *appears*, and every row under it moves down one the moment it does. Filming a dock while agents finish, the change you see is not the badge that just turned green — it is the whole list stepping.

The counts were already on the rows and rolled up on the folder headers, so the line was paying for a shift with a repetition. Gone; the dock's chrome is a fixed three rows now whatever is pending. The badges, the roll-ups, the status-bar notice and `Orchestrator: Jump to Attention` are untouched. §2.3 of the UX design note is marked **withdrawn** with the reason rather than deleted, and its wireframes regenerated from `gen-orchestrator-wireframes.py`, which is the source of truth for them.

## Three things the cleanup cuts needed that the other clips did not

**Scripting the editor from outside is refused by design.** `fresh --cmd script run` and `fresh --cmd init reload` both answer *"no capability token: script evaluation is not authorized"*. The token is minted per terminal and injected into the PTY, so the only two ways in are `init.ts`, which runs at startup with the full API, and a process the editor itself spawned. The clip uses both: `init.ts` stages the 25 workspaces, and `runAgent` launches the cleanup agent into a real pane where it inherits a real token and drives the dock over it — which is what makes the agent honest rather than a puppet.

**The agent follows the camera.** Pacing the two off separate clocks cannot be made to work: they are anchored by one constant — how long `Return` takes to become a running agent — and every error in it shifts the whole sequence. At 1.2s every shot came back a step late; corrected to 2.3s the middle lined up and the ends still did not. The error is not constant, so no value fixes it, and the failure is invisible in the stills because all the states are still *distinct*, just each one a step on. So `tui-clip` leads: `shot` writes its raw `.xwd` at the instant it grabs, `record.sh` hands the agent that directory, and the agent makes one dock change, waits for the shot that records it, then makes the next.

**The dock's compact row is `sessionNodeEntry`: glyph, name, tags, nothing else.** Branch and the git summary belong to card view, and 25 rows of cards do not fit. The fixture's branches and uncommitted work are real either way — `git-report.json` records what the probe saw — the compact dock simply does not draw them, and restyling it to would be filming something fresh does not do.

## …and four the badges cut needed

**The agent is the repo's own fake agent.** Every row runs `crates/fresh-editor/tests/fixtures/coding_agent.py`, copied in by `record.sh` rather than duplicated under `assets/`, so this clip and the e2e showcases film the same program. That matters beyond tidiness: the dock reads a *terminal*, and a bare line printer is not one.

**A state is a schedule, not a claim.** A row that goes quiet at `W + run` is still `working` for `IDLE_AFTER_MS` and repaints 100ms after that, so the `run` column in `rows.json` *is* the edit: six values 0.6s apart put six checks on the camera in turn, alternating between the two folders so both roll-ups tick, and the seventh lands the red dot after them. 0.6s is what leaves a shot room between two flips; nothing under ~2.0 works at all, because `coding_agent.py` waits 0.5s before its first line and the burst left over has to clear `WORK_MIN_MS` with room to spare. `gen-spec.py` derives every shot time from that same table.

**The take and the cut keep different clocks.** A badge cannot land sooner than `run` + `IDLE_AFTER_MS`, so the shots are spread over thirteen seconds of real dock; the cut plays them a third of a second apart. That works because exactly one thing changes per still and always the same kind of thing, in the same two places. The eleven rows are filed under two folders *before* filming — this clip is about what a badge says, and a list reorganising itself is the other clip.

**W — the moment all eleven start — is the camera's first shutter, not a clock.** Staging and working have to be two phases (eleven creates is eleven layout changes, and output from a not-yet-speaking session is written off for 1.5s after each), and the second phase has to line up with the camera to the tenth of a second. `bin/wait-for-start.py` holds every row until a raw `.xwd` appears in `CLIP_SHOTS_DIR`; the capture's absolute `at:` marks hang off the `mark` beside that same shot. Without a camera it falls back to a clock, which is what makes the fixture inspectable in tmux without recording anything.

At 2.6s the list has said everything it has to say, so the clip wipes to the toolbar and takes it apart: `[ + New ]`, the search affordance and the `⋯` menu lift out of the bar, each outlined and labelled, while the rest falls back. **Those three rects are the window's own numbers** — `Dump UI Tree` is bound to F9 in the clip's config (a key rather than the palette, because a palette invocation dumps the frame with the palette over everything), and `tui-tree ui-tree.json --list` reads out `widget:new-session` at `(0,1 9x1)`, `widget:search-toggle` at `(28,1 8x1)` and `widget:dock-menu` at `(38,1 1x1)`, which is what the spec says verbatim. A font or geometry change costs one re-dump and no arithmetic.

It also films *this tree's* plugin: plugins are embedded at build time, so a released `FRESH_BIN` carries the orchestrator it shipped with. `record.sh` copies `crates/fresh-editor/plugins/orchestrator.ts` and its `lib/` into the capture's config directory, where a plugin wins over the embedded copy — unedited, so what is on screen is the plugin as committed beside it.

## Isolation

The orchestrator API files and renames real rows, so a demo that ran against a live daemon would bury the user's own workspaces under the fixture. Everything runs under `$CLIP_HOME` with its own XDG root, so `listWorkspaces()` can only ever return the staged fixture — verified before the first mutation. Each `record.sh` tears down before it stages, because a worktree registered by a previous take makes git refuse the next `worktree add`, and `teardown.sh` removes the tree, the registry and the worktree registrations.

## Repeatability

```sh
FRESH_BIN=… CLIP_SPEC=scripts/clips/fresh-dock-cleanup-focus.json \
  scripts/clips/assets/fresh-dock-cleanup/record.sh

FRESH_BIN=… scripts/clips/assets/fresh-dock-ready/record.sh
```

`verify-shots.py` then checks the cleanup take photographed the states it was aimed at — `folders` before any row is filed, one more filed row per `mv<i>`, exactly sixteen rewritten rows at the end. It compares rows by how much of each moved rather than exactly: two screens of the same dock differ by a caret phase or a scrollbar segment without a word of text differing, and on a good take the rewritten rows score 11.5–20.1 mean absolute difference against 0.95–5.7 for those.

Rendered output lands in `tui-clips/out/`, which is gitignored — nothing binary here.

https://claude.ai/code/session_01EA4na1W7FKhmygpduwXK1Y


sinelaw:
`test ubuntu-latest` is red on `79f443833`, and it is not this PR's.

```
Summary [1637.777s] 9841 tests run: 9840 passed (1 slow), 1 timed out, 163 skipped
TIMEOUT [180.013s] (7225/9841) fresh-editor::all_tests
  e2e::vi_mode_bugs::test_vi_cw_dot_repeat_twice_stays_a_change
```

One test, hitting nextest's 180s per-test budget, in vi-mode dot-repeat — code this diff does not touch. Everything else passed, **including the new `e2e::orchestrator_done_badge`**, which is the check this PR actually needed.

**The base branch is red the same way.** Run [10407](https://github.com/sinelaw/fresh/actions/runs/35153871518) on `41fe3b98e` — this PR's own merge base, `CHANGELOG: prepare 0.5.2` — failed with:

```
Summary [1599.795s] 9784 tests run: 9783 passed (4 slow), 1 timed out, 165 skipped
TIMEOUT [180.014s] (4835/9784) fresh-editor::all_tests
  e2e::markdown_compose_scrollbar_markers::every_heading_is_marked_without_scrolling
```

Same shape, different test, different platform: exactly one e2e test out of ~9,800 hits the 180s ceiling in a ~27-minute run. That the *identity* of the test moves between runs — vi-mode here, markdown-compose on the base, and two more on macOS in #3278 — is the signature of runner contention against a fixed per-test timeout, not a defect in any one of them. The three most recent `master` runs before this PR branched (10407, 10343, 10338) are all red.

No fix exists to port: this is a timeout budget against a saturated runner, not a change anyone has landed. Re-running the failed job once, which is the one re-run these rules allow.

One thing I considered and rejected as the cause: this PR adds `orchestrator_done_badge`, which spends ~14s of real wall-clock holding a PTY, so it does add load to the same binary. But the base fails identically without it, so it is not what pushed a neighbour over the edge — and the test cannot be made cheaper without removing the thing it tests, which is a clock.

---

PR #3301, opened 2026-09-15
Orchestrator: make remote workspaces work, and stop the dock misleading people

sinelaw:
Fixes from two rounds of first-time-user testing of the Orchestrator dock. Reports and evidence are in `docs/internal/workspace-sessions-usability-test.md`.

## The main thing

**You can now create a workspace on a remote SSH machine.** Before, picking a host from `~/.ssh/config` and giving it a project path always failed. Two reasons:

- The dock resolved the host alias itself and passed the resolved address to `ssh`, so everything else in your `Host` block — the key to use, where known hosts live, proxy settings — was ignored. It now passes the alias and lets `ssh` read its own config.
- A host you'd never connected to was a dead end. Now you get a prompt showing the machine and its key fingerprint, with **Trust and connect** or **Cancel**. Accepting writes the key wherever your config says it should go.

## Fresh no longer writes a branch into your repository

Archiving or deleting a workspace used to maintain a `fresh/fresh-sessions` branch, through a `.sync-workspace` worktree registered in your own repo, and push it to your `origin` under your git identity — with no way to decline and no message when it failed.

All of it is gone. Archive and delete now touch the workspace's own worktree and Fresh's data directory, and nothing else.

It was the wrong shape as well as the wrong place. Fresh has no central session list by design — a session *is* a directory, and the registry is the per-directory workspace cache it scans at startup — and there's an older migration in-tree that moved *away* from a central file. This put one back, inside the working tree, which is where Fresh otherwise never keeps state. Nothing ever read it back.

## Saved machines survive concurrent windows

Machines were one `machines.json`, rewritten in full on every change from a list each process had cached forever. Two windows adding different machines lost one of them, and a connection test finishing in one window rewrote every machine from a snapshot that could be minutes old.

Three separate fixes, because they're three separate problems:

- **One file per machine** (`machines/<id>.json`), so a write touches only the machine it's about. Also bounds a bad write to one machine instead of the registry, and makes delete an `unlink`.
- **Atomic writes** — temp file beside the target, then rename. Splitting the file doesn't give you this on its own.
- **Cache invalidation** when the Machines dialog or the New Workspace form opens, so a machine added in another window is actually visible. The split doesn't give you this either.

## Things that said one thing and did another

- The create form claimed a blank branch field meant "use the branch above". It actually created a new branch named after the workspace. The form now shows what it will really do, live as you type, including where the files will go.
- A checkbox showed as ticked next to a caption reading "Not checked". The caption was about something else entirely and has been reworded.
- The delete confirmation said your branch survives only implicitly. It says so now.
- It also promised to remove the worktree for remote workspaces and didn't. Now it does — or says plainly that it can't, when the machine isn't reachable.
- On a workspace with no worktree, it promised to run `git worktree remove` and warned about losing changes. Neither was true, and both lines are now gone.

## Rows and dialogs that ran out of room

- A remote row named the machine twice and the workspace never. It now reads `⇅ demo-1  testbox`: the name, then the machine once.
- While a workspace was being created, its row showed the name and host but never what it was *doing* — the status was pushed off the end of the line. CI caught this one; the row is back to name, machine, status, all visible.
- The delete dialog took its heading from the row's display name, which for a live session includes the shell's title. A heading reading `Delete workspace demo-3 · bash — root@vm: ~/.local/share/…/dem` answered everything except which workspace. It now uses the workspace's own name.

## When something goes wrong

- A failed workspace used to show a chopped-off error in the dock and hide the real text in the status bar. The full reason now appears under the list, with **Retry** and **Dismiss** buttons, the moment it fails.
- While the dock waits for you to confirm a host key, the row says so instead of claiming it's still working.

## Smaller fixes

- Delete now offers a checkbox to keep the worktree and its files, shown only when there is one.
- The workspace name no longer repeats one that's already taken.
- Deleting a workspace no longer leaves a dead folder behind in Fresh's data directory.
- "Add machine…" is gone from the machine list. It looked like a button but needed a hidden extra keypress, and quietly undid itself otherwise. The two working options sit right next to it. (Adding machines is unaffected — the Machines manager has a real button.)
- The form stopped shifting position a second after opening.
- Left/right arrows now work in an open dropdown, as the on-screen hint always said.

## Reported, not fixed

Four findings are written up rather than fixed, because each needs a decision rather than a patch. All four were re-tested on the current head, so these are current readings, not carried-over notes:

- **Work doesn't survive closing the editor.** After `Ctrl+Q` and relaunch, the workspace is not restored as a session and anything running in it is gone. The worktree and branch survive on disk, and the editor even remembers which workspace was last active — it just doesn't bring it back.
- **`[ + New ]` is a form, names no target, and silently reuses the last machine.** All three reproduce; the machine is remembered in `orchestrator.last_machine` with nothing on screen saying so.
- **Clicking a row moves keyboard focus off the dock**, so `F2` opens nothing until you click the dock header.
- **Local and remote worktrees live in differently-named folders.** Remote ones have to go somewhere on the host, which has no Fresh data directory.

Reasons, and what each would cost, are in the report.

## Testing

Every fix was checked by driving the real editor in tmux against a test SSH server, with the key and known-hosts files deliberately in non-standard places so a connection can only succeed if the config is genuinely honoured. There are new automated tests for the config fix and for the removed branch mechanism; the host-key prompt and the remote worktree cleanup need a live SSH host and were checked by hand.

Two reported problems turned out not to exist, and the report says so with the evidence.

One note on CI: this suite currently has one arbitrary test stall at the 180s timeout most runs, on Windows especially — master is red with the same signature ([run 35014229660](https://github.com/sinelaw/fresh/actions/runs/35014229660)). It is not from this branch and is worth its own look; details in the comments.

https://claude.ai/code/session_01FMGNVxT37NHo1nS9Be6xwC


sinelaw:
Added in `d3523d2`: the Delete confirmation now asks whether the worktree goes with the workspace, instead of stating it as a fact.

Two things were wrong beyond the missing choice. On an in-place or shared-tree session — the project row itself, say — the pane still announced "run `git worktree remove`" and "keep the branch", and warned that uncommitted changes would be lost; none of that happens, since that delete drops the record and touches no files. And for a session that does own a worktree there was no way to say "I'm done with this session but I want to keep the files".

```
This will:                                    This will:
  • stop all workspace processes                • stop all workspace processes
  • run `git worktree remove`                   • leave the worktree and its files where they are
  • keep the branch — … does not delete it      • drop the workspace record
  • drop the workspace record                   • update fresh/fresh-sessions — …
  • update fresh/fresh-sessions — …
                                              [ ] Also remove the worktree and its files
Uncommitted changes will be lost.
                                              [ Cancel ]  [ Confirm Delete ]
[v] Also remove the worktree and its files

[ Cancel ]  [ Confirm Delete ]
```

Checked by default, because removing the worktree is what Delete has always done — this adds a way to keep the files rather than quietly changing what the button means. The consequence list is rebuilt from the checkbox so the dialog never describes an action other than the one the buttons will take, and the data-loss warning now appears only when files are really going.

The checkbox is absent entirely when there is no worktree, using the same test the removal itself uses (`ownsWorktree` locally; the linked-worktree-under-`~/.fresh/worktrees/` pair remotely). The choice reaches `deleteOne` as an argument rather than through dialog state, so the plugin API's delete keeps its existing behaviour.

Verified by hand across all four combinations — local and remote worktree, each checked (worktree and registration removed, branch surviving) and unchecked (worktree, files and registration intact, only the dock row gone) — plus a no-worktree session showing no checkbox and none of the worktree lines.

Two e2e tests asserted the data-loss warning on the project row, where it was never true, and now anchor on a consequence line that is; their intent (the pane renders unclipped down to its buttons) is unchanged.

One thing to flag rather than fix: `e2e::keybinding_editor::test_unsaved_changes_confirm_dialog` and `test_confirm_dialog_discard` are flaky under plain `cargo test` — a *different* one of the pair failed on each of two runs of a broad filter, and both pass in isolation. They belong to the keybinding editor, which nothing here touches; `cargo test` runs the whole suite in one process, so order-dependent pairs surface there where `cargo nextest` would not. Noted, not chased.

---

sinelaw:
Ran a code review over the branch. It found six problems; five were real and are fixed in `d9a99a1`. Three of them are things the manual testing pass walked past, so they're worth naming.

**The branch preview still lied in one of its three arms.** The create has three: a typed new-branch name cuts it off the base; a typed *Checkout branch* with no new-branch name checks that branch out and cuts nothing; both blank cuts `<workspace>` off the default. The preview keyed off "is there a base", so the middle arm — the common "put me on `feature`" case — read "blank: new branch acme-widgets-1, cut from feature". No branch is created there at all.

This is the uncomfortable one: the manual pass *saw* it. A session came out on `feature` while the preview had named a new branch, and I wrote that off as "the New branch name field must have been cleared". The evidence was on screen and I explained it away. The preview now takes the typed base and the placeholder default as separate arguments — which arm runs turns on whether Checkout branch was *typed*, not on what the field displays — and the third arm says `checks out feature — no new branch is created`. Re-verified against git: worktree on `feature`, no new branch.

**The remote worktree removal could have deleted local files.** It routes through the session's own authority by making its window active first, but `SetActiveWindow` on a *disconnected* remote session installs an empty local shell and starts the connect *afterwards* — so the spawn races it and is served by the **local** spawner, running `git worktree remove --force` against a remote absolute path on this machine. Not hypothetical in this rig, where the "remote" is `127.0.0.1` and the path exists locally too.

Now it refuses unless the host reports the session connected. My first attempt at that guard was itself wrong — it tested our own `remote.state`, which a freshly created session leaves at `"starting"` for ever since nothing promotes it, so it blocked exactly the live sessions it was meant to allow. It asks `WindowInfo.remote.connected` now. Verified both ways: connected deletes remove the worktree; a delete after the connection was killed leaves everything untouched and says so.

**The host-key gate only covered the worktree path.** With the worktree toggle off the create went straight to `attachRemoteAgent`, whose carrier trusts an unknown key on its own, while the form had already promised "you'll be asked to confirm it on create". The reviewer predicted a hard failure; what actually happened was worse in kind — it *succeeded*, silently trusting a host whose fingerprint the user never saw, which is the whole thing F1 was about. The gate now runs for every ssh create.

**The auto-name counter could be driven by a typed name** (`release-2026` would push every later auto-name in every project to `-2027`); it's now claimed only when the user left the field blank and the spec carries the generated default. **The failure panel's height estimate** divided width by columns while the host word-wraps with a hanging indent, so a real `Permission denied (publickey,…)` over-allocated the tree and clipped the dock's last row — the exact failure the panel exists to prevent. And **two doc comments** had been separated from their functions by my insertions; both restored.

Two things fixed while in there: the "worktree left on the host" message was written with `setStatus` inside `deleteOne` and then immediately overwritten by the batch's own "deleted 1 workspace(s)" summary, so the one case that most needed reporting said nothing — it travels out on the result now. And the confirm dialog no longer promises `git worktree remove` for a host it cannot reach; it says the worktree stays put, and why.

`cargo fmt` and `clippy --all-targets` clean; 216 orchestrator/i18n tests pass. The report's follow-up section records all of this, including the misdiagnosis.

---

sinelaw:
Added `068fa5c` — a second round of fixes from the cold comparative study against `herdr` 0.9.0 (`docs/internal/workspace-sessions-comparative-study-cold.md` on `claude/herdr-0.9-cold-comparative-study`). Fast-forward onto this branch; no history rewritten.

Eleven items re-tested by driving the TUI in tmux at 200x50 against the study's own fixture — `sshd` on `127.0.0.1:2222`, pubkey only, `testbox` alias with **both** `IdentityFile` and `UserKnownHostsFile` at non-default paths, `known_hosts` and editor state wiped before each run. Screens captured whole, never as a row range: two of the study's conclusions turn out to be artefacts of partial or colour-stripped captures.

**Two don't reproduce.** The Severity-4 "unknown host key is a dead end" is already fixed on this branch — the study ran at `da558dd`, and the error it quotes (`host key not trusted — nothing was connected to`) is the string shown when its trust prompt is *declined*, added in `cf4712d`. Driven cold, the modal appears with host, key type and SHA256 fingerprint, accepting yields `SSH_CONNECTION=[127.0.0.1 55074 127.0.0.1 2222]` + `# REMOTE-DEMO-REPO` + a live `sshd: root@pts/2`, the key lands in the configured non-default `known_hosts`, and `~/.ssh/known_hosts` is never created. Retry re-opens the prompt rather than failing identically. The "indistinguishable placeholder" is italic grey — the study read a colour-stripped capture.

**Five fixed.** The real half of "permanently in progress": while the trust prompt waits there is genuinely no ssh process, no worktree and no `known_hosts`, yet the row still said `Adding worktree…` — which accounts for every observation in that item. It now says it is waiting for the user. Plus: "Add machine…" removed (it wasn't dead, it *armed* and needed an undocumented Enter, then silently reverted on Tab); the declined-key message no longer reads as cut off; remote rows carry the workspace name (`alpha · ssh:testbox`); and `terminals/<root>/` is removed with the workspace record instead of accumulating one dead directory per delete.

**Four reported rather than guessed at**, with reasons in the report: daemon-backed persistence and one-click creation both change defaults the brief flags as contentious; the dock-focus item's obvious fix is precluded by `click_on_focused_dock_row_dives_focus_into_session`, and its alternative depends on a `⋯` menu that neither I nor the study could open by mouse (an in-process e2e test clicks the same glyph and passes — that disagreement is a finding of its own); and the two worktree path conventions differ because the remote host has no fresh data directory to nest under.

**One investigation, no behaviour change.** The delete dialog claims `fresh/fresh-sessions` is "pushed to origin" — never exercised, because the study's fixture had no `origin`. I added one. Deleting (not creating) a workspace pushes that branch to the user's remote, committing `sessions.json` under their own git identity; against an unreachable origin the delete still reports success and says nothing. My judgement is that it **should not push by default** — a local delete is not a network operation, the branch lands in a shared namespace where it can trip CI and branch protection for collaborators, and failure is invisible in both directions. Left unchanged and proposed instead, as instructed.

`cargo fmt` and `clippy --all-targets` clean locally; the suite is CI's.

---

sinelaw:
Added `988ef1d`, which settles the one investigation the previous comment left open: **the `fresh/fresh-sessions` push to `origin` is removed.** That comment said the behaviour was left unchanged and proposed instead — that no longer holds.

The snapshot itself stays. `syncSessions` still maintains the `.sync-workspace` worktree and still commits `sessions.json` to the local `<user>/fresh-sessions` branch, so a future cross-machine recovery feature has something to read. Only the push is gone. No setting was added: not writing to a shared remote when someone deletes a local row is the right default rather than a preference, and nothing reads the branch back yet anyway — the push had no consumer. When recovery is actually built, the push belongs with it, as something the user turns on and can watch fail.

The archive/delete confirmation line changed with it, in all 15 locales: `pushed to origin` → `kept on this machine`.

**The test was the awkward part, and the first version of it was worthless.** `deleting_a_workspace_does_not_push_anything_to_origin` gives the fixture a real bare `origin` — the thing the original study lacked — deletes two workspaces and asserts `origin` gains no refs. But the snapshot commit lands *during* the sync, so waiting for the local branch and then checking `origin` just wins a race against the push. That version passed against the pushing code, which is to say it proved nothing; I only caught it because I put the push back and re-ran rather than trusting a green result.

The wait is now "the branch exists **and** the footer's in-flight indicator has cleared", and that indicator is shown for exactly as long as the sync is running, so the condition can only become true after the push would have happened. Verified both directions: with the push restored the test fails with `refs there are now: ["refs/heads/fresh/fresh-sessions"]`, and passes with it removed.

`cargo fmt` and `clippy --all-targets` clean; i18n key parity intact across all 15 locales; the TypeScript baseline is unchanged. The suite is CI's.

---

sinelaw:
`test ubuntu-latest` is red on `0fca0a0`: `e2e::vi_mode_bugs::test_vi_visual_line_dedent` timed out at 180s. 9836 of 9837 passed; nothing else failed.

Both earlier failures on this PR were mine and are fixed (`21fbe57`, `0fca0a0`). This one I do not believe is, and here is the evidence rather than the assertion:

- **The PR touches no vi-mode code.** Against the merge base it changes 12 files: the orchestrator plugin and its i18n, `plugin_dispatch.rs` (removing the terminals dir on delete), one widget change, the orchestrator tests and fixtures, and the report. The widget change is `dropdown.rs`, and it is confined to the *open-list* branch of the Dropdown key handler — `Up|Left` and `Down|Right` move the selection. A vi-mode text-editing test never has a dropdown open, so it cannot reach that code.
- **`test macos-latest` passed on this exact commit**, running the same suite including this test.
- **It passes locally on this exact tree** in 0.59s.
- **It passed on ubuntu on the previous commit of this branch** (`988ef1d`), in the run where only the orchestrator test timed out.
- **It is a timeout, not an assertion failure.** The test ends in `wait_for_buffer_content`, which has no deadline of its own, so a wrong or late buffer state parks it until the harness kills it. That shape times out; it does not fail fast.

So: same commit green on macOS, same tree green locally, same test green on ubuntu one commit earlier, and no path from this diff to the code under test.

I have not changed anything for it — there is no fix to port, because there is no reproduction to fix. I am spending this PR's one re-run on the failed job now to see whether it repeats. **If it fails again it is real**, and I will treat it as this PR's to root-cause rather than re-run it a second time.

Worth noting for context: master's own latest commit (#3302) fixes a proptest that "turned up on macOS CI while ubuntu passed the same commit" — so single-platform, single-run failures have a recent precedent in this suite. That is context, not an excuse, and it does not change what happens if the re-run is red.

---

sinelaw:
Following up on my earlier comment about the vi-mode timeout, because I said that if it repeated I would treat it as real rather than re-run it. It did repeat — and the answer turned out to be checkable against master rather than arguable.

**The same failure is on master.** Run [35014229660](https://github.com/sinelaw/fresh/actions/runs/35014229660) — head `1786c264`, the current master tip, no PR involved — fails `test windows-latest` with:

```
TIMEOUT [180.199s] e2e::vi_mode_bugs::test_vi_visual_charwise_join_spans_selection
```

The shape across every run I have looked at is identical: **exactly one test per job stalls at the 180s nextest cap, and it is a different, unrelated test each time.**

| head | job | timed-out test |
|---|---|---|
| `1786c264` (**master**) | windows | `vi_mode_bugs::test_vi_visual_charwise_join_spans_selection` |
| `0fca0a0` | ubuntu | `vi_mode_bugs::test_vi_visual_line_dedent` |
| `0fca0a0` | windows | `vi_mode::test_vi_visual_line_delete` |
| `2d7a6dd` | windows | `markdown_compose_scrollbar_markers::heading_marks_survive_exploring_the_document` |

A defect introduced by this PR does not wander from vi dedent to vi delete to markdown scrollbar markers between runs, and does not appear on master where none of this branch exists. Three of the four are in code this PR does not touch; the fourth is master's own.

On the current head `2d7a6dd`, **12 of 13 checks pass** — including `test ubuntu-latest`, which is green after carrying the vi timeout on the previous head, and `test macos-latest`. Windows is the only red one, with the markdown test above.

There is no fix to port: master is currently red with this same condition, so there is nothing merged or open that resolves it. It is worth someone's attention as its own piece of work — a per-test timeout that one arbitrary test hits each run points at the harness or the runner rather than at any of these tests — but it is not this PR's to fix, and widening the PR to chase it would be the wrong call.

Spending the re-run on the Windows job now; the earlier attempt was refused with 403 because the run was still in flight.

For completeness, the two failures on this PR that **were** mine are fixed and verified: the dock row that pushed its own status off the line (`21fbe57`) and the missing Linux gate on the alias shim test (`0fca0a0`).

---

PR #3304, opened 2026-09-16
Find the agent sessions other tools are running

sinelaw:
Adds discovery of the coding-agent sessions, workspaces and worktrees that *other* tools are running — tmux panes, Claude Code and Codex transcripts, Orca and Herdr workspaces, and 19 more tools from a table — a dialog to find them, and a way back into any of them.

Today each tool knows only about its own. If you have a Claude Code session in a tmux pane and two Codex sessions in Orca worktrees, nothing shows you all of them in one place — let alone on the three machines you left them running on.

**The bulk is plugins**; the rest is the core surface they needed, and one change to how the editor holds a backend that fell out of making remote scanning work at all.

## The branch

Five commits, one per area, in dependency order. Each is reviewable on its own; the last two are the plugins.

| commit | area | size |
| --- | --- | --- |
| `filesystem: one general walk, with bounded and cancellable streamed reads` | `FileSystem::walk`, the agent's `walk_entries`, exec streaming | 12 files, +674 −122 |
| `remote: a bounded handshake, and a reconnect that owns its carrier` | SSH handshake deadline, `Carrier`, its test | 5 files, +273 −63 |
| `editor: one registry owns every connection; attach re-points a window; plugins can open a machine` | `ConnectionRegistry`, attach without restart, `openMachine` and its four primitives | 50 files, +2314 −1147 |
| `agent-sessions: scanner plugins for the sessions other tools are running` | the hub, seven scanner plugins, shared parsing, their tests | 22 files, +3089 |
| `orchestrator: the Everything dialog, to find those sessions and rejoin them` | the dialog plugin, its pure row model, the orchestrator seam, the web-UI scrim fix | 11 files, +2143 −80 |

## The feature

`Orchestrator: Everything` in the palette, or **Everything…** in the dock menu:

```
┌ Everything ─────────────────────────────────────────────────────────────────┐
│  Machine        : [All machines         ▼]    [ Scan ]  ⏎                   │
│  Find           : [payments                  ]  [grouped by project ▼]      │
│                                                                             │
│ ▼ payments  github.com/acme/payments                                        │
│       payments-alpha  tmux  ·  master  ·  agentbox                          │
│       payments-beta   tmux  ·  feature/checkout-flow  ·  lagbox             │
│ ▶ sinelaw/fresh  github.com/sinelaw/fresh                                   │
│ ▶ Not on this machine (25)                                                  │
│ ▶ Problems (1)                                                              │
│                                                                             │
│                                                          [ Close ]  Esc     │
└─────────────────────────────────────────────────────────────────────────────┘
```

Type a name to find it across everything scanned. **Enter on a row opens the New Workspace form filled in to rejoin that session** — `tmux -S … attach -t payments` for a live pane, `claude --resume <id>` for a transcript — on the machine the row is on, in the directory the session was in. A row that offers no way back says so rather than looking like every other row, and one whose resume is not addressed by id (Codex has none) says that too, because `codex resume --last` rejoins the newest session in a directory and not necessarily the one that was picked.

**Group by project** — the default — groups by the *repository* a directory is a checkout of (the canonical remote URL out of `.git/config`, read as a plain file so it works on a read-only machine), falling back to the checkout root and then the directory, and says which rung it landed on. So two clones of one repository, a linked worktree, and the same repo on a laptop and a build box are one heading. The branch from `.git/HEAD` is a second axis: shown on the row, and its own grouping. Group by tool answers the inventory question instead. Groups are sorted by name and rows by title within them. Answers are cached with their age, so reopening the dialog shows what it already knows, and filtering and regrouping never re-scan.

**All machines** reads every configured machine at once, concurrently, so the wall clock is the slowest of them rather than the sum, and one machine being down is one attributed problem line rather than a lost scan. Rows from different machines merge under one project heading, each labelled.

A tool that is not on the machine, or cannot be (super.engineering off macOS), is listed under **Not on this machine**, not under Problems — the scanner contract carries that status (`found` / `absent` / `unsupported` / `failed`), so the problems list holds only tools that are there and could not be read.

## Where the knowledge lives: plugins

```
 plugins/agent_discovery.ts         the Everything dialog: state, rendering,
                                    the scan lifecycle, its cache
 plugins/lib/discovery.ts           how scans become rows — pure, tested on
                                    its own; and DiscoveryHost, what the
                                    dialog asks the orchestrator for
 plugins/agent_sessions.ts          the hub: opens a machine, runs each
                                    registered scanner, correlates, resolves
                                    each directory to a project, reports
 plugins/lib/agent_scanner.ts       shared types, registerScanner, and the
                                    one header-read pipeline every store uses
 plugins/lib/project_identity.ts    the repo → root → directory ladder
 plugins/agent_sessions_tmux.ts     one plugin per third-party tool
 …_claude_code, _codex, _herdr,     (7 plugins; 10 bespoke tools plus 19
 _multiplexers, _orchestrators,     table-driven rows in lib/agent_catalog.ts)
 _catalog
```

Nothing in Rust knows what a tmux socket or a Codex rollout is. Adding a tool is a plugin, or for the table-driven ones a row in `lib/agent_catalog.ts`. The orchestrator's part is five exported methods (which machines exist and how to reach them, how an agent is rejoined, where a New Workspace form opens, the dock focus hand-off); the dialog looks them up when it opens, so load order does not matter.

A scanner also names the agent a row belongs to (`claude`, `codex`) and, where attaching is a thing only it knows, how to attach (a multiplexer's socket path and session name). It does not build the resume command: that is resolved through the agent registry the orchestrator already keeps, so there is one description of how each agent rejoins a conversation rather than a second copy free to drift.

Each scanner's file header states the on-disk facts its parser depends on (the path pattern, which record carries the working directory) in a few lines; there is no separate design document.

## What core gained

Four generic primitives — `walkTree`, `readFilePrefixes`, `runOnTarget`, `machineEnv` — bound to a machine by `editor.openMachine(spec)`, which takes a window to borrow, a transport to dial (`ssh` / `kubectl-exec`), or an `AuthorityPayload`. A scanner is handed a `FreshMachine` and never thinks about which computer it is talking to: the same scanner reads a local home directory and an SSH one. A machine handle is a plugin-owned resource like a terminal or a watch handle: attributed to the plugin that opened it and closed when that plugin is unloaded.

A machine opened *by transport* is **read-only** — built under `TrustLevel::Blocked`, so `run` and a scanner's `allowCommands` fail rather than executing on a machine the user never attached to. Filesystem reads still work, which finds every tool that writes a transcript.

`machineEnv` answers from this process only when the authority's commands actually run in it; a container (docker-exec) is asked with `printenv` inside the container, the same as an SSH host. On a machine opened read-only it cannot be asked at all — reading the environment is itself a spawn — and the scan says so rather than letting the silence read as "not installed" (see **Known gaps**).

### `FileSystem::walk`

`walk_files` could not serve this: it skips every entry starting with `.`, because it exists for project file search, and every store a scanner reads is a dotfile directory.

So it is generalised. `walk` takes `WalkOptions { skip_dirs, include_hidden, include_dirs, max_depth, max_entries }` and yields `WalkEntry { path, rel, entry_type, metadata }`; `walk_files` remains as a *provided* method written in terms of it, so **every existing caller compiles untouched** — only the 16 implementors moved, and each now writes one method instead of two. Metadata is always carried: a walk stats every entry to classify it, so returning it costs a few bytes in a response already being sent, while fetching it afterwards costs a round trip per file.

The remote agent gains a `walk_entries` command. A new command rather than new parameters: an agent predating these options would ignore `include_hidden` and stream a silently empty result, where an unknown method fails loudly.

## The change to existing behaviour

Scanning a machine with no window meant holding a connection with no window — and the editor could not. An `Authority` and the keepalive holding its transport open were two values travelling in parallel: `Window.authority` beside an editor-wide `session_keepalives` map keyed by `WindowId`. Six call sites had to keep them in step, and a path that forgot one left a window connected to a transport nothing was holding open.

They are one value now. `Connection { authority, keepalive }` is created where the transport is established, and `ConnectionRegistry` owns every open connection; windows and plugin machine handles hold `Arc` references into it, and a connection closes when the last reference goes.

That in turn removed the reason for the attach-time restart. `Editor::install_authority` used to stash the authority, set `should_quit`, and let the standalone loop or `EditorServer::rebuild_editor` drop the whole `Editor` and rebuild it — the safe way to invalidate every cache that had closed over the old filesystem or spawner. With windows owning their backends (and `change_working_dir` already re-homing rather than restarting), an attach now re-points the window showing that project, keeping its buffers, splits and layout, or opens one there. **Every other window keeps the machine it was on.** A window's terminals move with it, reusing each backing file so scrollback continues — which the restart could not do at all.

`pending_authority`, `request_restart`, `take_restart_dir`, `should_restart`, `rebuild_editor` and main.rs's rebuild loop go with it, and `should_quit` means what it says.

### ⚠️ One plugin contract changed

`setAuthority` promised the calling plugin a reload ("follow-up work belongs in its post-restart init code"). Nothing is torn down now, so nothing reloads — **follow-up work belongs in an `authority_changed` handler**. In-tree, devcontainer and env-manager are updated; the orchestrator never depended on it. Worth a look if you have out-of-tree plugins calling `setAuthority`.

## How it behaves

- Reads only. Runs no other program unless asked, and cannot at all on a machine opened read-only.
- Bounded prefix reads, never whole files; the newest `maxSessions` of a store, capped *before* the read rather than after it; never copies prompts or agent output into a result; does not read credential files.
- One tool changing format skips that tool, not the scan — a scanner that throws becomes a `problem` against its own id.
- Every problem is shown, not just the first; and only problems are — a tool that is absent, or cannot be here, is said separately.
- A conversation and the pane running it are one row, linked, not two.
- Rejoining goes through the New Workspace form rather than launching silently: that is where the editor already shows what it is about to run and on which machine, probes the path, and — for a machine with no window — dials it with the same trust decision in front of it.

## Not waiting forever

Scanning a machine means asking it questions it may not answer, and this pass went through every wait on that path. Three of them had no bound at all.

- **The carrier handshake.** The code said a timeout was unnecessary because every failure ends in ssh exiting and an EOF — true of every failure but one. A host that completes the TCP connect and then says nothing leaves ssh sitting there, so the scan sat on "scanning" indefinitely and the `ssh` child outlived the dialog that started it. Now `ConnectTimeout` (OpenSSH applies it to the banner exchange too), `ServerAliveInterval` for a carrier that goes quiet after connecting, and a 30s deadline around the handshake itself — covering *both* halves, since pushing the agent source through a slow link stalls in the write.
- **Streamed replies.** `walk_entries` and remote `exec` read with no deadline, unlike unary requests' ten seconds. Both are bounded by the *agent's* silence now, not by total time: a walk streams by nature, and `exec` forwards output as it appears and sends a keepalive chunk while a quiet process runs, so a legitimately long or quiet command is never cut off, and on expiry the request is cancelled at the agent rather than merely abandoned here.
- **Work nobody is waiting for.** Closing the dialog cancelled nothing — `FileSystem::walk` has always taken a cancellation flag and nothing ever passed it one that anybody set. The flag now belongs to the machine handle, `ScanOptions.onStarted` hands the caller a way to trip it, and the wait itself wakes on cancellation rather than only on data.

Measured against a machine whose proxy delays every chunk a quarter-second: a full scan takes 27s and is not cut off, and Escape mid-scan releases the carrier in 2s where before it was never released. Against a host that accepts and never answers: a clear failure in 10s, with no child left behind. A reconnect after a dropped link keeps the carrier it dials: the reconnect task owns the process behind the transport it installs, which a test pins with a real kill-on-drop process.

The editor thread was already clear of all this — with a dead host mid-scan, Tab still moves focus, Esc still closes the dialog, typing still reaches the buffer, and scanning another machine returns in under a second. One path could have reached it and no longer does: the keepalive's runtime was dropped under the UI, and dropping a `tokio::runtime::Runtime` joins its workers.

## Also in this diff

- **Web UI:** the tree layer's `.tree-scrim` is now recognised as chrome by the pointer router, so a click over the dock while a modal is up is swallowed by the modal instead of reaching the dock. The Playwright harness gains one assertion for the dock's `⋯` menu (below); the rest of its dock rewrite is master's.
- **Dock `⋯` menu:** its rows are padded to the widest, so the narrow column no longer clips them (`Discover agent ses`). It also gains the **Everything…** entry.
- **Dialog action rows:** the footer buttons of the orchestrator's dialogs are aligned the same way.

## Verified by hand

Driven interactively in tmux: against a seeded `$HOME`, against the repo's fake `kubectl`, and against a real `sshd` — a second account on the box whose agent stores exist only on its side of the connection, reached three ways:

| target | what it is | result |
| --- | --- | --- |
| `agentbox@localhost:2222` | a real sshd, its own `$HOME` | finds the remote-only Claude Code and Codex rows; the local rows are absent, so nothing was read locally |
| a port that accepts and never sends a banner | host up, sshd wedged | fails in 10s with "SSH could not connect", no child left behind |
| the same sshd behind a 0.25s-per-chunk `ProxyCommand` | a slow link | same rows, ~27s instead of ~10s, UI responsive throughout, and not cut off by any deadline |

The read-only rule was measured rather than assumed, by pointing both machines at one tmux server: **Local** (trusted) reports the session by name, `payments-shell`; the **same server over the transport** can only report the socket, `agentsock`, because `tmux list-sessions` is refused there.

The journeys were then walked end to end rather than reasoned about, which is what found the bugs worth having found:

- **Rejoin a detached tmux session.** Typed its name, Enter, Create — a tab showing that session's own output, with `tmux list-sessions` reporting `attached=1`. Rejoined, not restarted.
- **Rejoin work on another machine.** This is where the form's Project Path turned out to be a *different field per backend*: the prefill always wrote the local one, so an SSH machine silently dropped the directory and showed its own default as placeholder text that reads exactly like a filled field. And where a machine the reader was *attached to* turned out to seed the form as Local, because "no machine id" was being read as an answer; the form is now seeded by a value that has no null.
- **A session with no exact resume.** `DiscoverVerb.exact` existed "so the dialog can admit it" and nothing admitted it.
- **Find something and press Enter.** The cursor landed on the group heading above the hit, where Enter means fold — and folding means nothing while a filter is on, so it was a dead key on the one journey this dialog exists for.
- **Two clones of one repository**, at unrelated paths and on different branches: one `payments` heading with both, each row carrying its branch; a linked worktree of this repo resolves to the same identity as the main checkout. **Two tmux servers** each with a session `0`: two rows, where they had collided into one.
- **Four machines at once**, one of them dead and one of them slow: one project heading holding the same session from two hosts, each labelled; a filter searching across all of them; the dead machine as an attributed problem line rather than a lost scan; and Enter on each row opening the form on that row's own machine.

## Docs

- `docs/internal/remote-authority-trust.md` §1.4 — the attach path and why there is no longer a restart.

## Known gaps

- **A read-only machine cannot report its environment**, because that would be a spawn. Every scanner that locates its store through a variable (`CODEX_HOME`, `SCREENDIR`, `TMUX_TMPDIR`, `XDG_RUNTIME_DIR`, `ORCA_USER_DATA`, `XDG_DATA_HOME`) therefore reads the default path on such a machine. The scan says this once per scan instead of letting each tool read as "not installed"; lifting it means deciding that `printenv` is not "running something", which is a trust decision rather than a bug fix.
- **Worktree directories are not scanned** — `~/.codex/worktrees`, `~/.herdr/worktrees`, `~/.cursor/worktrees`, and Claude's in-repo `<repo>/.claude/worktrees/`. They are the largest stores by an order of magnitude, and whether a checked-out worktree with no live agent counts as a "session" is a product question.
- **Live multiplexer listing on an unattached machine** — a machine opened by transport is trusted to run nothing, so `tmux list-sessions` there is refused by design. File-based discovery is unaffected.
- **The Machines dialog's action row is not flush with its frame.** A wrapping row justifies against its own extent, which the enclosing column stretches to the panel only when a sibling does not fix it narrower — and that dialog's list does. Making it flush needs a full-width `List` in the widget layer, which is a core `WidgetSpec` change; every other dialog's footer is flush because its fields already stretch the column.

https://claude.ai/code/session_0197TDEGGPABZLDYVwKqQFYQ


sinelaw:
## `playwright suite` is red here, and it is not this PR's

**177 passed, 5 failed** on `59d22d8` ([job](https://github.com/sinelaw/fresh/actions/runs/35251025195/job/105303118203)). All five are `master`'s, exposed rather than caused by this branch.

```
FAIL dock toggle click flips real plugin state  null -> null
FAIL New Task dropdown rows are items below the trigger  {}
FAIL New Workspace dialog is a centered floating panel
FAIL a dimming scrim sits behind it
FAIL a click over the dock is eaten by the modal  {"stillModal":false,…}
```

### Why it is not this branch's

`web-ui.yml` is path-filtered on `web-ui/**`, `src/webui/**`, `scene.rs`, `scene_parity.rs` and `webui_server.rs`. This branch had touched none of them until `59d22d8` deleted the dead `rebound_editor` from `src/webui/mod.rs` — so **this is the first time the suite has run on this PR at all**, and the first time it has run anywhere since `a3e6722` (2026-09-13), which was green.

In between, `6f71400` — *"orchestrator: the dock and its dialogs, rebuilt on the widget library"*, already an ancestor of this PR's base `41fe3b9` — rebuilt the dock and left `web-ui/test/drive.mjs` aiming at the old one:

| what `drive.mjs` looks for | at `a3e6722` (green) | on `master` today |
| --- | --- | --- |
| `findLine(dock, /New Task/)` | `dock.new_btn` = `"New Task… ▾"` | gone; the trigger is `dock.new` = `"+ New"` |
| a dock row matching `/^\s*New Task…\s*$/` | `dock.new_menu_task` = `"New Task…"` | gone — there is no create dropdown any more |
| a dock row matching `/empty/i` | a dock row | moved into the `⋯` menu as `dock.show_empty` |

`findLine` returns null, the harness clicks nothing, and the four downstream assertions fail on a dialog that was never opened. The fifth (`null -> null`) is the same shape: no row to toggle.

Read off the catalogs rather than inferred:

```
$ git show a3e6722:…/orchestrator.i18n.json   → dock.new_btn = 'New Task… ▾', dock.new_menu_task = 'New Task…'
$ git show origin/master:…/orchestrator.i18n.json → dock.new = '+ New', no 'New Task' string anywhere
$ git merge-base --is-ancestor 6f71400 41fe3b9 → yes, already in this PR's base
```

### Why I have not fixed it here

It is not a label swap. `+ New` is now a `button` that opens the New Workspace form **directly**, so the two-step *trigger → pick "New Task…" → modal* no longer exists, and the `[dock dropdowns are layers over the column]` block has no dropdown left to assert against — it would need re-aiming at the `⋯` menu. That is a rewrite of three blocks of someone else's harness to match their dock, and it belongs with that change rather than inside a PR about agent-session discovery.

Shape of the fix, for whoever takes it:

* `[dock clicks are cells routed over the tree]` — drive `show empty` through the `⋯` menu (`main:empty`) instead of looking for a dock row.
* `[dock dropdowns are layers over the column]` — assert against the `⋯` menu's options, which is the dropdown the rebuilt dock actually has.
* `[New Workspace dialog is modal …]` — one click on `+ New` (`key: "new-session"`); drop the option-picking step.

No re-run spent: these are deterministic selector misses, not flakes, and a re-run cannot change them.

### The rest of this head

Green: `clippy`, `fmt`, `schema`, `doc`, `check (no plugins)`, `rust 1.98`, `installer`, `actionlint`, `supply-chain audit`, `build-musl`, and `scene parity (web/TUI)` — the other half of this same workflow, which passes.

Still running as of this comment: `test ubuntu-latest`, `test macos-latest`, `test windows-latest`. Those carry `orchestrator_discover_sessions`, which is what actually covers the two commits on this head; they were green on the previous head `3ae2849`. I will report their result rather than assume it.

---

sinelaw:
## Update: the stale-harness failure is fixed; a different one is left, and it is a real finding

[My earlier comment](https://github.com/sinelaw/fresh/pull/3304#issuecomment-5718445200) said `playwright suite` was failing on master's stale selectors and not this PR's to fix. The first half was right; the second is now out of date, and a reviewer reading it would be misled. Superseding it.

### Fixed

`b4fe3a1` re-aimed `web-ui/test/drive.mjs` at the dock `6f71400` actually built — `+ New` opens the New Workspace form directly, the `⋯` menu is the dock's one dropdown, and "show empty" lives inside it. Four of the five failures went with it, and **CI confirmed the suite green on that head**. That block also now asserts the `⋯` rows are not clipped, which is the regression that produced `Discover agent ses`.

### What is still red — one assertion, and it is not a flake

```
FAIL a click over the dock is eaten by the modal (row unchanged, modal still open)
{"stillModal":true,"before":"fresh","after":"fresh",
 "bgBefore":"#000000","bgAfter":"#0064c8",
 "clickedAt":{"x":4,"y":3},"yAfter":3,
 "modalRect":{"x":15,"y":2,"w":91,"h":31}}
```

Failed on CI on every head since `48b89eb`, and passed locally every time (183/0, three runs). Two commits added *only* diagnostic detail to that assertion's report — no condition was changed — and between them they settle it:

* `yAfter == clickedAt.y`, so the premise holds: it is the same row, not a re-ordered dock.
* `bgAfter` is `#0064c8`, the **selection** colour. The dock row went from unselected to selected — so the click was not eaten.
* `clickedAt.x` is 4; the modal spans x 15→106. The click lands on the dock, with only the scrim over it.

So the scrim is not absorbing clicks over the dock, which is exactly what this assertion exists to check.

### Why it passes locally, and why that matters

Most likely **the assertion is vacuous here**: the `/fresh/` row is already selected before the click on my machine, so its background does not change and the check passes without testing anything. On the runner the row starts unselected and the same behaviour becomes visible. If that is right, this is a pre-existing scrim hit-testing bug that fixing the harness made reachable again — not something the button-alignment change caused.

The one fact that resists that reading is that the suite passed on `b4fe3a1` and started failing at `48b89eb`, which touched only dialog action rows. So something about the dock's default selection may also have shifted, and I have not proven which.

### What blocks fixing it

A decision, not effort. Three options, none of which I want to take unilaterally on someone else's assertion:

1. **Confirm the vacuity** — print the values on success too and run it once here. Decisive.
2. **Fix the scrim's hit-testing** — the real bug, but shared web-UI behaviour well outside this PR's subject.
3. **Pin the row by key and assert the scrim explicitly** — makes the assertion honest either way, and it will then fail loudly until (2) is done.

I have not weakened the `bg` comparison to get green, and will not: it is the term doing the work.

### Everything else on this head

Green: `clippy`, `fmt`, `doc`, `schema`, `check (no plugins)`, `rust 1.98`, `installer`, `actionlint`, `supply-chain audit`, `scene parity (web/TUI)`. The three platform test jobs are running; the previous head's full matrix was **9856/9857 on ubuntu and 9800/9801 on macOS**, the single failure being a plugin-i18n completeness check for three strings I had added in English only — translated for all fourteen other locales in `7ecfe6e` and verified locally.

---

sinelaw:
**`playwright suite` — `a click over the dock is eaten by the modal` is unstable on CI.** This supersedes my earlier comments on it, one of which was wrong about the cause.

Everything else is green: clippy, fmt, doc, schema, check, rust 1.98, scene parity, installer, audit, actionlint, and the Rust matrix.

**What it found, which was real.** `onChrome` in `web-ui/js/80-input.js` is the one function deciding whether a pointer event is chrome or gets forwarded to the editor as a cell click. It listed the palette's `.modal-scrim` and not the tree layer's `.tree-scrim` — the dim drawn behind every centered floating panel, given `pointer-events:auto` for exactly this purpose. So a click over the dock hit the scrim, was not recognised as chrome, fell through to `cellAt`, and reached the dock: selecting the row the modal was meant to be covering. Fixed in fcba4ce, and worth having regardless of the test.

**What is left.** With that fix in, the assertion passed on `4253a6e` and failed on `5bee7de` — and `5bee7de` changes only Rust test code, so the same web-UI ran both times. It is therefore flaky rather than deterministically red. Locally it passes (183/0, repeatedly).

The diagnostics I added to the assertion rule out the two explanations I had. It is not geometry: `clickedAt {x:4,y:3}` against `modalRect {x:15,y:2,w:91,h:31}`, so the click is nowhere near the dialog either way. It is not the dock's own asynchronous repaint arriving on the clicked row: the baseline is now taken only after two identical reads of the dock, and `selBefore: []` / `selAfter: [1,3,3,3]` says nothing carried the selection before the click.

One environmental difference is visible and unexplained: the click lands on column 4 on CI and column 6 locally, because `clickLine`'s offset is `min(w-1, 2)` and the matched row item is one cell wide there and wider here. So the two runs are clicking different elements of the row, which would matter if one of them is a layer item (`z-index: 2`, later in DOM order than the scrim) and the other is not.

I have stopped here at the repo owner's direction rather than because it is finished. Anyone picking it up: the question to answer is what `e.target` actually is for that click on CI. Leaving the assertion as it is — it is testing something true, and weakening the background comparison would hide the defect it correctly caught once.

---

crates/fresh-editor/src/services/remote/connection.rs
```diff
@@ -644,6 +726,9 @@ async fn establish_ssh_transport(
     cmd.stdin(Stdio::piped());
     cmd.stdout(Stdio::piped());
     cmd.stderr(Stdio::null()); // No terminal for reconnection
+                               // Same as the initial connect: if whatever awaited this reconnect goes away
+                               // mid-handshake, the carrier goes with it rather than being orphaned.
+    cmd.kill_on_drop(true);
```
sinelaw, line 706: **`kill_on_drop(true)` SIGKILLs every *successful* reconnect carrier, so an SSH session can never recover.**

This was added (cb9d58e) "so a reconnect that is abandoned mid-handshake doesn't orphan the carrier", but its only caller is the reconnect factory around `connection.rs:395`:

```rust
let (reader, writer, _child) = establish_ssh_transport(&params).await?;
...
Ok::<_, SshError>((reader, writer))
```

`_child` is a *named* binding, so it drops when that async block completes — on the success path too, before `spawn_reconnect_task_with` reaches `channel.replace_transport(reader, writer)`. tokio's `ChildDropGuard::drop` calls `kill()` when `kill_on_drop` is set, so the freshly handshaked `ssh` is SIGKILLed. On master, dropping the `Child` without `kill_on_drop` was a no-op, which is why reconnect worked.

Failure: laptop sleeps on an SSH workspace. `is_connected()` goes false, the reconnect task backs off, dials, the agent sends its ready line — and the carrier is killed on the way out of `connect_fn`. `replace_transport` installs a reader already at EOF, `read_task` immediately sets `connected = false`, and the loop restarts. The remote session never comes back, and the editor spawns + kills a real `ssh` login (with auth) every 1–30s for as long as the window is open.

Not covered: the `spawn_reconnect_task_with` tests use a fake `connect_fn`.

Fix: keep the child alive for the life of the swapped-in transport (return it to the reconnect task and park it beside the connection), or clear kill-on-drop on the success path.

---

crates/fresh-editor/src/services/remote/spawner.rs
```diff
@@ -521,31 +521,66 @@ impl ProcessSpawner for RemoteProcessSpawner {
...
+        // run for as long as it likes and stream as it goes; what must not be
+        // unbounded is waiting on a machine that has stopped saying anything.
+        // Every chunk resets it, so a slow command is not a cancelled one.
+        //
+        // Unary `request` has always had its ten seconds. This path had
+        // nothing at all, which is why a scan could sit on a single
+        // `tmux list-sessions` for as long as the editor ran.
+        loop {
+            match tokio::time::timeout(
+                crate::services::remote::channel::STREAM_IDLE_TIMEOUT,
```
sinelaw, line 537: **The 45s idle bound on `exec` fires on healthy remotes, because the agent does not stream output the way the bound assumes.**

The premise — "every chunk resets it, so a slow command is not a cancelled one" — does not hold against `agent.py`'s `cmd_exec` (~`agent.py:494-520`): after `select()` reports the pipe readable it calls `fd.read(4096)` on a `BufferedReader`, which blocks until **4096 bytes or EOF**. So a remote command producing less than 4 KB total streams *nothing at all* until it exits; one producing nothing streams nothing by construction.

Failure: a plugin calls `runOnTarget` on a remote machine with something like `sh -c 'make test'` that runs 90s and prints 600 bytes at the end, over a perfectly healthy link. At t=45s the client gives up with `ChannelError::Timeout`, and the `cancel_detached(request_id)` makes the agent's `cmd_cancel` `proc.terminate()` the process — so the user's remote command is actually **killed**, and the plugin is told the machine timed out. Before this change the call simply waited and succeeded.

Either the agent needs `read1(4096)`/non-blocking reads so partial output actually reaches the client (which is what makes "idle, not total" true), or `exec` needs a much larger or caller-supplied bound. The same `STREAM_IDLE_TIMEOUT` on `walk_entries` is fine — that one really does stream per chunk.

---

crates/fresh-editor/src/app/plugin_commands.rs
```diff
@@ -3391,6 +3392,447 @@ impl Editor {
...
+    ) {
+        use crate::services::authority::SessionAuthoritySpec;
+
+        let Some(authority) = self.authority_for(machine, callback_id) else {
+            return;
+        };
+
+        // A local machine's environment is this process's own, so it is a map
+        // lookup and answers on the editor thread.
+        if matches!(authority.session_spec(), SessionAuthoritySpec::Local) {
```
sinelaw: **`machineEnv` returns the *host editor's* environment for a container (docker-exec) machine.**

`Authority::session_spec()` (`services/authority/mod.rs:449-492`) derives the spec from `command_wrap`, and **`CommandWrap::Direct | CommandWrap::Prefix(_) => SessionAuthoritySpec::Local`**. `CommandWrap::Prefix` is exactly what `from_plugin_payload` builds for `SpawnerSpec::DockerExec` — i.e. every devcontainer session (`devcontainer.ts:1626` installs it via `setAuthority({ kind: "docker-exec" })`). `openMachine`'s own docs advertise the payload form as the way to reach "a local **or container** backend".

So on a devcontainer window this fast path reads `std::env::var` from the *editor process*, while `runOnTarget`/`machine.run` on the same handle executes **inside the container** via the docker-exec prefix.

Failure: `agent_sessions_tmux.ts:81` reads `TMUX_TMPDIR` then scans/`run`s `tmux` on that machine; `agent_sessions_multiplexers.ts:36` reads `USER`/`LOGNAME` to build the screen socket dir. In a devcontainer the container user is typically `root`/`vscode` and the host user is not, so the scanner builds `/run/screen/S-<host-user>` and finds nothing, or looks in the host's `TMUX_TMPDIR` inside the container.

This is precisely the invariant the generated `.d.ts` promises: *"Never this computer's values for another machine."* The fast path should be gated on `CommandWrap::Direct` only.

---

crates/fresh-editor/src/app/plugin_commands.rs
```diff
@@ -3391,6 +3392,447 @@ impl Editor {
...
+                    return;
+                }
+            };
+
+        let id = self.next_machine_id;
+        self.next_machine_id += 1;
+        let info = Self::machine_info(&authority, id);
+        let connection =
+            self.open_connection(crate::services::authority::Connection::plain(authority));
+        self.open_machines.insert(
```
sinelaw, line 3538: **`openMachine` handles are never released on plugin unload/reload — leaked `ssh`/`kubectl` carriers.**

`open_machines` is only ever mutated by `handle_open_machine` (this insert) and `handle_close_machine` (remove) — a grep over the tree shows no other write. Every other plugin-owned resource of this shape *is* tracked for unload cleanup via `PluginTrackedState` (`quickjs_backend.rs:676-706`: background processes, terminals, watch handles, timers, virtual/composite buffers) and released by the unload path. `OpenMachine` was not added to it, and `PluginCommand::OpenMachine` carries no `plugin_name` to attribute it to.

Failure: `agent_sessions.ts` opens a machine by transport spec, which registers a `Connection` holding a live `Keepalive` over an `ssh`/`kubectl exec` child. The user edits and reloads the plugin (or the plugin throws and is unloaded) before `close()` runs. The handle id lived only in the JS heap that was just discarded, so nothing can ever call `closeMachine`: the entry stays in `open_machines` forever, `prune_connections` sees a live borrower, and the child stays up for the life of the editor — one more per reload. There is also no `Drop for Editor` path that drains it.

Given the scan's `finally { machine.close() }` this is mostly a plugin-development hazard rather than an end-user one, but it is unbounded and invisible.

---

crates/fresh-editor/plugins/orchestrator.ts
```diff
@@ -9663,6 +9744,1137 @@ function machineForHost(h: SshConfigHost): Machine {
...
+  }
+  // **The row's machine, not the picker's.** Under "All machines" the
+  // selected entry is not a machine at all, and even under one machine the
+  // reader can change the picker after a scan. A session is rejoined where it
+  // lives — the row is the only thing that knows where that is.
+  const targets = discoverTargets();
+  const target = targets.find((t) => t.key === machineKey)
+    ?? targets[st.machineIndex]
+    ?? null;
+  pendingFormMachine = target?.machineId ?? null;
```
sinelaw: **Rejoining a session discovered on a *remote* machine opens the New Workspace form on **Local**, with the remote path in the local Project Path field.**

`DiscoverTarget.machineId` is set in exactly one place — the `for (const m of loadMachines())` loop in `discoverTargets()` (~line 10004). The targets built by the *first* loop (~9957-9984, one per machine that has an open window) omit `machineId` entirely, and that loop is the **preferred** path: the second loop does `if (byMachine.has(machineFacetKey(m))) continue;`, so a saved machine you are currently attached to is *only* ever represented by a window-borrowed target with no `machineId`.

Failure: save machine `buildbox` (ssh `me@buildbox`), attach a workspace to it so it has an open window. Everything → the picker offers `me@buildbox · ssh` → Scan → Enter on a `claude` row whose cwd is `/srv/work/api`. `target.machineId` is `undefined`, so `pendingFormMachine = null`, which `seedFormMachine` reads as "explicitly Local" and returns without applying any machine pick (it also suppresses the `orchestrator.last_machine` fallback). `applyPrefillPath` then takes the `default:` branch and writes `/srv/work/api` into `form.projectPath`, the *local* slot. Submitting creates a local workspace rooted at a directory that does not exist here, running `claude --resume <uuid>` there — which cannot find the conversation.

Same class as 5cfdd71 ("Put the rejoined session's directory where the form will read it"), which fixed it for saved-machine targets; the window-borrowed ones were missed. The window target knows its session's `remote`, so it can carry the matching machine id.

---

crates/fresh-editor/plugins/orchestrator.ts
```diff
@@ -9663,6 +9744,1137 @@ function machineForHost(h: SshConfigHost): Machine {
...
+          };
+        }
+      }),
+    );
+    // **Whose answer is this?** An abandoned scan still settles — being
+    // abandoned is what makes it settle early — and by then the dialog may be
+    // closed, or reopened onto a different machine. Writing into the state it
+    // was started against would repaint the dialog with a stale answer under a
+    // new machine's name, which is the one mistake this dialog must not make.
+    if (discoverState !== st) return;
```
sinelaw: **Changing the machine picker while a scan is in flight makes the scan's answer land under the *other* machine's name — the exact mistake the comment two lines above says must not happen.**

The guard is identity of the `DiscoverState` **object**. But `handleDiscoverEvent` mutates `st.machineIndex` **in place** on a `change` from `discover-machine` — the object is never replaced — and there is no `st.scanning` guard there, no `discoverStop?.()`, and nothing locks the picker while scanning (`refreshDiscoverDialog()` only moves focus).

Failure: select `buildbox`, Scan (slow host, ~20s). While it runs, Tab back to the picker and switch to `laptop`. `loadDiscoverCache()` swaps `st.scans` to laptop's cache and reflows. When buildbox's `Promise.all` resolves, `discoverState === st` still holds, so it writes `st.scans = answers` (buildbox's sessions) and caches under `target.key = machine:buildbox` — while the dropdown reads `laptop` and the age label beside it is computed from `discoverCache.get(selected.key)` with `selected` = laptop. The reader sees buildbox's sessions labelled `laptop`, "scanned 4m ago". Nothing on screen contradicts it: a single-machine scan has `manyMachines === false`, so the rows carry no machine column either. Then Enter hands the form a directory that only exists on the other host.

Fix: capture `target.key` (or `st.machineIndex`) at scan start and compare that too, or call `discoverStop?.()` on a picker change.

---

crates/fresh-editor/plugins/agent_sessions_claude_code.ts
```diff
@@ -0,0 +1,161 @@
...
+      problems.push(`${projects}: stopped at the walk limit, so some sessions are not reported`);
+    }
+
+    const transcripts = walk.entries.filter((e) => e.rel.toLowerCase().endsWith(".jsonl"));
+    // Newest first before any cap, so what survives is the newest rather than
+    // whatever sorted first by path.
+    transcripts.sort((a, b) => b.mtime - a.mtime);
+
+    const prefixes = await machine.readFilePrefixes(
+      transcripts.map((e) => ({ path: e.path, maxBytes: HEADER_BYTES })),
```
sinelaw: **Every transcript header is read before any cap is applied, so one large store burns the hub's whole shared 60s budget and starves every scanner after it.** (Same bug at `agent_sessions_codex.ts:115` and `agent_sessions_catalog.ts:133`.)

The scanner sorts newest-first "before any cap" and then reads **all** `transcripts` — up to the walk's 5001 entries at `HEADER_BYTES = 256 KB` each. The only cap is `maxSessionsPerTool` (default 200), applied by the hub *after* `await scanner.scan(...)` returns, and `ScanContext` carries only `allowCommands`, so a scanner cannot learn the cap. `read_file_prefixes` (`plugin_offloop.rs:186`) is a serial loop — one `read_range` per file, i.e. one round trip per file on a remote machine.

Failure: a developer with a year of Claude Code use (a few thousand `<bucket>/<uuid>.jsonl`) opens Everything against an SSH machine. This scanner issues ~3000 sequential ranged reads of up to 256 KB. It does not finish inside `SCAN_BUDGET_MS`, so the race discards the *entire* report — zero Claude Code sessions — and because the budget is shared and scanners run in alphabetical id order, every scanner after `claude-code` (`codex-cli`, `cursor`, `gemini`, `goose`, `herdr`, `orca`, `screen`, `tmux`, `zellij`, …) takes the `remaining <= 0` branch and reports "not asked — the scan ran out of time first" with `installed: false`.

Net effect on a machine where everything is installed: an empty dialog plus ~20 problem rows.

The 200-row cap is knowable before the read. Slice `transcripts` before `readFilePrefixes` — and since the hub owns the cap, the cleanest fix is to put it on `ScanContext` and share one `readHeaders(machine, entries, { maxBytes, limit })` helper (see the duplication note on `lib/agent_scanner.ts`).

---

crates/fresh-editor/plugins/agent_sessions_tmux.ts
```diff
@@ -0,0 +1,239 @@
...
+    problems.push(`tmux -S ${socket} list-sessions: ${list.stderr.trim() || `exit ${list.code}`}`);
+    return null;
+  }
+
+  const sessions = new Map<string, CollectedSession>();
+  for (const line of list.stdout.split("\n")) {
+    if (!line.trim()) continue;
+    const [id, name, created, attached, path] = splitEscapedFields(line);
+    if (!id) continue;
+    sessions.set(id, {
```
sinelaw, line 91: **Session ids are not qualified by socket, so two tmux servers mint colliding `tool/id` rows.**

`socketPaths` returns every socket under `/tmp/tmux-<uid>/`, and `liveSessions` builds `id: name || id` from `session_name` alone (and pane rows as `${session.id}${paneId}`). The socket appears only in `attach.args`.

Failure: a user runs plain `tmux` (socket `.../default`) and `tmux -L work` (socket `.../work`); both servers have a first session named `0` and a pane `%0`. The scan emits two sessions with `id: "0"` and two panes with `id: "0%0"`, all under `tool: "tmux"`. The consumer keys rows as `session:${machineKey}/${session.tool}/${session.id}`, whose own comment says *"two rows with one key make the tree fold and select the wrong one"* — so the two servers collide in exactly the way that key was designed to avoid, and Enter can attach to the wrong server's socket. The hub's `disambiguate` only relabels **titles** (`0 (1)` / `0 (2)`); the ids, and so the row keys and the `tool/id` link endpoints, stay identical.

Qualifying the id with the socket basename would fix it at the source.

Same class, lower confidence because it depends on an external layout: `sessionIdFor("file-stem", rel)` in `lib/agent_catalog.ts:314` uses only the basename, so a store with a fixed filename in per-session directories — Gemini CLI's `~/.gemini/tmp/<hash>/logs.json`, walked at `maxDepth: 4` — gives every project's row the id `logs`.

---

crates/fresh-editor/plugins/lib/agent_scanner.ts
sinelaw, line 1: **Two load-bearing duplications that belong here, and that are already drifting.** This file exports only `joinPath`, `baseName`, `isAbsoluteOn`, `firstJsonRecord`, `jsonRecords`, `pickString`/`pickStringFrom`, `hub`, `registerScanner` — neither helper below exists.

**1. The "walk → filter → sort newest-first → `readFilePrefixes` → zip by index → build `CollectedSession`" pipeline is copy-pasted five times:** `agent_sessions_catalog.ts:81` (`scanStore`), `agent_sessions_claude_code.ts:77`, `agent_sessions_codex.ts:74`, `agent_sessions_herdr.ts:122`, `agent_sessions_orchestrators.ts:69` (orca).

This is not stylistic. The starvation bug above is present in **three** of the five copies and must be fixed three times. The copies already disagree on error policy: catalog/codex/claude-code push `prefix.error` and keep the row; orca pushes and drops it; herdr pushes and drops; and only `lib/project_identity.ts:308` defends against a short result array (the Rust side truncates with `take_while` on cancellation, so a short batch is reachable for all of them). One `readHeaders(machine, entries, { maxBytes, limit })` here would fix the cap, the short-batch pairing and the error policy in one place.

**2. Env-override store resolution** ("read the variable, reject a relative value, else default under `$HOME`") exists **eight** times: `agent_sessions_catalog.ts:47`, `agent_sessions_codex.ts:58`, `agent_sessions_tmux.ts:81` (`TMUX_TMPDIR`), `agent_sessions_multiplexers.ts:39` (`SCREENDIR`) and `:121` (`XDG_RUNTIME_DIR`), `agent_sessions_orchestrators.ts:73` (`ORCA_USER_DATA`/`XDG_DATA_HOME`), `agent_sessions_herdr.ts:104` (`XDG_CONFIG_HOME`), `agent_sessions_claude_desktop.ts:36` (`XDG_CONFIG_HOME`/`APPDATA`).

Already drifted: the first five push a "set to a relative path; ignoring it" problem, while herdr, claude-desktop and orca fall back **silently**. So with `XDG_CONFIG_HOME=config` exported, the catalog and codex scanners explain themselves and Herdr/Claude Desktop just report "not installed" with no problem line — which is precisely the failure mode the `problems` channel exists to prevent.

The per-tool split is the right architecture; these two are the parts that are *not* per-tool knowledge and shouldn't have been split with it.

---

sinelaw (commented):
Automated review of this PR's diff (`41fe3b9...d26c8b1`, 80 commits, 94 files). Nine inline findings, most severe first:

| # | where | what |
| --- | --- | --- |
| 1 | `remote/connection.rs:731` | `kill_on_drop(true)` SIGKILLs every **successful** reconnect carrier — remote sessions can never recover, and the editor re-dials every 1–30s forever |
| 2 | `remote/spawner.rs:543` | the new 45s idle bound fires on healthy remotes and **kills the user's command**, because `agent.py`'s `fd.read(4096)` streams nothing until 4 KB or exit |
| 3 | `plugin_commands.rs:3688` | `machineEnv` takes the "Local" fast path for `CommandWrap::Prefix` (docker-exec), returning the *host editor's* environment for a container machine |
| 4 | `orchestrator.ts:10844` | window-borrowed discover targets carry no `machineId`, so rejoining a session found on a remote machine opens the form on **Local** with the remote path in the local field |
| 5 | `orchestrator.ts:10753` | the staleness guard compares object identity while the picker mutates `st.machineIndex` in place — results land under the other machine's name |
| 6 | `agent_sessions_claude_code.ts:102` | transcript headers are read before any cap, so one large store exhausts the shared 60s budget and every later scanner reports "not installed" (same bug in 3 copies) |
| 7 | `agent_sessions_tmux.ts:140` | session ids omit the socket, so two tmux servers produce colliding row keys and Enter can attach to the wrong one |
| 8 | `plugin_commands.rs:3571` | `openMachine` handles are absent from `PluginTrackedState`, so a plugin reload leaks the handle and its `ssh`/`kubectl` carrier permanently |
| 9 | `lib/agent_scanner.ts` | the scan pipeline is copy-pasted into five scanners and the env-override rule into eight; #6 exists in three copies, and the copies already disagree on error policy |

**One claim checked and rejected.** A reader flagged the new `remove_dir_all` in `DeleteWorkspace` as able to wipe the shared `terminals/` directory via an empty-encoded root. It cannot: `encode_path_for_filename` ends with `if final_result.is_empty() { final_result = "root" }` (`path_encode.rs:154`), so `""` and `/` both encode to `root` and `terminal_dir_for` yields `terminals/root`, never `terminals/` itself.

**Scope note.** An earlier pass of this review ran against a stale local `master` and reported six further findings — in `shell_host.rs` (prose-press focus, drag-left selection), `server/mod.rs` (fixed daemon name), `editor_init.rs` (`pick_active_window_globally` vs Workspace-Trust roots), `plugin_commands.rs` (`handle_insert_text` seen-byte-ranges), and `window_actions.rs` (`last_focused_at`). None of those symbols appear as added lines in this PR's diff; they are already on `master` and belong in their own issues, not here.

**On the architecture**, for the record: the direction is good and unusually well argued — `LiveRuntime` replacing bare `tokio::runtime::Handle` with a workspace-wide clippy deny, one `ConnectionRegistry` replacing per-window keepalive bundles, deleting the duplicate widget text projection, the unified key vocabulary. Each removes a whole class of drift. Against that, `orchestrator.ts` is now ~17.6k lines: the discovery work was correctly split into a hub plus per-tool plugins, but the dialog driving it went into the monolith.

---

crates/fresh-editor/src/services/remote/connection.rs
sinelaw: Fixed in 7ae5e9b. The representable-bad-state was the root: `connect_fn` returned `(reader, writer)` and the process had nowhere to go. It now returns a `Carrier { reader, writer, process }` and the reconnect task parks the process for as long as that transport is installed — replaced by the next reconnect, dropped on abort. Kube's bootstrap gets the same kill-on-drop now that its reconnect child is held too. Pinned by `test_reconnect_task_holds_the_carrier_process`, which stands a real kill-on-drop process in for the carrier: alive after the reconnect, dead after the task is aborted.

---

crates/fresh-editor/src/services/remote/spawner.rs
sinelaw: Fixed in 87d6a91, on the agent side — the bound was fine, its premise wasn't. `stream_output` now reads with `os.read` (so it actually streams), drops an EOF'd pipe from the poll set (it used to spin `select()` at 20Hz), and while the process is quiet sends an empty chunk every `EXEC_KEEPALIVE_SECS` = 10s, a third of the client's 45s. That is what makes "bounded by silence" a bound on the agent's silence rather than on the command's verbosity: a chunk means the agent is there, nothing means it isn't. The two constants are documented against each other on both sides.

---

crates/fresh-editor/src/app/plugin_commands.rs
sinelaw: Fixed in f507047. `session_spec()` was answering the wrong question — it says what to reconnect to on restore, and for Direct and Prefix that is legitimately "nothing". The predicate that was actually needed is new: `Authority::env_is_this_process()`, true only for `CommandWrap::Direct`. A prefix (docker-exec) now goes through the existing remote path, whose `printenv` runs through the authority's own spawner and so inside the container.

---

crates/fresh-editor/src/app/plugin_commands.rs
sinelaw: Fixed in 385616e, through the same machinery as terminals and watch handles rather than a special case: the runtime attributes the `openMachine` request to the calling plugin, the editor answers with a new `PluginResponse::MachineOpened` (all three open paths — window-borrowed, payload, transport), the thread records the id in `PluginTrackedState.machine_ids`, and unload sends `CloseMachine` for each. That close carries no callback — the promise that would take the answer is in the discarded heap — so `CloseMachine.callback_id` is now `Option` and the editor resolves it only when present.

---

crates/fresh-editor/plugins/orchestrator.ts
sinelaw: Fixed in a768345, and the fix is to the type rather than the one site: `pendingFormMachine` was `string | null | undefined` with `null` meaning Local, so a target with no id resolved to `?? null` and an absence read as an answer. It is now a `FormSeed` — an option key, an unsaved ssh target, or a cluster pod — and every real `DiscoverTarget` carries a `reach`, window-borrowed ones included: a saved machine when its facet matches the window's, else the window's own host or pod, which the form opens as "Other host…" with the host already typed. The picker fallback is gone too; a row whose machine is no longer listed says so instead of rejoining on whatever the picker shows.

---

crates/fresh-editor/plugins/orchestrator.ts
sinelaw: Fixed in a768345. The guard was comparing the wrong identity — the dialog's, when the question is the scan's. Each scan now mints a `scanToken` only it holds; a picker change while scanning calls `discoverStop`, clears `scanning`, and disowns the token, so the settling scan finds a token that isn't its own and writes nothing. That also stops the abandoned scan's connection rather than letting it run to completion for nobody.

---

crates/fresh-editor/plugins/agent_sessions_claude_code.ts
sinelaw: Fixed in 513db34, the way this thread suggested: the cap is on `ScanContext.maxSessions` (the hub passes its limit), and one `readHeaders(machine, entries, { maxBytes, limit, problems })` in `lib/agent_scanner.ts` sorts newest-first, caps *before* the read, reads one batch, and pairs by position when the batch came back whole and by path when it was cut short. All three copies that read before capping (claude-code, codex, catalog) are on it, as are herdr and orca. It says what it left out (`reporting the newest 200 of 3417`). Pinned by `tests/agent_scanner_pipeline.test.ts` against a fake machine that records the request list.

---

crates/fresh-editor/plugins/agent_sessions_tmux.ts
sinelaw: Fixed in 513db34. tmux ids are qualified by socket basename (`default/0`, `work/0`; panes follow). Verified live with two servers each holding a session `0`: they now render as two rows under one project heading, with different branches, where before their row keys collided. The catalog case is handled generically rather than by guessing Gemini's layout (the stores doc has that row as unverified): `scanStore` re-keys only the rows whose ids actually collide, by their relative path minus extension. Cline's row, whose `maxDepth: 1` could never reach the parent directory its `parent-dir-name` layout needs, is at depth 3 like the other rows of that layout.

---

crates/fresh-editor/plugins/lib/agent_scanner.ts
sinelaw: Done in 513db34. `readHeaders` replaces the five pipeline copies and `envDirs` the eight env-override copies; both live here. The split that stays is deliberate: whether a file with no readable header is still a *row* differs by format (a transcript is, a state file isn't), so that decision stays with each caller — everything about *reading* does not. The drift is gone with the copies: every relative override is now rejected out loud, and Orca reports a state file larger than the prefix instead of parsing the cut. `tests/agent_scanner_pipeline.test.ts` pins both helpers.

---

PR #3318, opened 2026-09-17
Fix terminals inside the editor, two markdown compose glitches, and two code tour bugs

sinelaw:
Six bug fixes. Five were reported; one was found on the way. Each was reproduced first, and each has a test that fails on the code before the fix.

## What changes for you

| What you saw | What happens now |
| --- | --- |
| A terminal opened inside Fresh didn't know it was inside Fresh, so `fresh somefile` in it opened a **second editor** instead of opening the file in the one already on screen. Typing plain `fresh` there filled the pane with **garbage**. | The terminal knows its session. `fresh somefile` hands the file to the editor you are already in, and a plain `fresh` opens an ordinary editor instead of shredding the pane. |
| Typing in a markdown header made the line **flicker** — it briefly drew differently and snapped back. | No flicker. Typing appends the character and nothing else on the line moves. |
| On a bullet line, typing `- a` and pressing backspace made the **cursor jump two rows down** and to the left edge until you typed again. | The cursor stays where you are typing. A bullet also no longer draws a spurious extra blank row under itself. |
| In a code tour, clicking the description panel moved its caret but **left the keyboard in the file** — you typed into your source by mistake. | Clicking the description panel focuses it, so your keystrokes go where you clicked. |
| In a code tour, the caret **disappeared** when it was at the end of a line in the description panel. | The caret stays visible there. |
| Word highlighting (the other occurrences of the word under the cursor) **updated late** — it only caught up on your next keystroke, which read as the whole line flickering. | It updates on its own, a moment after the cursor lands, with no input needed. This also unsticks the LSP spinner and the paste indicator, which were frozen for the same reason. |

Details below, one section per bug.

## A terminal inside Fresh had no `FRESH_SESSION`

Reproduced in tmux: open a terminal buffer and the shell in it has `FRESH_BIN` set but `FRESH_SESSION` empty. Two causes.

**The daemon never bound its local control socket.** The TUI, the GUI and the web UI all call `local_control::start()`; `EditorServer::run` only ever *pumped* it. The terminal spawner reads `local_session_id()` to decide whether to advertise `FRESH_SESSION`, so in a daemon-backed editor it advertised nothing. Orchestrator mode, on by default, made the daemon the ordinary way to run Fresh, which turned a long-standing gap into "every terminal, on master". Agent terminals were unaffected because `agent_command_env` calls `start()` for itself.

**The garbage that empty variable was hiding.** A bare `fresh` takes the Orchestrator launch and attaches to the shared daemon — and a terminal buffer *inside that daemon* is exactly where it can be typed. The daemon then renders one shared screen into a client living inside its own output: a feedback loop, not a second editor. `wants_orchestrator_launch` now declines when `FRESH_SESSION` says we are already inside a Fresh terminal, falling through to the ordinary inline editor, which renders correctly.

Three tests drive the real binary, because both halves are about what a separate process does: a terminal opened in a daemon reads a `local-…` session back from its own shell, a bare `fresh` with `FRESH_SESSION` set starts no daemon, and — the control — the same launch without it still does.

## Typing in a markdown header made it flicker

Two independent causes, found in that order.

**The occurrence highlight, which is not compose at all.** It happens in any buffer; turning `highlight_occurrences` off removes it, turning compose off does not. `ReferenceHighlightOverlay::update` runs *inside* a render and does nothing until its 150 ms debounce has elapsed, so the update lands on whatever frame something else happens to cause. Sitting still after a cursor move causes none: a real daemon, instrumented, still held `current=Word("HeadWing")` with the cursor sitting in `Heading` five seconds after it stopped moving. The next keystroke was the first thing to force a frame, so the highlight jumped as you typed.

Two things were wrong, and they are the same thing said twice. *A deferral has to ask for the frame it needs* — `next_periodic_redraw_deadline` is where the LSP spinner, the animations and the async-paste fallback already declare exactly that, and the highlight declared nothing; it does now, across every buffer in the window. *And every loop has to honour that declaration* — the TUI and the web loop wait on it, while the daemon had a hand-copied subset (an active animation, a wheel gesture), so every other time-driven effect simply did not happen there. That is why this was far worse in a daemon, and why the spinner and the paste fallback were stranded there too.

**And a real compose defect underneath it.** From the report: an empty `bla.md`, type `# a`, then one more character — the line is drawn with the `# ` concealed and recentred, then redrawn as it was. Captured in tmux: `ESC[3;29H "ab"` then `ESC[3;29H "# ab"`, where the fix leaves one frame appending `b`.

Compose's conceals carry an `unless-cursor-in` rule scoped to the line: hide the syntax while the cursor is elsewhere, show it while you edit. The scope was stored as a **fixed length** relative to the decoration's anchor, on the reasoning that an edit inside the line re-fires `lines_changed` and the plugin rebuilds the conceal anyway. It does — but only after a round trip through the plugin thread, and frames are rendered before that lands. In between, the text had grown and the scope had not, so a cursor at the end of what it had just typed sat one byte past its own line's scope: no cursor in scope, conceal on, whole line re-laid out for a frame.

The scope is now anchored on markers, like every other range here — two **point** markers, not one span marker. `MarkerList::create_span` exists for overlap queries and its end is not reliable under editing: the interval tree is a BST on interval start and `adjust_recursive` descends right from a node whose start precedes the edit, so a span to that node's left but reaching past the edit never has its end shifted (markers at 0 and 1 plus a span `[0, 4)`, insert at 3, and the span is still `[0, 4)`). That was measured, after a first attempt with a span marker did not fix the flicker; `overlay.rs` keeps its own start/end markers for the same reason. The conceal manager and the soft-break manager own the two markers and release them where they already release the decoration's own.

## The caret jumped two rows down after deleting on a list line

In an empty `bla.md`, type `- a` and press backspace: the caret lands two rows below the line, at the page's left edge, until you type again. In tmux the frame after the delete ends `ESC[5;29H` where the frames before and after it end `ESC[3;31H`. A numbered list does it too; a heading does not.

What a list has that a heading has not is a virtual row hung under it. The renderer keeps `last_line_end` — the buffer's last *content* row and whether it ended with a newline — and that decides whether to draw the implicit empty line following a buffer's final newline, the line a cursor at the buffer end belongs on. Every rendered row updated it, virtual rows included, and a virtual row is rendered from an *injected* newline. So a buffer with no trailing newline at all looked newline-terminated, an implicit empty row was drawn that no line of the document owns, and its end-of-buffer clause took the caret.

It stayed hidden while the content row claimed the caret itself, and deleting the `a` is what stops that: the row then ends on the list marker's space, and the position past a trailing space is not a cell the row draws (`row_end_exclusive`) — it is the end-of-buffer resolution's to place, and `resolve_cursor_fallback` places it correctly from an honest `last_line_end`. A virtual row now leaves it alone, the same policy, for the same reason, as the iterator's trailing empty line beside it. The spurious row went with it.

Tested through the virtual-line seed rather than the markdown plugin: the virtual row is the mechanism, and a seeded one stays put instead of being rebuilt a frame later, which is what made this look like a flicker.

## A line edited at its very start was not offered to plugins again

Found while chasing the compose flicker and fixed on its own: `adjust_seen_byte_ranges_for_edit` treated an insert landing exactly on a line's first byte as being *before* that line, so the line kept its "already seen" range and `lines_changed` skipped it. Typing at the head of a line left its decorations built from the text before the keystroke until some later edit dislodged them.

## Clicking the code tour's description moved a caret but not the keyboard

The panel is mounted into a pane, and a press on its prose is answered by the prose run's own node, which calls `e.stop()`. The `WidgetHit` arm makes the pane half of such a press explicitly (`Slot::Pane(pane) => focus_pane(pane)`) because nothing else can — the pane's own pointer surface never sees a press a widget claimed. `WidgetProsePress` never made it. The test types into the file first, so it cannot pass by never having given the file focus at all.

## The prose caret vanished at the end of a line

It was drawn as a one-byte *selection* — the range from the caret to the next character, washed reversed. A selection is the cells its bytes are drawn in, and the byte at the end of a rendered line is the `\n` the wrap dropped: `selected_spans` clamps a range to each row's own source span and yields nothing for it, so the wash covered no cells; at the end of the document the range was empty outright. Clicking or dragging past the end of a line is exactly how a reader puts the caret there.

A caret is a point *between* cells, which a range cannot say and `cell_of` already answers. So the run gets a `block_caret` beside `cursor`: the same byte→cell placement, painted as a washed cell instead of moving the terminal's one. The prose needs exactly that, because it cannot take the hardware cursor — the panel buffer's viewport follows it and scrolls the panel's own header off the top (`test_prose_caret_never_scrolls_the_panel`).

## Testing

- Every new test was run against the code without its fix and fails there with the reported symptom; the controls pass in both states.
- All 13 CI checks are green on this head, including the full suite on Linux, macOS and Windows.
- Both markdown reports were re-checked the way they were reported: a real editor in tmux with raw frame capture, before and after — the flicker on a single keystroke and on a burst, the caret jump on `- a` and on `1. a`.

https://claude.ai/code/session_01EG7711E55sA6JYeD6trYuY


sinelaw:
`test ubuntu-latest` failed on `fe1a39d`, and the failure is not this PR's.

**What failed:** one test timed out; 9847 of 9848 passed.

```
TIMEOUT [ 180.009s] (7266/9848) fresh-editor::all_tests e2e::vi_mode_bugs::test_vi_visual_up_keeps_anchor_character
```

**Why it isn't this PR's:**

- It hung from its *first* wait — `wait_until(editor_mode() == "vi-visual")` after `j`,`v` — so the vi plugin never entered visual mode at all. The captured stderr is `wait_until still pending` at 10 s, 20 s … 170 s, then SIGTERM. That is a vi-mode plugin dispatch hang; nothing in this PR touches vi mode, key dispatch or plugin startup.
- `wait_until` is deliberately unbounded (it dumps the screen every 10 s and relies on nextest's slow-timeout to kill it), so any missed plugin event in any e2e test becomes exactly this 180 s timeout.
- **The base commit fails the same way.** `41fe3b98`, which this PR branches from, failed its own master CI run ([35153871518](https://github.com/sinelaw/fresh/actions/runs/35153871518)) with the identical shape — one e2e test timing out at 180 s, 9783 of 9784 passing:
  `TIMEOUT [180.014s] e2e::markdown_compose_scrollbar_markers::every_heading_is_marked_without_scrolling` (macOS). Several other recent master runs are red too.
- Locally the test passes 20/20 in ~0.35 s, and a full `all_tests` run on this branch ran it green.

**Fix ported:** none — there is no fix for this to port. It is an intermittent hang in an unrelated test, and the general fragility behind it (unbounded `wait_until`) is a repo-wide harness property, not something this PR should widen to take on.

Re-running the job once, which is the one re-run I'll spend on it. Every other check on this commit is green.

---

Issue #2806, opened 2026-07-26
support markdown render in web ui


asukaminato0721:

kokic:
## Recommended Plan: pulldown-cmark + KaTeX

- https://github.com/pulldown-cmark/pulldown-cmark

```rust
pub const OPTIONS: Options = Options::ENABLE_MATH
    .union(Options::ENABLE_YAML_STYLE_METADATA_BLOCKS)
    .union(Options::ENABLE_TABLES)
    .union(Options::ENABLE_SMART_PUNCTUATION)
    .union(Options::ENABLE_FOOTNOTES)
    .union(Options::ENABLE_GFM)
    .union(Options::ENABLE_STRIKETHROUGH)
    .union(Options::ENABLE_TASKLISTS)
    .union(Options::ENABLE_DEFINITION_LIST)
    .union(Options::ENABLE_HEADING_ATTRIBUTES);
// ...
let events = pulldown_cmark::Parser::new_ext(source, OPTIONS);
```

- https://github.com/KaTeX/KaTeX/

```html
<!-- Also see: https://katex.org/docs/autorender -->
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.18.1/dist/katex.min.css" integrity="sha384-1vdNCNel6Tx/NQa8IR1mGOGKsbGreCkOPfbtPPnUURJ5Tu2PRVfQ/7KLZC+Pi1p1" crossorigin="anonymous">
<script defer src="https://cdn.jsdelivr.net/npm/katex@0.18.1/dist/katex.min.js" integrity="sha384-ycJ6GAwiS15LoUPipwJOrWTvkUHl/YqELValBwI5I4awP1EeEQJYarj+w85ntcz7" crossorigin="anonymous"></script>
<script defer src="https://cdn.jsdelivr.net/npm/katex@0.18.1/dist/contrib/auto-render.min.js" integrity="sha384-bjyGPfbij8/NDKJhSGZNP/khQVgtHUE5exjm4Ydllo42FwIgYsdLO2lXGmRBf5Mz" crossorigin="anonymous"></script>
<!-- Also see: https://github.com/KaTeX/KaTeX/tree/main/contrib/copy-tex -->
[...]

Issue #2733, opened 2026-07-20
Web-UI bridge (webui/mod.rs) review: 1 crash, 1 hang, 2 CSRF/origin gaps, input-fidelity bugs — and the HTTP POST routes are now vestigial


sinelaw:
Code review of the Rust side of the web-UI bridge — `crates/fresh-editor/src/webui/mod.rs` (hand-rolled HTTP + RFC 6455 WebSocket, the region-diff protocol, input dispatch, scene serialization) — cross-checked against the frontend counterpart (`web-ui/index.html` `applyFrame`/`connect`) and `docs/internal/web-ui.md`.

Every finding below was independently verified against the code. Line numbers are against the current `claude/rust-web-ui-bridge-review-ccumq1` tip. Scope is understood: this is a documented single-client, localhost, no-auth dev prototype — the security findings are **not** "no auth is bad", they are places where a guard the code *claims to provide* is inconsistent or bypassable.

---

## 0. Are the HTTP POST routes still needed? (investigation)

**Conclusion: the POST routes are leftovers from before the WebSocket transport.** Nothing that ships or is tested uses them:

| Consumer | How it drives the bridge | Uses HTTP POST? |
|---|---|---|
| Frontend (`web-ui/index.html`) | all input via WebSocket `wsSend` — incl. `sendPaste` → `wsSend({type:"paste"})` | **No.** Only `GET /state` in the debug-only `refresh()` (`window.fresh.refresh`) |
| Playwright suite (`web-ui/test/drive.mjs`) | WebSocket; `run.sh` polls `GET /state` for readiness | **No** |
| Rust parity test (`crates/fresh-editor/tests/scene_parity.rs`) | in-process calls to `apply_step` / `scene_value` / `build_editor` — no socket at all | **No** |

So the routes `POST /key /paste /mouse /action /widget /settings /kbedit /resize` (superseded by the shared `apply_message` WS path) and `POST /step /reset` (the parity harness calls the functions in-process) have **no automated consumer** — only manual curl. The module comment at `mod.rs:47-57` and `docs/internal/web-ui.md:57` claiming "curl and the parity harness depend on them" / "the parity harness (`/step`, `/reset`) are untouched" is **factually stale**: neither test suite touches those HTTP routes.

**Why this matters:** these POST routes are the *entire* attack surface of finding #3. Two clean options:
[...]

Issue #534, opened 2025-12-29
crazy idea: wasm version


asukaminato0721:
as a preview maybe

asukaminato0721:
https://github.com/sinelaw/fresh/commit/7541d884b94d5767080d90065ac5d2f9d53bd194

sinelaw:
It's still a work in progress, need to continue refactoring the remaining modules to allow more code to work in wasm build

Issue #1577, opened 2026-04-15
Possibly incorrect rendering of variable width unicode


sinelaw:
A standard "i", a ＦＵＬＬＷＩＤＴＨ "Ｗ", a massive Arabic ligature "﷽", a ZWJ emoji sequence "👨‍👩‍👧‍👦", a Zalgo combining word "t̷e̷x̷t̷", and an ancient cuneiform "𒈙".

<img width="1920" height="1144" alt="Image" src="https://github.com/user-attachments/assets/6220499a-965f-4db1-b39a-3101724a3504" />

sinelaw:
Some of the problems are due to running in tmux. Can't fix that.

sinelaw:
For reference, only ghostty really renders everything correctly.

<img width="2992" height="1234" alt="Image" src="https://github.com/user-attachments/assets/e430227d-1c12-4aa5-92e5-6971859bae9d" />

polarathene:
Just to confirm with your comparison of Ghostty being the only one to render correctly, what is wrong with `windows-cmd` and `notepad`?

I tried both of those with `hello "👨🏿‍👩🏿‍👧🏿‍👦🏿" world` (_[darkest tone](https://emojipedia.org/family-man-dark-skin-tone-woman-dark-skin-tone-girl-dark-skin-tone-boy-dark-skin-tone) for better distinction that the tone is applied_) where I pasted the emoji in between the quotes and both seem to work correctly? (_no issue with horizontal navigation or removal_)

The white-space issue on Windows Terminal is valid though (_WSL2 tab, cursor is blinking in screenshot below right after the emoji that is highlighted from the paste_):

> <img width="309" height="22" alt="Image" src="https://github.com/user-attachments/assets/33095b5e-a74e-4fa2-a11f-7ce01e799713" />

So it's unclear from the screenshot you provided what the issue is with `notepad` and `windows-cmd` vs `ghostty`? 🤔 

---

## Reference `fresh` (Windows Terminal)

In WSL2 I ran `fresh` (latest `0.3.7` release) and pasted into quotes without the padding issue present 👍

> <img width="177" height="25" alt="Image" src="https://github.com/user-attachments/assets/da31249c-8ba9-4712-9c35-bf61458c216d" />

However if using another emoji composition like [`👩🏾‍❤️‍💋‍👨🏼`](https://emojipedia.org/kiss-woman-man-medium-dark-skin-tone-medium-light-skin-tone) `fresh` did run into issues.

[...]

PR #3319, opened 2026-09-17
Fix the open web-UI issues: styled popups, the digit row, shipped `--web`, the missing themes, and the black hero

### LINKED ISSUE #3106: Bug: Homepage terminal animation (screenshot-wrapper) is a blank black box on Chrome / Fedora
Hi, the terminal animation preview (`.screenshot-wrapper`) at the very top of https://getfresh.dev/ is broken for me. It only displays a solid black background; none of the text or animations inside are visible.

Environment:
- OS: Fedora Linux
- Browser: Google Chrome Version 150.0.7871.186 (Official Build) (64-bit)

Image

### LINKED ISSUE #3107: Bug: "Tokyo Night" theme listed on homepage but missing
Hi, the homepage (https://getfresh.dev) states that "Tokyo Night" is a built-in theme. However, after installing the latest fresh-editor v0.4.10, this theme is missing from the `Select Theme` list. 

Is this a packaging bug, or is there an official link where we can download the `tokyo-night.json` file manually? Thanks!

### LINKED ISSUE #3315: web ui hover missing highlight
Image

### LINKED ISSUE #3316: aur package add web feature

sinelaw:
Fixes the five reproducible issues from the web-UI backlog. Each was reproduced first (repro steps posted on the issues), then fixed and re-measured. Screenshots below are from those runs; every browser check is headless Chromium driving a real `fresh --web`, and the images live in `docs/internal/screenshots/`.

Closes #3315, closes #3316, closes #3107, closes #3106. Addresses the last live finding in #2733 and the hover half of #2806.

---

## #3315 — a hover popup lost every colour the editor gave it

`project_popup` flattened a markdown popup's styled spans into one `String` per line, so an LSP hover reached the browser as flat white text: no syntax highlighting, no code-block background, no bold or italic, and blank lines collapsed so paragraphs ran together. The terminal drew all of it, and the same markdown in a *buffer* (compose mode) rendered fully — the loss was specific to popups.

Popups now project runs in the shape cell runs already use (`t`/`fg`/`bg`/`b`/`i`/`u`/`r`), and the frontend paints them with the logic that paints a buffer run, REVERSED included.

| before | after |
|---|---|
| before | after |

The ink is inline style, so a web theme skins the popup's frame and never the editor colours inside it. Verified in all five web themes (Cosmos, macOS, macOS Dark, Compact, Winamp): 43 spans, 5 distinct colours, the code-block background, bold and italic, monospace and all 11 rows present in every one. Also verified across the new editor themes — here gruvbox ink inside the Winamp skin:

gruvbox in winamp

This is also the hover half of #2806: markdown in the web UI now renders with its styling. Native HTML markdown (the pulldown-cmark + KaTeX proposal in that thread) is a separate, larger decision and is **not** in this PR — buffer-side markdown already renders through compose mode.

## #2733 — `Ctrl+Shift+<digit>` never reached the editor

A browser reports the character a key *produces*, so `Ctrl+Shift+1` arrives as `{key:"!"}`, while bindings — and the kitty-protocol events a terminal sends — are written against the unshifted key plus SHIFT. All ten `set_bookmark` slots were unreachable from a browser and worked in the terminal.

The page now sends `code` alongside `key`, and with CONTROL held the bridge takes the digit from it. ALT is excluded so AltGr layouts keep their own characters, and without CONTROL the typed character is still the truth — typing `!1` inserts `!1`. This also fixes the digit row on layouts whose shifted digits are not the US symbols.

Same keystroke, same session — before (the old wire message) and after:

before
after

The other seven findings in #2733 are already fixed on `master`; I re-ran them all against a live bridge and posted the results on the issue.

## #3316 — no shipped package contained the web UI

`web` is opt-in in Cargo.toml and nothing that ships named it, so every installed `fresh` rejected `--web` with clap's bare "unexpected argument". The feature is now named where the shipped binaries are built — the AUR source PKGBUILD, `debian/rules`, the linux-packages job behind the .deb and .rpm, the musl tarballs, and the release job whose archives the AUR `-bin` package, the Homebrew formula and the npm package install.

Where a build still legitimately lacks it (`cargo install` without the feature, the min-size profile), `--web` is accepted and explains itself:

```
$ fresh --web 127.0.0.1:8137
this build of fresh was compiled without the web UI
rebuild with: cargo build --release --features web
(or install a package that ships it)
$ echo $?
1
```

The flag stays hidden from `--help` there. `cargo install fresh-editor` still needs `--features web`: making `web` a default feature is a binary-size decision beyond this issue, and the message now says what to do.

## #3107 — Tokyo Night (and Gruvbox) were advertised but not shipped

The homepage lists both as built-in; neither existed. Both now ship as complete theme files using the canonical upstream palettes, and pass the built-in theme invariants the suite enforces (schema match, blame-header band contrast, an occurrence highlight that survives 256-colour quantisation, highlighting that never erases syntax, a prominent palette indicator, a visible selected-whitespace indicator). `Select Theme` now offers ten.

| Tokyo Night | Gruvbox |
|---|---|
| tokyo night | gruvbox |

## #3106 — the homepage hero was a black box

asciinema-player 3.13.0 split the terminal into three stacked layers inside `div.ap-term`: a `<canvas>` painting every cell's *background* colour, an `<svg>`, and a `<pre>` holding only the text. On Chrome/Linux that canvas intermittently never paints, and since the cell backgrounds live nowhere else, the recorded TUI — dark text on light cells — collapses onto the host's own near-black. The text is present and visible the whole time, which is why it reads as blank rather than broken, and why no z-index or compositing override helps: there are no pixels to composite.

Vendored 3.12.1, the last release whose renderer is a single `<pre class="ap-terminal">` with per-span background colours — ordinary DOM paint, no separate layer to lose.

| before | after |
|---|---|
| before | after |

Measured over fresh page sessions, mean luminance of `.screenshot-wrapper` as the oracle (black box ≈ 1.4, good render ≈ 170+):

| | DPR 1 | DPR 2 |
|---|---|---|
| before (3.15.1) | 9/12 black | 10/12 black |
| after (3.12.1) | 0/20 black | 0/20 black |

Two things worth a maintainer's judgement: the pin goes **backwards**, so anyone bumping it reintroduces the bug (recorded in a comment next to the script tag), and the demo is now ~20% brighter because 3.12.1 hits the xterm-256 palette exactly where 3.15.1's canvas path rendered every background at ~0.8×. `filter: brightness(0.83)` on `.asciinema-host` restores the old look if you prefer it.

---

## Verification

- `cargo test -p fresh-editor --lib --features web` — 2542 passed, 0 failed. `cargo test -p fresh-editor-core --lib` — 1243 passed, 0 failed. Includes new regression tests for both bridge fixes: popup runs keep colour/emphasis and serialise in the cell-run shape, and the digit row is recovered from the physical key.
- `scene_parity` (the TUI-vs-web derivation test) passes.
- The bundled web-UI Playwright suite: **177 passed, 5 failed** — the same five as the pre-change baseline, all in the Orchestrator dock / New Workspace dialog, which this PR does not touch.
- `cargo fmt --all --check` clean; `cargo clippy -p fresh-editor --features web` introduces no new warnings.
- Homepage: no console errors and no failed local requests at 1440×900 and 400px wide, at DPR 1 and 2; autoplay/loop intact; the static `hero.png` fallback still works when the bundle is blocked.

Not addressed here: #534 (a wasm build) and #3272 (the legacy rendering doc), neither of which has reproducible behaviour to fix.

https://claude.ai/code/session_01Nm3GXF2oCYsrzQqAQRyVXN

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nm3GXF2oCYsrzQqAQRyVXN)_


sinelaw:
CI: 15 of 16 checks green. `playwright suite` is red, and it is not this PR's — the same failures are on `master`.

**On this PR** (run 35285040409): `177 passed, 5 failed`
```
FAIL dock toggle click flips real plugin state null -> null
FAIL New Task dropdown rows are items below the trigger {}
FAIL New Workspace dialog is a centered floating panel
FAIL a dimming scrim sits behind it
FAIL a click over the dock is eaten by the modal (row unchanged, modal still open)
```

**On `master` at 04de710**, same suite, same machine, no changes from this branch: `173 passed, 9 failed` — a superset containing all five above, plus `right-click on a dock row opens the context-menu panel`, the two context-menu assertions after it, and `a buffer click still lands on the right cell while zoomed`. So this branch fails strictly fewer checks than the base it merges into.

Why CI has not shown this before: `.github/workflows/web-ui.yml` only runs on changes under `crates/fresh-editor/web-ui/**`, `src/webui/**`, `view/scene.rs`, `tests/scene_parity.rs` and `examples/webui_server.rs`. The last run on `master` was 2026-09-13 (a3e6722, green). #3306 landed on 2026-09-16 and changed exactly the surfaces these assertions cover — Orchestrator/workspace behaviour, the dock, the New Workspace dialog — without touching a filtered path, so the suite never ran on it. This PR touches `view/scene.rs` and `web-ui/`, which is why the suite runs here and the standing breakage surfaces on it.

Nothing in this PR touches the dock, the Orchestrator, the New Workspace dialog or the zoom hit-testing path. The three surfaces it does change are covered and green: the popup checks, the whole web-theme section, and `scene parity (web/TUI)`.

I have not widened this PR to fix the dock regressions — they are a separate change to a separate surface, and they want their own diff and their own review. Happy to take them on next if you want; the fastest lead is a bisect of #3306 against the five assertions above.

---

crates/fresh-editor/web-ui/css/40-widgets.css
```diff
@@ -64,17 +64,17 @@
   .w-tree{display:flex;flex-direction:column;overflow:auto}
   .w-tree-row{padding:1px 6px;white-space:pre;overflow:hidden;text-overflow:ellipsis;cursor:default;border-radius:3px}
   .w-tree-row.sel{background:var(--menuhi);color:#fff}
-  .w-tree-disc,.w-tree-check{color:var(--muted)} .w-tree-row.sel .w-tree-disc,.w-tree-row.sel .w-tree-check{color:#dfe8ff}
+  .w-tree-disc,.w-tree-check{color:var(--muted)} .w-tree-row.sel .w-tree-disc,.w-tree-row.sel .w-tree-check{color:color-mix(in srgb,var(--on-sel) 62%,transparent)}
   /* Continuation rows of a fixed-height card (Tree.itemHeight > 1): muted
      detail lines that select + highlight with their card's primary row. */
   .w-tree-xrow{padding:0 6px;white-space:pre;overflow:hidden;text-overflow:ellipsis;cursor:default;border-radius:3px;color:var(--muted)}
-  .w-tree-xrow.sel{background:var(--menuhi);color:#dfe8ff}
+  .w-tree-xrow.sel{background:var(--menuhi);color:var(--on-sel)}
```
sinelaw: `--on-sel` is the wrong token for a **solid** `var(--menuhi)` fill, and this commit's own comment says why: `--on-sel` is documented (20-cells.js) as the ink for rows that *tint* their surface, where the theme foreground is the right answer. On a solid fill it measures worse than the `#dfe8ff` it replaced:

| theme | old `#dfe8ff` | new `--on-sel` |
|---|---|---|
| solarized-dark | 3.00 | **1.16** |
| tokyo-night | 2.05 | **1.56** |
| dark | 3.35 | **2.77** |
| nord | 3.29 | **2.98** |

This is exactly the case `--on-menuhi` was introduced for — it should be used here (and on `.w-tree-card.sel .w-tree-xrow` below).

---

crates/fresh-editor/web-ui/css/40-widgets.css
```diff
@@ -64,17 +64,17 @@
   .w-tree{display:flex;flex-direction:column;overflow:auto}
   .w-tree-row{padding:1px 6px;white-space:pre;overflow:hidden;text-overflow:ellipsis;cursor:default;border-radius:3px}
   .w-tree-row.sel{background:var(--menuhi);color:#fff}
-  .w-tree-disc,.w-tree-check{color:var(--muted)} .w-tree-row.sel .w-tree-disc,.w-tree-row.sel .w-tree-check{color:#dfe8ff}
+  .w-tree-disc,.w-tree-check{color:var(--muted)} .w-tree-row.sel .w-tree-disc,.w-tree-row.sel .w-tree-check{color:color-mix(in srgb,var(--on-sel) 62%,transparent)}
```
sinelaw: The row this line belongs to is the one the commit is about, and it was missed: `.w-tree-row.sel` (line 66) still hardcodes `color:#fff` on a solid `var(--menuhi)` fill — **1.72:1 on gruvbox, 1.13:1 on light** — while this line flips only its disc/check markers to a derived token.

So the selected dock row now derives its markers from the theme and leaves its label as fixed white: the two halves of one row disagree, and the label is the half that fails. Whatever token line 71 settles on should apply to line 66 as well.

---

crates/fresh-editor/src/webui/mod.rs
```diff
@@ -2077,6 +2088,16 @@ fn scene_json(editor: &mut Editor, cols: u16, rows: u16) -> Value {
...
+/// The digit a browser's physical `code` names — `"Digit1"` → `'1'`. Every
+/// other code (letters, punctuation, named keys) is left to the reported
+/// character, which already matches what a terminal sends.
+fn digit_row_char(code: &str) -> Option<char> {
+    let rest = code.strip_prefix("Digit")?;
+    let mut chars = rest.chars();
+    let d = chars.next()?;
+    (chars.next().is_none() && d.is_ascii_digit()).then_some(d)
+}
+
```
sinelaw, line 2118: Architectural: the theme already answers this question and the scene drops the answer.

`Theme` carries `menu_highlight_fg` — the ink a theme author chose for its own highlight fill (8.69:1 on gruvbox, 6.79:1 on tokyo-night). This payload exports `menuHi` but not `menuHighlightFg`, so the frontend has to *infer* ink it was already given, and the inference can only ever be black or white. The premise in the new `--on-menuhi` comment ("the theme names the fill and never the ink on it") is therefore not true.

Exporting `menuHighlightFg` here and using it as the primary value — with the computed black/white as the fallback for a theme that leaves it at terminal reset — removes the duplication, respects the theme author, and beats the computed answer on every bundled theme.

---

crates/fresh-editor/web-ui/css/20-chrome.css
```diff
@@ -18,16 +18,18 @@
   .dropdown.submenu{box-shadow:8px 14px 40px rgba(0,0,0,.45)}
   .mitem,.msep,.mlabel{position:absolute;z-index:81}
   .mitem{display:flex;align-items:center;gap:16px;padding:0 10px;white-space:nowrap;cursor:default;color:var(--fg)}
-  .mitem.hi{background:var(--menuhi);color:#fff}
-  .mitem.disabled{color:#6a6a6a}
+  .mitem.hi{background:var(--menuhi);color:var(--on-menuhi,#fff)}
```
sinelaw: Dead edits. The stylesheets are concatenated in filename order and `60-polish.css` re-points all three of the selectors touched in this hunk:

- `.mitem.hi` → `background:var(--sel)` (60-polish.css:207)
- `.mitem.hi .accel` → the `--on-sel` 62% mix (60-polish.css:244)
- `.kbedit .kb-field.focus` → `color:var(--on-sel)` (60-polish.css:241)

So `--on-menuhi` never reaches a rendered pixel from here. And if it did it would be wrong: `80-skin.css:78` makes those backgrounds *translucent* over the panel surface, which is the tinted case `--on-sel` exists for — not a solid fill.

Either drop this hunk, or move the change to where the cascade actually lands and decide per-selector whether the fill is solid or a tint.

---

crates/fresh-editor/web-ui/js/15-theme.js
```diff
@@ -231,6 +231,15 @@ function applyWebTheme() {
     else if (!THEME_KEYS.includes(k)) r.removeProperty(k);   // back to the :root default
     // keys applyTheme owns are left as it just set them (Cosmos wants the TUI value)
   }
+  // A skin may replace the fills themselves (Winamp's navy highlight, macOS's
+  // blue), and the ink that goes ON a fill is computed from it — so recompute
+  // after the overrides land, or a row keeps the ink picked for the editor
+  // theme's fill (gruvbox's amber ink on Winamp's navy measured 1.9:1).
+  const eff = getComputedStyle(document.documentElement);
+  for (const [fill, ink] of [["--menuhi", "--on-menuhi"], ["--accent", "--on-accent"]]) {
```
sinelaw, line 245: This loop clobbers a value a skin declared for itself. `WEB_THEME_VARS` sets `--on-accent` explicitly (macOS Dark declares `#ffffff`), and it is applied by the loop just above — then this recompute overwrites it from the accent, turning it into `#10131a` on every render.

A skin that states its own ink should win; the recompute is for the tokens it did *not* state. Skipping any key present in `vars` fixes it:

```js
for (const [fill, ink] of [["--menuhi", "--on-menuhi"], ["--accent", "--on-accent"]]) {
  if (ink in vars) continue;   // the skin stated it
  ...
}
```

---

crates/fresh-editor/web-ui/js/20-cells.js
```diff
@@ -112,17 +148,28 @@ function applyTheme(t){
...
   set("--bg",t.bg); set("--fg",t.fg); set("--accent",t.accent);
-  set("--muted",t.muted); set("--bg2",t.popupBg); set("--bg3",t.menuBg);
+  // `--muted` is chrome SECONDARY TEXT (tab names, status segments, setting
+  // descriptions), seeded from the theme's gutter grey. A gutter grey is
+  // chosen to recede behind code, not to be read as a label: tokyo-night's
+  // #565f89 lands at 2.5:1 on the panel surface, which took every muted string
+  // in the chrome under the 3:1 floor. Lift it toward the foreground until it
+  // clears that floor on both surfaces it is drawn on; a theme already above
+  // the floor is left exactly as it is.
+  set("--muted",legibleMuted(t.muted,t.fg,[t.popupBg,t.bg]));
```
sinelaw, line 221: The ground list is incomplete, so the floor reports success where it does not hold. Muted text is also drawn on `--bg3` (`menuBg`) — menu accelerators, the keybinding-editor title band — and that surface is not in the list:

- dark: accelerator ends at **2.18:1**
- nostalgia: **1.90:1**

…while `legibleMuted` returns "cleared the floor". Adding `t.menuBg` to the array is the fix, and it is worth asserting the loop's postcondition rather than trusting the caller to pass every ground: if no blend clears all of them, returning `fg` (as it does) is right, but silently missing a ground is the failure mode that matters.

---

crates/fresh-editor/web-ui/css/50-popups.css
```diff
@@ -7,15 +7,28 @@
...
+  /* A popup body is EDITOR content (LSP hover docs, the theme inspector), not
+     chrome: it keeps the monospace grid even under a proportional chrome theme
+     (macOS), the same rule the buffer interiors follow, and its runs bring
+     their own colour as inline style. The line only says what a run leaves
+     unsaid. */
+  /* A body of editor content stands on the EDITOR's popup ground, not the
+     chrome skin's surface: its runs carry the editor theme's colours, which
+     were chosen against that ground. Chrome-modelled popups (completion,
+     action, the status indicators) keep the skin's surface. */
+  .popup .popup-body.ink{background:var(--ed-popup-bg,var(--bg2));color:var(--ed-popup-fg,var(--fg))}
```
sinelaw: This puts an **opaque** body inside a deliberately translucent card. Under the macOS skins `.popup` is frosted glass — 74% alpha plus a backdrop blur — and the title and description bands stay translucent, so a hover popup now reads as a glass frame with a solid slab dropped into it.

The legibility goal doesn't require opacity, only that the ink's ground is the editor's rather than the skin's. Two ways to keep both:

- paint the ground with the same alpha the skin gives the card (`color-mix` the editor bg with `transparent`, or let the skin override `--ed-popup-bg` with its own alpha), or
- scope the opaque ground to the skins that don't frost.

Worth a look at the macOS Dark hover popup before/after — this is a visual regression on exactly the skin whose identity is the glass.

---

crates/fresh-editor/web-ui/css/40-widgets.css
```diff
@@ -130,8 +130,8 @@
   .settings-modal .set-list-head,.settings-modal .set-list-head .set-list-sub{color:var(--muted)}
   /* Keyboard focus inside a list control: highlight the focused entry / add
      row and show the TUI's dim `[Enter to edit]` hint. */
-  .settings-modal .set-list-row.sel,.settings-modal .set-list-add.sel{background:var(--menuhi);color:#fff}
-  .settings-modal .set-list-row.sel .set-list-sub{color:#dfe8ff}
+  .settings-modal .set-list-row.sel,.settings-modal .set-list-add.sel{background:var(--menuhi);color:var(--on-menuhi,#fff)}
+  .settings-modal .set-list-row.sel .set-list-sub{color:color-mix(in srgb,var(--on-sel) 62%,transparent)}
```
sinelaw: One selected row, one solid fill, three different derivations of the ink on it:

- the label → `var(--on-menuhi)` (line 133)
- the sub-label → the `--on-sel` 62% mix (this line)
- the `.set-list-sub` in the polish layer → still literal `#cfd8ea`

The mix on this line is computed against a token chosen for a *different* ground than the one the row is actually painted with, so the primary and secondary text can land on opposite sides of readable. Secondary ink on a solid fill should be derived from the same base as the primary — e.g. `color-mix(in srgb, var(--on-menuhi) 62%, transparent)` — so one decision drives the row.

Worth pulling the "ink on a solid fill" pair (primary + secondary) into one place rather than repeating the mix at each call site; there are five of them now.

---

crates/fresh-editor/src/webui/mod.rs
```diff
@@ -227,6 +227,17 @@ pub fn build_editor(cols: u16, rows: u16, files: &[PathBuf]) -> Result<Editor> {
...
+    // fire them: `plugins_loaded` once the registry and init.ts are in, then
+    // `ready` once the startup files are open. `ready` is the "we have
+    // finished starting up" signal — it is where the Orchestrator opens its
+    // dock and the welcome screen opens itself — so a bridge that fired only
+    // the first served a session with neither, while a directly-launched
+    // editor and `fresh -a` had both. Opening the files first is what lets a
+    // plugin branching on "is a real file open?" see them rather than race
+    // them.
+    editor.process_pending_file_opens();
+    editor.fire_ready_hook();
```
sinelaw: Firing `ready` here is right for a *served* session, but `build_editor` is shared with `POST /reset` and with the Rust parity runner — both of which exist precisely to hand a test a known editor.

`ready` is what makes the Orchestrator mount its dock and the welcome screen open itself, and those arrive asynchronously from the plugin thread. So after a `/reset` the harness now gets a buffer whose "known top-of-buffer view" can be changed by a dock/welcome that lands a tick or two later — a race the reset was there to remove. `scene_parity` drives the same constructor.

Two options, either fine: fire it in the serve path only and leave `build_editor` at `plugins_loaded`, or give `build_editor` a flag (`fire_ready: bool`) that the serve path passes `true` and `/reset` + parity pass `false`. Worth doing before this lands, since a flaky parity test is expensive to diagnose later.

---

crates/fresh-editor/web-ui/css/20-chrome.css
```diff
@@ -18,16 +18,18 @@
...
+  /* Dim, but derived from the theme rather than a fixed grey: the same
+     "disabled" reads on a light skin and a dark one. */
+  .mitem.disabled{color:color-mix(in srgb,var(--fg) 45%,var(--bg3))}
```
sinelaw: Wrong ground in the mix: a disabled dropdown row sits on the dropdown panel, which is `--bg2` (and translucent under the skins), not `--bg3`. The two sibling rules added in this same change — `.palette .prow.disabled` and `.popup .popup-row.disabled` — both correctly mix against `--bg2`, so this one is inconsistent with its own commit.

`color-mix(in srgb, var(--fg) 45%, var(--bg2))` matches them and lands on the surface the text is actually drawn on.

---

crates/fresh-editor/web-ui/js/20-cells.js
```diff
@@ -112,17 +148,28 @@ function applyTheme(t){
...
-  // white cursor) still shows its label instead of white-on-white.
-  const onColor=hex=>{ const m=/^#?([0-9a-fA-F]{6})$/.exec((hex||"").trim()); if(!m) return null;
-    const n=parseInt(m[1],16), lin=c=>{c/=255; return c<=.03928?c/12.92:Math.pow((c+.055)/1.055,2.4);};
-    const L=.2126*lin((n>>16)&255)+.7152*lin((n>>8)&255)+.0722*lin(n&255);
-    return L>.45?"#10131a":"#ffffff"; };
   set("--on-accent", onColor(t.accent));
+  // Same question for a row filled SOLID with the menu-highlight: a theme
+  // whose highlight is light (gruvbox's amber) left hardcoded white text at
+  // 1.7:1. Rows that merely *tint* their surface keep `--on-sel` below.
+  set("--on-menuhi", onColor(t.menuHi));
```
sinelaw: Duplication introduced by this change: `hexRgb` was extracted here as the one hex parser, but `mix()` further down still carries its own copy of the same regex-and-shift parsing. Now there are two parsers with the same job in one file, and only one of them handles the "theme left it at terminal reset" case.

`mix()` should call `hexRgb` (and return its input unchanged when the parse fails, as the new helpers do).

---

crates/fresh-editor/web-ui/js/20-cells.js
sinelaw, line 1: Separate from this PR, but found while sweeping the web UI and it belongs on this file: **the SVG cell renderer breaks every grapheme cluster longer than one UTF-16 code unit.**

`cellsSvg` measures and pins in `r.t.length` — JS string units — for both the background `<rect>` width (`px(r.t.length, CW)`) and the per-character `x` list on each `<tspan>`. The cell grid upstream is per *grapheme*, so anything that is one grapheme spanning several code units gets its pieces pinned to separate columns, and the browser can no longer shape it.

Measured against a scratch file (`x` list per tspan, from `window.fresh`):

```
ZWJ family  "👨‍👩‍👧‍👦"  x=[123.0 132.4 141.9 151.4 160.8 170.3 179.…]   ← 7+ columns for ONE grapheme
single      "👨"        x=[141.9 151.4]                                ← surrogate pair → 2 columns
flag        "🇮🇱"        x=[66.2 75.7 85.1 94.6]                        ← 4
skin tone   "👍🏽"        x=[113.5 123.0 132.4 141.9]                    ← 4
combining   "áèô"       x=[113.5 123.0 132.4 141.9 151.4 160.8]        ← 6 for 3 graphemes
CJK         "漢" "字"    x=[56.8] [75.7]                                 ← correct (wide cell + blank)
```

Only the CJK case is right, and it is right because the bridge already emits the trailing blank cell for a wide glyph. The ZWJ case is what shows up as a narrow glyph followed by an empty gap (the CHANGELOG line about #1577 renders that way in the browser — the very feature the entry announces as fixed is broken in the web UI).

The fix is to iterate grapheme clusters rather than code units when building both the rect width and the `x` list — `Intl.Segmenter('…', {granularity:'grapheme'})` is available in every browser the CSP already allows — and to advance `col` by the cell width the bridge assigned rather than by `.length`. Worth its own issue; happy to take it.

---

sinelaw (commented):
Self-review of the theme/contrast pass, plus one pre-existing rendering bug found while sweeping.

The honest summary: the `ready` hook and the Orchestrator dock fix hold up, but **the contrast work is half-applied and in two places makes things worse than what it replaced.** `--on-menuhi` was introduced for solid `var(--menuhi)` fills and then used on two of the roughly six selectors that carry one; three of the rest were pointed at `--on-sel`, which this same commit documents as the ink for *tinted* rows — a measurable regression (solarized-dark 3.00 → 1.16). Two of the edits are dead altogether, overridden later in the concatenation order.

Underneath that is an architectural mistake worth fixing before the surface ones: `Theme` already carries `menu_highlight_fg`, the ink a theme author chose for their own highlight. The scene never exports it, so the frontend computes a black-or-white approximation of data it was already given — and the theme's own value beats the computed one on every bundled theme. Exporting it (computed value as the fallback) removes both the duplication and most of the per-selector guesswork.

Also flagged: a skin-declared `--on-accent` clobbered on every render, an incomplete ground list that lets the muted floor report success at 2.18:1, an opaque popup body dropped into the macOS frosted-glass card, `ready` firing in the shared constructor that `/reset` and `scene_parity` also use, and a second hex parser left behind by the extraction.

Separately, the last comment is not about this PR: the SVG cell renderer pins per UTF-16 code unit rather than per grapheme, so ZWJ emoji, flags, skin tones and combining marks are torn across columns in the browser. That is the thing visible in the CHANGELOG line about #1577 — the feature the entry announces as fixed renders wrong in the web UI. It wants its own issue and fix.

---

crates/fresh-editor/web-ui/js/20-cells.js
sinelaw: Fixed in 6ee6925.

`cellsSvg` now walks `Intl.Segmenter(…, {granularity:"grapheme"})` (ASCII fast path, code-point fallback) for both the background rect width and the column advance. One extra thing turned up while doing it: an explicit `x` starts a new SVG text chunk and the browser will not shape across chunks, so a single tspan with one `x` per cluster still detached combining marks (`áèô` rendered as `a´èô` and pushed the rest of the line right). A multi-unit cluster is therefore emitted as its own tspan carrying a single `x`; consecutive single-unit cells stay batched in one tspan with an x list, so the common case is unchanged.

Same probe, after:

```
ZWJ family  "👨‍👩‍👧‍👦"  x=[123.0]   then the bridge's pad cell at 132.4, ")" at 141.9
single      "👨"        x=[141.9]
flag        "🇮🇱"        x=[66.2]
skin tone   "👍🏽"        x=[113.5]
combining   "á" "è" "ô"  x=[113.5] [123.0] [132.4]
CJK         "漢" "字"    x=[56.8] [75.7]   ← unchanged
```

The Playwright suite gains a regression check (the rendered unicode row must be exactly as wide, in cells, as an ASCII row of the same pane, and a multi-unit cluster must carry exactly one `x`): 185 passed, 0 failed, three consecutive runs.

---

sinelaw:
All twelve review findings are fixed and pushed. One commit per finding, each verified by measurement in a live headless browser rather than by eye:

| # | finding | commit |
|---|---|---|
| 3 | the theme's own `menu_highlight_fg` was dropped by the scene | dc180e2 |
| 1, 2, 4, 8 | one rule for ink on a solid highlight fill, applied to every row (label, sub-label, markers) | 4365ccd |
| 5 | the recompute clobbered an ink a skin declared for itself | dad069a |
| 6 | the muted floor did not hold on the menu surface (`--bg3`) | 40484b1 |
| 7 | an editor-ink popup body was an opaque slab inside frosted glass | 1f03083 |
| 9 | `ready` fired for `/reset` and the parity runner, not just a served session | 30e22e2 |
| 10 | a disabled dropdown row mixed against the wrong ground | 1aa1d56 |
| 11 | two hex parsers in one file | 28bd69d |
| 12 | the cell grid walked UTF-16 units instead of grapheme clusters | 6ee6925 |

Measurements after the ink change (WCAG contrast on the solid `--menuhi` fill, worst row per theme): dark 9.8, light 8.1, gruvbox 8.7, tokyo-night 6.8, nord 7.4, nostalgia 4.6 — previously as low as 1.13. The muted floor now clears 4.5:1 on the panel, menu and buffer grounds in all five web skins. #7 was re-measured with the macOS Dark hover popup specifically: the body takes the editor's ground at the card's own alpha (74% / 84%), so the glass survives, and the ink still clears the floor.

Verification: `cargo test -p fresh-editor --lib --features web` green including `scene_parity`; the bundled Playwright suite **185 passed, 0 failed** (three consecutive runs) — it was 177/5 when this PR opened and 173/9 on `master`, and those pre-existing failures are fixed here too.

---

PR #3320, opened 2026-09-18
Replace the bottom-row confirmation prompts with a modal dialog

sinelaw:
## Objective

Every confirmation in the editor asked on the terminal's last row, in the prompt's low-contrast colours, on the same line a transient status message uses — and stated its answers as parenthesised letters run together in a sentence. On a tall terminal that row is forty lines from where the user is looking, and on a narrow one the sentence is simply cut off:

```
 Trusted  Local  Ln 1, Col 5    Opened...  LF  ASCII  Rust   LSP (off)
1 buffer has unsaved changes. (s)ave and quit, (d)iscard and quit, (q)ui
```

That is a real 100-column capture — `Cancel` is not on screen at all. People reported the quit prompt as a hang, which is the right reading: a blocked exit with nothing visible to answer looks exactly like one.

The goal is that a question the editor is blocked on is impossible to miss, and that its possible answers are legible without decoding a legend.

## Solution

One modal dialog, shared by all fourteen confirmations: a centred card over a dimmed frame, with each outcome spelled out as its own button.

```
            ┌──────────────────────────────────────────────────────────────┐
            │ Unsaved Changes                                              │
            │──────────────────────────────────────────────────────────────│
            │                                                              │
            │ 1 buffer has unsaved changes.                                │
            │                                                              │
            │──────────────────────────────────────────────────────────────│
            │ [ Save and Quit ]  Discard and Quit  Quit (recoverable)      │
            │                                                     Cancel   │
            └──────────────────────────────────────────────────────────────┘
```

It is declared on the editor's existing retained-mode UI tree rather than drawn by a bespoke painter, reusing the primitives the workspace-trust modal already established: an exclusive layer with a dimming scrim. Doing it that way is what makes the dialog modal in substance rather than appearance — the frame behind it is dimmed, inert to the mouse, and takes no keystrokes — and it inherits the tree's mouse handling and the web projection without either being written again.

Above that sits one description of what a confirmation *is* — a question, some outcomes, and which of them is the retreat — so the fourteen prompts differ only in the content they supply, and cross-cutting rules live in one place instead of being restated at each call site.

Answering is deliberately untouched underneath. Each button produces exactly the answer the corresponding keystroke produced before, so every existing confirmation handler works unchanged.

## Prompts covered

Quitting with unsaved work · the opt-in quit safety net · closing a modified buffer · reverting · saving over a file that changed on disk · Save As onto an existing file · saving into a folder that does not exist · saving without permission · deleting one or several files · paste name conflicts, single and batched · loading a large file in a fixed encoding · scanning a large file for line numbers.

Each was driven by hand in a terminal and screenshotted, twice: once for the dialog itself, and again after the keyboard rework below.

Query-replace keeps its bottom-row prompt. It is the one confirmation whose job is to keep the buffer visible while it asks, and a card over the match would hide the thing being confirmed.

## Decisions and trade-offs

**Reuse the existing UI tree rather than build a dialog widget.** Cheaper, and consistency with the trust modal is free. The cost is that the dialog is only as capable as those primitives — there is no bespoke layout escape hatch — which so far has cost nothing beyond a wrapping rule for the button row on narrow terminals.

**One dialog for all fourteen, not fourteen dialogs.** Every prompt reads the same way and every safety rule has one home. The trade-off is that a prompt wanting a genuinely different shape would have to either fit this mould or leave it.

**Keep the answer protocol; derive the keyboard from the labels.** The buttons feed the handlers the same answers the old single letters did, so the change is confined to *asking*. The letters themselves are no longer carried around as data: each button's accelerator is picked from the text it displays and shown underlined in it, so nothing is advertised that does not work and nothing works that is not advertised. Letters are picked per dialog rather than from a fixed global table, with the retreat choosing first so `C`ancel is the same key everywhere; the price is that a label rename can move an accelerator, which the tests pin.

**Only ASCII is offered, and a label with none is numbered.** A terminal delivers `s` as a bare key press, but `保` or `ก` arrive through an input method if at all, so marking one would advertise a key nobody can press — and in ja/ko/zh-CN/th *every* label is written that way. Those buttons get a visible `(1)` instead, the digit underlined like any other accelerator, retreat numbered first so `1` is the way out. This also removes a trap in the Latin locales: nothing can hand `Ü` to one button and `ü` to another, which German's paste conflict did — the underline sat on Overwrite while the `ü` a keyboard produces answered Overwrite All. The visible cost is that German accelerators are now mid-word letters (`Ü`**b**`erschreiben`) rather than initials.

**Safety rules belong in the shared layer, not at the call sites.** Driving every prompt by hand found three that opened with a destructive button armed, because the rule against it was a convention each caller had to remember. It is now structural: a dialog refuses to open on an outcome that loses work. This is the main behavioural decision in the PR and the reason for the manual sweep.

**Arm the safe button rather than nothing.** NN/g advises against defaulting a confirmation to "yes", and the W3C dialog pattern requires focus to start inside the dialog and suggests it land on the least destructive action. A terminal has no pointer-only mode — there has to be somewhere for the keyboard to be — so the two are reconciled by always arming, but never arming something that loses work.

**The question scrolls; the answers do not.** The buttons sit outside the card's viewport, so a card too short for its message clips the message rather than the row of outcomes. A confirmation with no visible answer is the fault this whole change exists to remove, and a tall question on a short terminal is exactly where it would otherwise have come back.

**The scrim blocks but does not dismiss.** A press outside the card is swallowed rather than treated as an answer, matching the trust modal. A misaimed click should not answer a question about unsaved work.

**Rework the existing strings rather than add a parallel set.** The prompts' text is reused with the letter legends removed, which keeps fourteen prompts' worth of translation work rather than doubling it. Because the outcomes are buttons now, variants that existed only to offer one extra answer collapsed — the quit prompt went from eight phrasings to four.

## Behaviour changes worth a reviewer's attention

- Which button `Enter` takes changed on several prompts, and destructive outcomes are never armed.
- The single-letter answers changed: every dialog's keys now come from its own button labels and are shown in them. The old unadvertised aliases are gone.
- In locales whose labels carry no ASCII (ja, ko, zh-CN, th) the buttons are numbered `(1)`…`(n)` and answer to those digits. Previously those locales advertised keys no terminal can send.
- `Esc` on the large-file scan prompt now dismisses the question rather than choosing an answer.
- Buttons highlight under the mouse, and the dialog holds the keyboard against vi mode and the emacs keymap alike.
- All 15 locales are updated with real translations, not placeholders.

## Testing

23 end-to-end tests that drive keys and the mouse and assert only on rendered output — including two under vi mode and one under the emacs keymap — plus 24 unit tests on the shared model and the dialog's geometry. Every regression test here was checked to fail without its fix.

Beyond the automated suite, all fourteen prompts were driven by hand in a real terminal after the keyboard rework, and the German, Japanese and short-terminal cases again after the accelerator and layout fixes.

https://claude.ai/code/session_01FbFbBBtdQB8Kg6dFGatex4


sinelaw:
`test macos-latest` is red on `d79cfe3`, on a failure that isn't this PR's.

**What failed:** `e2e::plugins::review_diff_line_staging::test_review_visual_discard_status_is_localized`, TIMEOUT at nextest's 180s cap. 9840 of 9841 tests passed; that one hung rather than asserting.

**Why it isn't this PR's:**

- This branch touches no review-diff or plugin source. The only test-side change outside the confirmation dialog's own files is `+7` lines in `test_plugin_i18n.rs`, and the locale diff contains zero `status.lines_*` lines — which is the key family this test guards.
- The "confirmation" the test waits on is a **status-bar message** emitted by the review-diff plugin after a line-level discard, not one of the confirmation dialogs this PR reworked. No modal, no prompt, no accelerator anywhere in that path — so the keyboard changes here cannot reach it.
- `test ubuntu-latest` passed this same test on this same commit, and `test macos-latest` passed it on both preceding commits (`515eb5f`, `1c91ae0`). It hangs on an indefinite `wait_until` for an async plugin repaint, which is the shape of a timing flake.

**Re-run:** I tried to re-run the failed job and GitHub refused with `403 This workflow is already running` — `test windows-latest` is still in flight in run `35330238641`. I'll re-run the macOS job once that run completes and report the result here rather than leaving it.

No fix ported, because there is nothing here to port: the failure is a pre-existing timing sensitivity in a plugin test unrelated to this change, and widening this PR to harden it isn't warranted. Flagging it rather than staying quiet.

---

sinelaw:
Following up on the above: the re-run of `test macos-latest` **passed**, so `test_review_visual_discard_status_is_localized` was the timing flake the evidence pointed to and not a fault in this branch.

The PR is now green on all 13 checks on `d79cfe3`, with the full suite passing independently on ubuntu, macOS and Windows. No conflict with `master`.

That was the single re-run for this failure; nothing was changed in the branch to obtain it.

---

PR #3321, opened 2026-09-18
The dock is host chrome: declared by its plugin's manifest, remembered across launches

sinelaw:
## Before

Start `fresh` with the workspace dock on (Orchestrator mode, the default; or any launch with `autoOpenDock`, also the default). The editor comes up filling the whole terminal. Then the dock appears on the left and shoves everything over. In Orchestrator mode that's a couple of frames; on an ordinary `fresh .` it's however long the plugins take to load, which is hundreds of milliseconds.

## After

The dock's column is there from the very first frame, at the width the dock will have, and the dock's contents fill it in place. Nothing moves. The dock also remembers whether you left it open and how wide you dragged it, across launches.

## Why it happened

The dock isn't drawn by the editor itself — the orchestrator plugin mounts it from its `ready` hook. That hook is sent to the plugin thread and not waited on, and it runs only after every plugin has loaded. So the editor paints its first frame before the dock exists, and the dock's arrival re-flows the layout.

## The approach

The dock's *content* is the plugin's; the *column* is the editor's. The editor now decides, before any plugin runs, whether the column is open and how wide — the way it already restores the explorer before the first paint — from three things, in order:

1. **What the plugin declares.** A new sidecar, `orchestrator.manifest.json`, read by the host at plugin discovery (before any plugin code runs, on both the sync and the deferred load paths). It says: this plugin fills the dock, `autoOpenDock` is its switch, and here is the width rule. The same place a plugin already keeps its `.i18n.json` and `.schema.json`.
2. **What the user left.** A small `chrome.json` in the data directory remembers whether the dock was open and the width it was dragged to. `open` is recorded at quit (the one moment "the way you left it" is literally true — a plugin's own transient close-and-reopen is not a decision); the width is written as a drag ends. Both are read at construction.
3. **The launch.** A bare `fresh` in Orchestrator mode always opens it.

The plugin then mounts at `ready` iff the host says the slot is open (`editor.dockOpen()`), into a column that is already there, laid out to the width the host reports (`editor.dockCols()`). A column held for a dock that never arrives is handed back when the hook completes.

Two things got simpler along the way:

- The dock's width had three authors that disagreed: the plugin's responsive rule, a fixed 32 columns in the host's mount path (then re-widened by the plugin after mounting — a second, smaller reflow), and the panel placement's own copy. It is now one rule, owned by the host, derived in one place and read every frame. The plugin's copy and its resize→`dock_width` re-issue are deleted, and `PanelPlacement::LeftDock` no longer carries a width.
- The empty column paints its own background and divider, so it reads as an empty dock rather than a bare strip of terminal.

`autoOpenDock`'s meaning shifts slightly: on, the dock comes back the way you left it (first launch: open); off, it stays closed until you open it. Orchestrator mode overrides both, as before. A dragged width sticks — across resizes and launches — until the next drag.

## Tests

- Manifest parsing (defaults, unknown fields rejected, disabled plugins skipped, first directory wins) and the width rule.
- The startup decision: a declared dock is carved at construction at the declared width; no declaration carves nothing; the named setting switches it off (except in Orchestrator mode); a disabled plugin declares nothing; what the user leaves at quit — open, closed, dragged width — is what comes back, and a transient close undone before the quit is not remembered; the hook's sentinel releases a column nothing mounted into.
- The reserved column paints its ground and divider; an unreserved empty one paints neither.
- End-to-end with the real plugin: for both a bare `fresh` and an ordinary launch, the editor's chrome starts at the same column before and after the dock lands; and the dock is remembered across three launches (open → closed → open) against the same data directory, each ended with the harness's quit.
- The harness releases the startup column by default (it fires no startup hooks); tests that model a launch opt in with `with_startup_chrome()`.

## Review

A simplicity pass over the diff produced eleven inline findings (self-review on this PR); all are addressed in f59e6b5 except the `DockSlot` enum, which is left as a follow-up — the reachable inconsistent state it would prevent is closed by clearing the reservation in the slot-move op, with `dock_slot_reserved()` as the flag's one reader.

https://claude.ai/code/session_01QKg1eHNVVyUMi5T53w9xTz


sinelaw:
CI status on a6b0e84: `test ubuntu-latest` is red on one timeout — `e2e::plugins::review_diff_line_staging::test_review_visual_stage_only_selected_line_of_hunk` hit nextest's 180 s cap; the other 9871 tests passed, macOS is green, Windows is still running.

That test isn't this PR's: it loads only the `audit_mode` plugin (no orchestrator manifest, so no dock column) and never fires `ready`, so nothing in this diff is on its path. It's the same shape as the earlier per-run timeouts that hit a different pty-heavy test each time (two vi-mode tests on the first head, a markdown-compose test on macOS) — runner load, not the change. The one timeout that *was* this PR's (`stdin_spool_lifetime`, reproducible across heads) is fixed in a6b0e84.

I'll use the single re-run of the failed job once the workflow finishes (GitHub refuses a re-run while Windows is still in progress).

---

crates/fresh-editor/src/app/plugin_dispatch.rs
```diff
@@ -6072,6 +6101,8 @@ impl Editor {
         // it, so clearing on its close would only cause a visible flicker.
         if slot == super::PanelSlot::Dock {
             self.request_full_redraw();
+            // The user closed the dock: remember that across launches.
+            self.persist_dock_chrome();
```
sinelaw: **Bug: programmatic closes are remembered as the user's choice.** This runs on *every* dock unmount, but `orchestrator.ts` unmounts the dock programmatically and re-mounts it later (e.g. `closeOpenDialog()` before `attachToWorktree(...)`, with the dock coming back via `showDockUnfocused()` on success). Quit — or crash — inside that window and `chrome.json` says `open: false` for a dock the user never closed; the next `fresh .` neither reserves nor opens it.

The host can't tell a user's Toggle Dock from a plugin's transient close here. What it *can* know is the state at quit, which is literally "what the user left it as": record `open` in the quit path (next to `save_all_windows_workspaces`) instead of on mount/unmount, and keep the eager write only for the dragged width on `GripRelease`. That also removes this write site and the one on mount (L5947).

---

crates/fresh-editor/src/app/shell_host.rs
```diff
@@ -3086,15 +3086,13 @@ impl Editor {
...
+                    // responsive re-fit never snapped it back); the release
+                    // is when it is worth a write to disk.
                     Grip::DockWidth => {
                         self.dock_resizing = false;
-                        if let Some(crate::app::PanelPlacement::LeftDock { width_cols }) =
-                            self.dock.as_ref().map(|f| f.placement)
-                        {
-                            self.dock_width = Some(width_cols);
-                        }
+                        self.persist_dock_chrome();
```
sinelaw: **Bug: a press+release on the reserved column's grip writes `open: false`.** `grip_strip` always wraps `draggable` (`view/shell/dock.rs`), so while the column is merely held open (`dock_reserved`, `dock == None`) the pointer path still runs `DockResizeBegin` → `GripRelease` → `persist_dock_chrome()`, which records `open: Some(self.dock.is_some())` = `false`. If the plugin's mount then never arrives, `release_startup_dock_reservation` deliberately remembers nothing — but the file already says closed, contradicting "the column going away is not the user's doing".

Goes away with the quit-time `open` write suggested at `plugin_dispatch.rs` L6105: this release should persist only the width, and only when a dock is actually mounted.

---

crates/fresh-editor/src/app/mod.rs
```diff
@@ -1493,11 +1492,35 @@ pub struct Editor {
...
+    /// lands. So the host decides at construction whether the slot is open —
+    /// from the plugin's manifest, what the user left it as, and the launch
+    /// mode (`Editor::apply_startup_dock_chrome`) — and lays out for it from
+    /// the first frame. The plugin's mount then fills the column in place.
+    ///
+    /// Cleared by that mount, or by the `ready` hook's `HookCompleted`
+    /// sentinel when nothing mounted: every command the handlers sent is
+    /// ahead of it in the channel, so a column still empty then belongs to
+    /// nobody and goes back to the editor.
+    pub(crate) dock_reserved: bool,
```
sinelaw, line 1502: **Design: `dock_reserved: bool` beside `dock: Option<_>` is a three-state slot in disguise.** The invariant "reserved ⇒ `dock` is `None`" is enforced only at `MountFloatingWidget` (`plugin_dispatch.rs` L5946); the sidebar→dock slot-move op (`"dock"` on a `PanelSlot::Sidebar`, ~L6211) puts a panel into `self.dock` without clearing the flag, so `(Some, true)` is reachable. Every reader then re-guards: `render.rs` (frame description and `compute_dock_split`), `chrome/dock.rs` (`release_startup_dock_reservation`), and the snapshot update in `plugin_dispatch.rs`.

Minimum fix: clear `dock_reserved` wherever `dock` becomes `Some`. The real fix is a `DockSlot::{Closed, Reserved, Mounted(FloatingWidgetState)}` enum, which deletes the four guards and makes the invariant a type rather than a convention — reasonable as a follow-up.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
```diff
@@ -6113,6 +6144,21 @@ impl Editor {
...
+        // is answered before the panel is borrowed. An explicit width — it
+        // sticks, like a drag, until the dock is next opened from scratch
+        // with nothing remembered. The dock itself no longer needs this to
+        // be responsive: the host re-reads its width rule on every frame.
+        // A no-op unless the panel is the dock.
+        if op == "dock_width" {
+            if slot == super::PanelSlot::Dock {
+                let max_cols = self.terminal_width.max(20).saturating_sub(20).max(10);
+                self.dock_width = Some((arg.max(0.0) as u16).clamp(10, max_cols));
+                self.persist_dock_chrome();
```
sinelaw: **Semantics + a wrong comment.** An explicit width is now remembered across launches and nothing ever clears it back to `None`, so after a single drag (or one `dock_width` op) the manifest's responsive rule is permanently out of play — on a 300-column terminal the dock stays at the 38 it was dragged to on a 120-column one. That's a defensible choice (a dragged sidebar sticks in most editors), but the comment's "until the dock is next opened from scratch with nothing remembered" describes a state that never recurs; either say it sticks for good, or give the user a way back to the rule (e.g. a double-click on the divider clearing `dock_width`).

Separately, the clamp: `terminal_width.max(20).saturating_sub(20).max(10)` is a second bare-literal copy of `frame::dock_width`'s `frame_width - EDITOR_MIN` rule (the first is in `handle_dock_resize_drag`), and it's redundant — `compute_dock_split` clamps whatever `dock_width` holds on every frame. Store `arg as u16` (or share one `clamp_dock_width` with the drag path) and let the frame rule be the only one.

---

crates/fresh-editor/src/app/chrome/dock.rs
```diff
@@ -1,6 +1,203 @@
...
+            tracing::warn!("chrome state: failed to rename {tmp:?} → {path:?}: {e}");
+        }
+    }
+
+    /// The dock column's width as the host would carve it now — with a
+    /// panel in the slot, held open for one, or neither: what a dock
+    /// *would* get. `0` when the terminal is too narrow for one at all.
+    /// What the plugin lays its content out to, before its mount has
+    /// been processed as much as after.
+    pub(crate) fn dock_cols_if_open(&self) -> u16 {
```
sinelaw, line 200: **Duplication: the requested width is derived in two places.** `dock_width.unwrap_or_else(|| dock_width_rule.width(w))` is computed here *and* in `compute_dock_split` (`render.rs`), and `lifecycle.rs::dock_cols()` is a third accessor differing only by the slot-open check — while the comment above promises "one number from one place". A future tweak (say, clamping a remembered width against the rule's max) has to land in both or `dockCols()` and the painted column disagree by a frame, which is the exact class of bug this PR removes.

One `requested_dock_width(frame_width)` helper that both read, or `dock_cols_if_open` expressed as `compute_dock_split` with the slot forced open, keeps it single-sourced.

---

crates/fresh-editor/src/view/shell/frame.rs
```diff
@@ -381,6 +389,74 @@ pub fn fixed_rows(menu_bar: bool, status_bar: bool, search_options: bool, prompt
...
+        DOCK_MIN
+    }
+    fn default_max() -> u16 {
+        DOCK_MAX
+    }
+
+    /// The width this rule gives a frame `frame_width` wide. A frame of
+    /// unknown width (zero) gets the fixed default rather than the floor.
+    pub fn width(&self, frame_width: u16) -> u16 {
+        if frame_width == 0 {
```
sinelaw: **Dead branch.** Both production callers (`compute_dock_split`, `dock_cols_if_open`) pipe this result through `frame::dock_width`, which returns `None` for any frame narrower than `EDITOR_MIN + DOCK_MIN`, so a zero-width frame yields no dock whatever `width(0)` returns. The branch, the `DOCK_DEFAULT` constant it returns, and the `dock_default_width()` wrapper at L457 (no non-test caller — tests can call `DockWidthRule::default().width(..)`) are carried for a case that cannot occur. Ported from the plugin's `getScreenSize().width <= 0` guard, which the host doesn't need.

---

crates/fresh-editor/src/app/chrome/dock.rs
```diff
@@ -1,6 +1,203 @@
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub(crate) width: Option<u16>,
+}
+
+/// The file. Editor-global like the dock itself, and shared by every editor
+/// this data directory serves, the way plugin global state is.
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+struct ChromeState {
+    #[serde(default = "chrome_state_version")]
+    version: u32,
```
sinelaw: **Cruft.** `version` (and `chrome_state_version()`) is serialised, defaulted, and never read or compared anywhere — a version field with no migration path is speculative. `PersistedWindows` earns its `version` with an actual v1→v2 loader; this one doesn't. Serde defaults already handle additive changes to a two-field file.

---

crates/fresh-editor/src/app/chrome/dock.rs
```diff
@@ -1,6 +1,203 @@
...
+            read_dock_chrome_state(&*self.local_filesystem, &self.dir_context.data_dir);
+        self.dock_width = remembered.width;
+
+        // One slot, so one declaration; by name, so two plugins claiming it
+        // resolve the same way on every launch.
+        let mut declared: Vec<(&String, _)> = manifests
+            .iter()
+            .filter_map(|(name, m)| m.chrome.dock.as_ref().map(|d| (name, d)))
+            .collect();
+        declared.sort_by(|a, b| a.0.cmp(b.0));
```
sinelaw: **Simplify.** Twelve lines and a `Vec<(&String, _)>` with an explicit `sort_by` closure to say "the lowest-named declaration wins, warn if there was more than one":

```rust
let mut docks = manifests
    .iter()
    .filter_map(|(n, m)| m.chrome.dock.as_ref().map(|d| (n, d)));
let Some((name, decl)) = docks.clone().min_by_key(|(n, _)| *n) else { return };
if docks.count() > 1 { tracing::warn!(...) }
```

No allocation, and harder to get subtly wrong.

---

crates/fresh-editor/plugins/orchestrator.manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+  "chrome": {
+    "dock": {
+      "open": true,
```
sinelaw: **Nit.** `a_declared_rule_fills_in_from_the_default` establishes that a manifest states only what it changes, and `"open": true` is the default. Worth keeping the width rule spelled out — declaring the dock's geometry is the file's whole point — but this line adds nothing.

---

crates/fresh-editor/plugins/lib/fresh.d.ts
sinelaw, line 1: **Stale API docs.** `floatingPanelControl`'s doc (around L5022) still promises `"dock"` (`arg` = width in columns), but the host now ignores that argument for `"dock"` — the column is the editor's width — and the op that *does* set a width, `"dock_width"`, is documented only in Rust comments. A third-party plugin following the `.d.ts` would call `floatingPanelControl(id, "dock", 30)`, get the remembered/rule width, and have no documented way to ask for one. The source is the doc comment on `floating_panel_control` in `quickjs_backend.rs`; fix it there and regenerate with `cargo test -p fresh-plugin-runtime write_fresh_dts_file -- --ignored`.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
```diff
@@ -5914,8 +5938,13 @@ impl Editor {
         // of the chrome. Run the single layout funnel so terminals and
         // viewports reflow to the post-dock width right away (a centered
         // panel leaves `dock_cols` at 0, so this is a cheap no-op there).
+        // A column held open for this very mount was already carved, and the
+        // funnel finds nothing to move.
         if as_dock {
+            self.dock_reserved = false;
             self.relayout();
+            // The user has a dock: remember that across launches.
+            self.persist_dock_chrome();
```
sinelaw: **Asymmetric with the slot-move op.** A mount into the dock slot persists `open: true` here, but the sidebar→dock `"dock"` op (~L6211) also puts a panel into `self.dock` and persists nothing — so a dock reached that way isn't remembered. Moot if `open` moves to the quit-time write (see L6105), which reads the slot rather than tracking every way into it.

---

sinelaw (commented):
Self-review of the dock rework, from a simplicity/cruft pass (`/code-review`). Eleven inline notes on a6b0e84, grouped:

- **Two real bugs in the persistence** — every dock unmount (plugin-driven ones included) and even a press+release on the empty reserved column's grip record `open: false` as if the user had closed it. One fix covers both: record `open` at quit, keep the eager write only for the dragged width.
- **One reachable inconsistent state** — `(dock: Some, dock_reserved: true)` via the sidebar→dock slot-move op; cheap to close, and a `DockSlot` enum would remove the four re-guards for good.
- **Duplication and dead code** — the requested width derived twice, a clamp literal copied from the drag path, a zero-frame branch and wrapper only tests can reach, an unread `version` field, a Vec+sort where `min_by_key` will do, a redundant `"open": true`.
- **Stale API docs** — `.d.ts` still says `"dock"`'s argument is a width; `"dock_width"` is undocumented.

The one I'd push back on is "a dragged width is remembered forever": intended, but the comment describing when it resets is wrong. Everything except the enum refactor is small; happy to land it as one commit.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
sinelaw: Fixed in f59e6b5: `open` is recorded at quit (`Editor::save_dock_chrome`, on both quit paths and before a daemon rebuild); the mount/unmount writes are gone. A unit test now covers a transient close undone before the quit not being remembered.

---

crates/fresh-editor/src/app/shell_host.rs
sinelaw: Fixed in f59e6b5: the release writes only the width (`persist_dock_width`), and only when a dock is mounted; `open` is a quit-time fact now.

---

crates/fresh-editor/src/app/mod.rs
sinelaw: Minimum fix landed in f59e6b5: the slot-move op clears the reservation when it puts a panel in the dock slot, and `dock_slot_reserved()` is the flag's one reader (the four re-guards are gone). The `DockSlot` enum is left as a follow-up.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
sinelaw: Fixed in f59e6b5: the comment (and the field's doc) now says a dragged width sticks across resizes and launches until the next drag; the clamp is one `clamp_dock_width` against `EDITOR_MIN`, shared by the drag and the op.

---

crates/fresh-editor/src/app/chrome/dock.rs
sinelaw: Fixed in f59e6b5: `requested_dock_width(frame_width)` is the one derivation; `compute_dock_split` and `dock_cols_if_open` both read it.

---

crates/fresh-editor/src/view/shell/frame.rs
sinelaw: Removed in f59e6b5: the zero-frame branch, `DOCK_DEFAULT` and `dock_default_width()`; tests use `DockWidthRule::default().width(..)`.

---

crates/fresh-editor/src/app/chrome/dock.rs
sinelaw: Removed in f59e6b5.

---

crates/fresh-editor/src/app/chrome/dock.rs
sinelaw: Done in f59e6b5 — `min_by_key` plus a count, via a small iterator closure so it isn't built twice by hand.

---

crates/fresh-editor/plugins/orchestrator.manifest.json
sinelaw: Dropped in f59e6b5; the width rule stays spelled out.

---

crates/fresh-editor/plugins/lib/fresh.d.ts
sinelaw: Fixed in f59e6b5: the `floatingPanelControl` doc now says `"dock"` ignores its argument and documents `"dock_width"`; `fresh.d.ts` regenerated.

---

crates/fresh-editor/src/app/plugin_dispatch.rs
sinelaw: Moot as of f59e6b5: `open` is read off the slot at quit, so every way into the slot is covered.

---

PR #3327, opened 2026-09-19
Sidebar panels stay with the window and file they belong to; keyboard can reach them

### LINKED ISSUE #3326: Sidebar sections: unreachable by keyboard, editor-global across windows, and stale/dead panels
Seven defects in the sidebar-sections feature, found while asking a simple question ("can I reach the Contents panel of an open Markdown file with the keyboard?"). They share one root: a section is editor-global state with no keyboard entry point and no lifecycle tied to what the user is actually looking at. Filing together because the fixes overlap; each is independently actionable.

All reproduced interactively in tmux (200x45) against `ed212f6`, `cargo build --bin fresh` (debug), with an isolated `HOME` and a throwaway project (`notes.md`, `plain.txt`, `long.md`, `sub/deep.md`).

---

## A. No default keybinding for `focus_next_sidebar_section` — plugin sections are palette-only

`Action::FocusNextSidebarSection` (`crates/fresh-editor/src/input/keybindings.rs:697`, implemented at `crates/fresh-editor/src/app/sidebar.rs:484`) is the only way the keyboard enters a plugin section, and it is bound in none of the five shipped keymaps:

```
$ grep -rn focus_next_sidebar_section crates/fresh-editor-core/keymaps/
(no matches)   # default, vscode, macos, macos-gui, emacs
```

Every sibling chrome region has a key; this one doesn't. Command palette, filtered on "focus":

```
Focus File Explorer                        Ctrl+E   Move focus to the file explorer
Focus Next Sidebar Section                          Move the keyboard to the next section of the sidebar
Toggle Orchestrator Dock Focus             Alt+O    Move keyboard focus to or from the orchestrator workspace dock
Toggle Utility Dock                        Alt+J    Move keyboard focus to/from the shared bottom dock
```

It is also undocumented (no mention in `docs/`) and untested (`grep -rn FocusNextSidebarSection crates/fresh-editor/tests` → nothing).

**Consequence:** with `Contents` open next to the file tree, `Ctrl+E` only ping-pongs explorer ↔ editor; the section header alternates between `File Explorer` (focused) and `File Explorer (Ctrl+E)` (unfocused) and `Contents` is never reached. Once the palette command is run, the section behaves correctly — Down browses headings with the buffer cursor unmoved (Ln 1), Enter jumps (Ln 5) and hands the keyboard back — so this is purely a missing entry point.

## B. `markdown_toc` exposes no command to focus its own section

The plugin registers exactly two commands (`crates/fresh-editor/plugins/markdown_toc.ts:837-849`). Palette, filtered on "markdown":

```
Markdown: Set Compose Width
Markdown: Toggle Table of Contents
Markdown: Contents — Follow Cursor/Scroll
Markdown: Toggle Compose/Preview (buffer)
Markdown: Toggle Compose/Preview (all)
```

Nothing focuses `Contents`, and the plugin API it uses (`floatingPanelControl(id, "focus")`) is never surfaced as a command. Combined with (A), a keyboard-only user has no path to the panel at all.

## C. `Contents` reports ON while it is invisible

With the sidebar column hidden, the section is still mounted (auto-open) and the toggle reports state for something with no pixels on screen:

1. `fresh notes.md` (explorer hidden by default — `crates/fresh-editor/src/app/window/mod.rs:2435`)
2. Run `Markdown: Toggle Table of Contents` → status: `Markdown Contents: OFF`, although nothing was ever visible
3. Run it again → status: `Markdown Contents: ON`, still nothing on screen
4. `Ctrl+B` → `Contents` appears under the tree

`mountSidebarSection` has no equivalent of the `reveal_sidebar` + `reveal_sidebar_section` pair that `focus_sidebar_section` (`sidebar.rs:449-452`) performs, so "ON" means "mounted", not "visible".

## D. Sections are shared by every window — one window shows another's outline

`sidebar_sections` is a single list on the editor (`crates/fresh-editor/src/app/sidebar.rs`), so a section mounted while window A is active stays visible in every other window.

1. `fresh notes.md`, `Ctrl+B` → `Contents` shows `Title / Section A / Section B`
2. Run `Extract Tab to New Workspace` on the `notes.md` tab → lands in the new window `demo (2)`
3. Run `Next Window` → back in window A, whose only tab is now `Welcome` (not Markdown, and `notes.md` is not even open here)

Window A's sidebar still shows:

```
├ ▼ Contents ─────────────────────────────────×─┤
│▌ Title                                         │
│   Section A                                    │
│   Section B                                    │
```

Same symptom as the original report: a README's outline surviving in a window where no Markdown buffer is open.

## E. `set_active_window` never fires `buffer_activated`, so plugins can't clean up

`set_active_window` (`crates/fresh-editor/src/app/window_actions.rs:831`) fires `active_window_changed` only. `create_window_with_terminal` does fire `buffer_activated` right after it (`window_actions.rs:656-659`), so the two paths disagree.

`markdown_toc` removes its section only on `buffer_activated` with a non-Markdown buffer, or on `buffer_closed` (`plugins/markdown_toc.ts:768-780`) — neither fires on a plain window switch, which is the mechanism behind (D). Any plugin that tracks "the active buffer" has the same hole, and each one has to re-implement the bookkeeping.

## F. Restored placeholder sections never expire — "Panel unavailable" forever

`restore_sidebar_sections` (`crates/fresh-editor/src/app/sidebar.rs:668`) creates an empty placeholder for every plugin section named in the saved workspace. If the plugin never mounts it again, the placeholder stays forever; only closing it with its `×` removes it.

1. `fresh notes.md` (session persistence on), `Ctrl+B` → `Contents` mounts; quit with `Ctrl+Q`. The workspace records it:
   ```json
   {"kind": {"kind": "panel", "plugin": "markdown_toc", "id": 1}, "title": "Contents", "rows": 10, "collapsed": false}
   ```
2. Restart with the plugin absent (`fresh --restore --no-plugins`):
   ```
   ├ ▼ Contents ──────────────────────────────────┤
   │Panel unavailable
   ```
3. It survives hiding and re-showing the column, and it is saved again on quit, so it comes back on every subsequent restart.

The original report hit this with a dead `Welcome — Contents` (the welcome page wasn't reopened, so `welcome_screen.ts` had no reason to mount it); `--no-plugins` is just the shortest way to reproduce the same class.

## G. Viewport-follow blind spot while the file explorer holds the keyboard

`markdown_toc.ts:736-740` computes `paneFocused = !sectionFocused && data.split_id === getActiveSplitId()`. The host reports focus/blur for the *section*, but nothing reports that the *explorer* holds the keyboard, so browsing the tree looks identical to editing the file. The plugin documents this as the one residual gap against §5.5 (`markdown_toc.ts:61-64`); it is user-visible:

1. Open `long.md` (headings at lines 1/3/31/59/87), `Ctrl+B`, `Ctrl+E` → tree focused, `Contents` selection on `Long Doc`
2. Mouse-wheel the editor pane down until the viewport top is line 32 (inside `## Beta`)
3. `Contents` still points at `Long Doc`

Control: run `Focus Next Sidebar Section` so the section itself has focus, scroll the same way to line 75 — the selection now tracks the viewport (`Gamma`). The outline follows the reader only when the reader happens to be standing in the outline.

## H. Minor: the reveal is silent when it cannot reveal, and the status message forks

`focus_file_explorer` with the column hidden falls through to `toggle_file_explorer` (`crates/fresh-editor/src/app/file_explorer.rs:146-159`), which opens, focuses and reveals the active file. The reveal (`sync_file_explorer_to_active_file`, `:2152`) returns silently in two cases and the user gets no hint why the tree didn't move:

- active buffer has no path (a `[No Name]` buffer) — `:2161`; selection sits on the root row, `sub` stays collapsed
- active file is outside the window root (e.g. `/etc/hosts`) — `:2166`

Status wording also forks by branch: hidden → `File explorer opened` (`:123`), visible → `File explorer focused` (`:155`) — and when the tree has to be built, the async `File explorer ready` (`app/async_messages.rs:1551`) lands on top of `opened` anyway, so the opened/focused distinction is mostly invisible.

## Investigated and *not* a bug

The hidden branch skips `on_editor_focus_lost()` and `cancel_search_prompt_if_active()` that the visible branch runs. I tried to reach it by binding `focus_file_explorer` in the `searchPrompt` context: the key does nothing there, because prompt contexts only honor `Prompt*` actions (`app/chrome/prompt.rs:122-161`), and the palette doesn't offer the command in that context (`input/commands.rs:951`). Noting it so nobody else spends the time.

---

## Suggested direction

- **Bind the cycle.** Give `focus_next_sidebar_section` a default key in the keymaps, duplicated per context the way `toggle_dock_focus` is (`keymaps/default.json:1866-1901`), so it continues the cycle out of the tree and the dock. Add e2e coverage for explorer → sections → editor, which today has none.
- **Let a plugin focus its own section**, and let `markdown_toc` register `Markdown: Focus Contents`. Make `mountSidebarSection` (or a `reveal` control) able to show the column and its section, so "ON" always means visible.
- **One place that announces focus.** Have the editor compare the `(window, split, buffer)` triple after every state change and fire `buffer_activated` / `active_window_changed` when it differs, instead of sprinkling `run_hook` calls that individual paths can forget. That closes (D) and (E) together.
- **Scope sections to a window or a buffer** — store them per window next to the splits and explorer state they already sit beside in the saved workspace, or give `mountSidebarSection` a `scope: { window } | { buffer }` so the host hides/removes the section itself.
- **Expire placeholders**: keep a restored placeholder only until plugins finish loading, plus one pass after the owning plugin's first activation; keep its saved size/position for a later mount, and let a plugin mark a section non-persistent.
- **Tell plugins when chrome has the keyboard** (explorer/dock/section), so §5.5's "the pane does not have focus" is answerable without each plugin guessing.
- **Say something when the reveal can't happen** (unnamed buffer, out-of-root), and unify the explorer status wording.
- **Debug assertion + regression tests**: every visible buffer-scoped section refers to a buffer in the active window; no placeholder outlives plugin load. Tests: open `.md` in window A → switch to a non-Markdown window → outline gone → switch back → outline back; restore a workspace naming a section no plugin mounts → no lingering `Panel unavailable`.

sinelaw:
Fixes #3326.

## The problems

1. **A sidebar panel from one window showed up in every other window.** Open a Markdown file in one workspace and its "Contents" outline appeared in the sidebar of a different workspace — even one with no Markdown file open. Clicking a heading there opened a *second, separate copy* of the other window's file, so edits in the two copies could silently overwrite each other on save.
2. **Plugins didn't notice window switches.** Switching workspaces told plugins "the window changed" but not "the file you're looking at changed", so twelve bundled plugins kept acting on the previous window's file.
3. **Dead "Panel unavailable" placeholders never went away.** A restored session could keep an empty panel forever if the plugin that made it never made it again.
4. **You couldn't reach a sidebar panel with the keyboard.** The command to move focus into one existed but had no key in any keymap, and the Markdown plugin had no "focus the outline" command.
5. **Smaller rough edges.** The outline stopped following the scroll while the file tree had focus; toggling "Contents" reported ON while nothing was visible; focusing the explorer said nothing when it couldn't find the current file in the tree.

## What changes for the user

| Before | After |
|---|---|
| Another window's Markdown outline sits in your sidebar | Each window shows only its own panels; a panel about a file shows only while that file is active |
| Clicking a foreign heading opens a duplicate copy of the file | Can't happen — the foreign panel is not there |
| Switch workspace: status bar, outline, diagnostics may describe the old workspace | Plugins are told the active file changed and refresh |
| "Panel unavailable" stays after restart | Gone a few frames after startup if no plugin claims it; its size/collapsed state is remembered for when it comes back |
| No key reaches a sidebar panel | **Alt+Shift+N** / **Alt+Shift+P** cycle explorer → panels → editor; `Markdown: Focus Contents` jumps straight to the outline (and mounts it first if the file has none yet) |
| Outline stops following the scroll when the file tree is focused | Follows the viewport whenever the editor pane doesn't have the keyboard |
| Ctrl+E on an unnamed buffer or a file outside the project: tree doesn't move, no explanation | Status says why ("no file to reveal" / "… is outside the project") |
| Toggling Contents says ON with nothing on screen | Plugins can ask a panel to reveal the sidebar; "mounted" can mean "visible" |
| Switching sessions or previewing one in the picker could resize your sidebar from the other session's saved layout | A restore only lays out the window you are looking at; other windows' panels wait off screen until you switch to them |

## How it's fixed

- **One place announces focus.** Instead of ten separate code paths each remembering to fire `buffer_activated`, one function compares what's active now with what was last announced and fires the right hooks — including on window switches and window close, which used to be missed. A per-frame safety net catches any path nobody wired.
- **The window travels with every buffer id.** A buffer belongs to one window; now every hook payload and `BufferInfo` says which, and there's a single `active_buffer_changed` hook for "what the user is looking at changed". A plugin that uses another window's id gets a warning instead of a silent no-op.
- **Sidebar panels have a scope.** A plugin says whether a panel is about a buffer, a window, or the whole editor (default: the window it was made in). The editor parks panels that don't match the active window/file and removes them when their buffer or window closes. There is one panel per identity across the column and the parking list. Buffer-scoped panels are never saved to disk, and leftover placeholders expire after startup.
- **The editor says which chrome region has the keyboard** (`chrome_focus_changed`), so plugins don't have to guess; it is announced again once plugins have loaded.
- **Default keys for the sidebar cycle**, a reverse-cycle action, and the Markdown "Focus Contents" command.

Design note with the full reasoning: `docs/internal/plugin-window-scope-plan.md`. Plugin-facing changes are additive; the one behaviour change is that a sidebar panel mounted without a scope is now per-window rather than editor-wide.

Also carries a one-line port of the settings-test fix that #3328 left red on master (five Down presses instead of six on the File Explorer page); it no-ops once master has the same change.

https://claude.ai/code/session_01QEueU8rW1nZ7d81jaiUww4


sinelaw:
CI status on the first full run of this branch (`test ubuntu-latest` on 04ff621): five failures, four of them this branch's and fixed in fd4a84c (locale key parity, the focus-cycle probe, test F's plugin removal, the reveal-feedback status read).

The fifth, `e2e::settings::test_settings_file_explorer_toggles_propagate_to_runtime`, is not this PR's: it is red on master too ([run 35462901024](https://github.com/sinelaw/fresh/actions/runs/35462901024), all three OSes) since #3328 removed "Follow Active Buffer" from the File Explorer settings page and the test still counts six Down presses to reach "Show Gitignored". The one-line fix (five presses, comment updated) is ported into fd4a84c so this branch does not wait on it; it no-ops once master carries the same change.

afe9e04 on top acts on an independent review of the branch: a workspace restore for a non-active window no longer rewrites the live sidebar column, a section identity is unique across the column and the parking list, a remount keeps its section in place, reveal/focus run after the scope check, a divider drag survives the column shrinking, `window_id` is `number` in `fresh.d.ts`, and `markdown_toc` with `autoOpen` off no longer desyncs from a parked section.

---

PR #3331, opened 2026-09-19
Fix four editor gaps: explorer following, Pascal indenting, Odin support, untested word-drag

sinelaw:
## What's wrong today

1. **Double-click-and-drag selects whole words, but nothing tests it.** The feature works. If someone broke it, no test would notice.
2. **The file explorer doesn't follow the file you're editing.** Switch tabs and the sidebar stays where it was. You have to re-focus the sidebar to see where you are.
3. **Pascal indents badly.** Press Enter and you often land in the wrong column. Type `end` and it doesn't line up with its `begin`.
4. **Odin files are only half-supported.** They get colours, and nothing else: Toggle Comment does nothing, Enter after `{` doesn't indent, indentation ignores Odin's tab convention, and there's no way to reach the `ols` language server without writing config by hand.
5. **Scrollbar code left over from the UI rewrite.** Bars all work; the code behind them has a dead painter and a duplicated calculation.

## What changes for you

| | Before | After |
|---|---|---|
| **Explorer follows your file** | Sidebar stays put when you switch tabs | Turn on `file_explorer.follow_active_buffer` and the sidebar highlights whatever file you're in. Off by default. It leaves you alone while you're using the sidebar yourself. |
| **Pascal: Enter** | Copies the previous line's indent ± one level, so a line after a nested `if a then / if b then / X;` stays too deep | Lands at the level the enclosing block actually is |
| **Pascal: typing `end`** | Goes one level left of the line above | Lands on its `begin`. Same for `until`→`repeat`, `else`→`if`, `except`/`finally`→`try` |
| **Odin: Ctrl+/** | Nothing happens | Comments the line with `//` |
| **Odin: Enter after `{`** | No indent | Indents the body — procedures, `struct`/`union`/`enum`/`bit_field`, any block |
| **Odin: Tab** | Editor default | A tab, shown 8 wide, which is Odin's own convention |
| **Odin: language server** | Not offered | `ols` is preconfigured. It does *not* start on its own, because it has to be built from source — flip `lsp.odin.auto_start` once you have it |
| **Word-drag selection** | Works | Works, and now has a test |
| **Scrollbars** | — | No visible change |

Nothing else changes. Every new behaviour is either opt-in (the explorer setting, `ols`) or only reachable by editing a `.pas` or `.odin` file.

## How it's done

**Explorer following** — the editor already had the machinery to reveal a file in the tree, and a single function where the "should we follow?" conditions are checked. What was missing was the setting and anything to trigger it. The trigger is hooked to the one place in the code allowed to change which buffer a pane shows, so tab switches, jump-to-definition and plugins opening files all count without each having to remember. The conditions stay in that one function, which matters because revealing a file over SSH can take seconds — a request that arrives mid-reveal is queued and re-checked when it actually runs, so turning the setting off (or clicking into the sidebar) during a slow reveal is honoured.

**Pascal** — the old code guessed each line's indent from the line above it. Pascal can't be indented that way: how deep a line goes depends on which constructs are still open above it, and the line above can't tell you that. New `indent_pascal` module keeps a small stack of open constructs (`begin`, `case`, `record`, `try`, `repeat`, and the one-statement body a trailing `then`/`else`/`do` promises) and reads the indent off its depth. It's plain Rust — no new parser, no extra binary size — and only ever looks back as far as the nearest top-level declaration or 2000 bytes, so it costs the same in a huge file as a small one.

**Odin** — four table entries, no new machinery: a language entry (comment prefix, extension, tabs at 8), the existing curly-brace indent family, the grammar added to the runtime registry (the build already had it; the two lists had drifted), and an `ols` entry that's enabled but not auto-started.

**Word-drag** — a test that performs the actual gesture and reads the selection off the screen, so it catches a selection whose edges land mid-word. Checked that it fails when the feature is disabled.

**Scrollbars** — deleted a painter nothing called since the UI rewrite, and made the sidebar's thumb calculation call the shared one instead of keeping its own copy. One consequence worth knowing: the sidebar's copy shrank the thumb so it always had room to slide, and the shared one doesn't. A list one row taller than its panel now shows a full-length thumb you can't drag — the same as a buffer one line taller than its pane already does. Wheel and keys still scroll. Say the word if you'd rather the shared version grew the shrink instead.

**A third scrollbar change is not here.** Making the sidebar declare its bar the way every other surface does needs the UI library to change, not just the sidebar, so I stopped rather than bodging it. The sidebar owns its own scroll position and its own idea of how far down it can go (pinned parent folders push the last position past the usual limit), and it draws non-contiguous rows — pinned parents plus a run of children. The library's window owns the scroll position itself and assumes rows are contiguous. Details in the thread; happy to add either capability to the library if you want it.

## Notes

- One pre-existing test, `test_settings_file_explorer_toggles_propagate_to_runtime`, is red on `master` — it counts Down-key presses through the File Explorer settings list, and a removed entry left the count one short. Adding the follow setting back should restore it, since its own comment already expects "Follow Active Buffer" in that position.
- An independent review of this branch found four bugs in the Pascal work (two real mis-indentations, one wrong test, one latent trap in user config inheritance). All four are fixed in `70e8843`.
- Also noticed and left alone: `ScrollbarMouse` has no live caller — its one field is constructed and never read. Removing it means removing its tests, so it's flagged rather than done.

https://claude.ai/code/session_01HVvjkWX8WtjCrxXBgjSbGY


sinelaw:
**`test windows-latest` is red on a failure this PR does not cause, and I am not fixing it here.**

On head `52976dd` the only failure in the whole run is:

```
TIMEOUT [180.063s] (3823/9468) fresh-editor::all_tests
```

The same test times out the same way on **master itself**. Run [35471532339](https://github.com/sinelaw/fresh/actions/runs/35471532339) (master `9848f0c`, which this branch merged in), `test windows-latest`:

```
TIMEOUT [180.035s] (3829/9449) fresh-editor::all_tests
```

That master run is also red on ubuntu, on a different test:

```
TIMEOUT [180.009s] (7456/9941) fresh-editor::all_tests
  e2e::vi_mode_bugs::test_vi_visual_backward_keeps_anchor_character
```

So the shape is one plugin-driven e2e test hanging for the full 180s per test job, a different test per run and per platform, on master as well as on branches. Across the runs on this PR it has been `vi_mode_bugs::test_vi_buffer_switch_leaves_visual_mode` (ubuntu), `focus_announcer::a_window_switch_and_close_fire_the_buffer_hooks` (windows, twice) and none at all on macOS. None of them touches anything this branch changes — `a_window_switch_and_close_fire_the_buffer_hooks` never opens the file explorer and never enters vi insert mode, and it passes locally.

No fix for it exists anywhere yet that I could port — no open PR mentions the test — and the one permitted re-run is spent on the push that produced this run. I have not skipped, quarantined or relaxed anything.

**Everything else is green**, including both test jobs that were genuinely this branch's to fix:

- `test ubuntu-latest` ✅ and `test macos-latest` ✅ — `e2e::settings::test_settings_file_explorer_toggles_propagate_to_runtime` failed on all three platforms before, because this branch adds `file_explorer.follow_active_buffer` and that moves Show Gitignored one row down a list the test walked by counting Down presses. It now finds its rows by the `>` cursor marker the settings dialog draws beside the row that has the keyboard, so it is right under either list, and it reports a screen dump when a row is not where it expected.
- fmt, clippy, doc, schema, actionlint, installer, `check (no plugins)`, rust 1.98, build-musl and supply-chain audit ✅

---

PR #3332, opened 2026-09-19
Everything: a table with a button on every row

sinelaw:
The Everything dialog's rows were a name plus a run-on detail string, so nothing lined up — one row's tool sat under the next row's branch, and the eye had to read along every row. And the only way to act on a row was to select it and press Enter.

**Columns.** Rows now carry cells instead of a label and a detail, measured into columns across the whole answer: a tool under a tool, a branch under a branch. Widths are per family — a heading's columns are not a session's, and an absent tool's are neither. A cell too wide for its column is cut rather than allowed to push the columns after it off the panel, and which end goes is the cell's: a path keeps its tail, a name its head.

**A button per row.** Every row that can be rejoined now carries an `[ Import ]` button at its right edge, opening the same prefilled New Workspace form Enter does. Rows with no way back (an unknown agent, an absent tool, a problem line) carry none.

**The button is a host feature, not drawn text.** `TreeNode.action` renders `[ label ]` at the right edge of the width the tree is given, outside the window the body is fitted into, so a row too wide to fit slides *under* its button rather than pushing it off. A click fires a new `action` widget event with the row's index and key, having first moved the selection there; the keyboard still reaches the same thing through `activate`. The web UI draws and routes the same thing, and `fresh.d.ts` is regenerated.

**A scrollbar that covered it, and the missing third shape.** A revealed bar floats over its window's last column — deliberately, so a bar that comes and goes does not reflow the row being reached for. That is right where the row's last cell is padding and wrong where it is content, so the first cut of the button lost its `]` the moment the tree overflowed. `fresh-ui` had two shapes and needed three: reveal-on-attention (covers), always-drawn with a reserved column, and — missing — revealed *and* reserved, which neither covers nor moves anything. It was missing only because `scrollbar_revealed` cleared the stable gutter, so the two could not be asked for together; now they compose, and the layout reserves the column for an overlay bar that asks for one. A window that asks for neither is unchanged — which includes the panel's trees: reserving the column across a whole tree took one from *every* row, and a row that no longer fits the panel it was built for is worse than a bar over a blank cell (the file-review rows lost their status letter, the dock's hover band stopped a column short, and both tests were right to say so). So only a row carrying a host-drawn button is built a column short. The dock's card list, which has always asked for both, now gets the column it asked for, which stops a card's own right border being painted over.

**Docs.** `docs/internal/tmux-native-sessions.md` — what native tmux support would take, as three rungs: the rejoin flow that ships, adopting tmux sessions as dock rows (the discovered-worktree machinery keyed by socket and session name, so there is no import step — plugin-only), and control mode (`tmux -CC`), whose cost is not emulation but terminal ownership: today a terminal owns its PTY, and control mode has one PTY feeding N terminals with input going back as `send-keys`. Level 1 is not implemented here.

Covered by new cases in `plugins/tests/discovery.test.ts` (column widths measured per family, the rendered rows' second column starting at one offset, elision keeping the right end, which rows offer a button), a `render_tree_row` test pinning the button to the width's edge for a row that fits and one that overflows, and a `fresh-ui` test that a revealed bar with a gutter neither covers the rows nor moves them. Driven by hand in a real tmux pane throughout: the columns line up, a mouse click on a row's `[ Import ]` opens the form filled in for that row's session, and a 14-row tree keeps its brackets with the bar in its own lane.

CI on the current head is green everywhere except `test windows-latest`, which times out on `e2e::focus_announcer::a_window_switch_and_close_fire_the_buffer_hooks` — a master test that fails identically on the unrelated #3334; see the comments below.

Not in this PR: the multi-line text and prose viewports in panels still use a floating bar, so a long line's last glyph can be painted over there.

https://claude.ai/code/session_01Q27oaxsT6gTVyPqj8HAkHm


sinelaw:
`test macos-latest` came back red on `3c17ec0` with three failures. Two were mine and are fixed in `fcd5b9d`; the third is not this PR's.

**Mine** — `e2e::orchestrator_dock::dock_compact_row_hover_band_spans_the_dock` and `e2e::plugins::review_diff_ux_bugs::test_files_panel_rows_fit_a_narrow_panel`. Reserving the scrollbar a column across a whole tree took one from *every* row, so a plugin's rows — built to the width they are given — no longer fit the panel (the file-review rows lost their status letter to an ellipsis) and the dock's hover band stopped a column short. Both reproduce locally and both pass now: only a row carrying a host-drawn button is built a column short, and every other row is exactly as wide as before.

**Not mine** — `e2e::settings::test_settings_file_explorer_toggles_propagate_to_runtime`. It fails on this PR's base commit (`2570958`) with a clean tree: that commit removed `file_explorer.follow_active_buffer` from `plugins/config-schema.json`, while the test still counts it when it navigates the File Explorer category by key presses ("Auto Open On Last Buffer Close, Compact Directories, Custom Ignore Patterns, Follow Active Buffer, …"), so its six Downs now land on Show Hidden instead of Show Gitignored and the save never happens. No fix for it exists to port, and it is outside this PR's diff; the repair is on master — either restore the setting or correct the three key-press counts in `tests/e2e/settings.rs` (lines ~1857, ~1906, ~2017). Happy to do that as its own PR if you want it.

The five macOS timeouts in that run (`active_session_card_is_a_seamless_tab_and_follows_focus`, `dock_card_view_draws_card_borders`, `dock_compact_rows_drop_branch_name`, `dock_hint_bar_stays_pinned_after_folder_collapse`, `default_view_setting_card_opens_dock_card`) all pass locally in well under their 180s budget; they look like runner slowness and the new run will say.

---

sinelaw:
Correction to my previous comment, and where `fcd5b9d` stands.

**The settings test is green on CI.** `e2e::settings::test_settings_file_explorer_toggles_propagate_to_runtime` passes on both ubuntu and macOS at this head, so my "fails on the base branch, needs a master-side fix" was wrong as stated — please disregard that ask. What is true is narrower: the committed `plugins/config-schema.json` is missing `file_explorer.follow_active_buffer`, which the editor writes back into that file when it runs, and the test navigates that category by counting key presses. So it passes wherever the schema has been regenerated (CI) and fails against the committed copy verbatim — which is what my local checkouts kept restoring, and what sent me chasing a phantom. Worth knowing about, not a blocker here.

**Current head `fcd5b9d`:** 13 checks green, including macOS (previously red), scene parity, playwright, clippy, schema and musl; Windows still running. Ubuntu ran 9942 of 9943 green — the two regressions this push fixed are gone, and the ssh pending-row test that failed there on the last head passed.

**One red left, a timeout rather than a failure:** `e2e::terminal_resize::test_kitty_split_resize_in_focused_terminal_dock` hit its full 180s on the ubuntu runner. It passes locally in 2.1s and macOS ran it fine at this same commit, and it exercises kitty-protocol resize, which this diff does not touch. I'll spend one re-run on it once Windows finishes; if it times out again I'll treat it as real and investigate rather than call it a flake.

---

sinelaw:
Status on `81e8ea7` (this branch merged with current master): **14 of 15 checks green**, including ubuntu and macOS, whose earlier single-test timeouts did not recur. No re-run was spent.

The one red is `test windows-latest`, and it is the same 180s timeout as on the previous head: `e2e::focus_announcer::a_window_switch_and_close_fire_the_buffer_hooks`, 9450 of 9451 tests passing around it, no assertion failure. Twice on Windows and never on ubuntu or macOS, so I am not calling it a flake.

What I can say about ownership: that test does not exist on this branch's own commits — it arrived with master's `focus_announcer` work (#3326), and CI builds the PR merge commit, which is why it runs here at all. It waits on a fixture plugin rewriting `focus_log.txt` whole and polls for lines, which is the kind of thing that reads differently under Windows file semantics; nothing in this diff (tree row rendering, one `TreeNode` field, a scrollbar-gutter option) touches focus hooks, plugin hooks or file IO. I cannot reproduce it here — this session is Linux-only — so I am not able to confirm that by running it.

The decisive comparison is another PR on the same base: #3334's Windows job is running now. If it times out on the same test, this is master's to fix and not this PR's; if it passes there, I will treat the failure as mine and dig in rather than leave it. I will report either way and keep watching until this PR is merged or closed.

---

sinelaw:
The comparison came back: **#3334 fails on Windows with the identical timeout** — `e2e::focus_announcer::a_window_switch_and_close_fire_the_buffer_hooks`, 180s, 9483 of 9484 tests passing around it. That PR shares this base and touches nothing this one does, so the failure is master's, not this PR's. Nothing to port: it is the same commit on both sides.

Where this PR stands on `81e8ea7`: everything else green on all three platforms — ubuntu and macOS clean, and the two timeouts they showed earlier did not recur. No re-run spent on either.

One observation for whoever picks up that test, since I had to read it anyway. `wait_until` has no deadline of its own — it ticks and polls forever and relies on nextest's 180s slow-timeout to kill it, so a hook that never arrives costs the full three minutes and reports as a timeout rather than as a missing line. And the fixture plugin rewrites `focus_log.txt` *whole* on every hook (`editor.writeFile`) while the test reads the same path every 50ms. On POSIX that interleaves harmlessly, which is why ubuntu and macOS are green; on Windows a write while another handle holds the file open can fail with a sharing violation, the record is lost, and the awaited line never appears. If that is it, the cheap fixes are for the fixture to write to a temp file and rename over the log, or for the test to append rather than rewrite — and, separately, for `wait_until` to carry a deadline so a stuck wait fails with the log it was waiting on instead of hanging. I have not confirmed any of this: there is no Windows here to reproduce on.

Not touching this PR further on CI grounds. It is green apart from that shared failure and is waiting on reviewers.

---

PR #3334, opened 2026-09-19
explorer: declare the tree's window to fresh-ui (controlled offset + pinned rows)

sinelaw:
Rebased linearly onto `master` after #3331 merged (two commits, library first, then the migration).

**What a user can do before vs after.** Nothing changes on screen — the explorer's rows, sticky ancestors, search, compact directories, decorations/status slots, multi-select, preview-on-arrow and `follow_active_buffer` are all untouched, and the bar looks the same. The one new thing: the explorer's scrollbar can now be **pressed and dragged** (it was a painted column the pointer fell through). Wheel, keyboard and reveal behave as before.

## Two commits

**1. `feat(ui)` — the library.** The controlled/uncontrolled pattern `List` already has for selection (`Sel::{Own, At, Empty}` + `.selected()`/`.on_select()`) is applied to two more facts of a viewport:

- **Controlled offset:** `Node::scroll(offset)` + `Node::on_scroll(f)`. `ViewportProps::scroll` becomes a `Scroll` enum — `Own {x, y}` (initial value, then the framework's) or `At(offset)` (the owner's, re-applied at every layout). Every framework move — wheel chain, bar press/drag, anchor commands — now reports through `on_scroll`. Between frames the framework's copy still moves so a run of notches composes; at the next layout the owner's value wins. **A held offset is never clamped**: the ceiling can depend on facts only the owner can evaluate (which rows it pins at an offset it isn't at), so the bar's ceiling rises to meet it and clamping is the owner's job.
- **Pinned rows:** `Node::pinned(&[u32])` on an index-scrolled window. The owner names the items; the window makes room: the published window is the contiguous run under the pins, the ceiling is `count - (rows - pinned)`, and the builder inside is told how many pins the window honoured (`LayoutInfo::pinned`, capped at `rows - 1`) and draws exactly those above the run. Designed for the general sticky-header case, not the explorer specifically.

`List` gets `.scroll()`, `.on_scroll()`, `.pinned()` on the same terms. **Hit-testing:** a pinned row is an ordinary row for the pointer — layout put it where it is and its own handlers answer — so there's no window-row→index arithmetic to get wrong. The bar's content is `ceiling + window` (not the count) in both `paint_over` and `hit.rs`'s `bar_extents`, so track press and thumb agree by construction.

**2. `refactor(explorer)` — the migration.** The rows sit in a `viewport().items(total).scroll(offset).pinned(sticky).on_scroll(..).scrollbar()`. The model keeps everything that decides which rows exist and what they look like:
- `FileTreeView` still owns the offset: a wheel/drag arrives as `UiFact::ExplorerScrollTo`, clamped with `max_scroll_offset` and stored (`scroll_file_explorer_to`).
- The sticky set is the model's (`sticky_display_indices`), handed to the window as `pinned`; the window derives its ceiling from them, and that ceiling is exactly `max_scroll_offset` for the offset it's at.
- Rows are still described by the model for the indices it would window; the `layout_reader` looks each requested index up out of that set (the row builder can't be `'static` — it borrows the tree, caches, theme, config).
- Rows are keyed/named by **display index** rather than screen row, so a pinned ancestor names itself. Press handlers use `get_node_at_index`; the panel-level right-press no longer re-derives a row from the rectangle (it would have named a real off-screen node).
- The hand-built `fn scrollbar` and the `Scroll { max_offset, rows }` numbers are gone; `ExplorerScroll` splits into `ExplorerWheel` (plugin hook + popup dismissal, unclaimed) and `ExplorerScrollTo`.

Module docs in `view/shell/file_explorer.rs` ("What it does not measure") are rewritten to say what the window is and whose; the residue note in `docs/internal/retained-mode-ui.md` is updated.

## Design notes / what I'd flag
- I kept the design as specified; it fits the library's grain. One deviation: `Scroll` has two states, not three — a window is always *somewhere*, so "controlled and empty" has no meaning (the `At(0)` vs `Own` distinction is what matters and is preserved).
- The pinned count travels through `ScrollInfo`/`LayoutInfo` as a new field (`pinned: u16`) — a small public-struct change in `fresh-ui`.
- A stale explorer offset after the tree shrinks (collapse, search) is now clamped through the model in `explorer_body` before rows are described, so the rows and the declared window are one number. Previously the bar and the rows could disagree there.
- Rebase note: #3332 changed the item-scrolled gutter rule to `reserved || (bar && n > rows)`; this PR keeps that and measures the overflow against the run under the pins (`n > rows - pinned`).

## Tests
- `fresh-ui`: new tests in `tests/pointer.rs` (held window reports + snaps back, bar press reports, never clamped, owned window reports), `tests/layout.rs` (pinned window publishes the run and the pinned ceiling; pins never take the whole window; owned pinned window clamps to the pinned ceiling), `tests/widgets.rs` (List pinned rows answer the pointer at their own index; controlled list reports wheel; reveal reports for a controlled window; reveal counts only the run).
- Explorer: the three protected unit tests pass unmodified (`the_thumb_reaches_the_end_only_at_the_models_last_offset` is the ceiling case). New shell unit tests: wheel-over-rows reports, bar press reports, pinned ancestors sit above the run and a press names them. New e2e `tests/e2e/file_explorer_window.rs`: dragging the bar scrolls the tree; the thumb reaches the end exactly when the last row is on screen under the pinned root.
- Locally: `cargo check --workspace --all-targets`, `--no-default-features --features runtime --all-targets`, `fmt`, `clippy --all-features --all-targets` — no new warnings vs base. Drove the real binary in tmux: wheel, sticky `proj` + `src/deep/deeper` rows pinned, and a bar drag moved the tree and the thumb together.
- CI on the pre-rebase head `20d641f` was fully green on all three platforms. On the intermediate base-merged head, `test windows-latest` timed out on `e2e::focus_announcer::a_window_switch_and_close_fire_the_buffer_hooks` — identically to the base branch itself, in a file this PR doesn't touch (see comment below).

https://claude.ai/code/session_01JirV77WNAncHUYcuQ83pw2


sinelaw:
**CI status on `bd87bf7`** (base #3331 merged in): everything green except `test windows-latest`, which times out on exactly one test:

```
TIMEOUT [180.047s] fresh-editor::all_tests e2e::focus_announcer::a_window_switch_and_close_fire_the_buffer_hooks
```

This is not this PR's:
- The **same single test times out identically on the base branch's own head** (`52976dd`, #3331's `test windows-latest` run at 00:18 UTC: `9468 run, 9467 passed, 1 timed out` — same test, same 180 s).
- `tests/e2e/focus_announcer.rs` came in from `master` (`91e17c2` / `ef58f5a`) through #3331's merge; this PR doesn't touch it, `app::focus_announcer`, or the plugin hook path it exercises.
- All 16 tests this PR adds passed on Windows (9484 run vs 9468 on the base), as did ubuntu and macos in full.

No fix PR exists for it yet that I can find (no open PR references `focus_announcer` or the test name), so there is nothing to port. I don't have the means to re-run the job from this session. If the Windows timeout is fixed on `master` or #3331, merging that in is all this PR needs.

---

PR #3340, opened 2026-09-20
fix(cli): a pipe on stdin is input, never keystrokes (#3252)

### LINKED ISSUE #3252: hangs when receiving STDIN without passing args
version 0.5.1

this locks up 
```
echo 123 | fresh
```
_NOTE: This is a user error. When fresh is followed by a dash all is good._

strace shows it spinning on:
```
poll([{fd=0, events=POLLIN}], 1, 50)    = 1 ([{fd=0, revents=POLLHUP}])
stat("XXX/.venv", 0x7ffd113dbce0) = -1 ENOENT (No such file or directory)
stat("XXX/venv", 0x7ffd113dbce0) = -1 ENOENT (No such file or directory)
stat("XXX/.envrc", 0x7ffd113dbce0) = -1 ENOENT (No such file or directory)
stat("XXX/mise.toml", 0x7ffd113dbce0) = -1 ENOENT (No such file or directory)
stat("XXX/.mise.toml", 0x7ffd113dbce0) = -1 ENOENT (No such file or directory)
stat("XXX/.tool-versions", 0x7ffd113dbce0) = -1 ENOENT (No such file or directory)
stat("XXX/Pipfile", 0x7ffd113dbce0) = -1 ENOENT (No such file or directory)
stat("XXX/poetry.lock", 0x7ffd113dbce0) = -1 ENOENT (No such file or directory)
```

It seems fresh doesn't check whether STDIN is a terminal, in which case I would suggest to imply "fresh -".

sinelaw:
Fixes #3252.

`echo 123 | fresh` — a pipe on fd 0 with no file arguments and no `-` — fell between the cases. `fresh -` drains stdin into a buffer and reopens stdin from `/dev/tty`; a bare `fresh` on a terminal hands the launch to the orchestrator; this launch did neither and left the pipe on fd 0.

Two things followed, both fatal. The piped bytes reached the key decoder **as keystrokes**, so `123\n` was *typed* into a buffer — that is the reported `Ln 2, Col 1` — leaving it dirty. Then the writer closed and `poll()` began reporting `POLLHUP`: immediately, forever, and without `POLLIN`, which `services::tty_input::poll_readable` reads as a timeout and asks again. The editor burned a full core, never saw another key, and could not even be quit, so that dirty buffer could be neither saved nor discarded.

A bare `fresh` with stdin redirected now means `fresh -`.

### One correction to the issue's analysis

The report says the piped content "is read fine — that part works". It isn't: on master nothing reads the pipe as a buffer. The `123` on screen is the pipe's own bytes decoded as key presses, which is also why the cursor sat on line 2 (`1`, `2`, `3`, Enter).

### One case the issue doesn't mention

`echo 123 | fresh file.txt` has the **identical** defect on master — `fd 0` is the same dead pipe, the same ~114% CPU, the same dead keyboard, and `123` typed into the file's buffer. The file arguments say what to open, so stdin is still not read as a buffer there, but it is reopened from the terminal so the editor has somewhere to take keys from. Three lines for a second hang with the same root cause seemed worth it; say the word and I'll split it out.

### It is "stdin is not a terminal", not "a pipe"

The test is `!IsTerminal::is_terminal()`, so `fresh < file`, `fresh < /dev/null` and an inherited socket are treated the same way. That's deliberate rather than sloppy: each of them left the same unreadable fd 0 and spun identically, and none is a source of keys. The docs and CHANGELOG say this rather than claiming "pipe".

## What changed

- `initialize_app` (`crates/fresh-editor/src/main.rs`) decides what to do with stdin in one if/else chain: read it as a buffer, refuse the launch, or just take the terminal back off fd 0.
- `stdin_has_data()` → `stdin_is_terminal()`. It only ever tested `!is_terminal()`, and the old name plus the `!stdin_has_data()` call site made a double negative out of it.
- `reopen_stdin_from_tty()` names `/dev/tty` in its error, and no longer no-ops when fd 0 is already closed (see below).
- `crates/fresh-editor/tests/stdin_no_file_args.rs` — six pty-driven tests covering every branch. Plus a CHANGELOG entry under 0.5.2.

## Before / after

tmux 100x24, clean `HOME`, debug build, `echo 123 | fresh`:

| | before | after |
|---|---|---|
| `/proc/PID/fd/0` | `pipe:[7823]` | `/dev/tty` |
| `%CPU` | 114 | 1.4 (7 ticks over 5s, sampled from `/proc/PID/stat`) |
| typing `ZZZ` | nothing; status stuck at `Ln 2, Col 1` | `ZZZ123`; `Ln 1, Col 4` |
| `Ctrl+Q` | nothing, process still alive | quits |

Also checked by hand: `echo 123 | fresh -` unchanged; `echo 123 | fresh named.txt` opens the file with stdin unread and the keyboard live; a bare `fresh` on a TTY still spawns `--server --session-name orchestrator --orchestrator-mode` and attaches.

## The tests fail without the fix

Same compiled test binary, run against the pre-fix `fresh`:

| test | pre-fix | post-fix |
|---|---|---|
| `a_bare_fresh_reads_the_pipe_and_keeps_the_keyboard` | **hangs** (killed at 60s — the spinning editor never writes another frame) | ok |
| `a_file_argument_wins_over_the_pipe_and_still_takes_the_terminal` | **FAILED** | ok |
| `no_controlling_terminal_fails_with_a_message` | **FAILED** | ok |
| `an_explicit_dash_still_reads_the_pipe` | ok | ok |

That last row is the guard on `fresh -` behaving identically before and after.

## Two things a review caught

**The reopen on the file-argument path was best-effort, justified by a claim that is false.** The comment said that path "used to launch (badly) without a controlling terminal", so swallowing the error preserved a working invocation. It preserved nothing — crossterm's `TerminalModes::enable` opens `/dev/tty` itself a couple of hundred lines later, so the launch already died, just with a bare `No such device or address` naming no device. Verified, then made `?`. It was also the catch-all `Err(_)` recovery CONTRIBUTING §9 warns about.

**`reopen_stdin_from_tty` was a no-op when fd 0 was already closed** (`fresh file 0<&-`): `File::open` is handed fd 0 itself, `dup2(0, 0)` is a no-op that *succeeds*, and dropping the `File` closes fd 0 again — leaving stdin dead while reporting success, which `?` cannot catch. Pre-existing, but this branch routes more launches through it. No test: whether `/dev/tty` lands on fd 0 depends on which descriptors startup has already taken, which a test can't pin.

## On size

An earlier revision of this PR put the decision in a `StdinPlan` enum behind a pure `stdin_plan()` function with four unit tests. That was 111 lines to express three lines of boolean logic, and the function's only reason to exist separately was that it could then be unit-tested — circular. It's inlined now. The one thing those unit tests covered that the pty tests didn't, `fresh -` with nothing redirected in, is now an end-to-end test of the real binary refusing the launch, which observes the message and exit status rather than an enum variant.

## Checks

`cargo fmt --all --check`, `cargo clippy --all-features --all-targets` (0 errors), `cargo test --bin fresh` (73 passed), and `cargo test -p fresh-editor --test all_tests -- stdin` (21 passed — the six new ones plus all of `stdin_spool_lifetime` and `e2e::stdin_input`).

https://claude.ai/code/session_01C8Cghr8tR8W3c5ALCy4pwn


sinelaw:
`test windows-latest` is red here, and it is not this PR's.

The job failed with one timeout out of 9490 tests:

```
TIMEOUT [180.036s] (3843/9490) fresh-editor::all_tests
```

The same job fails on **master at this PR's base commit** (`06dafb9dd`, run [35491730595](https://github.com/sinelaw/fresh/actions/runs/35491730595)) with the same test, plus a second one this PR did not hit:

```
TIMEOUT [180.045s] (3791/9486) e2e::focus_announcer::a_window_switch_and_close_fire_the_buffer_hooks
TIMEOUT [180.079s] (6477/9486) e2e::vi_mode_bugs::test_vi_visual_line_indent_stops_at_selection
```

`test windows-latest` is the only red job in that master run — every other job there is green — and master has been red on its last six runs. So this branch's failure set is a strict subset of the base branch's.

It also can't be reached from this diff. `e2e::focus_announcer` drives `EditorTestHarness` in-process and polls a file a fixture plugin writes; it never spawns the `fresh` binary, so it never reaches `initialize_app`, which is the only place this PR changes behaviour. The four tests this PR adds on top of master's 9486 are the pure `stdin_plan` unit tests (microseconds each); the new e2e file is `#[cfg(target_os = "linux")]` and compiles to nothing on Windows.

No fix for the Windows timeout exists that I can point at yet. #3330 ("not re-preparing the bundled plugins every process") is the plausible candidate, since this is a plugin test blowing a 180s budget on a slow runner, but its own `test windows-latest` job is still in progress, so I have nothing verified to port. I'm not touching the test itself — skipping or quarantining it is not on the table, and raising its budget would be an unrelated change riding this PR.

Re-running the failed job once, since this branch missed one of the two timeouts the base branch hit and may well come back green. If it fails again I'll treat that as confirmation and leave it standing here, watched, rather than widen the PR.

---

sinelaw:
The re-run reproduced it exactly — same single test, same 180s, 9489/9490 passed:

```
TIMEOUT [180.024s] (3868/9490) e2e::focus_announcer::a_window_switch_and_close_fire_the_buffer_hooks
```

So it is not a flake, and my one re-run is spent. Two things are now settled that were open in my previous comment:

**#3330 is not the fix.** Its `test windows-latest` has since finished, and it fails on the *same test at the same index* (`(3868/9486)`, 180.063s). So there is nothing to port, and this isn't about plugin-prep wall clock.

**It looks like a real Windows-only bug in the fixture, not slowness.** Four runs across three different heads (master@`06dafb9dd`, this branch twice, #3330) all hang this one test for exactly 180s. A deterministic hang at the nextest budget is what a `wait_until` that can never be satisfied looks like.

The likely reason, from reading the fixture — I could not verify it, having only Linux here, so please treat it as a lead rather than a diagnosis. `tests/plugins/test_focus_log.ts` builds its log path by string concatenation:

```ts
```

`quickjs_backend.rs` documents that `editor.getCwd()` surfaces Rust's `Path::canonicalize` output to plugins verbatim, so on Windows it is `\\?\D:\…\project`. Windows does not translate `/` to `\` inside a `\\?\` verbatim path — that is the whole point of the prefix — so the plugin writes to `\\?\D:\…\project/focus_log.txt` while the test reads `project\focus_log.txt`. `log_lines()` swallows the miss with `unwrap_or_default()`, so the first `wait_until` waits forever instead of failing. `pathJoin` exists for exactly this and has `test_path_join_preserves_unc_prefix` covering the same prefix, so the fixture should go through it:

```ts
```

That is also what CONTRIBUTING §1 asks for — don't hard-code path separators.

I'm leaving that patch here rather than pushing it: it belongs to #3326's test, not to this PR, and I can't run Windows to confirm it before pushing. Happy to send it as its own PR if you'd like — say the word and I will. Everything else on this PR is green and there is no merge conflict; I'll keep watching it.

---

sinelaw:
Merged master in (`8d657af`) now that #3336 has landed, and re-ran. Short version: this branch's CI is identical to master's.

The Windows timeout survived both the merge and #3336's fixture change — same test, same 180s, on a base that already has `replaceFile`:

```
TIMEOUT [180.034s] (3812/9507) e2e::focus_announcer::a_window_switch_and_close_fire_the_buffer_hooks
```

Worth flagging, since #3336's commit message reads as if this test were fixed: the failure it diagnosed was the **macOS** one, caused by that branch making `writeFile` create-only. The Windows failure is older — it is on master at `06dafb9dd`, before #3336 — and `replaceFile` does not touch it. On master's own CI at `a0408d3` ([run 35516742659](https://github.com/sinelaw/fresh/actions/runs/35516742659)), `test windows-latest` is the *only* red job; ubuntu and macos are both green. So master is currently red on Windows on this test, and this PR just mirrors it.

That also leaves my earlier guess still standing as the only candidate I have: the fixture builds `LOG` by concatenating a forward slash onto `editor.getCwd()`, which is a `\\?\` verbatim path on Windows, where `/` is not translated. `editor.pathJoin(editor.getCwd(), "focus_log.txt")` is the one-line change. Still unverified — I have no Windows here — and still not something I'll push onto this PR uninvited.

Everything else here is green and there is no conflict. Nothing further from me on the Windows job unless you want that patch as its own PR.

---

sinelaw:
CI is fully green on `394976d` — all 13 checks including `test windows-latest`, and the PR is mergeable.

**Which means I should retract the lead I gave you earlier.** I said the `focus_announcer` Windows timeout was "not a flake" and pointed at `test_focus_log.ts` building `LOG` by concatenating a forward slash onto a `\\?\` verbatim `getCwd()`. Please don't spend time on that patch: if the path were genuinely wrong on Windows the test would fail every run, and it has now passed twice.

The record, so the next person doesn't have to reconstruct it:

| head | `test windows-latest` |
|---|---|
| master `06dafb9d` | TIMEOUT (+ `vi_mode_bugs`) |
| this PR `ab64100`, attempt 1 | TIMEOUT |
| this PR `ab64100`, attempt 2 (re-run) | TIMEOUT |
| #3330 | TIMEOUT, same test, same index |
| master `a0408d3` | TIMEOUT |
| this PR `8d657af` | TIMEOUT |
| this PR `86ee57c` | green |
| this PR `394976d` | green |

Six consecutive failures then two passes, with nothing in between that touches `focus_announcer` — my commits since `8d657af` are stdin handling, a rename, and tests. So it is load- or timing-sensitive on the Windows runner rather than deterministic, and "reproduces on a re-run" was not the evidence of a real failure I took it for. The one thing that still looks worth keeping from that dig: the test's `wait_until` has no way to fail, because `log_lines()` swallows a read error with `unwrap_or_default()`, so whatever the underlying flake is, it always presents as a 180s timeout with no diagnostic. That is fixable independently of this PR and I'm happy to do it separately if useful.

Nothing outstanding on this PR from my side: green, no conflict, no open threads. It's waiting on review.

---

PR #3341, opened 2026-09-20
fix(render): probe tab_starts by character index, not visual column (#3218)

### LINKED ISSUE #3218: Render issue: CJK characters disappear when using tab indentation
OS: Fedora 44 (Cosmic)
Fresh Version: fresh 0.4.10

When editing a file in fresh editor, there is a rendering bug involving tab indentation and CJK characters. 
If a tab is used right after CJK character (instead of spaces), the character right before the tab disappear from the screen. The text still actually exists in the file, but it becomes invisible in the fresh editor UI.

Please see the attached screencast for a detailed demonstration of the issue:

https://github.com/user-attachments/assets/8e00f003-1dcd-42dd-b1c6-c25f504eba84

sinelaw:
Fixes #3218.

`ViewLine::tab_starts` records the **character index** at which each tab expansion begins. The cell pass looked it up with `col_offset`, the **visual column**. The two agree only while every character on the row is one column wide, so a double-width glyph before a tab pushed the column ahead of the character index and the lookup matched an earlier CJK character instead of the tab — whose glyph `display_cell_text` then replaced with the `→` whitespace marker. The character was gone from the screen while sitting intact in the buffer and on disk.

One line, plus the comment saying which index this set is keyed by. Every neighbouring predicate in the same pass (`is_tab_padding`, `tab_run_continues`, `is_second_tab_column`, `ws_indicator_visible`) already uses `display_char_idx`; this was the one holdout — and `is_tab_padding`'s own doc comment already noted that `tab_starts` "is keyed by character index".

### Before / after

Captured with `tmux capture-pane -p`, 100x20, default config, `tab_size` 4, on the reporter's fixture plus two extra shapes:

| file line | before | after |
|---|---|---|
| `你好⇥world` | `你→    world` — **好 gone** | `你好→   world` |
| `测试文字⇥abc` | `测试→字    abc` — **文 gone**, 字 stranded | `测试文字→   abc` |
| `plain⇥tab` | `plain→  tab` | `plain→  tab` |
| `你⇥好` | `你  好` — **marker gone** | `你→ 好` |
| `a你⇥b` | `a你 b` — **marker gone** | `a你→b` |

The first three rows reproduce the issue's table character for character. The last two are a second shape the same arithmetic caused and the issue did not mention: where the tab's visual column ran past every index in `tab_starts`, the marker was not misplaced but dropped entirely, so a tab was indistinguishable from spaces.

### Also checked

- **`tab_size` other than the default.** Covered by a test at 8; the stops move, the characters stay.
- **Cursor column.** The painted caret was already correct — it comes from `cell_screen_x()`, which uses `col_offset` as a genuine visual column. Walking right along `你好⇥world` puts it at screen x +0, +2, +4, +8, matching the tab stop. The status bar's `Col` is a 1-based *character* index, not a visual column (`Col 3` after 你好, `Col 3` after `pl`), consistently on ASCII and CJK lines. That is a different quantity, not a defect of this bug, and I left it alone.

### Tests

New `crates/fresh-editor/tests/e2e/issue_3218_cjk_tab_marker.rs`, three tests reading painted cells (not buffer text — the characters this bug lost were intact in the file): the reporter's fixture verbatim, the mixed-width shapes above, and the fixture at `tab_size` 8. All three fail on master with `你→    world` against `你好→   world`, and pass with the fix.

`cargo fmt --all --check` clean; `cargo clippy --all-targets -p fresh-editor` reports nothing on the changed files; `cargo test -p fresh-editor --lib` 2566 passed; `--test all_tests` 4914 passed with one unrelated flake (`remote_channel_timeout_tests::test_reconnect_task_holds_the_carrier_process`, which passes in isolation both with and without this change).

Two pre-existing things I ran into and did not touch: the two `fresh-editor-core` doctest failures (`populate_line_cache`, `expand_snippet`), and `docs/visual-regression/screenshots/Comprehensive_UI_B_01_state_b.svg`, which the visual-regression test rewrites on every run with one cursor-cell fill colour changed — identically with and without this change, on an all-ASCII scenario.

https://claude.ai/code/session_014cAGi3DWGPdWwrK1DaPQT6

---
_Generated by [Claude Code](https://claude.ai/code/session_014cAGi3DWGPdWwrK1DaPQT6)_


sinelaw:
CI: 12 of 13 green. `test windows-latest` is red, and it is not this PR's failure.

**What failed.** 9488 of 9489 tests passed; one timed out at the 180s cap:

```
TIMEOUT [180.021s] (3808/9489) fresh-editor::all_tests
```

**Why it isn't this PR's.** The same check, with the same test, is red on this PR's base commit. Master's own CI run on `06dafb9dd` ([run 35491730595](https://github.com/sinelaw/fresh/actions/runs/35491730595)) failed with `test windows-latest` as its only failed job, and **two** timeouts:

```
TIMEOUT [180.045s] e2e::focus_announcer::a_window_switch_and_close_fire_the_buffer_hooks
TIMEOUT [180.079s] e2e::vi_mode_bugs::test_vi_visual_line_indent_stops_at_selection
```

So the failing test already times out on the base with no change from me, and this PR's Windows run has one timeout where the base has two. `test ubuntu-latest` and `test macos-latest` both pass here, including the three tests this PR adds.

The diff is one predicate in the render cell pass (`tab_starts` probed by character index instead of visual column) plus a new e2e test file. `focus_announcer` drives the TypeScript plugin runtime and polls a log file the fixture plugin writes; nothing in that path reads the tab-marker lookup, and the three tests added here run in 0.06s locally.

**No fix to port.** I looked for one. #3330 (plugin-prepare caching, ~23% off suite wall clock) is the nearest candidate and is not a fix for this: its own description records the same 180s timeout still present on its branch and calls it unrelated and passing in isolation. Nothing else open touches `focus_announcer`. Since no fix exists yet, I have not widened this PR with a speculative one — a plugin-runtime hook test hanging only on Windows wants its own root-cause, not a change riding a render fix.

Re-running the failed job once, per the usual one-re-run rule. I'll keep watching this PR either way.

---

sinelaw:
Correcting my comment above: I said no fix existed for the `focus_announcer` Windows timeout. I have since found the root cause, and it is a one-line bug in the test fixture — not a flake, and not a hang in the code under test.

`tests/plugins/test_focus_log.ts:20` builds its log path by string concatenation:

```ts
```

On Windows `getCwd()` returns a `\\?\`-prefixed verbatim path. This repo documents that explicitly, on `pathJoin`, which exists to handle it:

> Preserves up to 2 leading slashes, which matters on Windows: Rust's `Path::canonicalize` returns `\\?\`-prefixed paths, and `editor.getCwd()` surfaces that to plugin code verbatim.

Verbatim paths bypass Windows path normalization entirely, so the appended `/` is not a separator — it is a literal character in a filename. The fixture writes somewhere the test never reads, `wait_until` never sees `buffer_activated`, and the test sits until the 180s cap. That is why it fails only on Windows, and why it fails on master as well as here.

The fix is to use the API provided for exactly this:

```ts
```

I have not pushed it here. It is unrelated to this PR's one-line render change, it fixes Windows CI for every PR rather than this one, and I gather it is already being raised separately — so it wants its own PR, or a maintainer's call on where it goes. Happy to open that PR if useful. Verified only that it still passes on Linux; I have no Windows runner, so the Windows half is reasoned from the documented `getCwd()` behaviour, not observed.

---

sinelaw:
The re-run reproduced the timeout, which confirms it is deterministic rather than a flake — consistent with the root cause above.

So I have opened the fixture fix as its own PR, #3342, and ported that one line here (`76158f7`) so this PR can reach green. It is the same commit on both branches and no-ops once #3342 lands, whichever merges first.

This PR's diff is now the one-line render fix plus its tests, and that one-line test-fixture fix. Happy to drop the second commit if you would rather this PR stayed strictly to the render change and waited on #3342.

---

sinelaw:
All 13 checks green now, `test windows-latest` included — the ported fixture fix is what cleared it, and #3342 (the same one-line change on its own, off master) is green too, which is the cleaner proof.

So the choice I offered above is live: this PR can keep the fixture commit and go in green on its own, or drop it and wait on #3342, whichever you prefer. Your call — nothing else is outstanding here.

---

PR #3342, opened 2026-09-20
fix(test): build the focus-log fixture's path with pathJoin, not concatenation

sinelaw:
`test windows-latest` is red on master. One of its two 180s timeouts has a deterministic one-line cause, fixed here.

## The bug

`crates/fresh-editor/tests/plugins/test_focus_log.ts` built its log path by appending a forward slash to `editor.getCwd()`:

```ts
```

On Windows `getCwd()` returns a `\\?\`-prefixed verbatim path. The repo documents this itself, on `pathJoin`, which exists for exactly this reason:

> Preserves up to 2 leading slashes, which matters on Windows: Rust's `Path::canonicalize` returns `\\?\`-prefixed paths, and `editor.getCwd()` surfaces that to plugin code verbatim. After the backslash→slash normalization the prefix becomes `//?/C:/...`

Verbatim paths bypass Windows path normalization entirely, so that appended `/` is a literal character in a filename, not a separator. The fixture writes somewhere `focus_announcer.rs` never reads; its `wait_until` never sees `buffer_activated`; the test sits until nextest's 180s cap.

The fix is the API provided for this:

```ts
```

## Why this is worth its own PR

`e2e::focus_announcer::a_window_switch_and_close_fire_the_buffer_hooks` times out on master at `06dafb9dd` ([run 35491730595](https://github.com/sinelaw/fresh/actions/runs/35491730595)) and on every branch cut from it, while `test ubuntu-latest` and `test macos-latest` pass. It is not a flake — it reproduced on a re-run on #3341, and the cause above is deterministic. So it turns every PR's Windows job red for a reason unrelated to that PR.

I found it while standing down on it as a pre-existing failure on #3341 (a CJK tab-render fix). It does not belong in that PR's diff, and it helps every open PR rather than that one, so it is here on its own.

## Verification

The test passes on Linux before and after — that platform was never affected, so a local run cannot prove the fix. The Windows half is reasoned from the documented `getCwd()` behaviour quoted above; I have no Windows runner, so **this PR's own `test windows-latest` job is the check that matters**. If it comes back green where master's is red, that is the confirmation.

Master's Windows job has a second timeout, `e2e::vi_mode_bugs::test_vi_visual_line_indent_stops_at_selection`, which this does not address and which I have not diagnosed.

## Related, deliberately not fixed here

`plugins/vi_mode.ts:4950` and `:4984` build `:e` and `:w` paths the same way:

```ts
const path = args.startsWith("/") ? args : `${editor.getCwd()}/${args}`;
```

That is production code, so on Windows `:e somefile.txt` with a relative argument would be affected for real users, not just tests. It wants its own change and someone who can exercise vi mode on Windows — flagging it rather than folding it in.

https://claude.ai/code/session_014cAGi3DWGPdWwrK1DaPQT6

---
_Generated by [Claude Code](https://claude.ai/code/session_014cAGi3DWGPdWwrK1DaPQT6)_


sinelaw:
Confirmed: **all 13 checks green, including `test windows-latest`** ([job](https://github.com/sinelaw/fresh/actions/runs/35516550622/job/106093364676)).

That is the check that mattered. The same job is red on master at `06dafb9dd` and on every branch cut from it; with this one-line change it passes. The `\\?\`-verbatim-path diagnosis holds.

Two corrections to what I wrote in the PR body:

- I said master's second Windows timeout, `e2e::vi_mode_bugs::test_vi_visual_line_indent_stops_at_selection`, was not addressed here and might still fail the job. It did not recur — Windows passed outright. So that one looks like a genuine flake rather than a second deterministic bug, on this run at least.
- I said the Windows half was reasoned rather than observed. It is now observed.

`plugins/vi_mode.ts:4950` and `:4984` still build `:e` / `:w` paths the same way, which would affect real users on Windows. Still out of scope here, still worth someone's attention.

---

sinelaw:
Superseded — closing.

#3341 merged as `64da043` carrying this same commit, so master already has the fix:

```
$ git show origin/master:crates/fresh-editor/tests/plugins/test_focus_log.ts | grep 'const LOG'
```

Nothing here is lost — this PR's diff is now a no-op against master. It served its purpose as the clean proof of the diagnosis: its `test windows-latest` job went green on a one-line change where master's was red, which is what established that the `focus_announcer` timeout was the `\\?\`-verbatim-path bug and not a flake.

Still outstanding and not fixed by either PR: `plugins/vi_mode.ts:4950` and `:4984` build `:e` / `:w` paths by the same concatenation, so a relative-path `:e` on Windows is likely broken for real users. That is production code and wants someone who can exercise vi mode on Windows.

---

PR #3349, opened 2026-09-21
Fold case by default in every search UI, and let each one say otherwise

### LINKED ISSUE #3212: feature request: config preset for case-sensitive search
Workspaces already persist the choice whether a search shall be case-sensitive. For new workspaces, it would be great to define the default in config.json (possibly along with the other search options).

sinelaw:
A search for `todo` now finds `TODO`. Every search surface starts that way, keeps a toggle that overrides it, and remembers what it was told for the rest of the session.

Closes #3212, which asked for the **preset**: workspaces already persisted the find prompt's choice, but a new workspace had to be taught it again, with nowhere to say so once. `editor.search` is that place — `case_sensitive`, `whole_word`, `regex`, `confirm_each`, all off, seeding every surface. It's a default, not a lock: the toggles stay live, and a workspace that saved its own choice ignores the config.

| Surface | Was | Now | Toggle | Remembers |
| --- | --- | --- | --- | --- |
| Find / Replace prompt | case-sensitive | **folds case** | `Alt+C` *(existing)* | window + workspace |
| Universal Search (Live Grep) | smart-case | **folds case** | **new** Case toggle, `Alt+C` | session |
| Search & Replace panel | folded case | folded case | `Alt+C` *(existing)* | **new** — survives reopen |
| Git Grep | case-sensitive (`git grep`'s default) | **folds case** | **new** Case checkbox | session |

Live Grep's smart-case is gone. It derived case sensitivity from whether the query carried an uppercase letter — a guess the user could not see, could not turn off, and which made case-insensitive matching unreachable for a capitalised query. The new flag is threaded through `SearchOpts` to every backend (rg, ag, git-grep, ack, grep) and through the JS-side scopes, so the toolbar, the subprocess and the buffer scan agree.

Two things this needed underneath:

- **`Workspace::search_options` is now `Option<SearchOptions>`.** An all-false struct is indistinguishable from "the user turned everything off", so restoring one stamped over the preset — every workspace without a saved choice silently overruled the config, leaving it unreachable. `None` means "never saved". Old workspace files read as `None`; a saved choice still round-trips. Both directions are pinned by a test.
- **The Live Grep preview highlight kept its own copy of the smart-case rule.** It now reads the Case toggle off the toolbar the plugin mounted, so the plugin stays the one owner of the setting rather than two spellings drifting apart.

Four existing tests assumed the old default: the `Alt+C` test now starts from unchecked; the two Dracula checked-checkbox tests preset the config on, since what they need is a checked option to read the style of; and a large-file find-previous fixture whose filler read "gap between needles" is now "markers", so that test stays about find-previous rather than becoming a test of case folding.

Docs: `docs/configuration/index.md` (new Search section) and `docs/features/search-replace.md` (per-surface table).

- [ ] I agree my contributions are licensed under GPL-3.0-or-later and may be relicensed in the future at the maintainer's discretion (see [CONTRIBUTING.md](../CONTRIBUTING.md#license)).

https://claude.ai/code/session_01TXk9wiq8veLaomvoeovCsH

---
_Generated by [Claude Code](https://claude.ai/code/session_01TXk9wiq8veLaomvoeovCsH)_


sinelaw:
`test macos-latest` is red on 076fe55, and I don't believe it belongs to this PR.

```
Summary [1564.222s] 9942 tests run: 9941 passed (1 slow), 1 timed out, 165 skipped
TIMEOUT [180.012s] (4364/9942) fresh-editor::all_tests e2e::lsp_order::test_did_open_sent_before_hover
```

What I checked:

- **The diff touches no LSP code.** It is `editor.search` config plumbing, the workspace key, three plugins, the search-options row and the Live Grep preview highlight. `git diff --name-only origin/master...HEAD` has no LSP file in it.
- **The test passes locally on this branch in 0.20s** — three orders of magnitude under the 180s limit, so this is a stall, not a slowdown my change could have caused by degrees.
- **It is shaped like a stall.** `test_did_open_sent_before_hover` spawns a fake LSP server subprocess and waits for it to report; if the spawn never comes up, the test waits out `slow-timeout` (60s × `terminate-after = 3` in `.config/nextest.toml`) and nextest kills it. Everything else on that runner passed.

The one thing I could plausibly have made worse is plugin load time — `git_grep.ts` now imports `lib/widgets.ts` for its Case toggle, which every test harness pays. That is a few milliseconds of extra parse, it applies equally to the 9941 tests that passed on the same runner, and it cannot turn a 0.2s test into a 180s hang. I'll revisit it if this recurs rather than assume it away.

There is no retry configured for this suite, so a single stall on a slow runner fails the whole job. I have not pushed anything for this: there is nothing in the diff to fix, and disabling or quarantining the test is not on the table.

Next step is the one re-run, once `test ubuntu-latest` and `test windows-latest` have finished on this head so I'm not restarting jobs mid-flight. If macOS reproduces the same timeout on the re-run, I'll treat it as real and dig into it properly.

---

sinelaw:
`test windows-latest` also went red on 076fe55, on a different test from the macOS one. This one I can show is not this PR's from the base branch itself.

```
Summary [2633.304s] 9511 tests run: 9510 passed (9 slow), 1 timed out, 453 skipped
TIMEOUT [180.044s] fresh-editor::all_tests e2e::focus_announcer::a_window_switch_and_close_fire_the_buffer_hooks
```

Master's own CI at a0408d3 — the merge base of this branch — has `test windows-latest` as its **only** failing job, on that same test:

```
Summary [2513.424s] 9503 tests run: 9502 passed (9 slow, 4 leaky), 1 timed out, 453 skipped
TIMEOUT [180.036s] fresh-editor::all_tests e2e::focus_announcer::a_window_switch_and_close_fire_the_buffer_hooks
```

<https://github.com/sinelaw/fresh/actions/runs/35516742659/job/106093857389> — every other job on that run is green, including `test macos-latest` and `test ubuntu-latest`. So this PR mirrors the base on Windows and adds nothing red there. No fix exists on master to port; the four most recent master pushes are all red the same way.

Where that leaves the three test jobs on this head:

| job | result | verdict |
| --- | --- | --- |
| `test ubuntu-latest` | green, 10002 tests | — |
| `test macos-latest` | 1 timeout, `e2e::lsp_order::test_did_open_sent_before_hover` | not in the diff's reach; green on the base branch, so re-running |
| `test windows-latest` | 1 timeout, `e2e::focus_announcer::a_window_switch_and_close_fire_the_buffer_hooks` | red on the base branch too, on the same test |

Ubuntu passing the full 10002 is also what rules out the one way I could have contributed: `git_grep.ts` now imports `lib/widgets.ts`, which every test harness pays at plugin load. Same commit, same cost, green on Linux.

I've spent the single re-run on the failed jobs. macOS is the one it is really for — that test is green on master, so if it times out again it is worth treating as real and I will. Windows will most likely reproduce, since it reproduces on master; that is not something this PR can fix, and it means this PR cannot show fully green until master's Windows timeout is addressed.

---

PR #3353, opened 2026-09-22
Retire the retained-mode duplicates: the two surfaces that laid themselves out, and the code nothing calls

sinelaw:
Removes the mechanisms the retained-mode migration replaced but left in place — including several where the *old* mechanism was still the one the painter used, with the tree's answer demoted to a `debug_assert` beside it. Each commit is one concern; each "dead" claim was re-verified by whole-workspace grep before deleting.

Seventeen commits, none of which undoes another. (The history was rebuilt from a messier 25 — same tree, byte for byte.)

---

## What was still live

**[`d216416`] Paint a pane into the rectangle the frame placed it in.** The biggest offender. `split_layout` built a throwaway `fresh_ui::Ui`, laid `pane_interior` out again at the pane's box and read four rectangles back by key — once per pane per frame, unconditionally from `paint_leaf`. The tree's answer was in hand as `pass.rects.content(split_id)` and was read only by a `debug_assert_eq!`; the painter used the throwaway's. **In release builds the cells landed in rectangles a discarded tree computed**, and `geometry::stats` could not see it because `split_layout` never counted itself.

**[`5c2ff75`] Make the scroll math read the renderer's gutter width.** `estimated_gutter_width` was a third statement of the formula, under a doc naming the stake. It had diverged twice: it missed byte-offset mode, so a binary buffer's scroll math wrapped wider than the pane did, and its doc described a `show_line_numbers` behaviour the code did not have.

**[`4e97bb9`] Read popup geometry off the tree.** `popup_areas` and `global_popup_areas` both took the outer rect off `popup::rects_of` and then re-derived the content rect by hand, in two copy-pasted blocks of border arithmetic — a second *statement*, not a cache. Three of `PopupAreaLayout`'s seven fields had no reader at all.

**[`85dce79`] Delete the settings dialog's second copy of its own heights.** `ScrollItem for TreeRow` fed `ScrollablePanel::ensure_focused_visible`, which walked every row's height to compute an offset nothing read.

**[`4d54040`] and [`a00fbe4`]** delete the leftovers nothing called, and move a retained-vs-fresh layout claim to a test where both sides really are what the claim says.

## The two surfaces that were still laying themselves out

**[`99f8025`] Let layout fit the status bar.** `status_bar_description` measured every element, reserved the right side, truncated the one that straddled the boundary and dropped the rest — all over measured text, before a single node existed. The description states what is on the bar; who yields is `Node::priority` and where the cut falls is `Elide::Tail`, made at paint against the width layout settled on.

**[`91b338d`] `fresh-ui`: a window scrolls the way it scrolls, on either axis.** A correction to this PR's own earlier claim: `fresh-ui`'s viewport was never vertical-only. `scroll` and `scroll_max` were already `Point`s, the clamp was already per-axis, `arrange` already translated by `sc.x`. What did not exist was any way to move a window across *by reference to its content*. The axis is the window's now (`Node::scroll_axis`), not the command's, so every `Anchor` command became axis-correct at once with no new public command.

**[`1880adb`] The tab strip is a window onto its tabs.** `view::shell::tabs::lay_out` did not use the layout engine at all — it measured every piece, sliced each label by column against an offset the editor held, and emitted fixed `Sizing::Cells` nodes. Because layout never decided the widths, a second full measurement ran behind `ensure_active_tab_visible` (five call sites) and a third in `split_tabs_width`. All gone, with `SplitViewState::tab_scroll_offset` and its persistence.

`<`, `>` and `+` are one kind of button now: same width, same two theme names, all three lighting on hover. A cap is not an element, so `update_hover` asks it the same question the press asks rather than waiting for an `Enter` that cannot come. And the name cap is a *decision*, not a truncation — `Strip::cap_names` compares `natural_width` against the window's outer width from the frame before, so a 26-column name is not elided on a 160-column screen.

## Sweeps

**[`8934ac0`] `fresh-ui`: the API nothing asks for.** Nine `pub fn`s with no reference anywhere. Three were whole mechanisms: `Anchor::scroll_to_end` with its command and arm, `Event::release_pointer` with `Ctl::release_request` and the `apply_controls` branch that cleared the capture, and `Commands::is_bound`.

**[`9cfa0bb`] The migrated surfaces left their measuring behind.** Each the *other half* of something already deleted — `hit_test_width`, `layout_field_action_buttons` ("shared by the renderer and the click hit-tester" when neither existed), `Popup::{scroll_state, description_height}`, the prompt's own suggestion-list key handling.

**[`a30c60a`] Two reactions do not need a registry.** `ChromeComponent` was down to one method and two implementors, and its `false` default body had already lost one of them: the menu bar was registered but silently inert.

**[`94098da`] The paint-recorded geometry roster is empty.** The last two entries were a copy of `shell::prompt::{suggestions_list_rect, suggestions_rect}` kept for one consumer. What remains is `suggestions_window`, which is not a cache but feedback.

**[`c997c0f`] Sweep the view layer.** 881 lines, nineteen files: every `pub fn` under `view/` and `widgets/` with no reference anywhere. `pub` is why the compiler could not say so; the sweep runs to a fixed point against rustc's own dead-code warnings.

**[`47df667`] Delete the layout oracles**, and say what their tests were claiming — a grid is a tiling, a pane's interior is stated, the fold is compared against layout.

## The record

**[`5af2fa7`] and [`6102db3`]** keep `docs/internal/retained-mode-ui.md` in line, including an index of what is still open and corrections to three claims in it that were refuted by grep.

---

## What is left

- **Composite buffer panes never migrated at all** — `compute_pane_layout` writes `pane_widths`, four hand-rolled hit tests read it back, a fifth hardcodes `gutter_width = 4`. Largest piece left; wants its own branch.
- **The keyed geometry index** — `Ui::find_by_key` is still a depth-first walk with ~180 call sites. The live asymptotic hole.
- **The gutter** — line numbers, folds and diagnostics are painted, not nodes, so `click_geometry` re-derives the compose-mode gutter reclaim by hand.

## Checks

`cargo fmt --all`; `cargo check --workspace --all-targets` with the warning set diffed against master and identical; `cargo clippy` clean. Five of the seventeen commits were checked to compile on their own at the boundaries most likely to break.

**Checked against master screen for screen**: both binaries driven through the same scripted tmux scenarios — tab strip, menus, palette, find bar, explorer with git decorations, settings, keybinding editor, file browser, splits, scrollbar drag — capturing every screen with its colours, 98 screens at three widths. Every difference is one this branch makes on purpose.

An automated review of the full range found six defects; five are fixed and one rejected, with every thread replied to and resolved ([review](https://github.com/sinelaw/fresh/pull/3353#pullrequestreview-5283922530), [summary](https://github.com/sinelaw/fresh/pull/3353#issuecomment-5784685691)).

https://claude.ai/code/session_01QcWJ6w6r5yUpKt4ytke5Zi

---


crates/fresh-editor/src/view/shell/tabs.rs
```diff
@@ -393,135 +390,84 @@ fn cluster(pane: LeafId, c: Cluster, overflow: bool) -> Node<UiMsg> {
...
-        }
-    }
-    if draw_right {
-        cells.push(arrow(pane, ">", 1, scroll_right_key(pane)));
+    cells.push(
+        plus_node(pane, s.hover_plus, PLUS.to_string())
+            .w(Sizing::Cells(NEW_TAB_BUTTON_WIDTH as u16)),
+    );
+
+    let mut window = fresh_ui::viewport(row().children(cells))
```
sinelaw, line 508: **`rects()` (below, L488) now lists tabs that are scrolled out of the window.**

The strip's tabs are ordinary nodes inside a non-virtualising `ScrollMode::Cells` viewport, so *every* tab has a non-empty rect whether or not it is on screen. `shell::rect_of` only filters zero-sized rects, and `screen_rect` clamps a negative `x` to `0` — so a tab scrolled off the left reports a phantom box pinned at the strip's left edge.

That contradicts `rects()`'s own docstring ("A tab scrolled off the strip has no rectangle and is not listed") and both its consumers read it as geometry: the drag drop-zone index (`compute_tab_drop_zone` → `tab_insert_index`) and the web tab bar. It should also trip the `r.len() < 8` assertion in `an_overflowing_strip_caps_its_ends_and_reveals_on_request`.

Suggested fix: intersect each rect with the viewport's own box (the `window` node here) and drop the empties, rather than relying on `rect_of` to filter.

---

crates/fresh-ui/src/render/layout.rs
```diff
@@ -748,8 +776,16 @@ impl<M: 'static> Ui<M> {
...
         // still the ones the last arrange produced, and it is that offset they
         // have to be read against.
-        let top = child.data.rect.y - vp.data.rect.y + arranged_at.y;
-        Some((top, child.data.rect.h as i32))
+        Some(match axis {
+            crate::event::Axis::Vertical => (
+                child.data.rect.y - vp.data.rect.y + arranged_at.y,
+                child.data.rect.h as i32,
+            ),
+            crate::event::Axis::Horizontal => (
```
sinelaw, line 802: **`keyed_band`'s horizontal arm measures in the wrong coordinate space.**

It measures from the window's *outer* edge, but the child was placed past the reserved cap cells (`prim.rs`, `cx.place(k, Point::new(caps as i32, 0))`), while `here` and `ceiling` in the command loop are in content space. So every `reveal_key` / `top_key` / `reveal_key_at` on a horizontal window lands `cap` columns off.

This is invisible while `cap == 1` and the target is mid-strip, but it is exactly what `a_horizontal_window_reveals_by_key_across`'s final assertion checks — `assert_eq!(ui.scroll(vp).0.x, 0, "and back for the first")` should now get `1`. Widening the caps to 3 makes the error three columns.

Note this is the one piece of cap geometry that is *not* routed through the shared `overflow_caps` helper; folding it in would have caught this.

---

crates/fresh-ui/src/hit.rs
```diff
@@ -152,7 +152,16 @@ impl<M: 'static> Ui<M> {
                 // message it produces is the window's own report of where it
                 // went (`Node::on_scroll`), for an owner that holds the
                 // offset.
+                // A press on a horizontal window's overflow cap steps it
+                // that way. The cap is the window's own affordance — it knows
+                // whether there is more behind that edge, which is why it is
+                // the thing drawn there — so the move is the window's too, and
+                // produces no message beyond its own report.
                 if button == MouseButton::Left {
+                    if let Some((r, dir)) = self.overflow_cap_hit(pos) {
```
sinelaw, line 183: **The new overflow-cap branch returns before `dismiss_for_pointer`.**

A press on `<` / `>` takes this early return, so it no longer dismisses non-blocking `OUTSIDE_POINTER` layers — the LSP hover tooltip and signature help stay up. The arrow nodes this replaced were ordinary gesture nodes and went through the full press path, so this is a behaviour regression rather than a pre-existing gap.

The scrollbar-drag branch just below has the same shape, so whatever the right answer is (run the dismissal before the cap step, or let the cap step fall through) it probably applies to both.

---

crates/fresh-ui/src/render/layout.rs
```diff
@@ -816,91 +852,103 @@ impl<M: 'static> Ui<M> {
...
                             (top + h - rows).min(top)
                         } else {
-                            scroll.y
-                        };
-                        Point::new(scroll.x, y)
+                            here
+                        })
                     }
                 };
                 let next = Point::new(next.x.clamp(0, max.x.max(0)), next.y.clamp(0, max.y.max(0)));
```
sinelaw, line 972: **`apply_anchors` still reports `next.y` unconditionally** — a few lines below this hunk (`self.report_scroll(r, next.y, &mut told)`).

The axis generalisation stopped short of the report: a horizontal window's offset is `next.x`, so a *controlled* horizontal window (one whose owner holds the offset) is told `0` and snaps straight back. The tab strip doesn't hit this because it is uncontrolled, which is why it looks fine in the editor — but it makes `Anchor` commands on a controlled horizontal window silently inert.

Should be `main(axis, next)`, with `axis` read the same way the command loop above reads it.

---

crates/fresh-editor/src/app/window/mod.rs
```diff
@@ -2950,14 +2921,54 @@ impl Window {
...
                     active_pane: leaf == active_split,
                     hover: hover.and_then(|(t, pane, close)| (pane == leaf).then_some((t, close))),
-                    offset,
+                    hover_plus: hover_plus == Some(leaf),
+                    reveal: Some(self.tab_reveal_for(leaf)),
                     preview_label: preview_label.clone(),
                 },
             );
         }
         out
```
sinelaw, line 2950: **`reveal_active_tab` can queue commands on an `Anchor` that is never mounted.**

`tab_reveal_for(pane)` hands back an `Anchor` and `reveal_key` pushes onto its queue, but the anchor only ever drains if the strip's viewport node exists. A pane with no strip — `window.tabs` off, a group's inner leaf, suppressed chrome — never builds one, so the queue grows without bound for as long as that pane is active.

Separately, `Window::tab_reveal` is a `HashMap<LeafId, _>` that is never pruned when a pane closes, so it leaks an entry per closed split for the window's lifetime.

Both are small, but the first one also means the *next* time that pane does get a strip it will replay a backlog of stale reveals.

---

sinelaw (commented):
Automated review of this PR's full range (`8c7e1f1..9f19eb4`, 86 files, +2825/−4265), run without reference to the commit messages or to `docs/internal/retained-mode-ui.md` — both are the same author's claims and were treated as things to check, not evidence.

**Six findings, five inline below.** The sixth cannot be anchored inline because the file it concerns is not part of this PR's diff:

> **`crates/fresh-editor/tests/semantic/migrated_tab_scrolling.rs:128`** — `expected_tab_label` still implements the width-dependent name cap that this PR removed. The strip now caps at `TAB_NAME_MAX_COLS` unconditionally (a deliberate change, recorded in the design note), so the very first `assert_screen_contains` in `migrated_active_tab_visibility_with_scrolling` should fail. The helper needs to follow the new rule.

**Worth knowing while reading these:** the tab-strip work has never completed a test run. `0bd27f2` is the last commit on this branch whose three OS suites finished, and it predates the strip rewrite; every push since has cancelled its predecessor's run. Findings 1, 2, 3 and 6 all predict named existing tests failing, and the run now in flight on `9f19eb4` should adjudicate them directly.

### Areas that were called out and came back clean

- **`Ui::needs_frame` counting `layout_dirty`** — no repaint-loop risk. `drain_layout` takes the list with `mem::take` and loops to a fixpoint behind a 16-round assert, `flush_layout` is reached from both `frame` and `tick`, and the editor's only reader redraws on `true`, which drains it. The missed-frame case it fixes is real.
- **Overflow cap paint/hit agreement** — `paint_over` gates on `props.scrollbar && !bar_hidden` and the hit walk on `n.scrollbar`, so a cap cannot light where it cannot be pressed; `overflow_caps` is genuinely shared, and `step_window` reads the inner width rather than the outer rect. The one cap geometry *not* shared is `keyed_band` — finding 2.
- **Hover registry → two direct calls** — equivalent. The per-implementor `old == new` guard is now the caller's early return, `|` keeps both reactions running, and evaluation order matches the old `components()` order. No live references to the removed trait remain.
- **Cached suggestion rects → live tree reads** — `palette_view` is guarded by `prompt.as_ref()?` and `suggestions_window` is still reset per frame, so the no-popup case degrades to the same fallback as before. One residual difference worth a thought: with a prompt open whose suggestion list has just emptied, the tree still holds last frame's list rect, so `outer_rect`/`list_rect` can now be non-null with `total == 0` where both were previously `None`.
- **The bulk deletions under `view/**`** — every removed `fn`/`struct`/`enum`/`trait`/`const`/`type` name in the diff (210 of them) was extracted and grepped across the workspace. Every surviving hit is prose in a doc comment, or is in `crates/fresh-editor/tests/fixtures/large.rs`, which is not a compiled target. Nothing reachable by trait impl, macro or string name was missed. One leftover: `Viewport::needs_sync` is now a reader with no writer (`mark_needs_sync` and `sync_with_cursor` both went, and nothing else sets it `true`) — harmless, but dead.

---

crates/fresh-editor/src/view/shell/tabs.rs
sinelaw: **Confirmed and fixed** in `703d443`.

You're right about the mechanism: `ScrollMode::Cells` shows part of its content rather than building only the part it shows, so an off-screen tab keeps a full-width rect, `rect_of` keeps it (it drops only *empty* ones), and `screen_rect`'s `x.max(0)` folds a negative `x` back onto the frame. Phantom box at the strip's left edge, and the drag's drop zone resolves a drop there to the wrong tab.

The window now carries `tab_window_key(pane)` and `rects` clips each rect to it before converting, dropping the empties — your suggested fix. The press path never had this problem because a hit test asks the tree, which clips; the docstring now says so.

Two notes beyond the finding:

- `an_overflowing_strip_caps_its_ends_and_reveals_on_request` was indeed failing, and so was `the_cluster_takes_the_right_columns_and_the_overflow_arrow` — the latter asserted the `+` was pinned inside the window ahead of the cluster, which is the placement this PR deliberately removed. Both were already red before this review; the branch had never run them.
- Fixing the clip moved the first test's failure to its *next* assertion, which turned out to be finding #2 in disguise. See that thread.

---

crates/fresh-ui/src/render/layout.rs
sinelaw: **Confirmed, and it was a live failing test.** Fixed in `703d443`.

`a_horizontal_window_reveals_by_key_across` was red on this branch — `left: 1, right: 0`, exactly your prediction. I had only run the two tests I'd just written, not the one the cap-width change broke.

Your diagnosis is right down to the cause: the content's origin is *inside* the node's rectangle (placed past the leading cap), while `here`/`ceiling` count from the content's own zero. The fix subtracts the inset, read back as whatever the window did not keep for its view, halved — asking the layout that made the reservation rather than re-deriving the condition it was made under.

And you're right that routing this through `overflow_caps` would have caught it. It is the one cap geometry that was not shared.

This also explains something I had patched around interactively: revealing `tab_span_key` instead of `tab_key` "fixed" a strip that rested with the `×` past the edge. It was working because the overshoot happened to carry the `×` in. With the band correct, `an_overflowing_strip_caps_its_ends_and_reveals_on_request`'s one-cap expectation after revealing `tab_key` is simply wrong — the `×`, separator and `+` follow the label, so both ends have more — and that test was passing *because* of this bug. It now asserts both caps, and separately that revealing the `+` does reach the end, which is what `reveal_active_tab` asks for on the last tab and the reason it does.

---

crates/fresh-ui/src/hit.rs
sinelaw: **Confirmed and fixed** in `703d443`.

And you're right that the scrollbar branch has the same shape, so I took your second reading: the dismissal is a press *band* — it runs for every press wherever it lands — so it now sits above both early returns rather than below them. A click on `<`, `>` or a scrollbar gutter dismisses a transient surface again, which is what the arrows and bars these replaced did.

Whole `fresh-ui` suite green after the move, including the dismissal tests.

---

crates/fresh-ui/src/render/layout.rs
sinelaw: **Confirmed and fixed** in `703d443` — `report_scroll(r, main(axis, next), …)`.

Exactly as you describe: the axis generalisation stopped at the report, a horizontal window's `y` is always `0`, so an owner holding the offset was told the window had gone back to the start and put it there. Invisible on the tab strip only because it is uncontrolled.

---

crates/fresh-editor/src/app/window/mod.rs
sinelaw: **Half confirmed.** Second part fixed in `703d443`; first part I think is not reachable.

**The unbounded queue — rejecting, with reasoning.** `pane_interior` places `s.tabs` in its slot unconditionally; a pane with no tabs gets `h(Sizing::Cells(0))` on the strip *row*, not a missing node. So the viewport is still built and reconciled, `anchor_to` still binds it, and `a.take()` still drains each frame — a zero-height row does not unmount its children. The anchor's commands are spent on a degenerate window rather than piling up. Happy to be shown otherwise if you have a path where the strip node is genuinely absent while `reveal_active_tab` fires; every caller I traced (`active_focus`, `buffer_close`, `lifecycle`, `split_actions` ×2) passes a visible pane.

**The leak — confirmed and fixed.** `tab_reveal` is made on first sight of a pane exactly as `panes` is, so it is now dropped where `panes` is, in `retain_pane_handles`. Good catch — that map had no prune point at all.

---

sinelaw:
All six review findings addressed in `703d443` — five fixed, one rejected. The five inline threads have replies and are resolved; this is the sixth, which had no thread because its file is not in the diff.

> **`crates/fresh-editor/tests/semantic/migrated_tab_scrolling.rs:128`** — `expected_tab_label` still implements the width-dependent name cap this PR removed.

**Confirmed and fixed.** The helper measured the tabs, reserved the pinned `+` and returned the full name when the total fitted — mirroring a rule that is gone, because it was a measurement of the whole strip made before the description existed and a window can show what does not fit. It caps at `TAB_NAME_MAX_COLS` unconditionally now, and all three `migrated_tab_scrolling` tests pass.

### Summary

| # | Finding | Outcome |
|---|---|---|
| 1 | `rects()` lists tabs scrolled out of the window | **Fixed** — window carries `tab_window_key`, `rects` clips to it |
| 2 | `keyed_band`'s horizontal arm measures from the outer edge | **Fixed** — subtracts the reserved inset; was a live failing test |
| 3 | `expected_tab_label` implements the removed conditional cap | **Fixed** |
| 4 | Cap/scrollbar press skips `dismiss_for_pointer` | **Fixed** — dismissal hoisted above both early returns |
| 5 | `apply_anchors` reports `next.y` whatever the axis | **Fixed** — `main(axis, next)` |
| 6a | `reveal_active_tab` queues on an unmounted `Anchor` | **Rejected** — `pane_interior` always places the strip; a pane with no tabs gets `h(Cells(0))` on the row, not a missing node, so the anchor binds and drains |
| 6b | `tab_reveal` never pruned on pane close | **Fixed** — pruned in `retain_pane_handles` beside `panes` |

**The review was right that the branch had untested failures.** Two tests were already red before it ran — `a_horizontal_window_reveals_by_key_across` (finding 2, failing with exactly the predicted `1` vs `0`) and `an_overflowing_strip_caps_its_ends_and_reveals_on_request`. A third, `the_cluster_takes_the_right_columns_and_the_overflow_arrow`, was red for a related reason the review did not name.

Three test expectations turned out to encode the old behaviour and were corrected rather than worked around:

- `an_overflowing_strip_caps_its_ends_and_reveals_on_request` expected one cap after revealing `tab_key`, which names the label alone — the `×`, the separator and the `+` follow it, so both ends have more. It was passing *because* of finding 2's overshoot. It now also checks that revealing the `+` does reach the end, which is what `reveal_active_tab` asks for on the last tab and the reason it does.
- `the_cluster_takes_the_right_columns_and_the_overflow_arrow` asserted the `+` was pinned inside the window ahead of the cluster — the placement this PR deliberately removed, because it depended on the overflow. It now checks the cap is flush with the window's trailing edge and the window ends before the cluster's buttons.
- `expected_tab_label`, above.

Finding 2 also explains something I had patched around during interactive testing: revealing `tab_span_key` rather than `tab_key` appeared to fix a strip resting with the `×` past the edge. It was working because the overshoot happened to carry the `×` in.

Verified locally: whole `fresh-ui` suite (20 targets), `view::shell::tabs`, `migrated_tab_scrolling` — all green; no change to the warning set. The rest is CI's.

---

